PR #78, opened 2012-12-26
AMQP-283/284 Upgrade to RabbitMQ 3.0.0; AMQP-287 ConnectionTimeout

garyrussell:
## AMQP-284 Deprecate immediate (publish) flag, Deprecate Federated Exchanges.

Exchange Federation is no longer configured using
exchange declaration.

Federation: Pre 3.0
Configuration lives in the configuration file.
You must ensure all nodes in a cluster have the same configuration.
The broker must be restarted for changes to take effect.
Federated exchanges are of a special type, and must be declared as such by AMQP clients.

Federation: 3.0
Configuration is stored in the broker database, manipulated with rabbitmqctl or the management plugin.
All nodes in a cluster automatically have the same configuration.
Changes to federation take effect immediately.
Federated exchanges are transparent to clients; exchanges can become federated at any time.
## AMQP-287 Add Configurable Consumer Start Timeout

See https://jira.springsource.org/browse/AMQP-287

Allow the configuration of an increased timeout for when
SYN packets are dropped.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
@@ -501,16 +501,16 @@ public class SimpleMessageListenerContainer extends AbstractMessageListenerConta
 
 		/**
 		 * Retrieve the fatal startup exception if this processor completely failed to locate the broker resources it
-		 * needed. Blocks up to 60 seconds waiting (but should always return promptly in normal circumstances).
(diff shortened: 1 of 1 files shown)


veerareddyakkili:
what properties need to set in spring tool suite to retry only limited times in case if Rabbit MQ goes down.

artembilan:
@veerareddyakkili ,

Your question isn't related to the subject of this closed PR almost 4 years ago.

Please, consider to ask questions on the Community support forums, e.g. StackOverflow.

We use GitHub only for bugs and feature or improvement requests.

PR #159, opened 2014-01-30
AMQP-369: Fix AmqpAdmin-Declarable classes tangle

artembilan:
JIRA: https://jira.springsource.org/browse/AMQP-369


rvervaek:
@artembilan 

I'm trying to understand why this change has been made. Why has the type of for the declaringadmins changed from AmqpAdmin to Object?

garyrussell:
The declaringAdmins can either be a reference to an `AmqpAdmin` bean, or a bean name which is used to prevent early instantiation in a bean post processor.

PR #245, opened 2014-12-22
AMQP-457: ListenerContainer Provide Consumer Queue

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-457

If the listener container is configured to listen to multiple queues,
it would be useful for the listener to have access to which queue a
message was received from.

Instead of a collection of consumer tags, maintain a `Map` of consumer
tags to queue names and populate the message properties with the
tag and queue name.

In the spring-messaging header mapper, map the properties to headers
(inbound only).


artembilan:
LGTM
Merging...

artembilan:
Merged after rebase

PR #248, opened 2015-01-05
AMQP-463: Fix Publisher Confirms with JRockit

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-463

JRockit throws an `IllegalArgumentException` if a
`TreeMap$Entry` is used after removal via the `Iterator`.

The `PublishSubscribeChannelImpl` uses this technique.

Obtain the entry value before removing the entry.

Add a test to reproduce the problem with the JRockit JVM.

Currently sends 10k messages across 100 threads; also tested
with 1m messages across 100 threads with no lost confirms.


artembilan:
LGTM. Tested locally with `gradlew testAll`
Merging...

PR #250, opened 2015-01-07
AMQP-138: Add Content-Type Message Converter

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-138

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/support/converter/ContentTypeDelegatingMessageConverter.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/support/converter/ContentTypeDelegatingMessageConverter.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.amqp.support.converter;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.springframework.amqp.core.Message;
+import org.springframework.amqp.core.MessageProperties;
+
+ /**
+  * A composite {@link MessageConverter} that delegates to an actual {@link MessageConverter}
+  * based on the contentType header. Supports a default converter when no content type matches.
+  * Note: the {@link MessageProperties} requires a content type header to select a converter
+  * when used for outbound conversion, but the converter will (generally) override it to match
+  * the actual conversion.
+  *
+  * @author Eric Rizzo
+  * @author Gary Russell
+  * @since 1.4.2
+  *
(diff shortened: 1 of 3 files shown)


spring-amqp/src/main/java/org/springframework/amqp/support/converter/ContentTypeDelegatingMessageConverter.java
```diff
@@ -0,0 +1,102 @@
...
+	public ContentTypeDelegatingMessageConverter() {
+		this(new SimpleMessageConverter());
+	}
+
+	/**
+	 * Constructs an instance using a the supplied default converter (may be null).
+	 * @param defaultConverter the converter.
+	 */
+	public ContentTypeDelegatingMessageConverter(MessageConverter defaultConverter) {
+		this.defaultConverter = defaultConverter;
```
artembilan: Just to be sure that it is expected. 
Since we don't have here `Assert.notNull`, it means that we can configure it without the `defaultConverter` to make this `ContentTypeDelegatingMessageConverter` as strict one. And this logic I see in the `getConverterForContentType`. 
Is that what you have in mind?

spring-amqp/src/main/java/org/springframework/amqp/support/converter/ContentTypeDelegatingMessageConverter.java
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	@Override
+	public Message toMessage(Object object, MessageProperties messageProperties) {
+		String contentType = messageProperties.getContentType();
+		return getConverterForContentType(contentType).toMessage(object, messageProperties);
+	}
+
+	protected MessageConverter getConverterForContentType(String contentType) {
+		MessageConverter delegate = getDelegates().get(contentType);
```
artembilan: Since we don't have `Assert.hasText(contentType)`, it means that `contentType` can be `null`, hence we can end up with `no converter` from the `Map` and try to use the `defaultConverter` if present.
Otherwise we end up with that `No delegate converter is specified for content type null`.
Is that correct too?

spring-amqp/src/test/java/org/springframework/amqp/support/converter/ContentTypeDelegatingMessageConverterTests.java
```diff
@@ -0,0 +1,83 @@
...
+
+		converter = new ContentTypeDelegatingMessageConverter(null); // no default
+		try {
+			message = converter.toMessage(foo, props);
+			fail("Expected exception");
+		}
+		catch (Exception e) {
+			assertThat(e, instanceOf(MessageConversionException.class));
+			assertThat(e.getMessage(), Matchers.containsString("No delegate converter"));
+		}
```
artembilan: Yes, I see this test confirms my concerns.
Let me know if I haven't missed something and I'll go ahead with merge

spring-amqp/src/main/java/org/springframework/amqp/support/converter/ContentTypeDelegatingMessageConverter.java
garyrussell: Yes; that was the intention; maybe enhance the javadoc...

`... (may be null meaning a strict content-type match is required).`

artembilan: Good. Will be polished on merge

artembilan:
LGTM
Merging...

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/0b9b99ac10f567aad810430144010b6e53bcf3ed

PR #252, opened 2015-01-10
AMQP-466: Move Direct reply-to Decoding to Address

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-466

1.4.1 introduced direct reply to and reply to address decoding required
using `AddressUtils`. The code should have been put in the `Address` ctor.

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/core/Address.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/core/Address.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2014 the original author or authors.
+ * Copyright 2002-2015 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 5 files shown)


spring-amqp/src/main/java/org/springframework/amqp/core/Address.java
```diff
@@ -58,6 +60,10 @@ public Address(String address) {
 			this.exchangeName = "";
 			this.routingKey = "";
 		}
+		else if (address.startsWith(AMQ_RABBITMQ_REPLY_TO)) {
+			this.routingKey = address;
+			this.exchangeName = "";
```
artembilan: H-m-m. It is strange that we have missed this fix for the first implementation...
It isn't breaking change, unless the movement of that `AMQ_RABBITMQ_REPLY_TO` const.
Will hope it won't harm any upgraded app to the AMQP 1.4.2
Merging and cherry-picking...

PR #254, opened 2015-01-22
AMQP-469 Introduce `Connection#sendAndReceiveRPC`

artembilan:
JIRA: https://jira.spring.io/browse/AMQP-469

Since `OtpConnection` isn't thread-safe between `sendRPC` and `receiveRPC`, add a new `sendAndReceiveRPC`
and make its implementation in the `DefaultConnection` as `synchronized`.

Modify `ErlangTemplate` to use that new operation.

Having that even with `SingleConnectionFactory` we don't meet race condition when one thread may receive result of another one.


garyrussell:
Merged

PR #256, opened 2015-01-29
Added "static" to inner BlockingQueueConsumer.DeclarationException

panchenko:


artembilan:
@panchenko , would be good if you'd provide some explanation of reason for such a change.
If we take a look to the source code around this `BlockingQueueConsumer.DeclarationException`, we'll see that there is no point which rethrows it to the caller - to the main start application Thread.

The reason not having it as a `static` do not allow to instantiate it without `BlockingQueueConsumer` - it's outer class.

So, I don't see reason to accept this change.

Thanks

panchenko:
@artembilan Inner classes without `static` have automatically generated field, referencing the outer class (and constructor to initialize it), if present it is used to access members of the outer class, but is not needed in this case. In byte code it looks like `final BlockingQueueConsumer this$0`. So the reason is a small optimization.

If an idea is restricting usage of it - I would suggest also changing visibility to package scope.

artembilan:
Ah! Yes. Good reason. Thanks!

No one day without a new knowledge!

However I can't merge your commit because it is based on the `master` branch.
Since this change is so simple I'm committing it to the `master` directly with my privileges..

For the future be sure that you create a new branch for changes.
And it would be better to start from JIRA ticket, rather than direct PR.

Merged as https://github.com/spring-projects/spring-amqp/commit/bc902e1c54af2adc2cdbc69bb46e9bb7e462d6ee

PR #257, opened 2015-01-29
AMQP-472: Recover From Conn. Close During Init

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-472

Previously, any `IOException` during passive queue declaration would enter
declaration retry and eventually throw a `QueuesNotAvailableException`. Whether
or not that is recoverable depends on the container's `missingQueuesFatal` property.

If the `IOException` is due to a connection close, we should not try to redeclare
and, further, recovery should be unconditional.

When a queue declaration fails, check if the connection is open and, if not, throw
an `AmqpIOException`, causing container recovery to begin. If the connection is open
continue retrying queue declaration as before.

In addition, expose the queue declaration retry properties on the `SimpleMessageListenerContainer`.

__cherry-pick to 1.4.x, 1.3.x; do not resolve JIRA - a second PR is required for namespace support for the new properties in 1.5.x__

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
@@ -244,7 +244,8 @@ public class BlockingQueueConsumer {
 	/**
 	 * Set the number of retries after passive queue declaration fails.
 	 * @param declarationRetries The number of retries, default 3.
-	 * @see #setFailedDeclarationRetryInterval(int)
+	 * @see #setFailedDeclarationRetryInterval(long)
+	 * @since 1.3.9
 	 */
 	public void setDeclarationRetries(int declarationRetries) {
 		this.declarationRetries = declarationRetries;
@@ -254,6 +255,7 @@ public class BlockingQueueConsumer {
 	 * Set the interval between passive queue declaration attempts in milliseconds.
 	 * @param failedDeclarationRetryInterval the interval, default 5000.
 	 * @see #setDeclarationRetries(int)
+	 * @since 1.3.9
 	 */
 	public void setFailedDeclarationRetryInterval(long failedDeclarationRetryInterval) {
 		this.failedDeclarationRetryInterval = failedDeclarationRetryInterval;
@@ -263,6 +265,7 @@ public class BlockingQueueConsumer {
 	 * When consuming multiple queues, set the interval between declaration attempts when only
 	 * a subset of the queues were available (milliseconds).
 	 * @param retryDeclarationInterval the interval, default 60000.
+	 * @since 1.3.9
 	 */
 	public void setRetryDeclarationInterval(long retryDeclarationInterval) {
 		this.retryDeclarationInterval = retryDeclarationInterval;
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
@@ -53,7 +53,6 @@ import com.rabbitmq.client.AMQP;
 import com.rabbitmq.client.AMQP.BasicProperties;
 import com.rabbitmq.client.AlreadyClosedException;
 import com.rabbitmq.client.Channel;
-import com.rabbitmq.client.Connection;
 import com.rabbitmq.client.DefaultConsumer;
 import com.rabbitmq.client.Envelope;
 import com.rabbitmq.client.ShutdownSignalException;
@@ -86,8 +85,6 @@ public class BlockingQueueConsumer {
 	private final boolean transactional;
 
 	private Channel channel;
-
-	private Connection connection;
 
 	private RabbitResourceHolder resourceHolder;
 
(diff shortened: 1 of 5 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
```diff
@@ -233,6 +240,33 @@ public String getConsumerTag() {
 	public final void setQuiesce(long shutdownTimeout) {
 	}
 
+	/**
+	 * Set the number of retries after passive queue declaration fails.
+	 * @param declarationRetries The number of retries, default 3.
+	 * @see #setFailedDeclarationRetryInterval(int)
+	 */
+	public void setDeclarationRetries(int declarationRetries) {
```
artembilan: Need `@since 1.3.9` for any new public API.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerIntegration2Tests.java
```diff
@@ -318,6 +319,86 @@ public Channel answer(InvocationOnMock invocation) throws Throwable {
...
 		assertTrue(networkGlitch.get());
+
+		container.stop();
+	}
+
+	@Test
+	public void testRestartConsumerOnConnectionLossDuringQueueDeclare() throws Exception {
+		this.template.convertAndSend(queue.getName(), "foo");
+
+		ConnectionFactory connectionFactory = new CachingConnectionFactory("localhost", BrokerTestUtils.getPort());
```
artembilan: When I was writing that test for the `basicQos` I worried about the new `connection` for more than one `connectionFactory.createConnection()`. But now I see that it doesn't matter, because default mode is `CacheMode.CHANNEL`, so we share the single connection.
Hence your solution is OK. 
LGTM, merging with simple polishing for those `@since`, `@Copyright` and rebase to `master`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
```diff
@@ -483,8 +518,11 @@ private void attemptPassiveDeclarations() {
 				if (logger.isWarnEnabled()) {
 					logger.warn("Failed to declare queue:" + queueName);
 				}
+				if (!this.connection.isOpen()) {
```
artembilan: Here is the simple polishing https://github.com/artembilan/spring-amqp/commit/a2cd0782e6480240d5e78b7abfed99789654da51
However I want to know why `this.channel.isOpen()` isn't enough here?
Looks like closing connection causes close for all its channels. See `AMQConnection.finishShutdown`

garyrussell: Polish looks good.

Rationale; my main reason for checking the connection is because the underlying channel closes when a declaration fails (e.g. due to a 404 NOT_FOUND), and it is refreshed by the `CCF.CCIH` by catching the `InvocationTargetException` so we'd actually be testing a different channel.

Of course, if the error was due to a lost connection and the channel `isOpen()` then it means the connection/channel was successfully refreshed. If the channel could not be refreshed, it would be `null` and the proxy would return false anyway, so that would be ok.

I was originally looking at adding code to only retry if we got a 404 (or maybe a 405) but because the refresh logic is in the catch block we might not actually get the original exception - we might get a new exception from the `createBareChannel`.

Now I think about it some more (thanks to your question). I think it's better to test the channel because the connection we captured during `start()` could possibly be a different connection to the one the exception actually occurred on (due to the refresh logic), especially when declaring multiple queues.

Pushed

I will apply your polish and push another change

artembilan:
With the last change the `BlockingQueueConsumerTests` doesn't pass.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumerTests.java
```diff
@@ -113,6 +113,9 @@ public void testPrefetchIsSetOnFailedPassiveDeclaration() throws IOException {
 
 		when(connectionFactory.createConnection()).thenReturn(connection);
 		when(connection.createChannel(Mockito.anyBoolean())).thenReturn(channel);
+		com.rabbitmq.client.Connection rabbitConnection = mock(com.rabbitmq.client.Connection.class);
+		when(rabbitConnection.isOpen()).thenReturn(true);
+		when(channel.getConnection()).thenReturn(rabbitConnection);
```
artembilan: `when(channel.isOpen()).thenReturn(true);` instead.
And:
```
blockingQueueConsumer.setRetryDeclarationInterval(100);
blockingQueueConsumer.setFailedDeclarationRetryInterval(100);
```
to make this test quick as before.

artembilan: If you are OK with that, I go ahead with merge

garyrussell: Oops; yes, of course.

artembilan:
Polishing according our discussion: https://github.com/artembilan/spring-amqp/commit/7f7c60c079cf5e7533bdd0182fdb38fdcf4d072c

garyrussell:
:+1:

artembilan:
Merging...

artembilan:
Merged and cherry-picked: https://github.com/spring-projects/spring-amqp/commit/c40f233fdd1f6469581e7cb873a56fd9a513b6c1

PR #259, opened 2015-01-31
[DO NOT MERGE] WIP - ReST Abstraction

garyrussell:


garyrussell:
Hi Boris; I added a second commit that maps the bindings to `Binding` domain objects; it turned out to be not too bad.

I added a `properties` field to `Binding` to capture additional properties returned (such as `vhost`).

I changed it to a pull request so we can continue the discussion here instead of on individual commits.

cc/ @artembilan

garyrussell:
Replaced by #260

PR #260, opened 2015-02-03
AMQP-473: Add REST RabbitManagementTemplate

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-473

RabbitMQ ReST API - Initial PoC

ReST PoC Second Commit

- Improve URI encoding
- Add ctor with CredentialsProvider
- Convert to Binding
- Add properties field to Binding

Fix Encoding; Remove CredentialsProvider ctor

- See https://jira.spring.io/browse/SPR-12679
- No need for ctor - user can supply a RestTemplate

Add remaining First Class Support Paths

- Use preemptive basic auth
- Add remaining elements

__Replaces #259 __

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -74,6 +74,7 @@ subprojects { subproject ->
 		commonsIoVersion = '2.4'
 		erlangOtpVersion = '1.5.6'
 		hamcrestVersion = '1.3'
+		httpClientVersion = "4.3.6"
 		jacksonVersion = '1.9.13'
 		jackson2Version = '2.3.2'
 		junitVersion = '4.11'
@@ -175,6 +176,8 @@ project('spring-amqp') {
 	dependencies {
 
 		compile "org.springframework:spring-core:$springVersion"
+		compile "org.springframework:spring-web:$springVersion"
+		compile "org.apache.httpcomponents:httpclient:$httpClientVersion"
 		compile ("org.springframework:spring-messaging:$springVersion", optional)
 		compile ("org.springframework:spring-oxm:$springVersion", optional)
 		compile ("org.springframework:spring-context:$springVersion", optional)
--- a/build.gradle
+++ b/build.gradle
@@ -176,7 +176,6 @@ project('spring-amqp') {
 	dependencies {
 
 		compile "org.springframework:spring-core:$springVersion"
-		compile "org.springframework:spring-web:$springVersion"
 		compile ("org.springframework:spring-messaging:$springVersion", optional)
 		compile ("org.springframework:spring-oxm:$springVersion", optional)
 		compile ("org.springframework:spring-context:$springVersion", optional)
@@ -212,6 +211,7 @@ project('spring-rabbit') {
 		compile "com.rabbitmq:amqp-client:$rabbitmqVersion"
 
 		compile ("org.springframework:spring-aop:$springVersion", optional)
+		compile ("org.springframework:spring-web:$springVersion", optional)
 		compile ("org.apache.httpcomponents:httpclient:$httpClientVersion", optional)
 		compile "org.springframework:spring-context:$springVersion"
 		compile "org.springframework:spring-messaging:$springVersion"
--- a/build.gradle
+++ b/build.gradle
@@ -177,7 +177,6 @@ project('spring-amqp') {
 
 		compile "org.springframework:spring-core:$springVersion"
 		compile "org.springframework:spring-web:$springVersion"
-		compile "org.apache.httpcomponents:httpclient:$httpClientVersion"
 		compile ("org.springframework:spring-messaging:$springVersion", optional)
 		compile ("org.springframework:spring-oxm:$springVersion", optional)
 		compile ("org.springframework:spring-context:$springVersion", optional)
@@ -213,6 +212,7 @@ project('spring-rabbit') {
 		compile "com.rabbitmq:amqp-client:$rabbitmqVersion"
 
 		compile ("org.springframework:spring-aop:$springVersion", optional)
+		compile ("org.apache.httpcomponents:httpclient:$httpClientVersion", optional)
 		compile "org.springframework:spring-context:$springVersion"
 		compile "org.springframework:spring-messaging:$springVersion"
 		compile "org.springframework:spring-tx:$springVersion"
--- a/spring-amqp/src/main/java/org/springframework/amqp/core/Queue.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/core/Queue.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2013 the original author or authors.
+ * Copyright 2002-2015 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
  * the License. You may obtain a copy of the License at
@@ -77,6 +77,23 @@ public class Queue extends AbstractDeclarable {
 	 * @param arguments the arguments used to declare the queue
 	 */
 	public Queue(String name, boolean durable, boolean exclusive, boolean autoDelete, Map<String, Object> arguments) {
+		this(name, durable, exclusive, autoDelete, arguments, null);
+	}
+
+	/**
+	 * Construct a new queue, given a name, durability flag, and auto-delete flag, and arguments.
+	 * @param name the name of the queue - must not be null; set to "" to have the broker generate the name.
+	 * @param durable true if we are declaring a durable queue (the queue will survive a server restart)
+	 * @param exclusive true if we are declaring an exclusive queue (the queue will only be used by the declarer's
+	 * connection)
+	 * @param autoDelete true if the server should delete the queue when it is no longer in use
+	 * @param arguments the arguments used to declare the queue
+	 * @param properties addition properties returned over REST by the broker
+	 *
+	 */
+	public Queue(String name, boolean durable, boolean exclusive, boolean autoDelete, Map<String, Object> arguments,
+			Map<String, Object> properties) {
+		super(properties);
 		Assert.notNull(name, "'name' cannot be null");
 		this.name = name;
 		this.durable = durable;
--- a/spring-amqp/src/main/java/org/springframework/amqp/core/Queue.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/core/Queue.java
@@ -89,7 +89,7 @@ public class Queue extends AbstractDeclarable {
 	 * @param autoDelete true if the server should delete the queue when it is no longer in use
 	 * @param arguments the arguments used to declare the queue
 	 * @param properties addition properties returned over REST by the broker
-	 *
+	 * @since 1.5
 	 */
 	public Queue(String name, boolean durable, boolean exclusive, boolean autoDelete, Map<String, Object> arguments,
 			Map<String, Object> properties) {
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplate.java
@@ -330,10 +330,7 @@ public class RabbitManagementTemplate {
 	public static Queue queueMapToQueue(Map<String, Object> map) {
 		String name = (String) map.get("name");
 		boolean durable = (Boolean) map.get("durable");
-		Boolean exclusive = (Boolean) map.get("exclusive");
-		if (exclusive == null) {
-			exclusive = false;
-		}
+		boolean exclusive = map.get("owner_pid_details") != null;
 		boolean autoDelete = (Boolean) map.get("auto_delete");
 		@SuppressWarnings("unchecked")
 		Map<String, Object> arguments = (Map<String, Object>) map.get("arguments");
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplate.java
@@ -18,6 +18,7 @@ package org.springframework.amqp.rabbit.core;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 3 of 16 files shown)


garyrussell:
@artembilan @bmorris591 

I consider this to be the minimum viable REST abstraction for rabbit; as previously mentioned, it doesn't support adding queues/exchanges/bindings (already handled by RabbitAdmin) but supports listing same.

It also supports adding policies and has a generic method for GET methods (paths) that don't have first class support (e.g. connections, channels).

We can add more first class support if deemed necessary before GA.

TODO: Docs and javadocs

build.gradle
```diff
@@ -175,6 +176,8 @@ project('spring-amqp') {
 	dependencies {
 
 		compile "org.springframework:spring-core:$springVersion"
+		compile "org.springframework:spring-web:$springVersion"
+		compile "org.apache.httpcomponents:httpclient:$httpClientVersion"
```
artembilan: First of all `optional`. If I am not interested in that `RabbitManagementTemplate` I would not like to have these dependencies.
From other side: the wrong `project`. The `RabbitManagementTemplate` is really for RabbitMQ, so move these to the `spring-rabbit` project, please.

spring-amqp/src/main/java/org/springframework/amqp/core/Queue.java
```diff
@@ -77,6 +77,23 @@ public Queue(String name, boolean durable, boolean exclusive, boolean autoDelete
...
+	/**
+	 * Construct a new queue, given a name, durability flag, and auto-delete flag, and arguments.
+	 * @param name the name of the queue - must not be null; set to "" to have the broker generate the name.
+	 * @param durable true if we are declaring a durable queue (the queue will survive a server restart)
+	 * @param exclusive true if we are declaring an exclusive queue (the queue will only be used by the declarer's
+	 * connection)
+	 * @param autoDelete true if the server should delete the queue when it is no longer in use
+	 * @param arguments the arguments used to declare the queue
+	 * @param properties addition properties returned over REST by the broker
+	 *
```
artembilan: `@since 1.5`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplate.java
```diff
@@ -0,0 +1,360 @@
...
+			protected HttpContext createHttpContext(HttpMethod httpMethod, URI uri) {
+				return localContext;
+			}
+
+		});
+		this.restTemplate = restTemplate;
+		this.baseUri = baseUri;
+	}
+
+	public RabbitManagementTemplate(String baseUri, RestTemplate restTemplate) {
```
artembilan: I'm not sure in this ctor for generic `RestTemplate`. I think we should hide the target implementation and just allow to provide high-level options as it is done in the previous ctor.
From other side you do that `RestTemplate` customization for that `preemptive basic auth`. Having this ctor we don't have guarantees that end-user provide correct `RestTemplate`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplate.java
```diff
@@ -0,0 +1,360 @@
...
+		localContext.setAuthCache(authCache);
+		RestTemplate restTemplate = new RestTemplate(new HttpComponentsClientHttpRequestFactory(httpClient) {
+
+			@Override
+			protected HttpContext createHttpContext(HttpMethod httpMethod, URI uri) {
+				return localContext;
+			}
+
+		});
+		this.restTemplate = restTemplate;
```
artembilan: I think we should optimize it a bit:
````java
this.restTemplate.setMessageConverters(Collections.<HttpMessageConverter<?>>singletonList(
				new MappingJackson2HttpMessageConverter()));
````
Since we get deal only with `application/json` `content-type`

File: spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplate.java [hunk 5]
```diff
@@ -0,0 +1,360 @@
...
+		this.baseUri = baseUri;
+	}
+
+	public RabbitManagementTemplate(String baseUri, RestTemplate restTemplate) {
+		this.restTemplate = restTemplate;
+		this.baseUri = baseUri;
+	}
+
+	@SuppressWarnings("unchecked")
+	public Map<String, Object> aliveness(String vhost) {
```
artembilan: `/api/aliveness-test/vhost`:
> If everything is working correctly, will return HTTP status 200 with body:
{"status":"ok"}

So, this method should return just `boolean` and can use `getForEntity` to extract just that `statusCode`, or even be as `void` and throws an Exception if status isn't `200 OK`.

Do you know what they return if there is anything wrong ?

artembilan: In  addition: can't we have in addition a default `vhost` const to have methods without `vhost` arg?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplate.java
```diff
@@ -0,0 +1,360 @@
...
+		for (Map<String, Object> map : queueMaps) {
+			queues.add(queueMapToQueue(map));
+		}
+		return queues;
+	}
+
+	public static Queue queueMapToQueue(Map<String, Object> map) {
+		String name = (String) map.get("name");
+		boolean durable = (Boolean) map.get("durable");
+		boolean exclusive = false; // not returned (Boolean) map.get("exclusive");
```
artembilan: `exclusive_consumer_tag` ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplate.java
```diff
@@ -0,0 +1,360 @@
...
+
+	public List<Queue> queues(String vhost) {
+		URI uri = UriComponentsBuilder.fromUriString(this.baseUri).pathSegment("queues", "{vhost}")
+				.buildAndExpand(vhost).encode().toUri();
+		@SuppressWarnings("unchecked")
+		List<Map<String, Object>> queueMaps = this.restTemplate.getForObject(uri, List.class);
+		return queueMapsToQueues(queueMaps);
+	}
+
+	public Queue queue(String vhost, String exchange) {
```
artembilan: s/exchange/queue

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplateTests.java
```diff
@@ -0,0 +1,261 @@
...
+ *
+ */
+public class RabbitManagementTemplateTests {
+
+	private final CachingConnectionFactory connectionFactory = new CachingConnectionFactory("localhost");
+
+	private final RabbitManagementTemplate template = new RabbitManagementTemplate();
+
+	@ClassRule
+	public static BrokerRunning brokerRunning = BrokerRunning.isRunning();
```
artembilan: I think this rule isn't enough for this test, because `management-plugin` must be enabled on the Broker.
At least on Windows I have to switch on it manually, as well as MQTT and STOMP.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplateTests.java
```diff
@@ -0,0 +1,261 @@
...
+
+		this.template.removePolicy("/", name);
+		list = this.template.policies();
+		out = null;
+		for (Policy policyOut : list) {
+			if (policyOut.getName().equals(name)) {
+				out = policyOut;
+				break;
+			}
+		}
```
artembilan: Looks like this test ins't finished.
```
assertNull(out);
```
ins the end ?

artembilan:
@garyrussell , that's all.
Work looks straightforward. Thanks.
I was going to suggests to use 
```
<T> ResponseEntity<T> exchange(URI url, HttpMethod method, HttpEntity<?> requestEntity,
			ParameterizedTypeReference<T> responseType) throws RestClientException;
```
But for that case we really should introduce domain classes to allow for Jackson automatically populate entities. And after that we can convert them to our existing domain.
But, of course, we can live with that manual population from the maps!

build.gradle
garyrussell: doh

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplate.java
garyrussell: We would need to document the need for preemptive (unless the rabbit guys fix it); my concern here is someone might want to use, say, a custom credentials provider to use encrypted credentials. I originally had a ctor that took a `CredentialsProvider` for that reason, but decided the more generic `RestTemplate` ctor solved that problem.

We can discuss further.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplate.java
garyrussell: I thought about that (boolean result) but this will not require any changes if they later add more properties.

+1 for the default vhost version.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplate.java
garyrussell: I saw that but assumed it meant an exclusive consumer was using the queue (which is different to an exclusive queue). I'll investigate.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplateTests.java
garyrussell: +1

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplateTests.java
garyrussell: Hmmm - don't know what happened to that; it was there at one time; must have dropped it during clean up.

garyrussell:
Yes, I thought about that, but I would rather keep it flexible so rabbitmq can add / remove non-core properties without us having to change this code every time (hence the `properties` bucket in the existing entities).

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplate.java
garyrussell: I confirmed that `exclusive_consumer_tag` does not represent `exclusive` (in fact even with an exclusive consumer on the queue, the `exclusive_consumer_tag` was emtpy :frowning: )

artembilan: So, one more question to RabbitMQ guys!

michaelklishin: This sounds like a bug.

simonmacmullen: Hmm, I just tested and exclusive_consumer_tag works for me (i.e. it is filled if the queue has an exclusive consumer, as mentioned it's not related to the queue being exclusive).

simonmacmullen: Also: to determine if a queue is exclusive, look for the key `owner_pid_details`. Yes, that's a hideous name. Sorry.

garyrussell: Thanks, @simonmacmullen , looks like a timing issue; I invoked the API right after receiving the consumeOK and the tag was empty; when I added a short wait, it's now populated (as well as the `consumer_details`).

I still don't see the queue's `exclusive` bit in the returned json tho'.

simonmacmullen: Yeah, pretty much everything reporting-ish in the API is delayed, we periodically emit stats events (and `exclusive_consumer_tag` counts as a "statistic" internally).

I'm saying that if you declare an exclusive queue (not a queue with an exclusive consumer) then it will have an `owner_pid_details` key in the queue JSON. There's no `exclusive` key.

garyrussell:
Pushed polish to address comments; still TODO: docs and test with https.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplate.java
garyrussell: Thanks.

garyrussell: So if a queue is declared exclusive but nobody owns it yet, there's no way to tell it __will__ be exclusive. Right?

garyrussell: Hmmm - of course, that makes no sense; I am mixing my exclusives again :frowning:

garyrussell: I see the `owner_pid_details` key; thanks again for your help.

build.gradle
```diff
@@ -175,6 +176,7 @@ project('spring-amqp') {
 	dependencies {
 
 		compile "org.springframework:spring-core:$springVersion"
+		compile "org.springframework:spring-web:$springVersion"
```
artembilan: But why don't you move this one? And `optional`, of course

artembilan:
That's all. Waiting for TODOs

build.gradle
garyrussell: Doh; ok.

Tested with http; will do docs tomorrow or Friday.

garyrussell:
Pushed docs, javadoc.

garyrussell:
Closing for now; will pick it up again based on [hop](https://github.com/michaelklishin/hop).

michaelklishin:
An update: [Hop](https://github.com/rabbitmq/hop) is now under the RabbitMQ organisation and getting some love again. Once we feel it is more than 60% complete we'll involve the Spring team.

PR #261, opened 2015-02-06
AMQP-474: Fix Log Appenders

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-474

When logs are generated soon after initialization (as is the case for
the test cases), some log messages could (infrequently) be lost.

This is due to the `EventSender` seeing a `null` `applicationId`, generating
a routing key:

    [null.org.springframework.amqp.rabbit.logback.AmqpAppenderIntegrationTests.DEBUG]

instead of:

    [AmqpAppenderTest.org.springframework.amqp.rabbit.logback.AmqpAppenderIntegrationTests.DEBUG]

(in the case of the logback test case).

Since this does not match a binding, the log message is discarded and the test fails.

Since the `applicationId` setter is called before `start()`, on the same thread, the
only way this can happen is if the `EventSender` runs on a CPU that has a stale
copy of the `AmqpAppender` in its cache (the applicationId field is not marked `volatile`).

Rather than making all the log variables volatile, synchronized blocks are now
used during `EventSender` initialization, to ensure the senders see the variable
values correctly.

Also enable logback debug.

__cherry-pick to 1.4.x, 1.3.x__

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j/AmqpAppender.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j/AmqpAppender.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2014 by the original author(s).
+ * Copyright (c) 2011-2015 by the original author(s).
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
  * the License. You may obtain a copy of the License at
@@ -409,6 +409,8 @@ public class AmqpAppender extends AppenderSkeleton {
 	 */
 	protected void startSenders() {
 		senderPool = Executors.newCachedThreadPool();
+		synchronized(this) {
+		} // (logically) flush all variables to main memory
 		for (int i = 0; i < senderPoolSize; i++) {
 			senderPool.submit(new EventSender());
 		}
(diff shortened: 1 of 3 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j/AmqpAppender.java
```diff
@@ -409,6 +409,8 @@ public void setCharset(String charset) {
 	 */
 	protected void startSenders() {
 		senderPool = Executors.newCachedThreadPool();
+		synchronized(this) {
+		} // (logically) flush all variables to main memory
```
artembilan: Hey!
How have you came up with this?
Merging anyway.
Cool! :+1:

artembilan:
Merged and cherry-picked.
Note, to `1.3.x` cherry-picked only the `org.springframework.amqp.rabbit.log4j.AmqpAppender`, because `logback` has been introduced since `1.4`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j/AmqpAppender.java
garyrussell: My initial fix was to make all the fields volatile but then copy them into the `EventSender` via a ctor, as `final` fields, to avoid accessing the volatile fields at runtime.

I thought "there has to be a better way than this". So I came up with this after a bit more thought. I checked the bytecode to make sure the compiler didn't optimize out the block with no code but I figured it wouldn't because `synchronized` has these semantics.

This technique ok here because the fields are only set up once, during log initialization.

artembilan: H-m-m... Thanks for the explanation.
Don't we have anything similar with our `EnableIntegrationTests` and `InterceptedSharedConnectionTests`?

garyrussell: I don't know :wink: 

Issues like this are tricky to find/debug.

PR #263, opened 2015-02-08
AMQP-474:Log Appenders appId Replacement Once Only

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-474

Instead of replacing the placeholder on every log event,
replace it once during initialization.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j/AmqpAppender.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j/AmqpAppender.java
@@ -30,7 +30,6 @@ import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.log4j.AppenderSkeleton;
 import org.apache.log4j.Layout;
 import org.apache.log4j.Level;
-import org.apache.log4j.MDC;
 import org.apache.log4j.PatternLayout;
 import org.apache.log4j.spi.ErrorCode;
 import org.apache.log4j.spi.LocationInfo;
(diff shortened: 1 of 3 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j/AmqpAppender.java
```diff
@@ -527,12 +528,6 @@ public void run() {
 						amqpProps.setMessageId(UUID.randomUUID().toString());
 					}
 
-					// Set applicationId, if we're using one
-					if (null != applicationId) {
-						amqpProps.setAppId(applicationId);
```
artembilan: Not sure yet that your `replace` solution is good one, but we shouldn't remove AMQP property anyway

garyrussell: doh

artembilan:
@garyrussell , here is a polishing: https://github.com/artembilan/spring-amqp/commit/1e941a8f32f20f0f6da44c3e5ba1bc36c23908e0

garyrussell:
Merged your polish; removed ctor in log4j `EventSender`; squashed,

garyrussell:
Rebased.

artembilan:
Cherry-picked to `1.4.x` as https://github.com/spring-projects/spring-amqp/commit/23bd1dff38f2036141e7d6d6934e5780e9867eaf
To `1.3.x` as https://github.com/spring-projects/spring-amqp/commit/def93af36b5c58be5d28c7aea9c5627b25db2404

PR #267, opened 2015-03-07
AMQP-484: Selectable SSL Algo

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-484

(must remain at SSLv3 when backporting to 1.4.x)

Also add overridable method for additional context customization.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
@@ -31,16 +31,16 @@ import javax.net.ssl.KeyManagerFactory;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.TrustManagerFactory;
 
-import com.rabbitmq.client.ConnectionFactory;
-import com.rabbitmq.client.ExceptionHandler;
-import com.rabbitmq.client.SaslConfig;
-import com.rabbitmq.client.SocketConfigurator;
-
 import org.springframework.beans.factory.config.AbstractFactoryBean;
 import org.springframework.core.io.Resource;
 import org.springframework.core.io.support.PathMatchingResourcePatternResolver;
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
+
+import com.rabbitmq.client.ConnectionFactory;
+import com.rabbitmq.client.ExceptionHandler;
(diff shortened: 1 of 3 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
```diff
@@ -279,10 +291,14 @@ private void setUpSSL() throws Exception {
...
 			tmf.init(tks);
 
-			SSLContext context = SSLContext.getInstance("SSLv3");
+			SSLContext context = createSSLContext();
 			context.init(kmf.getKeyManagers(), tmf.getTrustManagers(), null);
 			this.connectionFactory.useSslProtocol(context);
 		}
 	}
 
+	protected SSLContext createSSLContext() throws NoSuchAlgorithmException {
```
artembilan: Nothing to complain about.
Not sure that we must document it, but the good JavaDoc here, as well as on the class level, would be enough for end-users.

garyrussell:
Pushed

artembilan:
Merging and cherry-picking with appropriate fix for the `SSLv3`

artembilan:
Merged and cherry-picked

kalpa7:
Hi Gary,
I am currently using version 1.4.5 of spring-rabbit. I need to be able to connect my rabbitmq client with the server with TLSv1.2. I see that with this version (1.4.5) I can achieve this by configuring the sslAlgotithm as TLSv1.2. However, to get this to work, using SSL certificates becomes mandatory (sslPropertiesLocation). Otherwise, it just ends up using the default, SSLv3. I would like to know if I could avoid using certificates and still use TLSv1.2. Thanks in advance.

artembilan:
Thank you, @kalpa7, for feedback!

Please, move you question to StackOverflow. From one side GitHub PullRequests aren't intended for community support, from other side we can't help you right now, but with StackOverflow you may get a help from other community members.

Best regards,
Artem

garyrussell:
Sorry for the delay, I was traveling.

Please open a [JIRA Issue](https://jira.spring.io/browse/AMQP).

Currently, you are correct.

You can work around this by creating a subclass of `RabbitConnectionFactoryBean` and override `setUpSSL()`...

		this.connectionFactory.useSslProtocol("TLSv1.2");
	}

garyrussell:
https://jira.spring.io/browse/AMQP-500

PR #270, opened 2015-03-19
AMQP-486: Fix Version in QuickStart

garyrussell:
Enable attribute substitution in the code blocks, per http://asciidoctor.org/docs/user-manual/#applying-substitutions


artembilan:
Merged.
And I hope we don't have any more similar places in the Manual which should be addressed the same `subs` way.
I guess all those cases in the SI Manual can be repaired with that, too.
Although we can live with that now or address them on demand.

garyrussell:
I think this was the only place; actually, I had to change the docbook gradle plugin to make this work for the old .xml docs so I am pretty sure there are no more.

PR #271, opened 2015-03-27
Add "Real" Routing Connection Factory Test

garyrussell:
Create a more realistic test case; used to answer http://stackoverflow.com/questions/29133902/spring-amqp-multiple-queues-with-different-routingkey/29134502#29134502


artembilan:
Merged after rebase

PR #273, opened 2015-04-03
AMQP-467: INFO Log When Connection is Opened

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-467

Log at INFO level when a new connection is established; useful
when recovering a connection after a broker loss.


artembilan:
Nothing to complain about.
Merging...

PR #277, opened 2015-04-22
AMQP-443: Auto-Declare @RabbitListener Queues

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-443

Provide a mechanism to auto-declare/bind the queues for a `@RabbitListener` annotation,
if a `RabbitAdmin` is defined in the application context.

__review only - needs docs__

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListener.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListener.java
@@ -26,7 +26,8 @@ import org.springframework.messaging.handler.annotation.MessageMapping;
 
 /**
  * Annotation that marks a method to be the target of a Rabbit message
- * listener on the specified {@link #queues()}. The {@link #containerFactory()}
+ * listener on the specified {@link #queues()} (or {@link #bindings()}).
+ * The {@link #containerFactory()}
  * identifies the {@link org.springframework.amqp.rabbit.listener.RabbitListenerContainerFactory
  * RabbitListenerContainerFactory} to use to build the rabbit listener container. If not
  * set, a <em>default</em> container factory is assumed to be available with a bean
@@ -67,6 +68,10 @@ import org.springframework.messaging.handler.annotation.MessageMapping;
  * adding @{@link org.springframework.messaging.handler.annotation.SendTo SendTo} to the method
  * declaration.
  *
+ * <p>When {@link #bindings()} are provided, and the application context contains a
+ * {@link org.springframework.amqp.rabbit.core.RabbitAdmin},
+ * the queue, exchange and binding will be automatically declared.
+ *
  * @author Stephane Nicoll
  * @since 1.4
  * @see EnableRabbit
@@ -99,10 +104,11 @@ public @interface RabbitListener {
 	 * The queues for this listener.
 	 * The entries can be 'queue name', 'property-placeholder keys' or 'expressions'.
 	 * Expression must be resolved to the queue name or {@code Queue} object.
+	 * Mutually exclusive with {@link #bindings()}
 	 * @return the queue names or expressions (SpEL) to listen to from target
 	 * {@link org.springframework.amqp.rabbit.listener.MessageListenerContainer}.
 	 */
-	String[] queues();
+	String[] queues() default {};
 
 	/**
 	 * When {@code true}, a single consumer in the container will have exclusive use of the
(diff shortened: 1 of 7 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListener.java
```diff
@@ -130,4 +136,11 @@
 	 */
 	String admin() default "";
 
+	/**
+	 * Array of {@link QueueBinding}s providing the listener's queue names, together
+	 * with the exchange and optional binding information.
+	 * @return the bindings.
```
artembilan: `@since 1.5`

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java
```diff
@@ -149,6 +155,16 @@ public void testInvalidPojoConversion() throws InterruptedException {
 
 	public static class MyService {
 
+		@RabbitListener(bindings = {
```
artembilan: No reason to make it as an array if you use only one element.
I mean just only for this test-case

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListener.java
```diff
@@ -67,6 +68,10 @@
  * adding @{@link org.springframework.messaging.handler.annotation.SendTo SendTo} to the method
  * declaration.
  *
+ * <p>When {@link #bindings()} are provided, and the application context contains a
+ * {@link org.springframework.amqp.rabbit.core.RabbitAdmin},
+ * the queue, exchange and binding will be automatically declared.
+ *
  * @author Stephane Nicoll
```
artembilan: Add `@author Gary Russell`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -334,6 +365,66 @@ private String resolve(String value) {
...
+					exchange = new TopicExchange(exchangeName,
+							resolveExpressionAsBoolean(bindingExchange.durable()),
+							resolveExpressionAsBoolean(bindingExchange.autoDelete()));
+					actualBinding = new Binding(queueName, DestinationType.QUEUE, exchangeName, resolvedKey, null);
+				}
+				else {
+					throw new BeanInitializationException("Unexpected exchange type: " + exchangeType);
+				}
+				((ConfigurableBeanFactory) this.beanFactory).registerSingleton(exchangeName + ++this.increment, exchange);
+				((ConfigurableBeanFactory) this.beanFactory).registerSingleton(exchangeName + ++this.increment, actualBinding);
```
artembilan: Would you mind explaining the reason of this `increment` ?

garyrussell: I didn't want to inadvertently overwrite some other bean definition, perhaps the bean name should be even more distinct.

artembilan: Can't we rely just only on provided Exchange name and check if bean exists already before registering a new one ?

garyrussell: Remember, at least with the queue (via the parser), the id doesn't have to be the same as the queue name.

artembilan: Good. Let it be.
For those who want to have more control over bean names there is always an alternative with `@Bean` definitions for all this binding and declaration stuff

garyrussell: Right - this is just a convenience - for more sophisticated configuration use `@Bean`s.

artembilan:
For me everything looks good.
You can go ahead with docs.
Shouldn't we ask the reporter if this solution fine with him?
How about this:
> I would prefer to not have to create the service method, and have Spring code handle setting up an anonymous queue binding to a topic exchange given just the exchange and a set of binding keys.

Seems for me we don't cover an `anonymous queue` case here. Isn't it ?

garyrussell:
I will add code for an anonymous queue while adding docs.

We can ask the users to comment when the milestone is released.

garyrussell:
Pushed

artembilan:
Polishing for your review: https://github.com/artembilan/spring-amqp/commit/53a9c1ab2e3692d75ecb33c56081f483508950df

garyrussell:
Polish looks good - thanks!!

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/113b528616a22d61df6d1b3660bd649f4c26bfbf

PR #278, opened 2015-04-22
AMQP-490: Suppress ERROR on Normal Close

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-490

Currently, we suppress the error log if the channel is closed normally.
However, we should also suppress if the channel is closed because the connection
is closed normally.

__cherry-pick to 1.4.x__


artembilan:
OK. Nothing to complain about!

artembilan:
Merged after rebase and chery-picked

PR #279, opened 2015-04-23
AMQP-493: Add Consumer Tag Naming Strategy

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-493

Allow the user to generate custom consumer tags instead of the default server-generated tag.

__cherry-pick to 1.4.x - namespace PR to follow - 1.5 only__

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/support/ServerGeneratesConsumerTagStrategy.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/support/ServerGeneratesConsumerTagStrategy.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.amqp.support;
+
+/**
+ * The default strategy for consumer tags - causes the server to generate the tag
+ * when processing the {@code basicConsume} operation.
+ *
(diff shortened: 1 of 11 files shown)


garyrussell:
Added setter; squashed.

garyrussell:
Added `@RabbitListener` support - squashed

garyrussell:
Added namespace support - do not cherry-pick the second commit.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/SimpleRabbitListenerContainerFactory.java
```diff
@@ -187,6 +191,14 @@ public void setMissingQueuesFatal(Boolean missingQueuesFatal) {
 		this.missingQueuesFatal = missingQueuesFatal;
 	}
 
+	/**
+	 * @param consumerTagStrategy the consumerTagStrategy to set
+	 * @see SimpleMessageListenerContainer#setConsumerTagStrategy(ConsumerTagStrategy)
```
artembilan: `@since 1.4.5`.
I see the `@see`, but it really doesn't say which version this method has been introduced with.

spring-amqp/src/main/java/org/springframework/amqp/support/ServerGeneratesConsumerTagStrategy.java
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @author Gary Russell
+ * @since 1.4.5
+ *
+ */
+public class ServerGeneratesConsumerTagStrategy implements ConsumerTagStrategy {
+
+	@Override
+	public String createConsumerTag(String queue) {
+		return "";
```
artembilan: Would you mind explaining the reason of such a simple impl?
Looks like the ternary operator on "null check" with fallback to the empty string will be enough.

artembilan:
Otherwise everything looks good.
And I understand that cherry-picking must be first, because of namespace support as separate commit.

garyrussell:
I am ok with your suggestions if you want to make the changes during merge.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/9ac2afe71b7d8dd790cb6e004e0fbd5a167d1193 and first commit cherry-picked.
Note: the `ServerGeneratesConsumerTagStrategy` has been removed and replaced with ternary operator as we discussed.

PR #281, opened 2015-04-23
AMQP-450: Support Reply Routing via Exchange

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-450

Support reply routing using an exchange/routing key as well as
the simple queue name via the default exchange.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/TemplateParser.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/TemplateParser.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2010-2014 the original author or authors.
+ * Copyright 2010-2015 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
  * the License. You may obtain a copy of the License at
@@ -51,6 +51,8 @@ class TemplateParser extends AbstractSingleBeanDefinitionParser {
 
 	private static final String REPLY_QUEUE_ATTRIBUTE = "reply-queue";
 
+	private static final String REPLY_ADDRESS_ATTRIBUTE = "reply-address";
+
 	private static final String LISTENER_ELEMENT = "reply-listener";
 
 	private static final String MANDATORY_ATTRIBUTE = "mandatory";
@@ -101,7 +103,11 @@ class TemplateParser extends AbstractSingleBeanDefinitionParser {
 		NamespaceUtils.setValueIfAttributeDefined(builder, element, REPLY_TIMEOUT_ATTRIBUTE);
 		NamespaceUtils.setValueIfAttributeDefined(builder, element, ENCODING_ATTRIBUTE);
 		NamespaceUtils.setReferenceIfAttributeDefined(builder, element, MESSAGE_CONVERTER_ATTRIBUTE);
-		NamespaceUtils.setReferenceIfAttributeDefined(builder, element, REPLY_QUEUE_ATTRIBUTE);
+		String replyAddress = element.getAttribute(REPLY_ADDRESS_ATTRIBUTE);
+		if (!StringUtils.hasText(replyAddress)) {
+			NamespaceUtils.setReferenceIfAttributeDefined(builder, element, REPLY_QUEUE_ATTRIBUTE);
+		}
+		NamespaceUtils.setValueIfAttributeDefined(builder, element, REPLY_ADDRESS_ATTRIBUTE);
 		NamespaceUtils.setReferenceIfAttributeDefined(builder, element, RETURN_CALLBACK_ATTRIBUTE);
 		NamespaceUtils.setReferenceIfAttributeDefined(builder, element, CONFIRM_CALLBACK_ATTRIBUTE);
 		NamespaceUtils.setValueIfAttributeDefined(builder, element, CORRELATION_KEY);
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
@@ -161,7 +161,7 @@ public class RabbitTemplate extends RabbitAccessor
 
 	private volatile String encoding = DEFAULT_ENCODING;
 
-	private volatile Queue replyQueue;
+	private volatile String replyAddress;
 
 	private volatile ConfirmCallback confirmCallback;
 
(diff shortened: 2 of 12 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -256,11 +256,25 @@ public void setEncoding(String encoding) {
...
+	}
+
+	/**
+	 * An address for replies; if not provided, a temporary exclusive, auto-delete queue will
+	 * be used for each reply, unless RabbitMQ supports 'amq.rabbitmq.reply-to' - see
+	 * http://www.rabbitmq.com/direct-reply-to.html
+	 * <p>The address can be a simple queue name (in which case the reply will be routed via the default
+	 * exchange), or with the form {@code exchange/routingKey} to route the reply using an explicit
+	 * exchange and routing key.
+	 * @param replyQueue the replyQueue to set
```
artembilan: s/replyQueue/replyAddress
and `@since 1.5`

Can be addressed during merge if I won't find anything else

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/TemplateParser.java
```diff
@@ -51,6 +51,8 @@
 
 	private static final String REPLY_QUEUE_ATTRIBUTE = "reply-queue";
```
artembilan: Since we don't use this attribute any more, this const can be removed.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/TemplateParser.java
```diff
@@ -101,7 +103,11 @@ protected void doParse(Element element, ParserContext parserContext, BeanDefinit
 		NamespaceUtils.setValueIfAttributeDefined(builder, element, REPLY_TIMEOUT_ATTRIBUTE);
 		NamespaceUtils.setValueIfAttributeDefined(builder, element, ENCODING_ATTRIBUTE);
 		NamespaceUtils.setReferenceIfAttributeDefined(builder, element, MESSAGE_CONVERTER_ATTRIBUTE);
-		NamespaceUtils.setReferenceIfAttributeDefined(builder, element, REPLY_QUEUE_ATTRIBUTE);
+		String replyAddress = element.getAttribute(REPLY_ADDRESS_ATTRIBUTE);
+		if (!StringUtils.hasText(replyAddress)) {
+			NamespaceUtils.setReferenceIfAttributeDefined(builder, element, REPLY_QUEUE_ATTRIBUTE);
+		}
+		NamespaceUtils.setValueIfAttributeDefined(builder, element, REPLY_ADDRESS_ATTRIBUTE);
```
artembilan: Doh. Nevermind. Have mixed words

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -256,11 +256,25 @@ public void setEncoding(String encoding) {
...
 	 * be used for each reply, unless RabbitMQ supports 'amq.rabbitmq.reply-to' - see
 	 * http://www.rabbitmq.com/direct-reply-to.html
-	 *
+	 * @deprecated - use #setReplyAddress(String replyAddress)
 	 * @param replyQueue the replyQueue to set
 	 */
+	@Deprecated
 	public void setReplyQueue(Queue replyQueue) {
-		this.replyQueue = replyQueue;
+		setReplyAddress(replyQueue.getName());
```
artembilan: Well, looks like `reply-queue` and `reply-address` are mutually exclusive. And according to the `TemplateParser` logic the last one wins.

spring-rabbit/src/main/resources/org/springframework/amqp/rabbit/config/spring-rabbit-1.5.xsd
```diff
@@ -994,7 +994,23 @@
 				<xsd:annotation>
 					<xsd:documentation><![CDATA[
 	A reference to a <queue/> for replies; optional; if not supplied, methods expecting replies
-	will use a temporary, exclusive, auto-delete queue.
+	will use a temporary, exclusive, auto-delete queue or, if the RabbitMQ version is 3.4 or higher,
+	using the 'direct reply-to' mechanism. See also 'reply-address'.
```
artembilan: I think we should add `[DEPRECATED]` hint to this docs as well, since `setReplyQueue` is like that.

artembilan:
@garyrussell , how does this solution correlate with with this:
```
replyContainer.getPropertyValues().add("queues", new RuntimeBeanReference(element.getAttribute(REPLY_QUEUE_ATTRIBUTE)));
```
?
How to be if I want to use `reply-address` only?
Looks like I have to declare a separate `ListenerContainer` for the `queue` anyway.
Correct?
Shouldn't we say anything on the matter in docs ?

spring-rabbit/src/main/resources/org/springframework/amqp/rabbit/config/spring-rabbit-1.5.xsd
artembilan: OK. According to the logic in parser this suggestion isn't valid.
Nevermind

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
artembilan: My bad. Nevermind

artembilan:
Polishing for your review: https://github.com/artembilan/spring-amqp/commit/18095ffcabddff57017e6da1a1df56f04a212946

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
garyrussell: Yes, `reply-queue` is needed for the listener part.

spring-rabbit/src/main/resources/org/springframework/amqp/rabbit/config/spring-rabbit-1.5.xsd
garyrussell: Yes, I originally had this as deprecated but, of course, it's needed for the listener part.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/78e7ac94081b0257e0f648f048af91a9c059640f

PR #285, opened 2015-04-29
AMQP-444: Multi-Method @RabbitListener

garyrussell:
Enable class-level `@RabbitHandler` annotations, allowing for selection
of indidividual `@RabbitListener` methods based on `Message<?>` payload
type. Method selection must be unambiguous.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitHandler.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitHandler.java
@@ -26,13 +26,13 @@ import org.springframework.messaging.handler.annotation.MessageMapping;
 
 /**
  * Annotation that marks a method to be the target of a Rabbit message
- * listener within a class that is annotated with {@code @RabbitListener}.
+ * listener within a class that is annotated with {@link RabbitListener}.
  *
  * <p>See the {@link RabbitListener} for information about permitted method signatures
  * and available parameters.
  * <p><b>It is important to understand that when a message arrives, the method selection
  * depends on the payload type. The type is matched with a single non-annotated parameter,
- * or one that is annotated with {@code Payload}.
+ * or one that is annotated with {@code @Payload}.
  * There must be no ambiguity - the system
  * must be able to select exactly one method based on the payload type.</b>
  *
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitHandler.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitHandler.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.amqp.rabbit.annotation;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.springframework.messaging.handler.annotation.MessageMapping;
+
+/**
+ * Annotation that marks a method to be the target of a Rabbit message
+ * listener within a class that is annotated with {@code @RabbitListener}.
+ *
+ * <p>See the {@link RabbitListener} for information about permitted method signatures
+ * and available parameters.
+ * <p><b>It is important to understand that when a message arrives, the method selection
+ * depends on the payload type. The type is matched with a single non-annotated parameter,
+ * or one that is annotated with {@code Payload}.
+ * There must be no ambiguity - the system
+ * must be able to select exactly one method based on the payload type.</b>
+ *
+ * @author Gary Russell
+ * @since 1.5
+ * @see EnableRabbit
+ * @see RabbitListener
+ * @see RabbitListenerAnnotationBeanPostProcessor
+ */
+@Target({ ElementType.METHOD, ElementType.ANNOTATION_TYPE })
+@Retention(RetentionPolicy.RUNTIME)
+@MessageMapping
+@Documented
+public @interface RabbitHandler {
+
+}
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
@@ -34,6 +34,7 @@ import org.springframework.amqp.core.TopicExchange;
 import org.springframework.amqp.rabbit.config.RabbitListenerConfigUtils;
 import org.springframework.amqp.rabbit.core.RabbitAdmin;
 import org.springframework.amqp.rabbit.listener.MethodRabbitListenerEndpoint;
+import org.springframework.amqp.rabbit.listener.MultiMethodRabbitListenerEndpoint;
 import org.springframework.amqp.rabbit.listener.RabbitListenerContainerFactory;
 import org.springframework.amqp.rabbit.listener.RabbitListenerEndpointRegistrar;
 import org.springframework.amqp.rabbit.listener.RabbitListenerEndpointRegistry;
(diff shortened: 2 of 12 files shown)


spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapterTests.java
```diff
@@ -119,7 +119,7 @@ protected MessagingMessageListenerAdapter getSimpleInstance(String methodName, C
 
 	protected MessagingMessageListenerAdapter createInstance(Method m) {
 		MessagingMessageListenerAdapter adapter = new MessagingMessageListenerAdapter();
-		adapter.setHandlerMethod(factory.createInvocableHandlerMethod(sample, m));
+		adapter.setHandlerMethod(new HandlerAdapter(factory.createInvocableHandlerMethod(sample, m)));
```
artembilan: Need the Copyright update

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java
```diff
@@ -306,6 +320,56 @@ public RabbitAdmin rabbitAdmin(ConnectionFactory connectionFactory) {
...
+
+	}
+
+	@RabbitListener(bindings = @QueueBinding
+			(value = @Queue,
+			exchange = @Exchange(value = "multi.exch", autoDelete = "true"),
+			key = "multi.rk"))
+	static class MultiListenerBean {
+
+		@RabbitListener
```
artembilan: Does this mean that only methods with `@RabbitListener` are candidates for multi-method listener ?

garyrussell: Yes, but see new commit - changed to `@RabbitHandler` per discussion with @snicoll

garyrussell:
Squashed, pushed, based on offline @snicoll comments.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitHandler.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Annotation that marks a method to be the target of a Rabbit message
+ * listener within a class that is annotated with {@code @RabbitListener}.
+ *
+ * <p>See the {@link RabbitListener} for information about permitted method signatures
+ * and available parameters.
+ * <p><b>It is important to understand that when a message arrives, the method selection
+ * depends on the payload type. The type is matched with a single non-annotated parameter,
+ * or one that is annotated with {@code Payload}.
```
artembilan: `annotated with  {@code @Payload}`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitHandler.java
```diff
@@ -0,0 +1,51 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.springframework.messaging.handler.annotation.MessageMapping;
+
+/**
+ * Annotation that marks a method to be the target of a Rabbit message
+ * listener within a class that is annotated with {@code @RabbitListener}.
```
artembilan: s/`@code`/`@link`
You already use that below.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -213,21 +214,50 @@ public Object postProcessBeforeInitialization(Object bean, String beanName) thro
...
 			@Override
 			public void doWith(Method method) throws IllegalArgumentException, IllegalAccessException {
 				RabbitListener rabbitListener = AnnotationUtils.getAnnotation(method, RabbitListener.class);
 				if (rabbitListener != null) {
-					processAmqpListener(rabbitListener, method, bean);
+					processAmqpListener(rabbitListener, method, bean, beanName);
+				}
+				if (classLevelListener != null) {
+					RabbitHandler rabbitHandler = AnnotationUtils.getAnnotation(method, RabbitHandler.class);
+					if (rabbitHandler != null) {
```
artembilan: You have to convince me that we really need this `@RabbitHandler` for the method level.
Spring Integration doesn't require that:
```
<service-activator ref="myService"/>
```

CC @snicoll

snicoll: What the behaviour might be? Detect all public methods in that class?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitHandler.java
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * @author Gary Russell
+ * @since 1.5
+ * @see EnableRabbit
+ * @see RabbitListener
+ * @see RabbitListenerAnnotationBeanPostProcessor
+ */
+@Target({ ElementType.METHOD, ElementType.ANNOTATION_TYPE })
+@Retention(RetentionPolicy.RUNTIME)
+@MessageMapping
```
artembilan: I may miss something, but I really don't see that we rely on the `@MessageMapping` logic from Spring Messaging.
Where do we use that?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
garyrussell: This is not Spring Integration - I think it's important to be able to specify which methods are candidates - to be honest, I really don't like it that SI considers all public methods candidates - it can causes problems when unintended methods are considered.

Further, as I discussed with @snicoll, in future, we might add more attributes to `@RabbitHandler` to assist with method matching (e.g. add a SpelExpression `matchFor = "headers['foo'] == 'bar'"` which could allow enhancement to matching instead of just matching on payload type).

artembilan: Right, excluding `Object` methods and setters and getters.
Although `@RabbitHandler` may have sense for priority...

artembilan: OK, @garyrussell , that makes sense.
Agreed.
So, this question is fixed.

src/reference/asciidoc/amqp.adoc
```diff
@@ -1272,6 +1274,41 @@ Also `@SendTo` can be used without a `value` attribute.
...
+====== Multi-Method Listeners
+
+Starting with _version 1.5_, the `@RabbitListener` annotation can now be specified at the class level.
+Together with the new `@RabbitHandler` annotation, this allows a single listener to invoke different methods, based on the payload type of the incoming message.
+This is best described using an example:
+
+[source, java]
+----
+@RabbitListener(queues = "someQueue"
+public class MultiListenerBean {
```
artembilan: Need to close the annotation with `)`

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java
```diff
@@ -306,6 +320,56 @@ public RabbitAdmin rabbitAdmin(ConnectionFactory connectionFactory) {
...
 
+		@Bean
+		public MultiListenerBean multiListener() {
+			return new MultiListenerBean();
+		}
+
+	}
+
+	@RabbitListener(bindings = @QueueBinding
+			(value = @Queue,
```
artembilan: Having this I'd say that `@QueueBinding` can be fixed like:
```
Queue value() default @Queue;
```
Which means that we really use an `AnonymousQueue`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -59,7 +59,7 @@
 	 * processing an incoming {@link org.springframework.amqp.core.Message}.
 	 * @param handlerMethod {@link InvocableHandlerMethod} instance.
 	 */
-	public void setHandlerMethod(InvocableHandlerMethod handlerMethod) {
+	public void setHandlerMethod(HandlerAdapter handlerMethod) {
```
artembilan: The JavaDocs for this method must be fixed.

artembilan: As well as the class-level JavaDocs

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java
garyrussell: True (default). I didn't think it was worth complicating the BPP with `AnonymousQueue` since it's just a subclass with specific settings - handled in the BPP.

artembilan: Yes, agreed.
I see your logic in BPP:
```
queueName = UUID.randomUUID().toString();
....
Queue queue = new Queue(queueName,
						resolveExpressionAsBoolean(bindingQueue.durable()),
						exclusive,
						autoDelete);
```
That's fine.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitHandler.java
artembilan: Just commented out this `@MessageMapping` as well as on the `@RabbitListener` and all our tests pass.
Have I missed anything?
Maybe to handle STOMP with the same method it would be better to have an additional `@MessageMapping` on the method with appropriate destination?

From other side we can even do something like this:
````java
@MessageMapping("/foo")
@RabbitListener(queues = "foo")
@JmsListener(destination = "foo")
@RequestMapping("/foo")
public void handleMessage(Object payload) {}
````

artembilan:
Well, that's all from my side.
I'm ready to merge after the fixes for those simple concerns

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitHandler.java
garyrussell: It does only seem to be used by Stomp - but I'll have to defer to @snicoll to answer this.

snicoll: The `MessageMapping` is pure meta-data at this point. There is no specific reason why it's on `JmsListener` and `RabbitListener` except to indicate that it supports `MessageMaping`. 

This came from the initial Stomp support indeed and maybe I made a mistake of adding it as it's not used at all for the moment indeed.

artembilan: Well, for the robust complex solution, like my "mega-annotated-method" above, we **must** remove it from our custom annotations.

Not sure how STOMP will react if it finds several `@MessageMapping` on the same method...

garyrussell: It won't look for meta-annotations if the method is directly annotated - see `AnnotationUtils.findAnnotation()`.

garyrussell: I am going to leave it for now; we can remove before GA if there's a compelling reason.

garyrussell:
Pushed

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java
artembilan: Polishing looks good, but you haven't addressed this one.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/1b0edac0da1ef15f9ed18d3cfe534a237d54d229

PR #288, opened 2015-05-06
AMQP-481: Register `@RabbitListener` for `@Lazy`

artembilan:
JIRA: https://jira.spring.io/browse/AMQP-481

Previously the `@Lazy` components haven't been registered with `MessageListenerContainer`,
because `RabbitListenerEndpointRegistrar` has done that from its `afterPropertiesSet()`.

Add logic to `registerListenerContainer` on demand from the `RabbitListenerEndpointRegistrar#registerEndpoint`

**Cherry-pick to 1.4.x**

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/RabbitListenerEndpointRegistry.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/RabbitListenerEndpointRegistry.java
@@ -18,8 +18,8 @@ package org.springframework.amqp.rabbit.listener;
 
 import java.util.Collection;
 import java.util.Collections;
-import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.commons.logging.Log;
@@ -46,6 +46,7 @@ import org.springframework.util.Assert;
  *
  * @author Stephane Nicoll
  * @author Juergen Hoeller
+ * @author Artem Bilan
  * @since 1.4
  * @see RabbitListenerEndpoint
  * @see MessageListenerContainer
@@ -56,7 +57,7 @@ public class RabbitListenerEndpointRegistry implements DisposableBean, SmartLife
 	protected final Log logger = LogFactory.getLog(getClass());
 
 	private final Map<String, MessageListenerContainer> listenerContainers =
-			new LinkedHashMap<String, MessageListenerContainer>();
+			new ConcurrentHashMap<String, MessageListenerContainer>();
 
 	private int phase = Integer.MAX_VALUE;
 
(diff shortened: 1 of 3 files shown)


artembilan:
@snicoll , would you mind sharing your wisdom about `this.mutex = cbf.getSingletonMutex();` concern?
Our assumption is around the several `RabbitListenerAnnotationBeanPostProcessor` instances in the ctx. We don't see issues with an internal `mutex` for the `endpointDescriptors`...

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/RabbitListenerEndpointRegistry.java
```diff
@@ -80,27 +81,46 @@ public MessageListenerContainer getListenerContainer(String id) {
 		return Collections.unmodifiableCollection(this.listenerContainers.values());
 	}
 
+	/**
+	 * Create a message listener container for the given {@link RabbitListenerEndpoint}.
+	 * <p>This create the necessary infrastructure to honor that endpoint
+	 * with regards to its configuration.
+	 * @param endpoint the endpoint to add
+	 * @param factory the listener factory to use
```
garyrussell: missing `@param`

garyrussell:
Merged as-is for now - will revisit when @snicoll returns - in any case, we can't incorporate the `getSingletonMutex()` until an SF milestone or release candidate is available.

snicoll:
I had the same discussion with @jhoeller and I can't remember why he asked me to use this (certainly to avoid some subtle potential race conditions). In any case, I'd advise you to use the same mechanism as it's no incident if it's implemented that way.

artembilan:
Thank you, @snicoll, for feed back!

Unfortunately we can't use `cbf.getSingletonMutex()` until upgrade to SF 4.2. From one side Spring AMQP isn't so aggressive in upgrading, as Spring Integration, because there is no benefits from that like support for new features. From other side it would be good to have that `getSingletonMutex()` as backported to SF `4.1.x`: it's pretty simple change.

Although... I see this fix independent of version:
````java
public void setBeanFactory(BeanFactory beanFactory) {
	this.beanFactory = beanFactory;
	if (beanFactory instanceof DefaultSingletonBeanRegistry) {
        DefaultSingletonBeanRegistry cbf = (DefaultSingletonBeanRegistry) beanFactory;
		this.mutex = cbf.getSingletonMutex();
	}
}
````

Anyway, it would be great to know the cause of that kind of race conditions, which should force us to protect our code with that mutex from BF.

CC: @jhoeller

garyrussell:
> From one side Spring AMQP isn't so aggressive in upgrading,...

It's not really that - I have no issues with making Spring AMQP 1.5 dependent on SF 4.2, it's just that we need to release a milestone in the next couple of days and there is no SF 4.2 milestone yet; we can't release our milestone based on a snapshot.

PR #290, opened 2015-05-06
AMQP-427: Add `BackOff` for listener container

artembilan:
JIRA: https://jira.spring.io/browse/AMQP-427

To have more control over `recoveryInterval` add `BackOff` injection support
 for the `SimpleMessageListenerContainer`.
When the `backOffExecution` reaches `BackOffExecution.STOP`, stop the container for further recovery to prevent infinity logs pollution.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
@@ -69,6 +69,9 @@ import org.springframework.transaction.support.TransactionCallback;
 import org.springframework.transaction.support.TransactionTemplate;
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
+import org.springframework.util.backoff.BackOff;
+import org.springframework.util.backoff.BackOffExecution;
+import org.springframework.util.backoff.FixedBackOff;
 
 import com.rabbitmq.client.Channel;
 import com.rabbitmq.client.ShutdownSignalException;
@@ -96,11 +99,6 @@ public class SimpleMessageListenerContainer extends AbstractMessageListenerConta
 	public static final int DEFAULT_PREFETCH_COUNT = 1;
 
 	public static final long DEFAULT_SHUTDOWN_TIMEOUT = 5000;
-
-	/**
-	 * The default recovery interval: 5000 ms = 5 seconds.
-	 */
-	public static final long DEFAULT_RECOVERY_INTERVAL = 5000;
 
 	private volatile int prefetchCount = DEFAULT_PREFETCH_COUNT;
 
@@ -130,7 +128,7 @@ public class SimpleMessageListenerContainer extends AbstractMessageListenerConta
 
 	private volatile long shutdownTimeout = DEFAULT_SHUTDOWN_TIMEOUT;
 
-	private long recoveryInterval = DEFAULT_RECOVERY_INTERVAL;
+	private BackOff recoveryBackOff = new FixedBackOff();
 
 	// Map entry value, when false, signals the consumer to terminate
 	private Map<BlockingQueueConsumer, Boolean> consumers;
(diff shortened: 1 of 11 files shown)


artembilan:
@garyrussell , @snicoll , let me know you position about the solution and I'll go ahead with docs.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -214,9 +212,21 @@ public void setAdviceChain(Advice[] adviceChain) {
 	 * @param recoveryInterval The recovery interval.
 	 */
 	public void setRecoveryInterval(long recoveryInterval) {
-		this.recoveryInterval = recoveryInterval;
+		this.recoveryBackOff = new FixedBackOff(recoveryInterval, FixedBackOff.UNLIMITED_ATTEMPTS);
+	}
+
+	/**
+	 * Specify the {@link BackOff} for interval between recovery attempts.
+	 * The default is 5000 ms, that is, 5 seconds.
```
garyrussell: Since that depends on the default `FixedBackOff`, we probably shouldn't say 5 seconds here (in case it changes in future).

Perhaps we should reinstate the constant and default to `new FixedBackOff(DEFAULT_RECOVERY_INTERVAL, FixedBackOff.UNLIMITED_ATTEMPTS)` so the defaults here are independent of the defaults in `FixedBackOff`.

artembilan: OK. Agreed. It would be more friendly and independent of the target implementation.
Although I guess @snicoll got those `5 sec` for `FixedBackOff` impl from the JMS `DefaultMessageListenerContainer` :smile: .
And moreover I see that JMS container continues to use the `DEFAULT_RECOVERY_INTERVAL` const.
Will fix together with Docs.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -1266,13 +1279,18 @@ protected void invokeListener(Channel channel, Message message) throws Exception
...
 	 * chance to recover from consumer startup failure, e.g. if the broker is down.
-	 * @param t the exception that stopped the startup
+	 * @param backOffExecution the BackOffExecution to get the {@code recoveryInterval}
 	 * @throws Exception if the shared connection still can't be established
 	 */
-	protected void handleStartupFailure(Throwable t) throws Exception {
+	protected void handleStartupFailure(BackOffExecution backOffExecution) throws Exception {
+		long recoveryInterval = backOffExecution.nextBackOff();
+		if (BackOffExecution.STOP == recoveryInterval) {
+			stop();
```
garyrussell: `logger.warn("stopping container - restart recovery attempts exhausted")`

spring-rabbit/src/main/resources/org/springframework/amqp/rabbit/config/spring-rabbit-1.5.xsd
```diff
@@ -697,9 +697,25 @@
...
 	See also 'missing-queues-fatal'. Fatal errors include authentication problems.
+	Mutually exclusive with 'recovery-back-off'.
 				]]></xsd:documentation>
 			</xsd:annotation>
 		</xsd:attribute>
+		<xsd:attribute name="recovery-back-off" type="xsd:string">
+			<xsd:annotation>
+				<xsd:documentation><![CDATA[
+	The BackOff bean reference for recovery interval between attempts to start a consumer
+	if it fails to start with a non-fatal error. Default 5000.
```
garyrussell: Default FixedBackOff with unlimited retries every 5 seconds.

garyrussell:
That's all; LGTM

garyrussell:
One more thing; we need to add some synchronization to avoid multiple

`logger.debug("Shutting down Rabbit listener container");`

and more logs.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerTests.java
```diff
@@ -564,6 +567,24 @@ public Object answer(InvocationOnMock invocation) throws Throwable {
...
 	}
 
+	@Test
+	public void testContainerNotRecoveredAfterExhaustingRecoveryBackOff() throws Exception {
+		SimpleMessageListenerContainer container = new SimpleMessageListenerContainer(mock(ConnectionFactory.class));
+		container.setQueueNames("foo");
+		container.setRecoveryBackOff(new FixedBackOff(100, 3));
+		container.setConcurrentConsumers(10);
+		container.afterPropertiesSet();
+		container.start();
```
garyrussell: This needs some polishing to avoid

> 2015-05-07 11:20:11,855 WARN SimpleAsyncTaskExecutor-1 [org.springframework.amqp.rabbit.listener.SimpleMessageListenerContainer] - <Consumer raised exception, processing can restart if the connection factory supports it>
java.lang.IllegalArgumentException: Connection must not be null

artembilan: Yeah... It was there with `Mocks` from the beginning, but after that I found that it works as is with such a bad config for `ConnectionFactory` :smile: 
Will fix

artembilan:
Pushed

garyrussell:
Merged as 1954dfa0771f78e3f867a2754da328838b8827cb

PR #292, opened 2015-05-08
AMQP-496: Prevent AMQP threads dead lock

artembilan:
JIRA: https://jira.spring.io/browse/AMQP-496

The long living message listener performs `channel.basicAck` (`channel.txCommit()`)
to late after stopping container and connectionFactory.
It may cause unexpected dead locks in the underlying AMQP thread.

Since this `late commit` does not make any sense already,
because `consumer` has been canceled already by the `container.stop()`,
just add a `if (this.cancelled.get()) {` check and return immediately from `BlockingQueueConsumer#commitIfNecessary`
to prevent undesired connections to the Broker.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
@@ -36,7 +36,12 @@ import org.springframework.amqp.AmqpException;
 import org.springframework.amqp.AmqpTimeoutException;
 import org.springframework.amqp.rabbit.support.PublisherCallbackChannel;
 import org.springframework.amqp.rabbit.support.PublisherCallbackChannelImpl;
+import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.InitializingBean;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
+import org.springframework.context.ApplicationListener;
+import org.springframework.context.event.ContextClosedEvent;
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 
@@ -72,7 +77,10 @@ import com.rabbitmq.client.ShutdownSignalException;
  * @author Gary Russell
  * @author Artem Bilan
  */
-public class CachingConnectionFactory extends AbstractConnectionFactory implements InitializingBean, ShutdownListener {
+public class CachingConnectionFactory extends AbstractConnectionFactory
+		implements InitializingBean, ShutdownListener, ApplicationContextAware, ApplicationListener<ContextClosedEvent> {
+
+	private ApplicationContext applicationContext;
 
 	public enum CacheMode {
 		/**
@@ -118,6 +126,8 @@ public class CachingConnectionFactory extends AbstractConnectionFactory implemen
 	private volatile boolean publisherReturns;
 
 	private volatile boolean initialized;
+
+	private volatile boolean stopped;
 
 	/** Synchronization monitor for the shared Connection */
 	private final Object connectionMonitor = new Object();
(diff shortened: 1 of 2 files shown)


spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/MessageListenerContainerLifecycleIntegrationTests.java
```diff
@@ -115,7 +119,7 @@ public int value() {
 		}
 	}
 
-	@Rule
+//	@Rule
```
garyrussell: Intentional?

artembilan: Oh! No! Just for testing from IDE.

garyrussell: Hmmm - I am seeing many other test failures in this class now.

garyrussell: We're now not acking processed messages during shutdown - we need to take this action only after we have given up and forced the close.

artembilan: Got it. Thank you!

garyrussell: I think we need to treat a container shutdown different to a context destroy.

It seems to me that the issue (creating a new connection) is only a problem if the connection factory has been destroyed by a context `close()`. Otherwise, it doesn't hurt to try to send a late ack.

Maybe we need to make the CCF look at the context state and only re-create the connection if the context is active.

artembilan: Actually `CCF` can have an internal `active` flag. Which becomes `false` on its `destroy()`

garyrussell: Yes, but it's only reset for a short time.

artembilan:
Pushed.
Since it is fully different solution, I've just squeshed the previous commit.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
```diff
@@ -438,6 +460,7 @@ private Channel doCreateBareChannel(ChannelCachingConnectionProxy connection, bo
 
 	@Override
 	public final Connection createConnection() throws AmqpException {
+		Assert.state(!this.stopped, "The ConnectionFactory isn't active any more for obtaining connection.");
```
garyrussell: `"The ApplicationContext is closed and the ConnectionFactory can no longer create connections."`

I was going to merge with this minor change, but it appears we might have introduced some instability...

`Class org.springframework.amqp.rabbit.listener.SimpleMessageListenerContainerIntegrationTests`

`testListenerSunnyDay[7]`

>java.lang.AssertionError: Timed out waiting for message
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.springframework.amqp.rabbit.listener.SimpleMessageListenerContainerIntegrationTests.doSunnyDayTest(SimpleMessageListenerContainerIntegrationTests.java:215)
	at org.springframework.amqp.rabbit.listener.SimpleMessageListenerContainerIntegrationTests.testListenerSunnyDay(SimpleMessageListenerContainerIntegrationTests.java:182)

garyrussell:
Merged as 6f932979c8a3f94f7b7ce78b976d41c8de9d4fb9

PR #293, opened 2015-05-12
AMQP-498: Adjust Consumers With ChannelTransacted

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-498

In earlier versions, the while loop to drain the queue took an early exit
when the channel was transacted. (This was actually incorrect since message
delivery is not transactional, only acks).

AMQP-388 introduced a new OR condition on the loop: `consumer.hasDelivery()`.

This effectively made the `continuable` boolean in the OR condition irrelevant.
This boolean was true if messages were received and the channel was not transactional.

Another side effect was the `continuable` boolean was incorrectly used to adjust the consumers
if the workload demanded. Instead, that decision should have depended just on whether
messages were received; whether the channel is transacted is irrelevent.

Remove the `continuable` boolean; fix the while loop and use a new boolean `receivedOk` which
indicates whether messages were received and so whether we should consider increasing the
consumers.

Add a test to verify consumers are adjusted for transactional channels.

### DIFF
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerLongTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerLongTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2014 the original author or authors.
+ * Copyright 2013-2015 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
  * the License. You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerLongTests.java
```diff
@@ -48,34 +48,49 @@
...
+
+	@Test
+	public void testChangeConsumerCountTransacted() throws Exception {
+		testChangeConsumerCountGuts(true);
+	}
+
+	private void testChangeConsumerCountGuts(boolean transacted) throws Exception {
 		final SingleConnectionFactory singleConnectionFactory = new SingleConnectionFactory("localhost");
 		SimpleMessageListenerContainer container = new SimpleMessageListenerContainer(singleConnectionFactory);
-		container.setMessageListener(new MessageListenerAdapter(this));
```
garyrussell: I suggest reviewing this with `?w=1`

PR #294, opened 2015-05-29
AMQP-500: Configurable SSL Algo, No Properties

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-500

Enable algorithm selection when not using certificates.

__cherry-pick to 1.4.x__

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014 the original author or authors.
+ * Copyright 2014-2015 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -72,6 +72,8 @@ public class RabbitConnectionFactoryBean extends AbstractFactoryBean<ConnectionF
 
 	private volatile String sslAlgorithm = TLS_V1_1;
 
+	private volatile boolean sslAlgorithmSet;
+
 	/**
 	 * Whether or not the factory should be configured to use SSL.
 	 * @param useSSL true to use SSL.
@@ -94,6 +96,7 @@ public class RabbitConnectionFactoryBean extends AbstractFactoryBean<ConnectionF
 	 */
 	public void setSslAlgorithm(String sslAlgorithm) {
 		this.sslAlgorithm = sslAlgorithm;
+		this.sslAlgorithmSet = true;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/SSLConnectionTests.java
```diff
@@ -44,4 +49,27 @@ public void test() throws Exception {
...
+
+	@Test
+	public void testNoAlgNoProps() throws Exception {
+		RabbitConnectionFactoryBean fb = new RabbitConnectionFactoryBean();
+		ConnectionFactory rabbitCf = spy(TestUtils.getPropertyValue(fb, "connectionFactory", ConnectionFactory.class));
+		new DirectFieldAccessor(fb).setPropertyValue("connectionFactory", rabbitCf);
+		fb.setUseSSL(true);
+		fb.afterPropertiesSet();
+		fb.getObject();
+		verify(rabbitCf).useSslProtocol();
```
artembilan: Nothing to complain about, but before I merge it, please, let me know if we should change our default algorithm from `TLSv1.1` to `TLSv1` as it is in the `com.rabbitmq.client.ConnectionFactory`.

Or have I missed anything and it was really intentional to have it like this?

garyrussell: IIRC, I made it 1.1 because of the [beast attack](https://www.rabbitmq.com/ssl.html).

artembilan: OK. I understand that. And it is good. So, see my comment on code changes.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
```diff
@@ -297,7 +300,12 @@ protected ConnectionFactory createInstance() throws Exception {
 	 */
 	protected void setUpSSL() throws Exception {
 		if (this.sslPropertiesLocation == null) {
-			this.connectionFactory.useSslProtocol();
+			if (this.sslAlgorithmSet) {
+				this.connectionFactory.useSslProtocol(this.sslAlgorithm);
+			}
+			else {
+				this.connectionFactory.useSslProtocol();
```
artembilan: How about to always propagate our `this.sslAlgorithm`, even if default, to override that `TLSv1` in the `connectionFactory`? At least for our `1.5`. And notify RabbitMQ team to change their default value.

CC @michaelklishin

garyrussell: I thought it better to provide an option to use the default supplied by the rabbit client (in case it changes in future).

i.e. if you don't specify it in Spring configuration, you get the rabbitmq client default.

I don't want to get into the business of synchronizing our default with the rabbit client's.

michaelklishin: 👍

artembilan:
Thank you, guys!
Merging and cherry-picking...

PR #296, opened 2015-06-19
AMQP-502: Fix Fanout @Binding Declaration

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-502

Queues are bound to fanout exchanges with no routing key.
The `RabbitListenerAnnotationBeanPostProcessor` incorrectly set the key to `null` instead of `""`.

Add a test for a fanout exchange.


artembilan:
Merged after rebase.

PR #297, opened 2015-06-23
AMQP-503: Add Blocking Receive to RabbitTemplate

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-503

Now that the `QueueingConsumer` is no longer deprecated, add
`receiveTimeout` to the `RabbitTemplate` to allow for blocking `receive` operations.


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -668,7 +685,22 @@ public Message receive() throws AmqpException {
...
+		if (this.receiveTimeout == 0) {
+			return doReceiveNoWait(queueName);
+		}
+		else {
+			return doReceive(queueName);
+		}
+	}
+
+	/**
+	 * Non-blocking receive with timeout.
```
artembilan: with**out**

src/reference/asciidoc/amqp.adoc
```diff
@@ -764,12 +764,22 @@ The simpler option is to poll for a single `Message` at a time with a polling me
...
 We will look at an example of each approach in the next two sub-sections.
 
+[[polling-consumer]]
 ===== Polling Consumer
 
 The `AmqpTemplate` itself can be used for polled Message reception.
-If no message is available, `null` is returned immediately; there is no blocking.
+By default, if no message is available, `null` is returned immediately; there is no blocking.
+Starting with _version 1.5_, you can now set a `receiveTimeout`, in milliseconds, and the receive methods will block for
+up to that long, waiting for a message. A value of `-1` means block indefinitely (or at least until the connection to the
```
artembilan: A value **less than zero** ?

src/reference/asciidoc/whats-new.adoc
```diff
@@ -86,6 +86,11 @@ Previously, the consumer tags for asynchronous consumers were generated by the b
 With this release, it is now possible to supply a naming strategy to the listener container.
 See <<consumerTags>>.
 
+===== RabbitTemplate Blocking Receive Methods
+
+The `RabbitTemplate` now supports blocking in `receive()` and `convertAndReceive` methods.
```
artembilan: Or brackets for both cases, or no one. Fortunately you have "methods" word, so no need to worry about brackets.
Nit-pick, of course. But it would be better always be on our toes do not bring more bad errors :smile:

garyrussell:
Pushed

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -694,6 +726,71 @@ else if (isChannelTransacted()) {
...
+					throw new AmqpException("Blocking receive, consumer failed to consume");
+				}
+				Delivery delivery;
+				if (receiveTimeout < 0) {
+					delivery = consumer.nextDelivery();
+				}
+				else {
+					delivery = consumer.nextDelivery(receiveTimeout);
+				}
+				channel.basicCancel(consumer.getConsumerTag());
```
artembilan: `basicCancel` to prevent the consumption for the next message ? 

From other side can't we use shared `QueueingConsumer` instance since this operation is blocked?
Or at least lazy-load (TTL removal) per `queueName`, but blocked anyway...

garyrussell: Yes, cancel with prefetch = 1 means we'll only fetch 1 message.

I thought about a permanent QC, maybe just for the configured (default) queue (bear in mind people can consume from arbitrary queues), but then we need logic to deal with canceled consumers, etc, etc.

I figured we can keep it simple for now and enhance later if the demand arises.

artembilan: :+1:

artembilan:
So, merging as is after rebase

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/c1d0cad

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
Xokker: Why should we set receive timeout globally? Of course, it is handy to have default `receiveTimeout`, but it would be useful to have method `receive(String queueName, long timeout, TimeUnit unit)` (or `receive(queueName, timeoutMillis)`), isn't it?

garyrussell: Feel free to open a [new feature request](https://jira.spring.io/browse/AMQP).

PR #299, opened 2015-06-23
AMQP-499: Improve Channel Close Logging

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-499

Reduce to INFO for exclusive queue consume errors.

Add strategy to customize logging.


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
```diff
@@ -882,4 +892,48 @@ public String toString() {
...
+
+	}
+
+	/**
+	 * Default implementation of {@link ConditionalExceptionLogger} for logging channel
+	 * close exceptions.
+	 * @since 1.5
+	 *
+	 */
+	private class DefaultChannelCloseLogger implements ConditionalExceptionLogger {
```
artembilan: Looks like both these new classes are candidates to the top level - `org.springframework.amqp.utils`.
Otherwise this class must `static`.

garyrussell: I thought about the raising the interface to the top level but this probably belongs here but, yes, `static`.

garyrussell: Thinking some more, I think perhaps the message should be a parameter too ("Channel Shutdown: ").

artembilan: Yes, that's good catch. Having this particular context it our responsibility to emit the proper message on the matter, but that is up to end-user how to log in his implementation or even do something else.
If he isn't interested in the our message, he always can ignore it.

garyrussell:
Pushed

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
```diff
@@ -253,6 +258,19 @@ public void setChannelCheckoutTimeout(long channelCheckoutTimeout) {
 		this.channelCheckoutTimeout = channelCheckoutTimeout;
 	}
 
+	/**
+	 * Set the strategy for logging close exceptions; by default, if a channgel is closed due to a failed
```
artembilan: Typo: chann~~g~~el

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
artembilan: Should be static anyway.
All those minor issues will be fixed during merge.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/9bc8adb

PR #301, opened 2015-06-25
Add setUri() method to ConnectionFactory classes

Zteve:
Previously the setUri() method on rabbitmq.ConnectionFactory was not exposed
to the CachingConnectionFactory nor the SingleConnectionFactory. This change
adds this method to the AbstractConnectionFactory, which is then inherited by
the two implementation classes.

A constructor taking the URI is added to CachingConnectionFactory and also to
the SingleConnectionFactory to short-circuit the construction even further.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
@@ -14,7 +14,10 @@ package org.springframework.amqp.rabbit.connection;
 
 import java.io.IOException;
 import java.net.InetAddress;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.UnknownHostException;
+import java.security.GeneralSecurityException;
 import java.util.List;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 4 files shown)


garyrussell:
Hi Steve, thanks for the contribution!

Please note that we track all changes via JIRA; I have [opened an issue for this change](https://jira.spring.io/browse/AMQP-505).

In future, please create an issue and refer to it in the commit comment.

We also have a few code style requirements and need an entry in the "what's new" section of the docs.

I will take care of polishing this up.

Thanks again.

Gary

garyrussell:
Merged as 6682f41dc17ea628808976f3563e0c13e4ce247b with [minor polish](https://github.com/garyrussell/spring-amqp/commit/ed6bd99a8a44a8af7da450349e0eb749ae37d1dd).

Thanks again.

Zteve:
*In future, please create an issue and refer to it in the commit comment.
We also have a few code style requirements and need an entry in the "what's new" section of the docs.*

Thanks Gary; I'll try to catch this in future. We (Paul Harris and I) had a Pivotal Project story id we could have referenced, too.
I see that some of the polishing involved changing things I didn't touch — I'm happy to take some of the credit for these improvements :-)

garyrussell:
Thanks; the main reason for tracking with JIRA is so we can summarize the contents by pointing to the JIRA changelog.

np; the original author always gets credit for any minor tweaks we do :smile:

garyrussell:
By the way, 1.5 will be released in Aug or early Sep.

In the meantime, the work around is to configure the rabbit connection factory directly and pass it into the `CachingConnectionFactory` via that constructor.

If you need it sooner, we can back-port it and release 1.4.6

Zteve:
Thanks Gary; no rush -- we already use the work-around and raised the PR to
pick up later.

Steve Powell          [M: +*44*-*7815*-*838*-*558* H:+*44*-*1962*-*775*-
*598*]
*  The value of advice may go down as well as up; your information may
vary; subject to terms and conditions.*
pivotal.io <http://www.pivotal.io>|SpringSource
<http://www.springsource.org>|RabbitMQ <http://www.rabbitmq.com>

PR #302, opened 2015-06-26
AMQP-506: Publisher Confirms Concurrency

garyrussell:
JIRA:https://jira.spring.io/browse/AMQP-506

`RabbitTemplate` provides a method `getUnconfirmed` to get aged
non-confirmed correlation data.

This was not thread-safe.

While the `pendingConfirms` objects in `RabbitTemplate` and `PublisherCallbackChannelImpl`
are concurrent hash maps, they are maps of maps with the inner maps being
`Collections.synchronizedSortedMap`. While iterating over these maps, they must be
synchronized.

Add a test case to reproduce the exception and verify the problem is resolved.

__cherry-pick to 1.4.x, 1.3.x__


artembilan:
LGTM
Merging...
Not sure in the cherry-picking, but we'll see.

PR #303, opened 2015-06-27
AMQP-507: Add AmqpMessageReturnedException

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-507

When `mandatory` is true, throw an exception on send and reply methods
if the message cannot be delivered.


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -1159,16 +1192,33 @@ public Message doInRabbit(Channel channel) throws Exception {
...
+				}
 				return reply;
 			}
 		}, obtainTargetConnectionFactoryIfNecessary(this.sendConnectionFactorySelectorExpression, message));
 	}
 
+	private Message exchangeMessages(final String exchange, final String routingKey, final Message message,
+			final CorrelationData correlationData, Channel channel, final PendingReply pendingReply, String messageTag)
+			throws Exception, InterruptedException {
+		Message reply;
```
artembilan: Minor, of  course, but anyway:
`InterruptedException` isn't thrown right now.
The `reply` is redundant.

Sorry, the solution isn't clear to me, so I try to simplify code to figure out.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -1159,16 +1192,33 @@ public Message doInRabbit(Channel channel) throws Exception {
...
+		boolean mandatory = mandatoryExpression.getValue(evaluationContext, message, Boolean.class);
```
artembilan: Why `returnCallback != null` isn't accounted here?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
garyrussell: Hopefully, to clarify...

With direct-replyTo (no listener), the thread handoff from the reply to the request was done via a BlockingQueue within the method scope. With a listener container, we had a map of pending replies, keyed by the correlation id because the reply does not come within the method scope.

Now, both use the map of pending replies so that if a return is received, an exception can be queued to the requestor.

Hope that helps.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -1159,16 +1192,33 @@ public Message doInRabbit(Channel channel) throws Exception {
...
+		if (mandatory && returnCallback == null) {
+			message.getMessageProperties().getHeaders().put(RETURN_CORRELATION_KEY, messageTag);
+		}
+		doSend(channel, exchange, routingKey, message, mandatory, correlationData);
```
artembilan: So, `doSend` has been changed to accept the `mandatory` param. OK, but the code there:
```
if (mandatory) {
	messageProperties.getHeaders().put(PublisherCallbackChannel.RETURN_CORRELATION_KEY, this.uuid);
}
```
says that it overrides the previous `messageTag` population. 
Maybe this `mandatory` really must be changed to the `returnCallback != null && mandatory` ?

artembilan: Doh. It's different constant...
Investigating further...

artembilan:
OK, Gary, now I see that you killed here two birds: added `AmqpMessageReturnedException` when it is `mandatory && this.returnCallback == null` and reworked `replyHandoff` for the `doSendAndReceiveWithTemporary` to the `PendingReply`.

So, point me out, please, where I should pay attention.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
garyrussell: Because we're request/reply; the return callback is internal to the return handling for this case. See `handleReturn`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
garyrussell: If the user supplies a `ReturnCallback` - he will handle the return; otherwise we'll handle it here.

garyrussell:
>So, point me out, please, where I should pay attention.

You are right, it's mainly a refactoring of `pendingReply` to support both use cases (real reply and returned request) and adding code to `handleReturn` to hand the exception to the requestor.

artembilan:
Good, merging after rebase and simple code polishing mentioned above.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/0a11c54328bf4d63beb8be609a84c0297b401ecc

PR #305, opened 2015-06-29
AMQP-509: Add Consumer Failure Events

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-509

Publish application events when listener container consumers fail.


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/ListenerContainerConsumerFailedEvent.java
```diff
@@ -0,0 +1,75 @@
...
+
+	/**
+	 * Construct an instance with the provided arguments.
+	 *
+	 * @param source the source.
+	 * @param reason the reason.
+	 * @param throwable the throwable.
+	 * @param fatal true if the startup failure was fatal (will not be retried).
+	 */
+	public ListenerContainerConsumerFailedEvent(SimpleMessageListenerContainer container, String reason,
```
artembilan: not `container` but `source`.
Will be fixed on merge, if all concerns are minor.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -1272,6 +1291,14 @@ private void logConsumerException(Throwable t) {
 				logger.warn("Consumer raised exception, processing can restart if the connection factory supports it. "
 						+ "Exception summary: " + t);
 			}
+			publishEvent("Consumer raised exception, attempting restart", false, t);
+		}
+
+		private void publishEvent(String reason, boolean fatal, Throwable t) {
```
artembilan: Why doesn't this method belong to the top-level `SimpleMessageListenerContainer` class ?

garyrussell: I suppose it could be, or we could move it if/when the container itself needs to publish an event.

artembilan: I don't have the gold ground answer, but looks like this method 3is enough specific just to those exceptions above.
So, I've changed my mind and merge it as is.
If we find other places to emit events, we always will be able to introduce other top-level methods or just use `applicationEventPublisher` directly.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/5d427966768dc2ecaa1905ed4301c5e63f31b6d9

PR #306, opened 2015-06-29
AMQP-499: Make WARN Log Configurable

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-499

The previous fix was incomplete. The listener container also
emits a WARN log when the queue cannot be consumed due to exclusivity.

Provide a mechanism to customize that log too.


artembilan:
Looks like nothing to complain about!
Merging...

PR #309, opened 2015-07-21
AMQP-512: Delete Test Queues; Close Connections

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-512

Ensure test queues are removed after tests.

Some tests relied on the queues existing already.

Also make sure connections are closed when tests complete.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: java
+jdk: oraclejdk8
+sudo: false
 services:
   - rabbitmq
 env:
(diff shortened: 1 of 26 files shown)


artembilan:
I think I'm fine with the fix.
Even if we have missed anything else, we always can fix them later.
So, merging as is after rebase...

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/2f5170f3f97203c1ead16c967b1bc8192d146b69

PR #312, opened 2015-07-27
AMQP-449: Verify Reply Container Configuration

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-449

- RabbitTemplate: Detect that it is a MessageListener
- SMLC: Detect the reply queue matches (if a simple replyAddress)

### DIFF
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/JavaConfigFixedReplyQueueTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/JavaConfigFixedReplyQueueTests.java
@@ -120,9 +120,9 @@ public class JavaConfigFixedReplyQueueTests {
 			Throwable t = e.getCause();
 			assertThat(t, Matchers.instanceOf(IllegalStateException.class));
 			assertThat(t.getMessage(),
-					Matchers.containsString("Listener expects us to be listening on '"
+					Matchers.containsString("Listener expects us to be listening on '["
 							+ TestUtils.getPropertyValue(this.fixedReplyQRabbitTemplateWrongQueue, "replyAddress")
-							+ "'; our queues: " + Arrays.asList(this.replyListenerContainerWrongQueue.getQueueNames())));
+							+ "]'; our queues: " + Arrays.asList(this.replyListenerContainerWrongQueue.getQueueNames())));
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/JavaConfigFixedReplyQueueTests.java
```diff
@@ -129,6 +197,19 @@ public SimpleMessageListenerContainer serviceListenerContainer() {
...
 
 		/**
+		 * @return The reply listener container - the rabbit template with the wrong queue is the listener.
+		 */
+		@Bean
+		public SimpleMessageListenerContainer replyListenerContainerWrongQueue() {
+			SimpleMessageListenerContainer container = new SimpleMessageListenerContainer();
+			container.setConnectionFactory(rabbitConnectionFactory());
+			container.setQueues(replyQueue());
+			container.setMessageListener(fixedReplyQRabbitTemplateWrongQueue());
```
artembilan: I don't mind of your solution, but won't it looks better to override the `replyAddress` in the provided `ListenerContainerAware` with the value from this `queues`, instead of supporting the same value from both side?

Of course make a WARN if the `replyAddress` is presented already and isn't equal to the value from container.
From other side it looks bad to listen to the several queues, but use `RabbitTemplate` as a listener. The code there on the matter looks like:
```
PendingReply pendingReply = this.replyHolder.get(messageTag);
	if (pendingReply == null) {
		if (logger.isWarnEnabled()) {
			logger.warn("Reply received after timeout for " + messageTag);
		}
	throw new AmqpRejectAndDontRequeueException("Reply received after timeout");
}
```
But seems for me it isn't always true, because we may receive message from other queues on the listener container.

Have I missed anything?

garyrussell: I don't think we should replace/warn; failing to start is more appropriate.

However, I agree that we should also fail to start if there is more than one queue; I will polish.

garyrussell:
Pushed

artembilan:
I think I'm fine with the polishing.
Merging...

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/c3edce12fce7cc3cf572fb8e8e4efae9c4d23980

PR #314, opened 2015-08-06
AMQP-516: Fix Auto-Declare Anonymous Queue

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-516

Previously, anonymous queues (no value attribute) were only declared
auto-delete and exclusive if those attributes were not provided.

Still default to true, but also use the value if provided.


artembilan:
LGTM
Merging...

PR #316, opened 2015-08-07
AMQP-518: Remove <listener-container/> id Attrib.

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-518

Previously, we deprecated the ID (AMQP-416) but simply ignoring it
is not really a good solution because the user "discovers" the bean
name changes (e.g. when autowiring).

Remove the attribute instead to force a fast failure.

Leave the attribute in the schema as a comment, in case users look there
first after seeing the problem after an upgrade.

Beef up the what's new to an `IMPORTANT` block.


artembilan:
@garyrussell , I get the same bad results locally as Travis.
Sorry, it can't be merged yet...

garyrussell:
Doh - rebased, pushed

artembilan:
Merged after squesh

PR #318, opened 2015-08-10
AMQP-520: MessageListenerAdapter Method Map

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-520

Support mapping from queue or consumer tag to method name.


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/MessageListenerAdapter.java
```diff
@@ -208,6 +213,45 @@ protected String getDefaultListenerMethod() {
...
+	 * @param queueOrTag The queue name or consumer tag.
+	 * @param methodName The method name.
+	 * @since 1.5
+	 */
+	public void addQueueOrTagToMethodName(String queueOrTag, String methodName) {
+		this.queueOrTagToMethodName.put(queueOrTag, methodName);
+	}
+
+	/**
+	 * Add the mapping of a queue name or consumer tag to a method name. The first lookup
```
artembilan: s/Add/Remove and so on.
In other words: the wrong JavaDocs.
Can be fixed on merge if that's all.

artembilan:
OK. Merging with the JavaDocs fix to the `remove` method.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/80dc1955034f9c62be385b0b1deb3856cdc7a10a

PR #319, opened 2015-08-11
AMQP-521: LocalizedQueueConnectionFactory

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-521

Connect to the broker where an HA mirrored queue is mastered.

__ported from XD with some polishing__

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/LocalizedQueueConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/LocalizedQueueConnectionFactory.java
@@ -1,0 +1,303 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.amqp.rabbit.connection;
+
+import java.net.MalformedURLException;
+import java.net.URISyntaxException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import org.springframework.amqp.AmqpException;
+import org.springframework.beans.factory.DisposableBean;
+import org.springframework.core.io.Resource;
+import org.springframework.util.Assert;
+
+import com.rabbitmq.http.client.Client;
+import com.rabbitmq.http.client.domain.QueueInfo;
+
+
+/**
+ * A {@link RoutingConnectionFactory} that determines the node on which a queue is located and
+ * returns a factory that connects directly to that node.
+ * The RabbitMQ management plugin is called over REST to determine the node and the corresponding
+ * address for that node is injected into the connection factory.
+ * A single instance of each connection factory is retained in a cache.
+ * If the location cannot be determined, the default connection factory is returned. This connection
+ * factory is typically configured to connect to all the servers in a fail-over mode.
+ * <p>{@link #getTargetConnectionFactory(Object)} is invoked by the
+ * {@code SimpleMessageListenerContainer}, when establishing a connection, with the lookup key having
+ * the format {@code '[queueName]'}.
+ * <p>All {@link ConnectionFactory} methods delegate to the default
+ *
+ * @author Gary Russell
+ * @since 1.2
+ */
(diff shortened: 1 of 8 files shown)


artembilan:
Travis says that there is something wrong with the code:

> /home/travis/build/spring-projects/spring-amqp/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/LocalizedQueueConnectionFactoryTests.java:108: error: method does not override or implement a method from a supertype

>			@Override
			^

But if it is only an issue, I'll address on merge.

garyrussell:
Hmmm - something was lost with me switching computers; I changed that test yesterday. Let me look

garyrussell:
Oh, I see, I just failed to remove it.

garyrussell:
Pushed

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/LocalizedQueueConnectionFactory.java
```diff
@@ -0,0 +1,358 @@
...
+		}
+		logger.warn("Failed to determine queue location for: " + queue + ", using default connection factory");
+		return null;
+	}
+
+	/**
+	 * Create a client instance.
+	 * @param adminUri the admin URI.
+	 * @param username the username
+	 * @param password the pasword.
```
artembilan: Typo: pas**s**word

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/LocalizedQueueConnectionFactory.java
```diff
@@ -0,0 +1,358 @@
...
+			}
+			try {
+				Client client = createClient(adminUri, this.username, this.password);
+				QueueInfo queueInfo = client.getQueue(this.vhost, queue);
+				if (queueInfo != null) {
+					String node = queueInfo.getNode();
+					if (node != null) {
+						String uri = this.nodeToAddress.get(node);
+						if (uri != null) {
+							return nodeConnectionFactory(queue, node, uri);
```
artembilan: Why don't we cache `queue -> connectionFactory` as well?
I just don't like that `client.getQueue()` HTTP request for each message...

Or is this fully intentional because of HA failover aim?

garyrussell: That's the point - it's not for each message, it's only called when we initiate (or lose) the connection.

This factory should __not__ be used with a `RabbitTemplate`, only a listener container - where the consumers are long-lived.

When the current connection is lost, the consumer recovers by creating a new connection; we have to check where the HA queue moved to so that we connect to the right node.

artembilan: Excellent!
And thank you!
I guess I must read docs first before making unrelated questions :smile:

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/LocalizedQueueConnectionFactoryTests.java
```diff
@@ -0,0 +1,215 @@
...
+import com.rabbitmq.http.client.domain.QueueInfo;
+
+
+/**
+ *
+ * @author Gary Russell
+ */
+public class LocalizedQueueConnectionFactoryTests {
+
+	private final Map<String, ConnectionFactory> cfs = new HashMap<String, ConnectionFactory>();
```
artembilan: redundant property

I won't mind to leave it here, but I'm curious if you have missed anything in the logic...

src/reference/asciidoc/amqp.adoc
```diff
@@ -410,7 +416,65 @@ The Namespace support also provides the `send-connection-factory-selector-expres
...
+    cf.setAddresses(this.props.getAddresses());
+    cf.setUsername(this.props.getUsername());
+    cf.setPassword(this.props.getPassword());
+    cf.setVirtualHost(this.props.getVirtualHost());
+    return cf;
+}
+
+@Bean
+public ConnectionFactory queueAffinityCF(
+        @Qualifier("defaultConnectionFactory") ConnectionFactory defaultCF) throws Exception{
```
artembilan: I don't see any point from where we throw `Exception` here. So, let me remove it on merge!

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/LocalizedQueueConnectionFactoryTests.java
artembilan: That's all for me here and I can merge it with the fixes for those two minor concerns.
I'm just confused with this one.
Just give me an answer.

Thanks

garyrussell: Yes, it's not used; I want to add a caution to the docs, so let me address your comments and I'll push another commit.

garyrussell:
Rebased, pushed.

artembilan:
That's fine merging, if that `CAUTION` looks good in the final Manual :smile:

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/62cf949e78704c7e376ca6c268da550c9f684c25

PR #321, opened 2015-08-12
AMQP-523: Fix Class Tangle

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-523

`SimpleMessageListenerContainer` and `ListenerContainerConsumerFailedEvent`
referenced each other.

Revert to using `Object` as the event `source`.


artembilan:
Merged after rebase

PR #322, opened 2015-08-13
Update SF, rabbit amqp and http Client Versions

garyrussell:
- amqp-client: 3.5.4
- http-client: 1.0.0.RELEASE
- http-client is no longer optional - needed by the LQCF
 - exclude it's spring-web dep and use one consistent with others
- spring framework 4.1.7

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -87,10 +87,10 @@ subprojects { subproject ->
 		log4jVersion = '1.2.17'
 		logbackVersion = '1.1.2'
 		mockitoVersion = '1.9.5'
-		rabbitmqVersion = project.hasProperty('rabbitmqVersion') ? project.rabbitmqVersion : '3.5.3'
-		rabbitmqHttpClientVersion = '1.0.0.M1'
+		rabbitmqVersion = project.hasProperty('rabbitmqVersion') ? project.rabbitmqVersion : '3.5.4'
+		rabbitmqHttpClientVersion = '1.0.0.RELEASE'
 
-		springVersion = project.hasProperty('springVersion') ? project.springVersion : '4.1.6.RELEASE'
+		springVersion = project.hasProperty('springVersion') ? project.springVersion : '4.1.7.RELEASE'
 
 		springRetryVersion = '1.1.2.RELEASE'
 	}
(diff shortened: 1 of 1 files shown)


garyrussell:
Updated SF too

build.gradle
```diff
@@ -202,12 +202,15 @@ project('spring-rabbit') {
 		compile project(":spring-amqp")
 
 		compile "com.rabbitmq:amqp-client:$rabbitmqVersion"
-		compile ("com.rabbitmq:http-client:$rabbitmqHttpClientVersion", optional)
+		compile ("com.rabbitmq:http-client:$rabbitmqHttpClientVersion") {
+			exclude group: 'org.springframework'
+		}
```
artembilan: Wait...
This was `optional` before.
I think you meant this:
```
compile ("com.rabbitmq:http-client:$rabbitmqHttpClientVersion") { dep ->
			optional dep
	exclude group: 'org.springframework'
}
```

build.gradle
```diff
@@ -202,12 +202,15 @@ project('spring-rabbit') {
...
 
 		compile ("org.springframework:spring-aop:$springVersion", optional)
 		compile "org.springframework:spring-context:$springVersion"
 		compile "org.springframework:spring-messaging:$springVersion"
 		compile "org.springframework:spring-tx:$springVersion"
+		compile "org.springframework:spring-web:$springVersion"
```
artembilan: I really not sure that we should worry about this dependency...
Anyway it must be `optional`.

build.gradle
artembilan: OK. I see your comment:
> http-client is no longer optional - needed by the LQCF 

But that doesn't matter that we can't live with `optional`. Who wants to use `LQCF`, that must add that dep for his purposes.
WDYT?

garyrussell: My biggest concern is that it has a dependency on `spring-web` and we don't want mixed spring versions on the CP.

artembilan: OK. I see your point.
From other side HOP artifact is so small that I don't see any big deal to pull it automatically.
Any guy can `exclude` it manually if that.

PR #325, opened 2015-08-31
AMQP-524: Publisher CF Extract Interface

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-524

Enable mocking of a connection factory that supports publisher confirms and
returns callbacks.


artembilan:
Nothing to complain about.
Merging...

PR #326, opened 2015-08-31
AMQP-525: Blocking receiveAndReply()

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-525

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
@@ -776,30 +776,7 @@ public class RabbitTemplate extends RabbitAccessor implements BeanFactoryAware, 
 
 			@Override
 			public Message doInRabbit(Channel channel) throws Exception {
-				channel.basicQos(1);
-				final CountDownLatch latch = new CountDownLatch(1);
-				QueueingConsumer consumer = new QueueingConsumer(channel) {
-
-					@Override
-					public void handleCancel(String consumerTag) throws IOException {
-						super.handleCancel(consumerTag);
-						latch.countDown();
-					}
-
-					@Override
-					public void handleConsumeOk(String consumerTag) {
-						super.handleConsumeOk(consumerTag);
-						latch.countDown();
-					}
-
-				};
-				channel.basicConsume(queueName, consumer);
-				if (!latch.await(10, TimeUnit.SECONDS)) {
-					if (channel instanceof ChannelProxy) {
-						((ChannelProxy) channel).getTargetChannel().close();
-					}
-					throw new AmqpException("Blocking receive, consumer failed to consume");
-				}
+				QueueingConsumer consumer = createQueueingConsumer(queueName, channel);
 				Delivery delivery;
 				if (receiveTimeout < 0) {
 					delivery = consumer.nextDelivery();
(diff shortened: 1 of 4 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -1594,6 +1598,34 @@ public void onMessage(Message message) {
...
+
+			@Override
+			public void handleConsumeOk(String consumerTag) {
+				super.handleConsumeOk(consumerTag);
+				latch.countDown();
+			}
+
+		};
+		channel.basicConsume(queueName, consumer);
+		if (!latch.await(10, TimeUnit.SECONDS)) {
```
artembilan: @garyrussell , I wonder if we can get rid of this blocker.
For example `gradlew clean testAll --parallel` passes for me with this `if`.

Having this the code
```
if (receiveTimeout < 0) {
    delivery = consumer.nextDelivery();
}
```
does not make sense.
Because we end up with the `AmqpException("Blocking receive, consumer failed to consume")` exactly after this `10 sec`.

Have I missed anything?

garyrussell: Sorry, I don't understand; this code is before the nextDelivery() call and is there to force the channel closed if we don't properly set up the consumer.

artembilan: OK. Sorry. Now I see that `basicConsume` and 
```
if (method instanceof Basic.Deliver) {
                processDelivery(command, (Basic.Deliver) method);
                return true;
```
are fully different stories.

It's like two-way WebSocket protocol, when we can stream to both side. But first of all we must make hand shake  - register our consumer for delivery.

I'll come back to you soon about `handleCancel` because I'm afraid that we don't handle `QueueingConsumer.POISON` properly, but let me merge your other PRs first of all.

artembilan: @garyrussell , here is a polishing to your PR: https://github.com/artembilan/spring-amqp/commit/6ff39d359993b7c21cdad3457fed3de6a08df136

I don't see reason to raise a new JIRA for that test, just because we don't have any concern.

If the `handleCancel()` is invoked on the `QueueingConsumer`, its `handle(Delivery delivery)` will take care about that `POISON` and the `com.rabbitmq.client.ConsumerCancelledException` is thrown to the `RabbitTemplate.receive()` thread to be handled as normal in the `execute()`'s `catch` block.

Note: pay attention to the `TODO` in my commit. I'd like to hear your point.

artembilan: Will be merged after your confirmation for the polishing.

garyrussell:
Polishing looks good; thanks!

I didn't really understand your TODO comment and I think we need to capture the `cause` - more polish added.

If you mean we could use the rabbit exception directly, I don't think we should do that now.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -1594,6 +1598,34 @@ public void onMessage(Message message) {
...
 
+	private QueueingConsumer createQueueingConsumer(final String queueName, Channel channel) throws Exception {
+		channel.basicQos(1);
+		final CountDownLatch latch = new CountDownLatch(1);
+		QueueingConsumer consumer = new QueueingConsumer(channel) {
+
+			@Override
+			public void handleCancel(String consumerTag) throws IOException {
+				super.handleCancel(consumerTag);
+				latch.countDown();
```
artembilan: Looking to the `QueueingConsumer.handle(Delivery delivery)` logic, I'm not sure that we need this implementation. 
I may understand why we wait for the `handleConsumeOk`, but this `cancel` isn't clear to me.
It just don't make sense, because `nextDelivery()` is throwing exception in case of `cancel` anyway.
WDYT?

artembilan: By the way, looking to the `basicConsume` logic we can't receive `cancel` before the obtaining `consumerTag` as a result of `basicConsume`. Just because there is yet `Consumer` on the Broker for the `tag` to cancel.

garyrussell: We need to force close the channel if the consume doesn't work.

See `((ChannelProxy) channel).getTargetChannel().close();` below.

You are probably right about the cancel.

artembilan: Good. Sorry, if I wasn't clear here.
I just mean that we don't need this `handleCancel` and interaction on the `latch`.
I agreed about the `channel.close()` after some reasonable timeout.

From other side this `handleCancel()` doesn't hurt and we can still live with it because of GA dead line :smile: .
So, merging after squash and rebase!

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/b4e33d477b74a8594c5eae2b21fc2cf5b107197a

PR #331, opened 2015-09-11
AMQP-532: Fix Pub.Conf. Nacks on close()

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-532

Properly use the iterator to remove the entries.

__cherry-pick to 1.4.x__


artembilan:
... and cherry-picked

PR #334, opened 2015-10-09
AMQP-540: Fix @RabbitHandler on Proxy

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-540

Previously, class-level `@RabbitListener` was not detected if the class was proxied.

This solves the reported problem but it is incomplete - see AMQP-541.


artembilan:
Looks like nothing to complain about!
Merging...

PR #335, opened 2015-10-12
AMQP-541: Fix @RabbitListener Proxy Param Match

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-541

__ @artembilan  do not merge until @snicoll has reviewed__

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
@@ -39,6 +39,7 @@ import org.springframework.amqp.rabbit.listener.MultiMethodRabbitListenerEndpoin
 import org.springframework.amqp.rabbit.listener.RabbitListenerContainerFactory;
 import org.springframework.amqp.rabbit.listener.RabbitListenerEndpointRegistrar;
 import org.springframework.amqp.rabbit.listener.RabbitListenerEndpointRegistry;
+import org.springframework.aop.framework.Advised;
 import org.springframework.aop.support.AopUtils;
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.BeanFactory;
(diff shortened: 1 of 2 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -265,6 +266,15 @@ private Method checkProxy(Method method, Object bean) {
...
 				// is it also present on the proxy itself?
 				method = bean.getClass().getMethod(method.getName(), method.getParameterTypes());
+				Class<?>[] proxiedInterfaces = ((Advised) bean).getProxiedInterfaces();
+				for (Class<?> iface : proxiedInterfaces) {
+					try {
+						method = iface.getMethod(method.getName(), method.getParameterTypes());
+						break;
+					}
+					catch (NoSuchMethodException noMethod) {
+					}
```
artembilan: Well, if there is no other way to do the same (and I guess no :smile: ), I don't see reason to wait for @snicoll .
Although it looks like need to debug the stuff. Because you have changed the previous method for the `MethodRabbitListenerEndpoint` and thereafter for the `processListener()`...

artembilan:
@garyrussell , @snicoll , what is the final decision for this issue?
Thanks in advance

garyrussell:
Let's wait to see how @snicoll fixes [SPR-13576](https://jira.spring.io/browse/SPR-13576) - we should be consistent.

snicoll:
alright I am putting this one higher in the stack.

garyrussell:
We can't make 1.5.x require 4.2.3 (it currently only requires 4.1) so let's go with this approach for 1.5.x and move to the new `MethodIntrospector` in 1.6.

artembilan:
I can't find any point to complain about.
LGTM.
Merging...
Will revise in favor of `MethodIntrospector` in the `1.6`

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/8955cba632055ce6121ed194a433d4d12e483763 after rebase.

PR #336, opened 2015-10-16
AMQP-542: @SendTo with Class Level @RabbitListener

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-542

Spurious NPE logged in `AnnotationUtils` when using class-level annotation.

Detect a null `method` and add support for `@SendTo` on `@RabbitHandler`.

Class-level `@SendTo` is not supported due to annotation restrictions.
See SPR-13578.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/MultiMethodRabbitListenerEndpoint.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/MultiMethodRabbitListenerEndpoint.java
@@ -69,7 +69,10 @@ public class MultiMethodRabbitListenerEndpoint extends MethodRabbitListenerEndpo
 		@Override
 		protected Address getReplyToAddress(Message request) throws Exception {
 			Address replyTo = request.getMessageProperties().getReplyToAddress();
-			Address defaultReplyTo = delegatingHandler.getDefaultReplyTo();
+			Address defaultReplyTo = null;
+			if (delegatingHandler != null) {
+				defaultReplyTo = delegatingHandler.getDefaultReplyTo();
+			}
 			if (replyTo == null && defaultReplyTo == null) {
 				throw new AmqpException(
 						"Cannot determine ReplyTo message property value: " +
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/MultiMethodRabbitListenerEndpoint.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/MultiMethodRabbitListenerEndpoint.java
@@ -19,6 +19,9 @@ import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.springframework.amqp.AmqpException;
+import org.springframework.amqp.core.Address;
+import org.springframework.amqp.core.Message;
 import org.springframework.amqp.rabbit.listener.adapter.DelegatingInvocableHandler;
 import org.springframework.amqp.rabbit.listener.adapter.HandlerAdapter;
 import org.springframework.amqp.rabbit.listener.adapter.MessagingMessageListenerAdapter;
@@ -35,6 +38,8 @@ public class MultiMethodRabbitListenerEndpoint extends MethodRabbitListenerEndpo
 
 	private final Object bean;
 
+	private DelegatingInvocableHandler delegatingHandler;
+
 	public MultiMethodRabbitListenerEndpoint(List<Method> methods, Object bean) {
 		this.methods = methods;
 		this.bean = bean;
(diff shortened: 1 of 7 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/MultiMethodRabbitListenerEndpoint.java
```diff
@@ -47,7 +52,33 @@ protected HandlerAdapter configureListenerAdapter(MessagingMessageListenerAdapte
...
+		return new MultiMethodMessageListenerAdapter();
+	}
+
+
+	private final class MultiMethodMessageListenerAdapter extends MessagingMessageListenerAdapter {
+
+		@Override
+		protected Address getReplyToAddress(Message request) throws Exception {
+			Address replyTo = request.getMessageProperties().getReplyToAddress();
+			Address defaultReplyTo = delegatingHandler.getDefaultReplyTo();
```
artembilan: The solution looks weak.
I understand that all these classes are the part of the infrastructure and everything is called in the right order, but this filed can be `null`, if we don't call `configureListenerAdapter`

garyrussell:
Pushed

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/MultiMethodRabbitListenerEndpoint.java
```diff
@@ -47,7 +52,36 @@ protected HandlerAdapter configureListenerAdapter(MessagingMessageListenerAdapte
 			invocableHandlerMethods.add(getMessageHandlerMethodFactory()
 					.createInvocableHandlerMethod(getBean(), method));
```
artembilan: I wonder if we could end up with some issue, having `bean` as a field here and the setter for the same in the super class...
I'd say that we really should use for the `MultiMethodRabbitListenerEndpoint` the same setters model. Even for the `methods`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/DelegatingInvocableHandler.java
```diff
@@ -40,20 +49,35 @@
  */
 public class DelegatingInvocableHandler {
 
+	private static final ThreadLocal<Address> defaultReplyTo = new ThreadLocal<Address>();
```
artembilan: I don't like this solution, but, yes: I can't find the other way to deal with the `replyTo`...

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/MultiMethodRabbitListenerEndpoint.java
garyrussell: Good catch; in fact we don't even need the field; we can `setBean()` from the ctor and use `getBean()` everywhere else.

Not sure why I would have added the field to the subclass.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/DelegatingInvocableHandler.java
garyrussell: Right - to avoid it would require major rework in the adapter.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/MultiMethodRabbitListenerEndpoint.java
artembilan: Yes, that works, too. Since we shot it already, we can't remove ctor. So, yes, delegate to the setter is good compromise.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/DelegatingInvocableHandler.java
garyrussell: We could consider it in 1.6.

artembilan: Yep! Makes sense. 
So, I'm waiting for your final polishing and merge

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/MultiMethodRabbitListenerEndpoint.java
artembilan: OK. I can address this on merge.
Never mind

garyrussell:
Pushed - I was waiting for a `testAll`.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/d11c1161766585fdb9cad32921ffe43f8f89ef35

PR #337, opened 2015-10-19
Fix x-expires Test Case

garyrussell:
Due to the small `x-expires`, it's possible for both consumers to
redeclare the queue; the test only expects one to do so.

Also, change the queue to have a random name to avoid collisions with other builds.

Also, noticed an NPE occasionally when running the test (null `BackOffExecution`).

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
@@ -1353,7 +1353,7 @@ public class SimpleMessageListenerContainer extends AbstractMessageListenerConta
 	 * @throws Exception if the shared connection still can't be established
 	 */
 	protected void handleStartupFailure(BackOffExecution backOffExecution) throws Exception {
-		long recoveryInterval = backOffExecution.nextBackOff();
+		long recoveryInterval = backOffExecution != null ? backOffExecution.nextBackOff() : DEFAULT_RECOVERY_INTERVAL;
 		if (BackOffExecution.STOP == recoveryInterval) {
 			synchronized (this) {
 				if (isActive()) {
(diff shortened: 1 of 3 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -1353,7 +1353,7 @@ protected void invokeListener(Channel channel, Message message) throws Exception
 	 * @throws Exception if the shared connection still can't be established
 	 */
 	protected void handleStartupFailure(BackOffExecution backOffExecution) throws Exception {
-		long recoveryInterval = backOffExecution.nextBackOff();
+		long recoveryInterval = backOffExecution != null ? backOffExecution.nextBackOff() : DEFAULT_RECOVERY_INTERVAL;
```
artembilan: Regarding this one - `ListenFromAutoDeleteQueueTests.testAutoDeclareFalse()`:
```
BlockingQueueConsumer consumer = mock(BlockingQueueConsumer.class);
doThrow(RuntimeException.class).when(consumer).start();
when(listenerContainer.createBlockingQueueConsumer()).thenReturn(consumer);
```
But `SimpleMessageListenerContainer.createBlockingQueueConsumer()` has this piece of code:
```
consumer.setBackOffExecution(this.recoveryBackOff.start());
```

Shouldn't we just fix that test-case instead of the strange NPE check here?

garyrussell: Good; it wasn't obvious to me where the issue was so I just added the protection. 

Pushed

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/ListenFromAutoDeleteQueueTests.java
```diff
@@ -141,6 +147,7 @@ public void testAutoDeclareFalse() throws Exception {
 		//Prevent a long 'passiveDeclare' process
 		BlockingQueueConsumer consumer = mock(BlockingQueueConsumer.class);
 		doThrow(RuntimeException.class).when(consumer).start();
+		when(consumer.getBackOffExecution()).thenReturn(mock(BackOffExecution.class));
```
artembilan: Cool! This empty mock returns `0` for the `backOffExecution.nextBackOff()` :smile: 
Merging...

artembilan:
Merged after squash

PR #339, opened 2015-11-02
AMQP-544: API to Obtain @RabbitListener ids

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-544

Also fix a test that used a queue named `foo`. It failed
if the queue existed and had incompatible messages.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/RabbitListenerEndpointRegistry.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/RabbitListenerEndpointRegistry.java
@@ -21,6 +21,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 
@@ -54,6 +55,7 @@ import org.springframework.util.StringUtils;
  * @author Stephane Nicoll
  * @author Juergen Hoeller
  * @author Artem Bilan
+ * @author Gary Russell
  * @since 1.4
  * @see RabbitListenerEndpoint
  * @see MessageListenerContainer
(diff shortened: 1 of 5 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/RabbitListenerEndpointRegistry.java
```diff
@@ -91,6 +94,15 @@ public MessageListenerContainer getListenerContainer(String id) {
 	}
 
 	/**
+	 * Return the ids of the managed {@link MessageListenerContainer} instance(s).
+	 * @return the ids.
+	 * @see #getListenerContainer(String)
+	 */
```
artembilan: `@since  1.5.2`.
Minor, of course, and can be fixed on merge

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactoryIntegrationTests.java
```diff
@@ -92,6 +94,9 @@ public void open() {
 
 	@After
 	public void close() {
+		if (!this.connectionFactory.getVirtualHost().equals("non-existent")) {
```
artembilan: What is the point of this check ?
Seems for me we don't use other virtual hosts, unless the default one.

garyrussell: Without this check, `testReceiveFromNonExistentVirtualHost()` fails - it changes the virtual host in the connection factory.

artembilan: Doh... Need to read the code locally before asking such a stupid questions...
Thanks.
So, merging with that simple `@since` fix

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/453b2a8b088e1c2c3131459bc9b573cd9fc80e31

PR #341, opened 2015-11-14
AMQP-546: Add SmartLifecycle to Caching Conn Fact

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-546

Code was added to prevent connection use after the context is closed. However,
this was too early in the lifecycle because other beans might not have been
stopped yet; they could no longer use the connection.

Add `SmartLifecyle` to `CachingConnectionFactory` and put it in a late phase,
by default, so that beans in earlier phases can still use the connection in
`stop()`.

Add test case to reproduce the problem.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
@@ -139,6 +139,8 @@ public class CachingConnectionFactory extends AbstractConnectionFactory
 
 	private volatile boolean stopped;
 
+	private volatile boolean running;
+
 	private int phase = Integer.MIN_VALUE + 1000;
 
 	private volatile ConditionalExceptionLogger closeExceptionLogger = new DefaultChannelCloseLogger();
@@ -585,11 +587,12 @@ public class CachingConnectionFactory extends AbstractConnectionFactory
 
 	@Override
 	public void start() {
+		this.running = true;
 	}
 
 	@Override
 	public boolean isRunning() {
-		return true;
+		return this.running;
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
```diff
@@ -579,6 +583,50 @@ public void resetConnection() {
 		}
 	}
 
+	@Override
+	public void start() {
+	}
+
+	@Override
+	public boolean isRunning() {
+		return true;
```
artembilan: no, `this.running` must there anyway.
Or we can rely on the `this.contextStopped` variable...

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
```diff
@@ -579,6 +583,50 @@ public void resetConnection() {
...
+		this.phase = phase;
+	}
+
+	@Override
+	public boolean isAutoStartup() {
+		return true;
+	}
+
+	public void stop() {
+		if (this.contextStopped) {
```
artembilan: We should add some note to the JavaDocs and/or Reference to say:
> you can't stop() `CachingConnectionFactory` at runtime. Only applicationContext lifecycle

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/ConnectionFactoryLifecycleTests.java
```diff
@@ -0,0 +1,127 @@
...
+
+		@Override
+		public void start() {
+			this.running = true;
+			this.admin.declareQueue(this.queue);
+		}
+
+		@Override
+		public void stop() {
+			this.admin.deleteQueue(this.queue.getName());
```
artembilan: I guess, this is the main point in this test.
Previously (on current `master`) we end up with the error.

garyrussell:
Pushed

artembilan:
LGTM
Merging...

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/d9d0963f3abd43585a484d02ba874625423193d1 after squash.

PR #342, opened 2015-11-24
@EnableRabbitIntegrationTests Improvements

garyrussell:
Demo ClassMapper in @RabbitListener

Add example of how to use a custom ClassMapper with JSON.

Also, use a `TestExecutionListener` to defer queue deletion until after the
context has been stopped; avoids shutdown delays and noise in logs.

### DIFF
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/test/TestExecutionListenerAdapter.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/test/TestExecutionListenerAdapter.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


spring-rabbit/src/test/java/org/springframework/amqp/rabbit/test/TestExecutionListenerAdapter.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.test.context.TestExecutionListener;
+
+/**
+ * Base class for {@link TestExecutionListener}s.
+ *
+ * @author Gary Russell
+ * @since 1.5.3
+ *
+ */
+public class TestExecutionListenerAdapter implements TestExecutionListener {
```
artembilan: Seems for me `org.springframework.test.context.support.AbstractTestExecutionListener` should be enough for you to avoid this extra class.

artembilan: Otherwise LGTM and you can merge it after the fix for this my concern.

garyrussell: Cool; I didn't find that.

garyrussell:
Merged as a604f74e5dbda10aac165fe2cd387e6540815310

PR #343, opened 2015-11-26
Support @RabbitListener on interface methods

Chumper:
At our project we are using modules to define our interfaces between services.
In this module we put our hystrix interfaces as well as our amqp events that the services emits.
For a particular event we create interfaces that services can implement to listen on those events:
```java
interface UserCreatedEventListener {

    /**
     * Will be sent as soon as a new User was created
     */
    @RabbitListener(
            bindings = @QueueBinding(
                    value = @Queue(
                            value = '${spring.application.name}',
                            autoDelete = "true",
                            durable = "false"
                    ),
                    exchange = @Exchange(
                            type = "fanout",
                            value = "all-events",
                            durable = "true"
                    )
            )
    )
    void onEvent(UserCreated event);
}

//--- other class on the service
class UserService implements UserCreatedEventListener {
  void onEvent(UserCreated event) {
    // do stuff
  }
}
```

Now a service can implement this method and should auto register on the bus.
This does not work at the moment, because Annotations can not be inherited.

Luckily the `RabbitListenerAnnotationBeanPostProcessor` is responsible for the creation of the listener and does this in the method `postProcessAfterInitialization`.

Unfortunately the method `getAnnotation` does not consider superclasses or interfaces, but luckily the `findAnnotation` does.

So i propose to use the latter instead and enable us to provide the binding in another class (e.g. interface) than the implementing class.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
@@ -224,7 +224,7 @@ public class RabbitListenerAnnotationBeanPostProcessor
 
 			@Override
 			public void doWith(Method method) throws IllegalArgumentException, IllegalAccessException {
-				RabbitListener rabbitListener = AnnotationUtils.getAnnotation(method, RabbitListener.class);
+				RabbitListener rabbitListener = AnnotationUtils.findAnnotation(method, RabbitListener.class);
 				if (rabbitListener != null) {
 					processAmqpListener(rabbitListener, method, bean, beanName);
 				}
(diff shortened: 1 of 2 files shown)


Chumper:
I have added a test case to support my proposed changes

garyrussell:
Thanks for the contribution; before we can look at it; please confirm you have "signed" the CLA. See the [contribution guidelines](https://github.com/spring-projects/spring-amqp#contributing-to-spring-amqp).

Please note, also, that this should be addressed in 1.6 with [AMQP-545](https://jira.spring.io/browse/AMQP-545) by using Spring's new `MethodIntrospector` (requires Spring 4.2.3).

But if your change doesn't break anything we can consider ading it in a 1.5.x point release.

Thanks again.

Chumper:
I signed the CLA. 
As far as i can see it should contain any breaking changes, but i may be wrong.
Thanks for the hint with the `MethodInspector`, i will check it out.

garyrussell:
Thanks again for the contribution; merged as d976a776d599dc7397aeb5b4d7abad9f6c6e9835 after a bit of polish and another test.

In future, please submit a PR from a topic branch; you will need to `reset --hard upstream/master` your master because other commits have been applied before this (and I needed to squash your commits).

We also like to track changes like this via JIRA: https://jira.spring.io/browse/AMQP-552

See https://github.com/spring-projects/spring-amqp#contributing-to-spring-amqp and https://github.com/spring-projects/spring-integration/blob/master/CONTRIBUTING.md#a-day-in-the-life-of-a-contributor

Thanks again!

PR #344, opened 2015-11-30
AMQP-551: Fix Mockito Usage in Failover Tests

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-551

Perform all mock creation/interaction on the main thread.

### DIFF
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/LocalizedQueueConnectionFactoryTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/LocalizedQueueConnectionFactoryTests.java
@@ -73,14 +73,10 @@ public class LocalizedQueueConnectionFactoryTests {
 
 	private final Map<String, String> consumerTags = new HashMap<String, String>();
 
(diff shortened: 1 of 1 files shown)


spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/LocalizedQueueConnectionFactoryTests.java
```diff
@@ -90,22 +86,25 @@ public void testFailOver() throws Exception {
...
 			protected Client createClient(String adminUri, String username, String password)
 					throws MalformedURLException, URISyntaxException {
-				return doCreateClient(adminUri, username, password, firstServer.get() ? nodes[0] : nodes[1]);
+				return firstServer.get() ? client1 : client2;
 			}
 
-
 			@Override
 			protected ConnectionFactory createConnectionFactory(String address, String node) throws Exception {
-				return mockCF(address);
```
artembilan: This is the root of cause...
Yeah, need to take into account that any mock must be created before the actual test work.
Feel free to merge, if we don't have such a vulnerability somewhere else :smile:

PR #346, opened 2015-12-05
Code Polishing (Sonar)

garyrussell:
- remove redundant modifiers
- overridable methods called from ctors

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/ConnectionFactoryUtils.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/ConnectionFactoryUtils.java
@@ -15,8 +15,6 @@ package org.springframework.amqp.rabbit.connection;
 
 import java.io.IOException;
 
-import com.rabbitmq.client.Channel;
-
 import org.springframework.amqp.AmqpIOException;
 import org.springframework.transaction.NoTransactionException;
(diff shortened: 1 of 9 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/ConnectionFactoryUtils.java
```diff
@@ -25,6 +23,8 @@
 import org.springframework.transaction.support.TransactionSynchronizationManager;
 import org.springframework.util.Assert;
 
+import com.rabbitmq.client.Channel;
```
artembilan: Looks like you have different import config for AMQP project.
Or am I missing anything here again?
imports

artembilan:
Everything else looks good.
Just let me know about that epic import issue.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/ConnectionFactoryUtils.java
garyrussell: I haven't changed anything in my STS; I have always had these settings, which (I think) I got from Mark:

screen shot 2015-12-08 at 10 12 25 am

artembilan: OK. Doing the same settings and merging...
Good morning, BTW!

PR #349, opened 2015-12-18
AMQP-556: NPE in SMLC After Shutdown Timeout

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-556

If a listener doesn't shut down within the timeout, when it
next checks whether it is active, it gets an NPE.

Since it is a terminating listener, it is benign.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
@@ -818,7 +818,7 @@ public class SimpleMessageListenerContainer extends AbstractMessageListenerConta
 				logger.info("Successfully waited for workers to finish.");
 			}
 			else {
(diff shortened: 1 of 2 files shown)


artembilan:
Nothing to complain about.
Merging...
Just let me know if we should backport it.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -835,7 +835,13 @@ protected void doShutdown() {
 	private boolean isActive(BlockingQueueConsumer consumer) {
 		Boolean consumerActive;
 		synchronized(consumersMonitor) {
-			consumerActive = this.consumers != null && this.consumers.get(consumer);
```
artembilan: Yeah... That is really nasty bug from our side: do not care about outboxing for `null` :cry: .
Anyway, good to know and thanks for fix!

PR #350, opened 2015-12-23
AMQP-558: amqp-client 3.5.7 + 3.6.0 Compatibility

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-558

New methods on `Channel`.

__do not resolve JIRA__

After this is merged, we'll release 1.5.3, then make master 1.6, then bump amqp-client to 3.6.0.


artembilan:
I think I'm fine with the fix.
Pulling locally to `testAll` and merge....

PR #352, opened 2015-12-23
AMPQ-554: Namespace Additions

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-554

- `ThreadFactory` on connection factory
- `NamingStrategy` on anonymous queues

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/ConnectionFactoryParser.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/ConnectionFactoryParser.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2010-2014 the original author or authors.
+ * Copyright 2010-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
  * the License. You may obtain a copy of the License at
@@ -56,6 +56,8 @@ class ConnectionFactoryParser extends AbstractSingleBeanDefinitionParser {
 
 	private static final String CONNECTION_CACHE_SIZE_ATTRIBUTE = "connection-cache-size";
 
+	private static final String THREAD_FACTORY = "thread-factory";
+
 	@Override
 	protected Class<?> getBeanClass(Element element) {
 		return CachingConnectionFactory.class;
@@ -94,6 +96,7 @@ class ConnectionFactoryParser extends AbstractSingleBeanDefinitionParser {
 		NamespaceUtils.setValueIfAttributeDefined(builder, element, CONNECTION_TIMEOUT);
 		NamespaceUtils.setValueIfAttributeDefined(builder, element, CACHE_MODE);
 		NamespaceUtils.setValueIfAttributeDefined(builder, element, CONNECTION_CACHE_SIZE_ATTRIBUTE);
+		NamespaceUtils.setReferenceIfAttributeDefined(builder, element, THREAD_FACTORY, "connectionThreadFactory");
 
 	}
 
--- a/spring-rabbit/src/main/resources/org/springframework/amqp/rabbit/config/spring-rabbit-1.6.xsd
+++ b/spring-rabbit/src/main/resources/org/springframework/amqp/rabbit/config/spring-rabbit-1.6.xsd
@@ -31,8 +31,23 @@
 				<xsd:annotation>
 					<xsd:documentation><![CDATA[
 						The name of the queue.  Clients can receive or listen for messages by referring to the
-						queue itself, or to its name.
+						queue itself, or to its name. When no 'name' is provided an 'AnonymousQueue' with a
+						generated name is used. Also see naming strategy.
 					]]></xsd:documentation>
+				</xsd:annotation>
+			</xsd:attribute>
+			<xsd:attribute name="naming-strategy" type="xsd:string">
+				<xsd:annotation>
+					<xsd:documentation><![CDATA[
+						When no 'name' is provided, a naming strategy is used to create the anonymous queue name.
+						Default (when not provided) is the String representation of a UUID. To customize,
+						provide a reference to a bean implementing 'AnonymousQueue.NamingStrategy'.
+					]]></xsd:documentation>
+					<xsd:appinfo>
+						<tool:annotation kind="ref">
+							<tool:expected-type type="org.springframework.amqp.core.AnonymousQueue.NamingStrategy" />
+						</tool:annotation>
+					</xsd:appinfo>
 				</xsd:annotation>
 			</xsd:attribute>
 			<xsd:attribute name="auto-delete" use="optional" type="xsd:string" default="false">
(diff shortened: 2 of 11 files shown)


spring-amqp/src/main/java/org/springframework/amqp/core/QueueFactoryBean.java
```diff
@@ -0,0 +1,113 @@
...
+	@Override
+	public Class<?> getObjectType() {
+		return AnonymousQueue.class;
+	}
+
+	@Override
+	protected Queue createInstance() throws Exception {
+		Queue queue;
+		if (StringUtils.hasText(this.name)) {
+			queue = new Queue(this.name, this.durable, this.exclusive, this.autoDelete, this.arguments);
```
artembilan: I think the `AnonymousQueue` restriction in the JavaDoc above and in the `getObjectType()` is by mistake and we simply can represent this `QueueFactoryBean` as general factory for `Queue` implementations.

spring-amqp/src/main/java/org/springframework/amqp/core/QueueFactoryBean.java
```diff
@@ -0,0 +1,113 @@
...
+		}
+		else {
+			Assert.state(!this.durable && this.exclusive && this.autoDelete, "If there is no 'name' attribute, "
+					+ "'durable' must be false, 'exclusive' and 'auto-delete' must be 'true'");
```
artembilan: I think this restriction we can reach with the mutually exclusive constructors.
Or... or let's make this class non-`public`!

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/ConnectionFactoryParser.java
```diff
@@ -94,6 +96,7 @@ protected void doParse(Element element, ParserContext parserContext, BeanDefinit
 		NamespaceUtils.setValueIfAttributeDefined(builder, element, CONNECTION_TIMEOUT);
 		NamespaceUtils.setValueIfAttributeDefined(builder, element, CACHE_MODE);
 		NamespaceUtils.setValueIfAttributeDefined(builder, element, CONNECTION_CACHE_SIZE_ATTRIBUTE);
+		NamespaceUtils.setReferenceIfAttributeDefined(builder, element, THREAD_FACTORY, "connectionThreadFactory");
```
artembilan: Such a change requires Copyright update.

spring-rabbit/src/main/resources/org/springframework/amqp/rabbit/config/spring-rabbit-1.6.xsd
```diff
@@ -31,25 +31,41 @@
...
+				<xsd:annotation>
+					<xsd:documentation><![CDATA[
+						When no 'name' is provided, a naming strategy is used to create the anonymous queue name.
+						Default (when not provided) is the String representation of a UUID. Options are
+						'uuid' - the String representation of a UUID, 'spring' - 'spring.gen-<uniqueBase64String>',
+						or a reference to an 'AnonymousQueue.NamingStrategy' implementation.
 					]]></xsd:documentation>
+					<xsd:appinfo>
+						<tool:annotation kind="ref">
+							<tool:expected-type type="java.util.Map" />
```
artembilan: Must be `org.springframework.amqp.core.AnonymousQueue.NamingStrategy` or nothing since it accepts the raw string ('spring', 'uuid') as well.

spring-rabbit/src/main/resources/org/springframework/amqp/rabbit/config/spring-rabbit-1.6.xsd
```diff
@@ -31,25 +31,41 @@
...
 			</xsd:attribute>
-			<xsd:attribute name="exclusive" use="optional" type="xsd:string" default="false">
+			<xsd:attribute name="exclusive" use="optional" type="xsd:string">
 				<xsd:annotation>
 					<xsd:documentation><![CDATA[
 					Flag indicating that the queue is exclusive to this connection.  Default is false.
 				]]></xsd:documentation>
 				</xsd:annotation>
 			</xsd:attribute>
-			<xsd:attribute name="durable" use="optional" type="xsd:string" default="true">
```
artembilan: I think these can be reinstated if we introduce mutually exclusive constructors for the `QueueFactoryBean`.
Or do you worry here about something like `${queueName}` resolved to the empty string?
We can ignore these defaults if we have `naming-strategy` or don't have `name` attribute.

Seems for me that `QueueFactoryBean` adds more questions than answers to existing concerns...

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/config/ConnectionFactoryParserTests.java
```diff
@@ -114,6 +114,8 @@ public void testMultiHost() throws Exception {
 		assertEquals(-1, addresses[1].getPort());
 		assertEquals("host3", addresses[2].getHost());
 		assertEquals(4567, addresses[2].getPort());
+		assertSame(beanFactory.getBean("tf"), TestUtils.getPropertyValue(connectionFactory,
```
artembilan: I can update all Copyright violations on merge, of course.

spring-amqp/src/main/java/org/springframework/amqp/core/QueueFactoryBean.java
garyrussell: Yes; it started life as just for anonymous queues but I realized I needed it for all queues.

spring-rabbit/src/main/resources/org/springframework/amqp/rabbit/config/spring-rabbit-1.6.xsd
garyrussell: Doh - it should be omitted.

spring-rabbit/src/main/resources/org/springframework/amqp/rabbit/config/spring-rabbit-1.6.xsd
garyrussell: I had to go to a factory bean to support `spring`, `uuid` or a strategy; if you have a better idea; share it.

artembilan: How about just rely on the `bean reference`?
The `spring` is just only simple `Base64UrlNamingStrategy` bean definition.
The `uuid` can be used by default when we don't have both `name` and `naming-strategy` attributes.
You should agree with me that it is enough complex to workaround only `spring` word.
Right, it is important word anyway, but that doesn't help us to have clear robust solution :smile:

garyrussell:
Reverted parser; removed new factory bean; squashed; pushed.

src/reference/asciidoc/amqp.adoc
```diff
@@ -2516,6 +2530,30 @@ removed.
 You can provide your own naming strategy, whereby you can include other information (e.g. application, client host) in
 the queue name.
 
+Starting with _version 1.6_, the naming strategy can be specified when using XML configuration; the strategy can
+be `uuid`, `spring`, or a reference to a bean that implements `AnonymousQueue.NamingStrategy`.
```
artembilan: This expression isn't valid anymore.
Will fix on merge.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/a3412d554fd9c7eaa81fad0d5dcf934c3ec922aa after the Docs fix and `What's New` conflict fix.

PR #355, opened 2016-01-14
AMQP-561: Log Appender Exchange Declaration

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-561

When `declareExchange` is `true`, defer exchange declaration until
the connection is established.

__cherry-pick to 1.5.x__

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/LogAppenderConnectionListener.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/LogAppenderConnectionListener.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.amqp.rabbit.support;
+
+import org.springframework.amqp.core.Exchange;
+import org.springframework.amqp.rabbit.connection.Connection;
+import org.springframework.amqp.rabbit.connection.ConnectionListener;
+import org.springframework.amqp.rabbit.core.RabbitAdmin;
+
+/**
+ * A {@link ConnectionListener} that will declare the exchange for the
+ * appender, if so configured.
+ *
+ * @author Gary Russell
+ * @since 1.5.4
+ *
+ */
+public class LogAppenderConnectionListener implements ConnectionListener {
+
+	private final Exchange exchange;
+
+	private final RabbitAdmin admin;
+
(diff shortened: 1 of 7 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/LogAppenderConnectionListener.java
```diff
@@ -0,0 +1,56 @@
...
+			this.admin.declareExchange(this.exchange);
+		}
+		catch (Exception e) {
+
+		}
+	}
+
+	@Override
+	public void onClose(Connection connection) {
+	}
```
artembilan: Unfortunately I don't want this class :cry: 
It is very simple and very weak. 
I mean the provided `admin` maybe autowired with another `connectionFactory` and this event adds more noise, rather than helps.
So, my position is just hide this fix as an inline implementation:
````java
this.connectionFactory.addConnectionListener(new ConnectionListener() {

	@Override
	public void onCreate(Connection connection) {
		try {
	        	admin.declareExchange(exchange);
		}
		catch (Exception e) {
			// Ignore any admin failures
		}
	}

	@Override
	public void onClose(Connection connection) {

	}

});
````
WDYT?

Comment (garyrussell) [inline spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/LogAppenderConnectionListener.java]: >simple and very weak. 

Not sure what you mean by "weak".

I would have done as you suggest if we only had one appender.

Comment (artembilan) [inline spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/LogAppenderConnectionListener.java]: "Weak" meaning:"
    1.  lacking the power to perform physically demanding tasks; lacking physical strength and energy.
    2. liable to break or give way under pressure; easily damaged.

...
:smile: 

Having this class as public, end user may decide to use it for his on reason.
In this case the provided `admin` might be configured for absolutely different `connectionFactory`.
The `onCreate` for this `connectionFactory` doesn't help that wrong `admin` to live properly.
And so on.

I can be so nit-picking, of course, but the name for this class like `DeclareExchangeConnectionListener` may make me tolerant having this class in the `org.springframework.amqp.rabbit.support` and don't provide any `log` logic inside.
From other side I don't see reason to request the `admin`  at all:
```
public void onCreate(Connection connection) {
        try {
	      connection.createChannel(false).exchangeDeclare(this.exchange.getName(), this.exchange.getType(), this.exchange.isDurable(),
						this.exchange.isAutoDelete(), this.exchange.getArguments());
        }
	catch (Exception e) {

	}
}
```

Won't it be better?

garyrussell: That would work, but the admin reliably closes the channel - with this code you'd need a finally block etc, etc.

That's like doing your own JDBC work outside of a `JdbcTemplate` - I am rather surprised you would suggest something like that :smile:

artembilan: Doh. Yeah... Sorry, have missed the `RabbitTemplate.doExecute()`.

So, let's rename the class and merge!
It does nothing with `log`, so I don't see reason to have such a name far away from the `log` package.

garyrussell:
Yeah - I guess we could move the appenders to a common `log` package and make this package protected, but that would be too disruptive.

Renamed; pushed.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/bd075610f7b649c0d30da97db584144eb4cbf6dd and cherry-picked.

PR #357, opened 2016-01-20
AMQP-562: Check for Queue Declaration Mismatches

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-562

Add an option to the listener container to stop it running
if the queue declarations don't match what's on the broker.

Remove reflection on `ShutdownSignalException` now that the
mininmum client version is 3.4.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/SimpleRabbitListenerContainerFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/SimpleRabbitListenerContainerFactory.java
@@ -23,6 +23,9 @@ import org.aopalliance.aop.Advice;
 import org.springframework.amqp.rabbit.listener.RabbitListenerContainerFactory;
 import org.springframework.amqp.rabbit.listener.SimpleMessageListenerContainer;
 import org.springframework.amqp.support.ConsumerTagStrategy;
+import org.springframework.beans.BeansException;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
 import org.springframework.context.ApplicationEventPublisher;
 import org.springframework.context.ApplicationEventPublisherAware;
 import org.springframework.transaction.PlatformTransactionManager;
@@ -43,7 +46,7 @@ import org.springframework.util.backoff.FixedBackOff;
  */
 public class SimpleRabbitListenerContainerFactory
 		extends AbstractRabbitListenerContainerFactory<SimpleMessageListenerContainer>
-		implements ApplicationEventPublisherAware {
+		implements ApplicationContextAware, ApplicationEventPublisherAware {
 
 	private Executor taskExecutor;
 
@@ -82,6 +85,8 @@ public class SimpleRabbitListenerContainerFactory
 	private Long idleEventInterval;
 
 	private ApplicationEventPublisher applicationEventPublisher;
+
+	private ApplicationContext applicationContext;
 
 	/**
 	 * @param taskExecutor the {@link Executor} to use.
@@ -206,7 +211,6 @@ public class SimpleRabbitListenerContainerFactory
 
 	/**
 	 * @param missingQueuesFatal the missingQueuesFatal to set.
-	 * @since 1.6
 	 * @see SimpleMessageListenerContainer#setMissingQueuesFatal
 	 */
 	public void setMissingQueuesFatal(Boolean missingQueuesFatal) {
@@ -215,6 +219,7 @@ public class SimpleRabbitListenerContainerFactory
 
 	/**
 	 * @param mismatchedQueuesFatal the mismatchedQueuesFatal to set.
+	 * @since 1.6
 	 * @see SimpleMessageListenerContainer#setMismatchedQueuesFatal(boolean)
 	 */
 	public void setMismatchedQueuesFatal(Boolean mismatchedQueuesFatal) {
@@ -242,6 +247,11 @@ public class SimpleRabbitListenerContainerFactory
 	}
 
 	@Override
+	public void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
+		this.applicationContext = applicationContext;
+	}
+
+	@Override
 	protected SimpleMessageListenerContainer createContainerInstance() {
 		return new SimpleMessageListenerContainer();
 	}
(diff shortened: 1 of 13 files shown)


src/reference/asciidoc/amqp.adoc
```diff
@@ -3078,6 +3078,26 @@ This global property will not be applied to any containers that have an explicit
...
+If the problem is detected during recovery (e.g. after a lost connection), the container will be stopped.
+
+NOTE: If the broker is not available during initial startup, the container will start and the conditions will be
+checked when the connection is established.
+
+IMPORTANT: the check is done against all queues in the context, not just the queues that a particular listener
+is configured to use.
+If you wish to limit the checks to just those queues used by a container, you should configure a separate
+`RabbitAdmin` for the container, and provide a reference to it using the `admin` property.
+See <<conditional-declaration>> for more information.
```
artembilan: I like this last paragraph: it answers to my concerns before :smile:

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/SimpleRabbitListenerContainerFactory.java
```diff
@@ -211,6 +213,14 @@ public void setMissingQueuesFatal(Boolean missingQueuesFatal) {
 	}
 
 	/**
+	 * @param mismatchedQueuesFatal the mismatchedQueuesFatal to set.
+	 * @see SimpleMessageListenerContainer#setMismatchedQueuesFatal(boolean)
+	 */
```
artembilan: `@since 1.6` ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitUtils.java
```diff
@@ -208,16 +195,25 @@ public static boolean isExclusiveUseChannelClose(ShutdownSignalException sig) {
 				&& ((AMQP.Channel.Close) shutdownReason).getReplyText().contains("exclusive");
 	}
 
-	public static Object determineShutdownReason(ShutdownSignalException sig) {
-		if (shutDownSignalReasonMethod == null) {
-			return false;
-		}
-		try {
-			return ReflectionUtils.invokeMethod(shutDownSignalReasonMethod, sig);
+	public static boolean isMismatchedQueueArgs(Exception e) {
```
artembilan: How about some JavaDocs for such a public API ?
Yes, I see that many methods here are without them, but we can address that on demand.
Right now I vote for this one :smile: 
I think something like log sample for the `AMQP.PRECONDITION_FAILED` in the JavaDoc would be useful, too.

garyrussell:
Pushed

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/SimpleRabbitListenerContainerFactory.java
```diff
@@ -204,6 +206,7 @@ public void setRecoveryBackOff(BackOff recoveryBackOff) {
 
 	/**
 	 * @param missingQueuesFatal the missingQueuesFatal to set.
+	 * @since 1.6
```
artembilan: Haven't hit the aim :smile: 
Will fix on merge.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitUtils.java
```diff
@@ -199,8 +204,15 @@ public static boolean isPassiveDeclarationChannelClose(ShutdownSignalException s
 					&& ((AMQP.Channel.Close) shutdownReason).getMethodId() == 10); // declare
 	}
 
+	/**
+	 * Return true if the {@link ShutdownSignalException} reason is AMQP.Channel.Close
+	 * and the operation that failed was basicConsumer and the failure text contains
+	 * "exlusive".
```
artembilan: Typo: ex**c**lusive

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -764,6 +776,7 @@ protected void doStart() throws Exception {
 						+ Arrays.asList(queueNames));
 			}
 		}
+		checkMismatchedQueues();
```
artembilan: Let me share my mind here with you!
From the beginning I thought that we should move this call after the `this.rabbitAdmin` initialization. But after that I noticed that we don't call its `afterPropertiesSet()`.
So, according your documentation note about admin injection to restrict declaration by the `declared-by`, it looks like we don't need move it there, because we rely on the shared `RabbitAdmin` bean for the provided `ConnectionFactory`.
The same happens even in the `redeclareElementsIfNecessary()`.

Correct?

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/ContainerInitializationTests.java
```diff
@@ -0,0 +1,128 @@
...
+		@Bean
+		public Queue queue() {
+			return new Queue("test.mismatch", false, false, true); // mismatched
+		}
+
+		@Bean
+		public SimpleMessageListenerContainer container() {
+			SimpleMessageListenerContainer container = new SimpleMessageListenerContainer(connectionFactory());
+			container.setQueueNames("test.mismatch");
+			container.setMessageListener(new MessageListenerAdapter(new Object() {
```
artembilan: What is the purpose of this extra `new Object()`?
Why regular `new MessageListener() {}` isn't enough ?

Minor, of course, but curious.

garyrussell: Yeah - I guess a simple `MessageListener` is all we need.

artembilan:
OK. A couple minor polishing and merging.
And one comment to answer.
LGTM

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
garyrussell: Actually, the code in `onInit()` is a bit dubious.

First it randomly chooses the first admin if there are more than one.

Second, if there are none, it creates one, sets the AC but doesn't call `afterPropertieSet()` - which is likely a good thing since we shouldn't declare all the queues etc in that case.

I will rework that code in another commit.

artembilan: Good. Fingers crossed...

garyrussell: But yes, correct.

garyrussell:
Pushed

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/9ac8aa36ae2de00d3fcfcc7ccc1ab5b00eb09adc

PR #358, opened 2016-01-21
AMQP-563: Support Explicit Temporary Reply Queue

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-563

Allow user to override the use of direct reply-to when the broker supports it and use a
temporary reply queue for request/reply.

__cherry-pick to 1.5.x__ - what's new will need fixup.

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/core/Address.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/core/Address.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2015 the original author or authors.
+ * Copyright 2002-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
  * the License. You may obtain a copy of the License at
@@ -39,13 +39,24 @@ import org.springframework.util.StringUtils;
  */
 public class Address {
 
+	/**
+	 * Use this value in {@code RabbitTemplate#setReplyAddress(String)} to use a temporary reply queue
+	 * for each request, regardless of whether the broker supports direct reply-to.
+	 */
+	public static final String TEMPORARY_REPLY_QUEUE_TOKEN = "##TEMPORARY_QUEUE##";
+
+	/**
+	 * Use this value in {@code RabbitTemplate#setReplyAddress(String)} to explicitly
+	 * indicate that direct reply-to is to be used. When there is no replyAddress, the
+	 * template will use direct reply-to if the broker supports it. See {@link #TEMPORARY_REPLY_QUEUE_TOKEN}.
+	 */
+	public static final String AMQ_RABBITMQ_REPLY_TO = "amq.rabbitmq.reply-to";
+
 	private static final Pattern pattern = Pattern.compile("^(?:.*://)?([^/]*)/?(.*)$");
 
 	private final String exchangeName;
 
 	private final String routingKey;
-
-	public static final String AMQ_RABBITMQ_REPLY_TO = "amq.rabbitmq.reply-to";
 
 	/**
 	 * Create an Address instance from a structured String in the form
--- a/spring-amqp/src/main/java/org/springframework/amqp/core/Address.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/core/Address.java
@@ -40,15 +40,8 @@ import org.springframework.util.StringUtils;
 public class Address {
 
 	/**
-	 * Use this value in {@code RabbitTemplate#setReplyAddress(String)} to use a temporary reply queue
-	 * for each request, regardless of whether the broker supports direct reply-to.
-	 */
-	public static final String TEMPORARY_REPLY_QUEUE_TOKEN = "##TEMPORARY_QUEUE##";
-
-	/**
 	 * Use this value in {@code RabbitTemplate#setReplyAddress(String)} to explicitly
-	 * indicate that direct reply-to is to be used. When there is no replyAddress, the
-	 * template will use direct reply-to if the broker supports it. See {@link #TEMPORARY_REPLY_QUEUE_TOKEN}.
+	 * indicate that direct reply-to is to be used.
 	 */
 	public static final String AMQ_RABBITMQ_REPLY_TO = "amq.rabbitmq.reply-to";
 
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
@@ -569,7 +569,25 @@ public class RabbitTemplate extends RabbitAccessor implements BeanFactoryAware, 
 	}
 
 	private void evaluateFastReplyTo() {
-		this.usingFastReplyTo = false;
+		this.usingFastReplyTo = useDirectReplyTo();
+		this.evaluatedFastReplyTo = true;
+	}
+
+	/**
+	 * Override this method use some other criteria to decide whether or not to use
+	 * direct reply-to (https://www.rabbitmq.com/direct-reply-to.html).
+	 * The default implementation returns true if the broker supports it and there
+	 * is no {@link #setReplyAddress(String) replyAddress} set and the replyAddress
+	 * is not "##TEMPORARY_QUEUE##". When direct reply-to is not used, the template
+	 * will create a temporary, auto-delete queue for the reply.
+	 * <p>
+	 * This method is invoked once only - when the first message is sent.
+	 * @return true to use direct reply-to.
+	 */
+	protected boolean useDirectReplyTo() {
+		if (Address.TEMPORARY_REPLY_QUEUE_TOKEN.equals(this.replyAddress)) {
+			return false;
+		}
 		if (this.replyAddress == null || Address.AMQ_RABBITMQ_REPLY_TO.equals(this.replyAddress)) {
 			try {
 				execute(new ChannelCallback<Void>() {
@@ -580,7 +598,7 @@ public class RabbitTemplate extends RabbitAccessor implements BeanFactoryAware, 
 						return null;
 					}
 				});
-				this.usingFastReplyTo = true;
+				return true;
 			}
 			catch (Exception e) {
 				if (this.replyAddress != null) {
@@ -596,7 +614,7 @@ public class RabbitTemplate extends RabbitAccessor implements BeanFactoryAware, 
 				this.replyAddress = null;
 			}
 		}
-		this.evaluatedFastReplyTo = true;
+		return false;
 	}
 
 	@Override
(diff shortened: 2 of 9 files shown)


artembilan:
I wonder if Travis failure tells anything to you...

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitTemplateIntegrationTests.java
```diff
@@ -1324,7 +1325,17 @@ public void testSendAndReceiveFastExplicit() {
 		sendAndReceiveFastGuts();
 	}
 
+	@Test
+	public void testSendAndReceiveNeverFastt() {
```
artembilan: Typo: double `t` in the end of name.

src/reference/asciidoc/whats-new.adoc
```diff
@@ -36,12 +36,20 @@ See <<containerAttributes>> for more information.
 See <<containerAttributes>> (`autoDeclare`) for some changes to the semantics of that option with respect to the use
 of `RabbitAdmin` s in the application context.
 
-===== AmqpTemple: receive with timeout
+===== AmqpTemplate: receive with timeout
```
artembilan: The `AmqpTemple` was very sacred though :smile:

spring-amqp/src/main/java/org/springframework/amqp/core/Address.java
```diff
@@ -39,14 +39,25 @@
  */
 public class Address {
 
+	/**
+	 * Use this value in {@code RabbitTemplate#setReplyAddress(String)} to use a temporary reply queue
+	 * for each request, regardless of whether the broker supports direct reply-to.
+	 */
+	public static final String TEMPORARY_REPLY_QUEUE_TOKEN = "##TEMPORARY_QUEUE##";
```
tomaszszlek: What do you think about make this as a boolean option accessible on RabbitTemplate interface?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -569,7 +569,25 @@ public void setAfterReceivePostProcessors(MessagePostProcessor... afterReceivePo
...
+	 * direct reply-to (https://www.rabbitmq.com/direct-reply-to.html).
+	 * The default implementation returns true if the broker supports it and there
+	 * is no {@link #setReplyAddress(String) replyAddress} set and the replyAddress
+	 * is not "##TEMPORARY_QUEUE##". When direct reply-to is not used, the template
+	 * will create a temporary, auto-delete queue for the reply.
+	 * <p>
+	 * This method is invoked once only - when the first message is sent.
+	 * @return true to use direct reply-to.
+	 */
+	protected boolean useDirectReplyTo() {
```
tomaszszlek: It looks like this method is not a thread safe, shouldn't we have synchronization here ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -580,7 +598,7 @@ public Void doInRabbit(Channel channel) throws Exception {
 						return null;
 					}
 				});
-				this.usingFastReplyTo = true;
+				return true;
 			}
 			catch (Exception e) {
 				if (this.replyAddress != null) {
```
tomaszszlek: this line will be invoked only if replyAddress is null (condition at line 591) so replyAddress will be always null (assuming that this is thread safe)

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitTemplateIntegrationTests.java
```diff
@@ -1355,9 +1366,15 @@ public Message handleMessage(Message message) {
 			Object result = this.template.convertSendAndReceive("foo");
 			container.stop();
 			assertEquals("FOO", result);
-			assertThat(replyToWas.get(), startsWith(Address.AMQ_RABBITMQ_REPLY_TO));
+			if (tempQueue) {
```
tomaszszlek: Assertion could be moved to testSendAndReceiveNeverFastt test method and we could have a 2 tests one which test with tempQueue = false and one with tempQueue = true.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitTemplateIntegrationTests.java
```diff
@@ -1355,9 +1366,15 @@ public Message handleMessage(Message message) {
...
+				assertThat(replyToWas.get(), not(startsWith(Address.AMQ_RABBITMQ_REPLY_TO)));
+			}
+			else {
+				assertThat(replyToWas.get(), startsWith(Address.AMQ_RABBITMQ_REPLY_TO));
+			}
 		}
 		catch (Exception e) {
+			e.printStackTrace();
```
tomaszszlek: I would rather do not print stacktrace, you can pass exception as a logger parameter

spring-amqp/src/main/java/org/springframework/amqp/core/Address.java
garyrussell: We could, but my aim to was to make this have minimal impact - this solution requires no schema changes (to `<rabbit:template/>`); it also allows us to backport it to 1.5.x. (Our general rule is to not make XML schema changes in maintenance releases).

If we make it a boolean, it would only be available using Java configuration - or you'd have to set it programmatically after the context is initialized (with 1.5). We'd do the schema/parser change in 1.6.

To be honest, I doubt many people will use this (which is why I changed the logic to use direct reply-to when it's available in the first place). I don't really understand the use case for creating a temporary queue for each request when a much more efficient mechanism is available.

Can you explain your use case further? If you can convince me that this should be a first-class option, I'm happy to reconsider.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
garyrussell: It is called from `evaluatedFastReplyTo` which is called from a `synchronized` block.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
garyrussell: No, it's also called if `|| Address.AMQ_RABBITMQ_REPLY_TO.equals(this.replyAddress)`, per line 591.

i.e. if the user explicitly attempted to configure direct reply-to while using a broker that doesn't support it.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitTemplateIntegrationTests.java
garyrussell: Good catch; that's just some junk left over from my testing.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitTemplateIntegrationTests.java
garyrussell: :+1:

garyrussell: But then we'd need to change the other tests that call this.

spring-amqp/src/main/java/org/springframework/amqp/core/Address.java
artembilan: I don't like this option as well.
How about allow to provide `setReplyAddress(null)` and have some option there like:
`replyAddressExplicitlySet = true/false (by default)`. And consult it to determine the direct_reply_to or not?

garyrussell: Null won't work from XML.

OK, enough back pressure - :smile: - I'll make it a boolean (with namespace support in 1.6 only).

artembilan: That's fine with me, too.
Although I don't see reason to backport it to `1.5.x`.
Just a new feature in the new version!

tomaszszlek: I agree with Artem - Let's do it in 1.6 in a clean way so everybody will be happy ;)
Gary, I need such option because I am making RPC call which is processed by some non java app. This app expects temporary queue. Because such queue doesn't exist it throws exception and do not work correctly - with temporary queue everything works. I agree that when you have control over client and server then direct reply-to is the best, efficient solution.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
tomaszszlek: Fine, thanks for explanation.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
tomaszszlek: I didn't notice it, thanks!

spring-amqp/src/main/java/org/springframework/amqp/core/Address.java
garyrussell: Ah, now it makes more sense; thanks.

The reason I was going to backport it is because I assumed you would need it before May (the planned release for 1.6).

If you can work around it with a named queue with 1.5 we can just apply it to 1.6.

But I have no issues putting it in 1.5.4 (sans namespace change) if you want.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitTemplateIntegrationTests.java
tomaszszlek: Understood, I see that big part of code should be changed in that case. I just prefer BDD approach(given, when, then), it looks much cleaner.

spring-amqp/src/main/java/org/springframework/amqp/core/Address.java
tomaszszlek: I did some workaround for the time being so for me it's fine to schedule it for 1.6. Thank you very much that you took under consideration my specific case, appreciate!

garyrussell:
Pushed polishing to address review comments.

artembilan:
I'm fine with the fix and can merge.

@tomaszszlek , any objection?
We appreciate your feedback!

tomaszszlek:
Artem, looks good to me. Thank you both !

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/cd0a0df50d53cf3d7a8a185c7150783781a6149c

PR #359, opened 2016-01-22
AMQP-535: PublisherCallbackChannel Improvements

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-535

Internal state from the `PublisherCallbackChannel` (pending confirms) leaked
into the `RabbitTemplate` for the sole purpose of supporting `getUnconfirmed()`
which expires old entries and returns them to the caller.

Encapsulate this state within the channel and do not expose externally.

- Change `addListener` to return `void`
- Delete the unused `removeListener()` method from `PublisherCallbackChannel`
- Add `expire()` to `PublisherCallbackChannel`
- Change `getUnconfirmed` to invoke `expire` on each channel
- Change `PublisherCallbackChannel.Listener.removePendingConfirmsReference` to `...revoke()`
  (used to tell the Listener that it will receive no more confirms or returns from this channel
- Fix a problem in the `CachingConnectionFactory` where the pending confirms are not nacked
  if the underlying channel was detected as closed in `logicalClose()`

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2015 the original author or authors.
+ * Copyright 2002-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
  * the License. You may obtain a copy of the License at
@@ -801,6 +801,9 @@ public class CachingConnectionFactory extends AbstractConnectionFactory
 			if (this.target != null && !this.target.isOpen()) {
 				synchronized (targetMonitor) {
 					if (this.target != null && !this.target.isOpen()) {
+						if (this.target instanceof PublisherCallbackChannel) {
+							this.target.close(); // emit nacks if necessary
+						}
 						if (this.channelList.contains(proxy)) {
 							this.channelList.remove(proxy);
 						}
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
@@ -19,15 +19,13 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.Map;
-import java.util.Map.Entry;
 import java.util.Set;
-import java.util.SortedMap;
 import java.util.UUID;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -139,8 +137,8 @@ public class RabbitTemplate extends RabbitAccessor implements BeanFactoryAware, 
 
 	private static final String DEFAULT_ENCODING = "UTF-8";
 
-	private final ConcurrentHashMap<Object, SortedMap<Long, PendingConfirm>> pendingConfirms =
-			new ConcurrentHashMap<Object, SortedMap<Long, PendingConfirm>>();
+	private final ConcurrentMap<Channel, RabbitTemplate> publisherConfirmChannels =
+			new ConcurrentHashMap<Channel, RabbitTemplate>();
 
 	private final Map<String, PendingReply> replyHolder = new ConcurrentHashMap<String, PendingReply>();
 
@@ -541,27 +539,16 @@ public class RabbitTemplate extends RabbitAccessor implements BeanFactoryAware, 
 	 * Gets unconfirmed correlation data older than age and removes them.
 	 * @param age in milliseconds
 	 * @return the collection of correlation data for which confirms have
-	 * not been received.
+	 * not been received or null if no such confirms exist.
 	 */
 	public Collection<CorrelationData> getUnconfirmed(long age) {
 		Set<CorrelationData> unconfirmed = new HashSet<CorrelationData>();
-		synchronized (this.pendingConfirms) {
-			long threshold = System.currentTimeMillis() - age;
-			for (Entry<Object, SortedMap<Long, PendingConfirm>> channelPendingConfirmEntry : this.pendingConfirms.entrySet()) {
-				SortedMap<Long, PendingConfirm> channelPendingConfirms = channelPendingConfirmEntry.getValue();
-				synchronized(channelPendingConfirmEntry.getKey()) { // channel
-					Iterator<Entry<Long, PendingConfirm>> iterator = channelPendingConfirms.entrySet().iterator();
-					PendingConfirm pendingConfirm;
-					while (iterator.hasNext()) {
-						pendingConfirm = iterator.next().getValue();
-						if (pendingConfirm.getTimestamp() < threshold) {
-							unconfirmed.add(pendingConfirm.getCorrelationData());
-							iterator.remove();
-						}
-						else {
-							break;
-						}
-					}
+		synchronized (this.publisherConfirmChannels) {
+			long cutoffTime = System.currentTimeMillis() - age;
+			for (Channel channel : this.publisherConfirmChannels.keySet()) {
+				Collection<PendingConfirm> confirms = ((PublisherCallbackChannel) channel).expire(this, cutoffTime);
+				for (PendingConfirm confirm : confirms) {
+					unconfirmed.add(confirm.getCorrelationData());
 				}
 			}
 		}
@@ -1461,11 +1448,12 @@ public class RabbitTemplate extends RabbitAccessor implements BeanFactoryAware, 
 	private void addListener(Channel channel) {
 		if (channel instanceof PublisherCallbackChannel) {
 			PublisherCallbackChannel publisherCallbackChannel = (PublisherCallbackChannel) channel;
-			SortedMap<Long, PendingConfirm> pendingConfirms = publisherCallbackChannel.addListener(this);
 			Channel key = channel instanceof ChannelProxy ? ((ChannelProxy) channel).getTargetChannel() : channel;
-			if (this.pendingConfirms.putIfAbsent(key, pendingConfirms) == null
-					&& logger.isDebugEnabled()) {
-				logger.debug("Added pending confirms for " + channel + " to map, size now " + this.pendingConfirms.size());
+			if (this.publisherConfirmChannels.putIfAbsent(key, this) == null) {
+				publisherCallbackChannel.addListener(this);
+				if (logger.isDebugEnabled()) {
+					logger.debug("Added pubsub channel: " + channel + " to map, size now " + this.publisherConfirmChannels.size());
+				}
 			}
 		}
 		else {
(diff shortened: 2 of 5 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -1461,11 +1448,12 @@ private Address getReplyToAddress(Message request) throws AmqpException {
 	private void addListener(Channel channel) {
 		if (channel instanceof PublisherCallbackChannel) {
 			PublisherCallbackChannel publisherCallbackChannel = (PublisherCallbackChannel) channel;
-			SortedMap<Long, PendingConfirm> pendingConfirms = publisherCallbackChannel.addListener(this);
 			Channel key = channel instanceof ChannelProxy ? ((ChannelProxy) channel).getTargetChannel() : channel;
-			if (this.pendingConfirms.putIfAbsent(key, pendingConfirms) == null
-					&& logger.isDebugEnabled()) {
-				logger.debug("Added pending confirms for " + channel + " to map, size now " + this.pendingConfirms.size());
+			if (this.publisherConfirmChannels.putIfAbsent(key, this) == null) {
```
artembilan: Can't we use `Collections.newSetFromMap()` instead ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/PublisherCallbackChannel.java
```diff
@@ -34,26 +34,19 @@
...
-	 * The client must <b>NOT</b> modify the contents of
-	 * this array, and must synchronize on it when
-	 * iterating over its collections.
-	 *
+	 * Adds a {@link Listener}
 	 * @param listener The Listener.
-	 * @return A reference to pending confirms for the listener
 	 */
-	SortedMap<Long, PendingConfirm> addListener(Listener listener);
+	void addListener(Listener listener);
```
artembilan: M-m-m. Seems for me Andy Wilkinson won't be happy to have such a braking change in the minor release...
Although we may hope that no one has tests for this `PublisherCallbackChannel` :smile:

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
garyrussell: I looked at that, but we don't get a `putIfAbsent()`.

artembilan: ??? Do we need it?
We don't care here about the old value (value is always `this` BTW), so, simple `SetFromMap`:
```
public boolean add(E e) { return m.put(e, Boolean.TRUE) == null; }
```
is really enough for us.

Am I missing anything?

Although I agree: it is our internals and it will work the same way as right now. And it is always is based on the `this`, so no memory and GC impact as well.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/PublisherCallbackChannelImpl.java
```diff
@@ -691,32 +687,33 @@ private synchronized void generateNacksForPendingAcks(String cause) {
...
+		if (pendingConfirmsForListener == null) {
+			return Collections.<PendingConfirm>emptyList();
 		}
-		Iterator<Entry<Long, Listener>> iterator = this.listenerForSeq.entrySet().iterator();
-		while (iterator.hasNext()) {
-			Entry<Long, Listener> entry = iterator.next();
-			if (entry.getValue() == listener) {
-				iterator.remove();
+		else {
+			synchronized (pendingConfirmsForListener) {
```
artembilan: Can you explain this?
It always has been strange for me to synchronize on the local variable.
But here it is more suspicious because the whole method is `synchronized`.
As well as many other methods here which deal with the same `this.pendingConfirms.get(listener)`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/PublisherCallbackChannelImpl.java
```diff
@@ -691,32 +687,33 @@ private synchronized void generateNacksForPendingAcks(String cause) {
...
+				List<PendingConfirm> expired = new ArrayList<PendingConfirm>();
+				Iterator<Entry<Long, PendingConfirm>> iterator = pendingConfirmsForListener.entrySet().iterator();
+				while (iterator.hasNext()) {
+					PendingConfirm pendingConfirm = iterator.next().getValue();
+					if (pendingConfirm.getTimestamp() < cutoffTime) {
+						expired.add(pendingConfirm);
+						iterator.remove();
+					}
+					else {
+						break;
```
artembilan: Don't we need to remove *all* expired pendingConfirms?
```
* Expire (remove) any {@link PendingConfirm}s created before cutoffTime for the
* supplied listener and return them to the caller.
* @param listener the listener.
* @param cutoffTime the time before which expired messages were created.
* @return the list of expired confirms.
```

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/PublisherCallbackChannelImpl.java
```diff
@@ -658,7 +659,7 @@ private synchronized void generateNacksForPendingAcks(String cause) {
 				processAck(confirmEntry.getKey(), false, false, false);
 				iterator.remove();
```
artembilan: Not related to the PR, but this `remove()` is redundant: we `this.pendingConfirms.clear();` in the end of the method anyway.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/PublisherCallbackChannelImpl.java
```diff
@@ -691,32 +687,33 @@ private synchronized void generateNacksForPendingAcks(String cause) {
 				logger.debug("Added listener " + listener);
```
artembilan: The code not shown here on GH because of without changes, but:
```
if (!this.listeners.values().contains(listener)){
	this.listeners.put(listener.getUUID(), listener);
```
So, listener can change its `UUID` at runtime?
Why `puIfAbsent` isn't good here?
At least by the key hash nature we get a better performance.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
```diff
@@ -801,6 +801,9 @@ private void logicalClose(ChannelProxy proxy) throws Exception {
 			if (this.target != null && !this.target.isOpen()) {
 				synchronized (targetMonitor) {
 					if (this.target != null && !this.target.isOpen()) {
+						if (this.target instanceof PublisherCallbackChannel) {
+							this.target.close(); // emit nacks if necessary
```
artembilan: Isn't it a bug and we should consider to backport it?

artembilan:
That's all for now.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
garyrussell: I want to avoid calling `channel.addListener()` on every send (which is what we do now). It has protection against multiple calls, but it's synchronized. This way, we avoid that overhead.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/PublisherCallbackChannelImpl.java
garyrussell: We're not synchronized on the local variable, we are synchronized on the `ConcurrentHashmap` it holds a reference to. 

However, you are correct; I was thinking of `SynchronizedMap` where you have to synchronize while iterating.

It is not necessary to synchronize while iterating over a `ConcurrentHashMap`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/PublisherCallbackChannelImpl.java
garyrussell: ? that's what it's doing

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/PublisherCallbackChannelImpl.java
garyrussell: Oops, I should not have removed the `synchronized` here. But, yes, I can use putIfAbsent() instead - much better.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/PublisherCallbackChannelImpl.java
artembilan: Right, but we return from the loop exactly on the **first** unexpired (fresh) `pendingConfirm` and don't iterate to the end of the collection.
Or does that mean that the first fresh `pendingConfirm` assumes that all other are more fresh and that it is signal for us really to return ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
garyrussell: Yes, this should be backported.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/PublisherCallbackChannelImpl.java
garyrussell: Yes, it's a `SortedMap` keyed by the sequence number which is incremented by the rabbit client on each send.

See `RabbitTemplate.setupConfirm()`.

So, as soon as we hit a sequence that's not expired, the rest won't be either.

We have to keep them sorted because we might get a confirm that spans multiple sequence numbers (and listeners). See `processAck()`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
artembilan: Good. Can I just push this two lines of code into `1.5.x`?
Or is there anything to test against?

garyrussell: 3 lines :smile: 

I can try to come up with a test if you prefer.

artembilan: M-m-m. From one side I have asked about the backport for exactly these 3 lines.
But from other we always try to keep each change covered with test, especially if it is a bug.
Like this?

Anyway you have came here somehow to find this and fix

garyrussell: Yes; I couldn't figure out why a test was failing with the changes I made and I found it was because an `isOpen()` returned false here instead of above.

I should really try to understand what changed. I'll take a look next week (if my old body survives shoveling 600mm + snow :frowning: ).

artembilan:
@garyrussell , looks like something else should be fixed.
See Travis results.

garyrussell:
Pushed.

I also figured out what changed that exposed the problem that I fixed in CCF - since addListener is no longer called after the first time, it changed when the `isOpen()` returned false in the test and (luckily) exposed the bug.

I will prepare a fix for backporting and create a new PR for that.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/PublisherCallbackChannelImpl.java
```diff
@@ -670,53 +668,46 @@ private synchronized void generateNacksForPendingAcks(String cause) {
...
 		Assert.notNull(listener, "Listener cannot be null");
 		if (this.listeners.size() == 0) {
 			this.delegate.addConfirmListener(this);
 			this.delegate.addReturnListener(this);
 		}
-		if (!this.listeners.values().contains(listener)){
-			this.listeners.put(listener.getUUID(), listener);
-			this.pendingConfirms.put(listener, Collections.synchronizedSortedMap(new TreeMap<Long, PendingConfirm>()));
+		if (this.listeners.putIfAbsent(listener.getUUID(), listener) == null) {
+			this.pendingConfirms.put(listener, new ConcurrentSkipListMap<Long, PendingConfirm>());
```
artembilan: I can't still understand (find) why they use 4 index levels https://en.wikipedia.org/wiki/Skip_list, but I'm fine with the fix. 
Merging...

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/880c076ce1b25cbabe2702ea90a67e81df321b4b.
Also fixed the test since the last nightly CI: https://build.spring.io/browse/AMQP-MEIGHT-595

PR #360, opened 2016-01-22
AMQP-559: @RabbitListener Test Harness

garyrussell:
Captures method arguments and results (when non-null return) in blocking queues to facilite
validation in tests.

Overrides the BPP and proxies the listener instance with CGLIB (listener methods must not be final).


garyrussell:
PoC only, for review - needs docs.

spring-rabbit-test/src/test/java/org/springframework/amqp/rabbit/test/ExampleListenerIntegrationTest.java
```diff
@@ -0,0 +1,159 @@
...
+ * @since 1.6
+ *
+ */
+@ContextConfiguration
+@RunWith(SpringJUnit4ClassRunner.class)
+@DirtiesContext
+public class ExampleListenerIntegrationTest {
+
+	@Rule
+	public BrokerRunning brokerRunning = BrokerRunning.isRunning();
```
artembilan: There is no `import` for this.
From other it won't be at all because there is no:
```
testCompile project(":spring-rabbit").sourceSets.test.output
```
dependency in the Gradle config for this project.

spring-rabbit-test/src/test/java/org/springframework/amqp/rabbit/test/ExampleListenerIntegrationTest.java
```diff
@@ -0,0 +1,159 @@
...
+
+		BlockingQueue<Object[]> argumentQueue = this.harness.getArgumentQueueFor("bar");
+		assertNotNull(argumentQueue);
+		Object[] args = argumentQueue.poll(10, TimeUnit.SECONDS);
+		assertThat((String) args[0], equalTo("bar"));
+		assertThat((String) args[1], equalTo(queue2.getName()));
+
+		args = argumentQueue.poll(10, TimeUnit.SECONDS);
+		assertThat((String) args[0], equalTo("baz"));
+		assertThat((String) args[1], equalTo(queue2.getName()));	}
```
artembilan: The reformat code will be enough to fix this.

spring-rabbit-test/src/test/java/org/springframework/amqp/rabbit/test/ExampleListenerIntegrationTest.java
```diff
@@ -0,0 +1,159 @@
...
+
+	@Configuration
+	@EnableRabbit
+	@RabbitListenerIntegrationTest
```
artembilan: Wouldn't it be enough to place `@EnableRabbit` on the `@RabbitListenerIntegrationTest`?
Then the last one has a bad name...
`@EnableTestRabbit` ?

Maybe we will be able to provide more testing stuff in the future, not only this listener proxying?..

garyrussell: >Maybe ...

Yes - that's my goal - I already started writing a `MockConnectionFactory` - but it's quite a bit more work than I thought it would be. Maybe over time...

spring-rabbit-test/src/test/java/org/springframework/amqp/rabbit/test/ExampleListenerIntegrationTest.java
garyrussell: Oh; good catch - since this is intended for user testing, we can't do that - we'll have to move it to src/main (or create a fourth jar).

artembilan: I'd say we don't have choice, unless copy/paste the `BrokerRunning` here `spring-rabbit-test` under `src/main`.

spring-rabbit-test/src/main/java/org/springframework/amqp/rabbit/test/RabbitListenerIntegrationTest.java
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright 2015 the original author or authors.
```
artembilan: IDEA has option to reflect the current year here automatically :smile:

spring-rabbit-test/src/main/java/org/springframework/amqp/rabbit/test/RabbitListenerTestBootstrap.java
```diff
@@ -0,0 +1,41 @@
...
+ * @since 1.6
+ *
+ */
+@Configuration
+public class RabbitListenerTestBootstrap {
+
+	@Bean(name = RabbitListenerConfigUtils.RABBIT_LISTENER_ANNOTATION_PROCESSOR_BEAN_NAME)
+	@Role(BeanDefinition.ROLE_INFRASTRUCTURE)
+	public static RabbitListenerAnnotationBeanPostProcessor rabbitListenerAnnotationProcessor() {
+		return new RabbitListenerTestHarness();
```
artembilan: Funny word, but I see it standard terminology: https://en.wikipedia.org/wiki/Test_harness :smile:

spring-rabbit-test/src/main/java/org/springframework/amqp/rabbit/test/RabbitListenerIntegrationTest.java
garyrussell: I wrote this last year - see the comment in the JIRA.

spring-rabbit-test/src/test/java/org/springframework/amqp/rabbit/test/ExampleListenerIntegrationTest.java
```diff
@@ -0,0 +1,159 @@
...
+	private RabbitListenerTestHarness harness;
+
+	@Test
+	public void testTwoWay() throws Exception {
+		assertEquals("FOO", this.rabbitTemplate.convertSendAndReceive(this.queue1.getName(), "foo"));
+
+		BlockingQueue<Object[]> argumentQueue = this.harness.getArgumentQueueFor("foo");
+		assertNotNull(argumentQueue);
+		Object[] args = argumentQueue.poll(10, TimeUnit.SECONDS);
+		assertThat((String) args[0], equalTo("foo"));
```
artembilan: Yeah... I'm not well with all those testing frameworks and if you really think that the `queue` is better than `ArgumentCaptor`, `verify` etc., I can live with this and you are free to go ahead with the PR stuff.

garyrussell:
Added a `spy` version. In some ways, the first version is better because you don't need a latch for one-way tests.

However, we can offer both; the other can also be used when spy won't work (e.g. final methods).

build.gradle
```diff
@@ -228,6 +228,26 @@ project('spring-rabbit') {
...
+
+		compile project(":spring-rabbit")
+		compile ("junit:junit:$junitVersion") {
+			exclude group: 'org.hamcrest'
+		}
+		compile "org.hamcrest:hamcrest-all:$hamcrestVersion"
+		compile ("org.mockito:mockito-core:$mockitoVersion") {
+			exclude group: 'org.hamcrest'
+		}
+		testCompile project(":spring-rabbit").sourceSets.test.output
```
artembilan: So, you decided do not copy/paste `BrokerRunning` and therefore do not expose it as a high-level testing util.

spring-rabbit-test/src/main/java/org/springframework/amqp/rabbit/test/RabbitListenerAnnotationSpyBeanPostProcessor.java
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.amqp.rabbit.test;
+
+import static org.mockito.Mockito.spy;
```
artembilan: Even if it is testing framework I still think that we shouldn't use `import static` in the maven artifact.

spring-rabbit-test/src/main/java/org/springframework/amqp/rabbit/test/RabbitListenerSpyTest.java
```diff
@@ -0,0 +1,42 @@
...
+ * @author Gary Russell
+ * @since 1.6
+ *
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@EnableRabbit
+@Import(RabbitListenerSpyTestBootstrap.class)
+public @interface RabbitListenerSpyTest {
```
artembilan: This might be like a `spy = true/false` option in the `RabbitListenerTest`.
Just my five fingers.

spring-rabbit-test/src/test/java/org/springframework/amqp/rabbit/test/ExampleRabbitListenerTest.java
```diff
@@ -0,0 +1,158 @@
...
+	public void testTwoWay() throws Exception {
+		assertEquals("FOO", this.rabbitTemplate.convertSendAndReceive(this.queue1.getName(), "foo"));
+
+		BlockingQueue<Object[]> argumentQueue = this.harness.getArgumentQueueFor("foo");
+		assertNotNull(argumentQueue);
+		Object[] args = argumentQueue.poll(10, TimeUnit.SECONDS);
+		assertThat((String) args[0], equalTo("foo"));
+		BlockingQueue<Object> resultQueue = this.harness.getResultQueueFor("foo");
+		assertNotNull(resultQueue);
+		Object result = resultQueue.poll(10, TimeUnit.SECONDS);
```
artembilan: Wouldn't it be more friendlier to have something like:
```
ArgumentsAndResult argumentsAndResult = this.harness.getArgumentsAndResultFor("foo");
ArgumentsAndResult argumentsAndResult = this.harness.getArgumentsAndResultFor("foo", 10, TimeUnit.SECONDS);
```
?
To avoid the `BlockingQueue` noise.

spring-rabbit-test/src/main/java/org/springframework/amqp/rabbit/test/RabbitListenerTestHarness.java
```diff
@@ -0,0 +1,113 @@
...
+		public Object invoke(MethodInvocation invocation) throws Throwable {
+			Object[] args = invocation.getArguments();
+			if (args.length > 0) {
+				arguments.add(args);
+			}
+			Object result = invocation.proceed();
+			if (args.length > 0 && result != null) {
+				this.returns.put(result);
+			}
+			return result;
```
artembilan: Continuing the `ArgumentsAndResult` discussion, I think here we should be robust and even don't expose arguments until the `invocation.proceed()` returns.
Anyway there should be some hook to expect an `Exception`- `try...catch` around `invocation.proceed()`.

artembilan:
That's all for now.
I think I will be fine with the solution, so you can go ahead with anything what you have in mind on the matter.
Of course, I don't mean bypassing the farther nit-pick review :wink:

spring-rabbit-test/src/main/java/org/springframework/amqp/rabbit/test/RabbitListenerSpyTest.java
garyrussell: That's what I did initially but the bootstrap class doesn't have access to the annotation.

If there's another way, I'm open to it.

build.gradle
garyrussell: Yes, at least for now; we can revisit later.

spring-rabbit-test/src/main/java/org/springframework/amqp/rabbit/test/RabbitListenerSpyTest.java
artembilan: `ImportAware`?
See `EnableIntegrationManagement`. 
Or am I missing anything else here ?

garyrussell:
Thanks for the tip on `ImportAware` - pushed.

I still need to do the other suggestions you made, (which are good).

garyrussell:
Pushed; let me know if you have further suggestions before I start the docs.

spring-rabbit-test/src/main/java/org/springframework/amqp/rabbit/test/RabbitListenerTestHarness.java
```diff
@@ -0,0 +1,176 @@
...
+			Object adminTarget, String beanName) {
+		String id = rabbitListener.id();
+		if (StringUtils.hasText(id)) {
+			if (this.attributes.getBoolean("spy")) {
+				Object spy = Mockito.spy(bean);
+				this.listeners.put(id, spy);
+				super.processListener(endpoint, rabbitListener, spy, adminTarget, beanName);
+				return;
+			}
+			else {
```
artembilan: I would like to have them together, not `if...else`.
The `spy` option assumes that we add or not the `.spy()` around our listeners.

Right the `ProxyFactoryBean` and `InvocationData` can be achieved with the `Spy` (somehow) as well, but why just don't expose it as is right now and wrap to `spy` if that?

garyrussell: That makes sense; the only issue I see is the memory usage if they choose not to consume the `InvocationData`. But I guess that's not going to be excessive in a test case.

spring-rabbit-test/src/main/java/org/springframework/amqp/rabbit/test/RabbitListenerTestHarness.java
```diff
@@ -0,0 +1,176 @@
...
+		if (advice != null) {
+			return advice.invocationData.poll(wait, unit);
+		}
+		return null;
+	}
+
+	@SuppressWarnings("unchecked")
+	<T> T getSpy(String id) {
+		return (T) this.listeners.get(id);
+	}
```
artembilan: How about to override `postProcessAfterInitialization()` and return `spy/proxy` directly instead of `bean`?

With that we still get a gain of `@Autowired MyListener listener;` and no need to invoke such a method from test.

As well as we could register `CaptureAdvice` (or anything else) as a bean for the listener.

I don't mind the current solution, just some brainstorm ideas.

spring-rabbit-test/src/main/java/org/springframework/amqp/rabbit/test/RabbitListenerTestBootstrap.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.amqp.rabbit.config.RabbitListenerConfigUtils;
+import org.springframework.beans.factory.config.BeanDefinition;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.ImportAware;
+import org.springframework.context.annotation.Role;
+import org.springframework.core.type.AnnotationMetadata;
+
+/**
+ * Overrides the default BPP with a {@link RabbitListenerTestHarness}.
```
artembilan: The JavaDoc should be improved because of public maven-aware code.

spring-rabbit-test/src/main/java/org/springframework/amqp/rabbit/test/RabbitListenerTestHarness.java
garyrussell: Agreed - that way we can spy listeners that don't have an `id`.

spring-rabbit-test/src/test/java/org/springframework/amqp/rabbit/test/ExampleRabbitListenerSpyTest.java
```diff
@@ -0,0 +1,160 @@
...
+		@Bean
+		public RabbitTemplate template(ConnectionFactory cf) {
+			return new RabbitTemplate(cf);
+		}
+
+		@Bean
+		public SimpleRabbitListenerContainerFactory rabbitListenerContainerFactory(ConnectionFactory cf) {
+			SimpleRabbitListenerContainerFactory containerFactory = new SimpleRabbitListenerContainerFactory();
+			containerFactory.setConnectionFactory(cf);
+			return containerFactory;
```
artembilan: I guess we can't register such a bean by default because of the required `ConnectionFactory`, which should be injected somehow.
But let's introduce the ctor at least to simplify the `SimpleRabbitListenerContainerFactory` `@Bean` definition!

Yes, isn't related to the current PR...

artembilan:
That's all for now.
No more ideas yet...

spring-rabbit-test/src/main/java/org/springframework/amqp/rabbit/test/RabbitListenerTestHarness.java
garyrussell: You can't spy a CGLIB proxy (or vice versa). NPEs all over the place.

artembilan: Uh! OK. 
Thank you for new knowledges!

garyrussell:
Pushed the change to support both spy and capture advice.

I got all kinds of NPEs when moving to after initialization (null `connectionMonitor` in the CF - presumably some side-effect of Mockito spy. Mockito does say you shouldn't use the spied object directly.

So I'm inclined to not spend more time. Any more thoughts?

spring-rabbit-test/src/main/java/org/springframework/amqp/rabbit/test/RabbitListenerTestHarness.java
```diff
@@ -0,0 +1,175 @@
...
+	protected void processListener(MethodRabbitListenerEndpoint endpoint, RabbitListener rabbitListener, Object bean,
+			Object adminTarget, String beanName) {
+		Object proxy = bean;
+		String id = rabbitListener.id();
+		if (StringUtils.hasText(id)) {
+			if (this.attributes.getBoolean("spy")) {
+				proxy = Mockito.spy(proxy);
+				this.listeners.put(id, proxy);
+				super.processListener(endpoint, rabbitListener, proxy, adminTarget, beanName);
+				return;
```
artembilan: But you `return` in case of this `if`, so any further `pfb.setTarget(proxy);` isn't reachable for `spy`.

garyrussell: doh

garyrussell:
Pushed

build.gradle
```diff
@@ -228,6 +228,32 @@ project('spring-rabbit') {
...
+		testCompile "log4j:log4j:$log4jVersion"
```
artembilan: This is redundant. See `// dependencies that are common across all java projects`.
But yes, it is minor, of course.

As you said you would like do not spend more time on this and push it to public and get some feedback to go ahead.
So, feel free to document it and I'll review and merge.

garyrussell:
Pushed

garyrussell:
Doc fix.

src/reference/asciidoc/testing.adoc
```diff
@@ -0,0 +1,279 @@
+[[testing]]
+=== Testing Support
+
+==== Introduction
+
+Writing integration for asynchronous applications is necessarily more complex than testing simpler applications.
+This is made more complex when abstractions such as the `@RabbitListner` annotations come into the picture.
```
artembilan: Typo: RabbitList**e**ner.
Fixing on merge.

src/reference/asciidoc/testing.adoc
```diff
@@ -0,0 +1,279 @@
...
+The framework itself has many unit and integration tests; some using mocks, others using integration testing with
+a live RabbitMQ broker.
+You can consult those tests for some ideas for testing scenarios.
+
+Spring AMQP _version 1.6_ introduced the `spring-rabbit-test` jar which provides support for testing some of these more
+complex scenarios.
+It is anticipated that this project will expand over time but we need community feedback to make suggestions for
+features needed to help with testing.
+Please use https://jira.spring.io/browse/AMQP[JIRA] or
+https://github.com/spring-projects/spring-amqp/issues[GitHub Issues] to provide such feedback.
```
artembilan: Hey! You have opened the `GH issues` :smile: 
How about the SI ?

src/reference/asciidoc/testing.adoc
```diff
@@ -0,0 +1,279 @@
...
+
+==== @RabbitListenerTest and RabbitListenerTestHarness
+
+Annotating one of your `@Configuration` classes with `@RabbitListenerTest` will cause the framework to replace the
+standard `RabbitListenerAnnotationBeanPostProcessor` with a subclass `RabbitListenerTestHarness` (it will also enable
+`@RabbitListener` detection via `@EnableRabbit`).
+
+The `RabbitListenerTestHarness` enhances the listener in two ways - it wraps it in a `Mockito Spy`, enabling normal
+`Mockito` stubbing and verification operations.
+It can also add an advice to the listener enabling access to to the arguments, result and or exceptions thrown.
```
artembilan: tautology: to to
:smile:

artembilan:
Fixing all Copyrights on merge...

src/reference/asciidoc/testing.adoc
garyrussell: Done

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/f9ad86b9de8fc04056c2989063318077cf8f6937

PR #361, opened 2016-01-25
AMQP-564: Fix Confirm Callback Nacks

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-564

Nacks for pending confirms were not generated if the connection was
detected as closed in logicalClose.

__cherry-pick to master - new tests and comment in CCF__

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2015 the original author or authors.
+ * Copyright 2002-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
  * the License. You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


garyrussell:
When cherry-picking to master, change

     * For a cached channel an additional call is made in getChannel().

to 

     * For a cached channel an additional call is made in getChannel() but addListener() is not called.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
```diff
@@ -386,7 +386,11 @@ private Channel getChannel(ChannelCachingConnectionProxy connection, boolean tra
 						try {
 							Channel target = channel.getTargetChannel();
 							if (target != null) {
-								target.close(); // to remove it from auto-recovery if so configured
+								target.close();
+								/*
+								 *  To remove it from auto-recovery if so configured,
+								 *  and nack any pending confirms if PublisherCallbackChannel.
+								 */
```
artembilan: I'd move this comment before the line of code, but that doesn't hurt, I think.

artembilan:
... and cherry-picked as https://github.com/spring-projects/spring-amqp/commit/ea4c42e2f44aeb01177961e44f29e02b1885ad39 after the conflicts fix.

PR #362, opened 2016-01-26
AMQP-539: Add AsyncRabbitTemplate

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-539

Async `sendAndReceive` operations returning a `ListenableFuture`.

__Initial commit for review - needs docs.__

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
@@ -44,7 +44,9 @@ import org.springframework.util.concurrent.SettableListenableFuture;
  * allowing the caller to obtain the reply later, using {@code get()} or a callback.
  * <p>
  * When confirms are enabled, the future has a confirm property which is itself a
- * {@link ListenableFuture}.
+ * {@link ListenableFuture}. If the reply is received before the publisher confirm,
+ * the confirm is discarded since the reply implicitly indicates the message was
+ * published.
  * <p>
  * Returned (undeliverable) request messages are presented as a
  * {@link AmqpMessageReturnedException} cause of an {@link ExecutionException}.
@@ -204,7 +206,9 @@ public class AsyncRabbitTemplate<C> implements SmartLifecycle, MessageListener, 
 	 */
 	public void setEnableConfirms(boolean enableConfirms) {
 		this.enableConfirms = enableConfirms;
-		this.template.setConfirmCallback(this);
+		if (enableConfirms) {
+			this.template.setConfirmCallback(this);
+		}
 	}
 
 	/**
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
@@ -340,7 +340,7 @@ public class AsyncRabbitTemplate<C> implements SmartLifecycle, MessageListener, 
 			MessagePostProcessor messagePostProcessor) throws AmqpException {
 		CorrelationData correlationData = null;
 		if (this.enableConfirms) {
-			correlationData = new MutableCorrelationData(null);
+			correlationData = new CorrelationData(null);
 		}
 		CorrelationMessagePostProcessor correlationPostProcessor = new CorrelationMessagePostProcessor(
 				messagePostProcessor, correlationData);
@@ -566,8 +566,8 @@ public class AsyncRabbitTemplate<C> implements SmartLifecycle, MessageListener, 
 			}
 			String correlationId = getOrSetCorrelationIdAndSetReplyTo(messageToSend);
 			this.future = new RabbitConverterFuture(correlationId);
-			if (this.correlationData instanceof MutableCorrelationData) {
-				((MutableCorrelationData) correlationData).setId(correlationId);
+			if (this.correlationData != null && this.correlationData.getId() == null) {
+				this.correlationData.setId(correlationId);
 				future.setConfirm(new SettableListenableFuture<Boolean>());
 			}
 			AsyncRabbitTemplate.this.pending.put(correlationId, future);
@@ -576,22 +576,6 @@ public class AsyncRabbitTemplate<C> implements SmartLifecycle, MessageListener, 
 
 		private RabbitConverterFuture getFuture() {
 			return this.future;
-		}
-
-	}
-
-	/**
-	 * This class exists simply to increase the visibility of the id setter.
-	 */
-	private static final class MutableCorrelationData extends CorrelationData {
-
-		public MutableCorrelationData(String id) {
-			super(id);
-		}
-
-		@Override
-		protected void setId(String id) {
-			super.setId(id);
 		}
 
 	}
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
@@ -29,6 +29,7 @@ import org.springframework.amqp.core.AmqpMessageReturnedException;
 import org.springframework.amqp.core.Message;
 import org.springframework.amqp.core.MessageListener;
 import org.springframework.amqp.core.MessagePostProcessor;
+import org.springframework.amqp.core.MessageProperties;
 import org.springframework.amqp.rabbit.connection.ConnectionFactory;
 import org.springframework.amqp.rabbit.core.RabbitTemplate.ConfirmCallback;
 import org.springframework.amqp.rabbit.core.RabbitTemplate.ReturnCallback;
@@ -387,19 +388,22 @@ public class AsyncRabbitTemplate<C> implements SmartLifecycle, MessageListener, 
 	@SuppressWarnings("unchecked")
 	@Override
 	public void onMessage(Message message) {
-		String correlationId = new String(message.getMessageProperties().getCorrelationId(), this.charset);
-		if (correlationId != null) {
-			if (logger.isDebugEnabled()) {
-				logger.debug("onMessage: " + message);
-			}
-			RabbitFuture<?> future = this.pending.remove(correlationId);
-			if (future != null) {
-				if (future instanceof AsyncRabbitTemplate.RabbitConverterFuture) {
-					C converted = (C) this.template.getMessageConverter().fromMessage(message);
-					((RabbitConverterFuture) future).set(converted);
+		MessageProperties messageProperties = message.getMessageProperties();
+		if (messageProperties != null) {
+			byte[] correlationId = messageProperties.getCorrelationId();
+			if (correlationId != null) {
+				if (logger.isDebugEnabled()) {
+					logger.debug("onMessage: " + message);
 				}
-				else {
-					((RabbitMessageFuture) future).set(message);
+				RabbitFuture<?> future = this.pending.remove(new String(correlationId, this.charset));
+				if (future != null) {
+					if (future instanceof AsyncRabbitTemplate.RabbitConverterFuture) {
+						C converted = (C) this.template.getMessageConverter().fromMessage(message);
+						((RabbitConverterFuture) future).set(converted);
+					}
+					else {
+						((RabbitMessageFuture) future).set(message);
+					}
 				}
 			}
 		}
@@ -428,7 +432,7 @@ public class AsyncRabbitTemplate<C> implements SmartLifecycle, MessageListener, 
 		String correlationId = correlationData.getId();
 		if (correlationId != null) {
 			RabbitFuture<?> future = this.pending.get(correlationId);
-			if (future instanceof SettableListenableFuture) {
+			if (future != null) {
 				future.setNackCause(cause);
 				((SettableListenableFuture<Boolean>) future.getConfirm()).set(ack);
 			}
@@ -444,15 +448,18 @@ public class AsyncRabbitTemplate<C> implements SmartLifecycle, MessageListener, 
 
 	private String getOrSetCorrelationIdAndSetReplyTo(Message message) {
 		String correlationId;
-		byte[] currentCorrelationId = message.getMessageProperties().getCorrelationId();
+		MessageProperties messageProperties = message.getMessageProperties();
+		Assert.notNull(messageProperties, "the message properties cannot be null");
+		byte[] currentCorrelationId = messageProperties.getCorrelationId();
 		if (currentCorrelationId == null) {
 			correlationId = UUID.randomUUID().toString();
-			message.getMessageProperties().setCorrelationId(correlationId.getBytes(this.charset));
-			message.getMessageProperties().setReplyTo(this.replyAddress);
+			messageProperties.setCorrelationId(correlationId.getBytes(this.charset));
+			Assert.isNull(messageProperties.getReplyTo(), "'replyTo' property must be null");
 		}
 		else {
 			correlationId = new String(currentCorrelationId, this.charset);
 		}
+		messageProperties.setReplyTo(this.replyAddress);
 		return correlationId;
 	}
 
@@ -475,8 +482,8 @@ public class AsyncRabbitTemplate<C> implements SmartLifecycle, MessageListener, 
 
 		@Override
 		public boolean cancel(boolean mayInterruptIfRunning) {
-			boolean cancelled = super.cancel(mayInterruptIfRunning);
 			pending.remove(this.correlationId);
+			boolean cancelled = super.cancel(mayInterruptIfRunning);
 			return cancelled;
 		}
 
@@ -573,6 +580,9 @@ public class AsyncRabbitTemplate<C> implements SmartLifecycle, MessageListener, 
 
 	}
 
+	/**
+	 * This class exists simply to increase the visibility of the id setter.
+	 */
 	private static final class MutableCorrelationData extends CorrelationData {
 
 		public MutableCorrelationData(String id) {
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
@@ -62,7 +62,7 @@ import org.springframework.util.concurrent.SettableListenableFuture;
  * @since 1.6
  *
  */
-public class AsyncRabbitTemplate<C> implements SmartLifecycle, MessageListener, ReturnCallback, ConfirmCallback {
+public class AsyncRabbitTemplate implements SmartLifecycle, MessageListener, ReturnCallback, ConfirmCallback {
 
 	private final Log logger = LogFactory.getLog(this.getClass());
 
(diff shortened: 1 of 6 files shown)


artembilan:
The Travis already said you what to do even without my review :smile:

artembilan:
That is cool! :+1: 

Looks like you enough close to the Spring Reactive AMQP Client :wink: 

CC @smaldini

garyrussell:
Pushed javadoc fixes.

garyrussell:
Pushed docs.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
```diff
@@ -0,0 +1,589 @@
...
+	public void stop(Runnable callback) {
+		stop();
+		callback.run();
+	}
+
+	@SuppressWarnings("unchecked")
+	@Override
+	public void onMessage(Message message) {
+		String correlationId = new String(message.getMessageProperties().getCorrelationId(), this.charset);
+		if (correlationId != null) {
```
artembilan: All these `if`s should be applied for the `message.getMessageProperties().getCorrelationId()`.
Otherwise we are `NPE` before this assertion.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
```diff
@@ -0,0 +1,589 @@
...
+	@Override
+	public void confirm(CorrelationData correlationData, boolean ack, String cause) {
+		if (logger.isDebugEnabled()) {
+			logger.debug("Confirm: " + correlationData + ", ack=" + ack
+					+ (cause == null ? "" : (", cause: " + cause)));
+		}
+		String correlationId = correlationData.getId();
+		if (correlationId != null) {
+			RabbitFuture<?> future = this.pending.get(correlationId);
+			if (future instanceof SettableListenableFuture) {
```
artembilan: `RabbitFuture` is exactly `SettableListenableFuture`, but right it may be `null`, 
so `if (future != null)` would be better.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
```diff
@@ -0,0 +1,589 @@
...
+		}
+	}
+
+	private String getOrSetCorrelationIdAndSetReplyTo(Message message) {
+		String correlationId;
+		byte[] currentCorrelationId = message.getMessageProperties().getCorrelationId();
+		if (currentCorrelationId == null) {
+			correlationId = UUID.randomUUID().toString();
+			message.getMessageProperties().setCorrelationId(correlationId.getBytes(this.charset));
+			message.getMessageProperties().setReplyTo(this.replyAddress);
```
artembilan: So, are we OK do not set `replyTo` when we have already `currentCorrelationId` ?

artembilan: At least the method name assumes a bit different story...

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
```diff
@@ -0,0 +1,589 @@
...
+		private String nackCause;
+
+		public RabbitFuture(String correlationId) {
+			this.correlationId = correlationId;
+		}
+
+		@Override
+		public boolean cancel(boolean mayInterruptIfRunning) {
+			boolean cancelled = super.cancel(mayInterruptIfRunning);
+			pending.remove(this.correlationId);
```
artembilan: Would you mind explaining the decision for the order of these operators ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
```diff
@@ -0,0 +1,589 @@
...
+	}
+
+	private static final class MutableCorrelationData extends CorrelationData {
+
+		public MutableCorrelationData(String id) {
+			super(id);
+		}
+
+		@Override
+		protected void setId(String id) {
```
artembilan: We need to document that we have this class to increase the `setId()` visibility here.
I was going to comment fully different... :flushed:

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/CorrelationData.java
```diff
@@ -28,18 +28,23 @@
...
 	public CorrelationData(String id) {
 		this.id = id;
 	}
 
 	public String getId() {
-		return id;
+		return this.id;
+	}
+
+	protected void setId(String id) {
```
artembilan: Oh! No! 
You have just introduced this setter :innocent: 
Why don't make it just `public` and avoid that `MutableCorrelationData` and just have the `if (this.correlationData != null) {` in the `CorrelationMessagePostProcessor`?

Although that may be just my tired mind to be so nit-pick...

src/reference/asciidoc/whats-new.adoc
```diff
@@ -38,10 +38,17 @@ of `RabbitAdmin` s in the application context.
 
 ===== AmqpTemplate: receive with timeout
 
-A bunch of of new `receive()` methods with `timeout` have been introduced for the `AmqpTemple`
+A number of new `receive()` methods with `timeout` have been introduced for the `AmqpTemplate`
```
artembilan: Oh! Sorry. That was my fault to make it so sacred :wink:

artembilan:
That's all for now.
Would be glad to hear back from you, because I'm not convinced yet to merge it.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
garyrussell: A message from the listener container will always have properties, but I agree with your comment.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
garyrussell: Bug - added a test

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
garyrussell: I don't think it makes a lot of difference, but I'll swap them.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/CorrelationData.java
garyrussell: I think this is better, but you can try to convince me otherwise if you want.

garyrussell:
Rebased, pushed.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/CorrelationData.java
artembilan: Well, I'm not going to convince you on the matter, just share my mind. Of course we can live with your current solution.
Just some extra noise for review.

Since this class is `public`, no one is prohibited to extend it to make access to this setter, therefore no need to mock of the hide-and-seek and just have this as `public` setter and change the `if` in the `CorrelationMessagePostProcessor` to `!= null` because it true by the `convertSendAndReceive()` logic.

Just my five fingers and even if you don't change the code, I'll merge your solution as is.

garyrussell: ok; I'll change it :smile:

garyrussell:
Pushed

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
```diff
@@ -0,0 +1,583 @@
...
+	 * @param routingKey the default routing key.
+	 * @param replyQueue the name of the reply queue to listen for replies.
+	 * @param replyAddress the reply address (exchange/routingKey).
+	 */
+	public AsyncRabbitTemplate(ConnectionFactory connectionFactory, String exchange, String routingKey,
+			String replyQueue, String replyAddress) {
+		Assert.notNull(connectionFactory, "'connectionFactory' cannot be null");
+		this.template = new RabbitTemplate(connectionFactory);
+		Assert.notNull(routingKey, "'exchange' cannot be null, use \"\" for the default exchange");
+		this.template.setExchange(exchange == null ? "" : exchange);
```
artembilan: The previous `notNull` says correct words, but it is is for the `routingKey`. Easy to fix though.
But this ternary operator does confuse me.
What is the reason in that `notNull`, if we just treat `null` as an empty String ?

I prefer to place all asserts in the beginning to avoid object creation, but that is just IMO.

garyrussell: Yeah - that assert is for the wrong parameter - should be exchange.

I don't care either way, feel free to move them to the top during merge.

artembilan: ??? 
I mean that `Assert.notNull(exchange)` is redundant, because of exactly this ternary operator.

Can I remove it on merge ?

garyrussell: Right, sorry that's what I meant.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
```diff
@@ -0,0 +1,583 @@
...
+	}
+
+	/**
+	 * Convert the object to a message and send it to the default exchange with the
+	 * provided routing key.
+	 * @param routingKey the routing key.
+	 * @param message the message.
+	 * @return the {@link RabbitConverterFuture}.
+	 */
+	public RabbitConverterFuture convertSendAndReceive(String routingKey, Object message) throws AmqpException {
```
artembilan: `AmqpException extends RuntimeException`, so no need to expose it in the `throws`.
Would you mind to explain the approach to do that or not for `RuntimeException`s?

I'm asking because the previous method doesn't do that.

Thanks.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
```diff
@@ -0,0 +1,583 @@
...
+					}
+				}
+			}
+		}
+	}
+
+	@Override
+	public void returnedMessage(Message message, int replyCode, String replyText, String exchange, String routingKey) {
+		String correlationId = new String(message.getMessageProperties().getCorrelationId(), this.charset);
+		if (correlationId != null) {
```
artembilan: The same `correlationId` assertion concern is here, too.
Like we discussed in the `onMessage()`.

garyrussell: Well, since it's a returned message the issue is not at all likely (unless some user calls this method himself).

Feel free to fix, though, if you want.

artembilan: From other side: can't we just don't care about the `String <-> byte[]` conversion for the `correlationId` and just rely on it as is?
I mean that it looks strange for me to convert to `String` just because of `this.pending.remove(correlationId);`

Although maybe never mind, just because of `String id` in the `CorrelationData`...

garyrussell: It's actually unfortunate that `MessageProperties` uses `byte[]` since `BasicProperties` uses `String` so we end up converting back - see `DefaultMessagePropertiesConverter` - I thought we had a JIRA to fix that (support a String as an alternative to byte[]) but I haven't found it yet.

artembilan: H-m. Even if you can't find it let's fix it with the current 1.6 milestone!

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplateTests.java
```diff
@@ -0,0 +1,298 @@
+/*
+ * Copyright 2014 the original author or authors.
```
artembilan: Something wrong with your IDE config?
Or is that just a copy/paste artefact?

Will fix on merge

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
```diff
@@ -0,0 +1,583 @@
...
+ * {@link SimpleMessageListenerContainer} either provided or constructed internally.
+ * If an external {@link RabbitTemplate} is provided and confirms/returns are enabled,
+ * it must not previously have had callbacks registered because this object needs to
+ * be the callback.
+ *
+ * @author Gary Russell
+ * @since 1.6
+ *
+ */
+public class AsyncRabbitTemplate<C> implements SmartLifecycle, MessageListener, ReturnCallback, ConfirmCallback {
```
artembilan: Shouldn't we introduce a new interface and extend it from `AmqpTemplate` to meet `AmqpOutboundEndpoint` requirements ?

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplateTests.java
```diff
@@ -0,0 +1,298 @@
...
+		future.cancel(false);
+		assertEquals(0, TestUtils.getPropertyValue(template, "pending", Map.class).size());
+	}
+
+	@Test
+	public void testMessageCustomCorrelation() throws Exception {
+		this.fooMessage.getMessageProperties().setCorrelationId("foo".getBytes());
+		ListenableFuture<Message> future = this.template.sendAndReceive(this.fooMessage);
+		Message result = checkMessageResult(future, "FOO");
+		assertEquals(new String("foo"), new String(result.getMessageProperties().getCorrelationId()));
```
artembilan: Overdid: "foo" is String already :smile:

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
```diff
@@ -0,0 +1,583 @@
...
+		String correlationId = correlationData.getId();
+		if (correlationId != null) {
+			RabbitFuture<?> future = this.pending.get(correlationId);
+			if (future != null) {
+				future.setNackCause(cause);
+				((SettableListenableFuture<Boolean>) future.getConfirm()).set(ack);
+			}
+			else {
+				if (logger.isWarnEnabled()) {
+					logger.warn("Confirm: " + correlationData + ", ack=" + ack
```
artembilan: Maybe `info()`?

That really doesn't matter already when we have received or canceled reply.

garyrussell: True - maybe even debug? We already documented that confirms are discarded if the reply has already been received.

artembilan: :+1: 
That's why I decided to ask for the more lower logging level.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
garyrussell: I thought about doing that, but it's not really an extension of `AmqpTemplate` it wraps one.

I thought about extracting `AsyncRabbitOperations`.

I suppose we could extend `AmqpTemplate` and delegate the other methods to the `RabbitTemplate` but that won't be much use to us because in most places we use `RabbitTemplate` directly - it supports features (confirms) that are extensions to Amqp.

We could extend `RabbitTemplate` but that would introduce a number of difficulties - particularly around confirm and return handling we'd have to add a bunch more asserts, or override the methods with `UnsupportedOperationException`.

So I decided to keep it simple.

Let me know if any of these options appeals to you.

artembilan: Well, I totally agreed with you!

We can just end up with the:
```
public AmqpOutboundEndpoint(AsyncRabbitTemplate asyncRabbitTemplate) {
...
}
```
And go to drink beer! :smile: 

Out of topic: do we really need that `<C>` generic?

It does not make so much sense for the class level, because we are in the messaging, so any message may carry its own object.
The contract like:
```
public <C> RabbitConverterFuture<C> convertSendAndReceive()
```
makes more sense for me.

garyrussell: Agreed - will add another commit

garyrussell:
Pushed - hopefully the last :smile:

src/reference/asciidoc/amqp.adoc
```diff
@@ -2077,6 +2079,81 @@ Simply bind a queue to the configured dead letter exchange with a routing key eq
...
+The `sendAndReceive` methods return a `RabbitMessageFuture`; the `convertSendAndReceive` methods return a
+`RabbitConverterFuture`.
+
+You can either synchronously retrieve the result later, by invoking `get()` on the future, or you can register a
+callback which will be called asynchronously with the result.
+
+[source, java]
+----
+@Autowired
+private AsyncRabbitTemplate<String> template;
```
artembilan: This should be fixed according the last polishing.
Will address on merge.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
```diff
@@ -0,0 +1,581 @@
...
+		return sendAndReceive(this.template.getExchange(), routingKey, message);
+	}
+
+	/**
+	 * Convert the object to a message and send it to the default exchange with the
+	 * default routing key.
+	 * @param message the message.
+	 * @return the {@link RabbitConverterFuture}.
+	 */
+	public <C> RabbitConverterFuture<C> convertSendAndReceive(Object message) {
```
artembilan:
```
warning: no @param for <C>
```
Will address on merge.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/d0f6e61308a020bdeb06992ece63fa3c4bad99d7.

Good job, of course!
I wonder when I start already do something similar :smile:

GrapeBaBa:
Thanks, Russell. But as ListentableFuture in spring seems not support composition, it is not better to handle multiple rabbit invocation sequence. I saw a new project in Reactor that make rabbitmq invocation with reactive stream, is it already for using in production? Is it also support fault tolerance which is done in spring-amqp?

garyrussell:
This was a relatively simple effort to address a specific user request.

We are looking at Reactor integration as a separate effort with the reactor guys; I don't know the current state of the rabbitmq work over there.

PR #363, opened 2016-01-26
AMQP-565: Fix NPE in RabbitAdmin

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-565

When `setIgnoreDeclarationExceptions(true)` a PRECONDITION_FAILED
declaration failure causes an NPE in `declareQueue`.

`declareQueues` leaves `null` entries in the `declareOks` array which
is not handled by `declareQueue`.

__cherry-pick to 1.5.x__


artembilan:
Merged after rebase and cherry-picked.

PR #365, opened 2016-01-28
Include module in exclusions so they are included in generated pom

wilkinsona:
This change fixes the generated pom and gets the Spring IO checks passing again


artembilan:
Thank you @wilkinsona !

Nothing to complain about. Merging...

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/59710724e5df4414139f67145152fc943d45baf7.
Note: I made the same even for `testCompile` excludes. Just to be consistent.

PR #366, opened 2016-01-28
AMQP-557: @Repeatable @RabbitListener

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-557

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -224,12 +224,7 @@ project('spring-rabbit') {
 		}
 
 		compile ("ch.qos.logback:logback-classic:$logbackVersion", optional)
-		
-		compileTestJava {
-			sourceCompatibility = 1.8
-			targetCompatibility = 1.8
-		}
-		
+
 	}
 
 }
@@ -249,6 +244,33 @@ project('spring-rabbit-test') {
 		}
 		testCompile project(":spring-rabbit").sourceSets.test.output
 
+	}
+
+	project.tasks.withType(Test).all { t ->
+		if (System.getProperty('java.version').startsWith('1.7')) {
+
+			// exclude just the java 8 tests when building with java 7
+
+			exclude	'org/springframework/amqp/rabbit/repeatable/**',
+					'org/springframework/amqp/rabbit/test/mockito/**'
+		}
+		else if (t.name.equals('springIoJdk7Test')) {
+
+			// exclude everything here because the classes were compiled with JDK 8
+
+			exclude '**'
+		}
+	}
+
+	compileTestJava {
+		if (!System.getProperty('java.version').startsWith('1.7')) {
+			sourceCompatibility = 1.8
+			targetCompatibility = 1.8
+		}
+		else {
+			exclude	'org/springframework/amqp/rabbit/repeatable/**',
+					'org/springframework/amqp/rabbit/test/mockito/**'
+		}
 	}
 
 }
--- a/build.gradle
+++ b/build.gradle
@@ -224,6 +224,12 @@ project('spring-rabbit') {
 		}
 
 		compile ("ch.qos.logback:logback-classic:$logbackVersion", optional)
+		
+		compileTestJava {
+			sourceCompatibility = 1.8
+			targetCompatibility = 1.8
+		}
+		
 	}
 
 }
--- a/spring-rabbit-test/src/test/java/org/springframework/amqp/rabbit/repeatable/AbstractRabbitAnnotationDrivenTests.java
+++ b/spring-rabbit-test/src/test/java/org/springframework/amqp/rabbit/repeatable/AbstractRabbitAnnotationDrivenTests.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright 2002-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.amqp.rabbit.repeatable;
+
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+import org.springframework.amqp.rabbit.annotation.RabbitHandler;
+import org.springframework.amqp.rabbit.annotation.RabbitListener;
+import org.springframework.amqp.rabbit.config.RabbitListenerContainerTestFactory;
+import org.springframework.amqp.rabbit.listener.MethodRabbitListenerEndpoint;
+import org.springframework.context.ApplicationContext;
+import org.springframework.stereotype.Component;
+
+/**
+ *
+ * @author Stephane Nicoll
+ * @author Gary Russell
+ */
+public abstract class AbstractRabbitAnnotationDrivenTests {
+
+	@Rule
+	public final ExpectedException thrown = ExpectedException.none();
+
+	@Test
+	public abstract void rabbitListenerIsRepeatable();
+
+	/**
+	 * Test for {@link RabbitListenerRepeatableBean} that validates that the
+	 * {@code @RabbitListener} annotation is repeatable and generate one specific container per annotation.
+	 */
+	public void testRabbitListenerRepeatable(ApplicationContext context) {
+		RabbitListenerContainerTestFactory simpleFactory =
+				context.getBean("rabbitListenerContainerFactory", RabbitListenerContainerTestFactory.class);
+		assertEquals(4, simpleFactory.getListenerContainers().size());
+
+		MethodRabbitListenerEndpoint first = (MethodRabbitListenerEndpoint)
+				simpleFactory.getListenerContainer("first").getEndpoint();
+		assertEquals("first", first.getId());
+		assertEquals("myQueue", first.getQueueNames().iterator().next());
+
+		MethodRabbitListenerEndpoint second = (MethodRabbitListenerEndpoint)
+				simpleFactory.getListenerContainer("second").getEndpoint();
+		assertEquals("second", second.getId());
+		assertEquals("anotherQueue", second.getQueueNames().iterator().next());
+
+		MethodRabbitListenerEndpoint third = (MethodRabbitListenerEndpoint)
+				simpleFactory.getListenerContainer("third").getEndpoint();
+		assertEquals("third", third.getId());
+		assertEquals("class1", third.getQueueNames().iterator().next());
+
+		MethodRabbitListenerEndpoint fourth = (MethodRabbitListenerEndpoint)
+				simpleFactory.getListenerContainer("fourth").getEndpoint();
+		assertEquals("fourth", fourth.getId());
+		assertEquals("class2", fourth.getQueueNames().iterator().next());
+	}
+
+	@Component
+	static class RabbitListenerRepeatableBean {
+
+		@RabbitListener(id = "first", queues = "myQueue")
+		@RabbitListener(id = "second", queues = "anotherQueue")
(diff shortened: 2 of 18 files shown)


build.gradle
```diff
@@ -224,6 +224,12 @@ project('spring-rabbit') {
 		}
 
 		compile ("ch.qos.logback:logback-classic:$logbackVersion", optional)
+		
+		compileTestJava {
+			sourceCompatibility = 1.8
+			targetCompatibility = 1.8
```
artembilan: Have you decided to move out of Java 7 finally?..

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/AbstractRabbitAnnotationDrivenTests.java
```diff
@@ -260,6 +266,50 @@ public void defaultHandle(@Validated String msg) {
...
+		assertEquals("second", second.getId());
+		assertEquals("anotherQueue", second.getQueueNames().iterator().next());
+	}
+
+	@Component
+	static class RabbitListenerRepeatableBean {
+
+		@RabbitListener(id = "first", queues = "myQueue")
+		@RabbitListener(id = "second", queues = "anotherQueue")
+		public void repeatableHandle(String msg) {
```
garyrussell: @artembilan @wilkinsona This will break the Spring IO JDK 7 compatibility build.

I guess we have two options; disable the JDK7 build or comment out this test. I think I'm leaning towards the latter since the class below is functionally equivalent.

Thoughts?

artembilan: I'm fine with just commenting out this test with appropriate Java 8 compatibility marker.

wilkinsona: It's a shame that you're considering removing a test just to keep the compatibility build happy. The only alternative that I can think of is to separate out the tests that require Java 8, either into a separate task or even a separate project, and somehow tell the IO JDK 7 build to ignore them.

garyrussell: Thanks, that's actually a great suggestion - we recently added a new project `spring-rabbit-test` (adds a test harness to help users with testing `@RabbitListener`). I had to back-off lambda usage and [convert to anonymous classes](https://github.com/spring-projects/spring-amqp/commit/8afa05fd6c262047dca1cea131e69979b44f725a) in a test case there for the same reason.

Perhaps we can move this test there and somehow exclude it from the jdk7 compat build.

garyrussell:
Pushed - moved the JDK8 tests to the `-test` project and skip tests in the JDK 7 IO Compatibility build.

Thanks again @wilkinsona

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitTests.java
```diff
@@ -60,6 +61,8 @@
 	@Override
 	@Test
 	public void sampleConfiguration() {
+		String property = System.getProperty("java.version");
+		System.out.println(property);
```
artembilan: I think it is an overhead.
I believe you that you have had a good testing around there.
Although I'll recheck test tasks with IO during merge.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/config/RabbitListenerContainerTestFactory.java
```diff
@@ -27,22 +29,27 @@
...
 	}
 
 	@Override
 	public MessageListenerTestContainer createListenerContainer(RabbitListenerEndpoint endpoint) {
 		MessageListenerTestContainer container = new MessageListenerTestContainer(endpoint);
-		this.listenerContainers.add(container);
+		this.listenerContainers.put(endpoint.getId(), container);
 
 		// resolve the id
 		if (endpoint.getId() == null && endpoint instanceof AbstractRabbitListenerEndpoint) {
```
artembilan: Looks like the previous line of code should be moved after this `if`, otherwise we end up with the `null` key in the map.

garyrussell: Good catch :) except all my tests have an id :smile:

spring-rabbit-test/src/test/java/org/springframework/amqp/rabbit/repeatable/AbstractRabbitAnnotationDrivenTests.java
```diff
@@ -0,0 +1,96 @@
+/*
+ * Copyright 2002-2016 the original author or authors.
```
artembilan: New classes, so only the current year, plus `@since 1.6`.

spring-rabbit-test/src/test/java/org/springframework/amqp/rabbit/test/mockito/AnswerTests.java
```diff
@@ -32,47 +29,17 @@
...
-//		assertEquals("FOOfoo", foo.foo("foo"));
-//		doAnswer(new LambdaAnswer<String>(false, (i, r) ->
-//			"" + i.getArguments()[0] + i.getArguments()[0])).when(foo).foo(anyString());
-//		assertEquals("foofoo", foo.foo("foo"));
-//	}
-
 	@Test
 	public void testLambda() {
+		String property = System.getProperty("java.version");
+		System.out.println(property);
```
artembilan: Deja Vu? Seems for me I've already commented about these two lines...
Different classes :smile:

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListener.java
```diff
@@ -89,11 +90,13 @@
  * @since 1.4
  * @see EnableRabbit
  * @see RabbitListenerAnnotationBeanPostProcessor
+ * @see RabbitListeners
  */
 @Target({ ElementType.TYPE, ElementType.METHOD, ElementType.ANNOTATION_TYPE })
 @Retention(RetentionPolicy.RUNTIME)
 @MessageMapping
 @Documented
+@Repeatable(RabbitListeners.class)
```
artembilan: The `@Repeatable` is `@since 1.8`.
How does it compile on Java < 8?
Or have I just forgotten that Java doesn't care about unknown annotations?..

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -236,21 +239,55 @@ public void doWith(Method method) throws IllegalArgumentException, IllegalAccess
...
+		if (hasClassLevelListeners) {
+			processMultiMethodListeners(classLevelListeners, multiMethods, bean, beanName);
 		}
 		return bean;
 	}
 
-	private void processMultiMethodListener(RabbitListener classLevelListener, List<Method> multiMethods, Object bean,
-			String beanName) {
+	/*
+	 * AnnotationUtils.getRepeatableAnnotations does not look at interfaces
```
artembilan: Is it a know SF issue?
Can't we just use `TODO` for their JIRA ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListeners.java
```diff
@@ -0,0 +1,43 @@
...
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Container annotation that aggregates several {@link RabbitListener} annotations.
+ *
+ * <p>Can be used natively, declaring several nested {@link RabbitListener} annotations.
```
artembilan: The `<p>` on its own line. 
At least I like it that way on the class level.
On the method level I'm fine with everything compact.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListeners.java
```diff
@@ -0,0 +1,43 @@
...
+ * Can also be used in conjunction with Java 8's support for repeatable annotations,
+ * where {@link RabbitListener} can simply be declared several times on the same method,
```
artembilan: on the same method/**class**.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListener.java
garyrussell: It doesn't compile on Java < 8, but it's ok because we always compile with Java 8 - even though the target level is 6.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListeners.java
garyrussell: :+1:

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
garyrussell: I didn't open a JIRA issue - the `JmsListener` used `getAttribute()` not `findAttribute()` before this change. So it's not an issue for them (I guess JMS doesn't support `JmsListener` on interfaces).

I guess we should bring it to Stephane and/or Juergen's attention.

spring-rabbit-test/src/test/java/org/springframework/amqp/rabbit/test/mockito/AnswerTests.java
garyrussell: hehe - I thought I'd removed these; they were from when I was debugging the IO compatibility build.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListener.java
artembilan: Oh! Yes, that's true. It doesn't matter where it is used, because Java just ignores (and warns) unknown annotations:
```
 warning: Cannot find annotation method 'value()' in type 'SuppressWarnings': class file for edu.umd.cs.findbugs.annotations.SuppressWarnings not found
```

It's only important that we always compile it with those classes on place.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
artembilan: It doesn't matter how `JmsListener` behaves there, I think that some generic approach would be good to have in the core of cores :smile: 

CC @sbrannen

artembilan:
That's all. 
Just a couple minor concerns.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
sbrannen: If you guys need an `AnnotationUtils.findRepeatableAnnotations(...)` variant (note "find" vs "get"), please just open a JIRA issue against Spring Framework, and feel free to assign it to me.

artembilan: Here we are: https://jira.spring.io/browse/SPR-13911

Thank you @sbrannen for attention!

garyrussell:
Rebased, pushed.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/0c7294f187799f4c5490f16f8055352108f311ac.
Pay attention to the PDF overflow fix.

PR #368, opened 2016-02-01
AMQP-531: Log4j2 AmqpAppender

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-531

Initial Commit.

__review only - needs more tests and docs__


build.gradle
```diff
@@ -228,6 +231,20 @@ project('spring-rabbit') {
 
 }
 
+project('spring-rabbit-ext') {
```
artembilan: Why just don't use the normal full name:
`spring-rabbit-extensions`
?

spring-rabbit-ext/README.adoc
```diff
@@ -0,0 +1,5 @@
+= spring-rabbit-ext
+
+This project provides extensions to `spring-rabbit`.
+
+In particular, the `log4j2 Appender` is contained here, to avoid classpath conflicts with `spring-rabbit`.
```
artembilan: What is the problem with "classpath conflicts"?
Both Log4jS should be `optional` there in the `spring-rabbit` and it is up to end-user to choose an appropriate dep for his requirements.

Please, share your experience and I may agree with you to go ahead with `ext` .
From other side seeing `ext` in the Gradle config I was going to suggest to move other our `Appender`s there as well...

spring-rabbit-ext/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
```diff
@@ -0,0 +1,495 @@
...
+import org.springframework.amqp.core.TopicExchange;
+import org.springframework.amqp.rabbit.connection.AbstractConnectionFactory;
+import org.springframework.amqp.rabbit.connection.CachingConnectionFactory;
+import org.springframework.amqp.rabbit.core.RabbitAdmin;
+import org.springframework.amqp.rabbit.core.RabbitTemplate;
+import org.springframework.amqp.rabbit.support.DeclareExchangeConnectionListener;
+
+/**
+ * A Log4J2 appender that publishes logging events to an AMQP Exchange.
+ * <p>
```
artembilan: I think this `<p>` is auto-generated, but redundant here anyway.

spring-rabbit-ext/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
```diff
@@ -0,0 +1,495 @@
...
+		appender.startSenders();
+		return appender;
+	}
+
+	/**
+	 * Submit the required number of senders into the pool.
+	 */
+	private void startSenders() {
+		for (int i = 0; i < manager.senderPoolSize; i++) {
+			manager.senderPool.submit(new EventSender());
```
artembilan: I wonder if we can avoid this in favor of built-in async feature: https://logging.apache.org/log4j/2.x/manual/async.html

build.gradle
garyrussell: Just seemed too long; I am ok either way.

spring-rabbit-ext/README.adoc
garyrussell: I read somewhere that you can't have both log4j and log4j2 on the classpath so I was concerned that adding it to the main project would break the log4j Appender tests. But we should try it I suppose; I think that discussion might have been about the log4j1 -> log4j2 bridge.

spring-rabbit-ext/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
garyrussell: I thought about doing that but I noticed log4j1 has a similar appender [here](https://logging.apache.org/log4j/1.2/apidocs/org/apache/log4j/AsyncAppender.html) so I left it async with retries to be consistent.

Of course, we could change that one too.

OTOH, do we want to subject callers to RabbitMQ connectivity problems when they don't configure an async appender?

spring-rabbit-ext/README.adoc
artembilan: Yes, I think it is worth to try to avoid extra module just because of `log4j`.

garyrussell:
Pushed

build.gradle
```diff
@@ -111,7 +112,9 @@ subprojects { subproject ->
 		testCompile ("junit:junit:$junitVersion") {
 			exclude group: 'org.hamcrest', module: 'hamcrest-core'
 		}
-		testCompile "log4j:log4j:$log4jVersion"
+		if (!(subproject.name ==~ /spring-rabbit-ext/)) {
```
artembilan: There is no `ext` module anymore.
Will fix on merge.

build.gradle
```diff
@@ -224,6 +225,12 @@ project('spring-rabbit') {
 		}
 
 		compile ("ch.qos.logback:logback-classic:$logbackVersion", optional)
+
+		compile ("org.apache.logging.log4j:log4j-core:$log4j2Version", optional)
+		compile ("org.apache.logging.log4j:log4j-jcl:$log4j2Version", optional)
+
+		testCompile "commons-cli:commons-cli:1.2"
```
artembilan: I'll extract the version variable on merge anyway.

settings.gradle
```diff
@@ -2,5 +2,5 @@ rootProject.name = 'spring-amqp-dist'
 
 include 'spring-amqp'
 include 'spring-rabbit'
+include 'spring-rabbit-ext'
```
artembilan: DITO re `ext` module

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/package-info.java
```diff
@@ -0,0 +1,4 @@
+/**
+ * Provides classes supporting Log4J2 appenders.
```
artembilan: Minor of course, but brand sounds like ` Log4j 2`.
Will fix on merge.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
```diff
@@ -0,0 +1,495 @@
...
+		appender.startSenders();
+		return appender;
+	}
+
+	/**
+	 * Submit the required number of senders into the pool.
+	 */
+	private void startSenders() {
+		for (int i = 0; i < manager.senderPoolSize; i++) {
+			manager.senderPool.submit(new EventSender());
```
artembilan: Re.: https://github.com/spring-projects/spring-amqp/pull/368#discussion_r51629216

OK. Let it be consistent with others.
No, I don't think that it is good to notify caller about the `appender` problem. It must robust from the the caller perspective. At least `logger.debug()` should have the same effect for any Broker state or even appender configuration.

From other side the 12-factor requires just log to the SOUT and the target platform (PCF ?) should take care about that to aggregate it somewhere like Splunk.
So, I don't think that we should worry a lot about such a stuff like AMQP Appender.

build.gradle
```diff
@@ -224,6 +225,12 @@ project('spring-rabbit') {
 		}
 
 		compile ("ch.qos.logback:logback-classic:$logbackVersion", optional)
+
+		compile ("org.apache.logging.log4j:log4j-core:$log4j2Version", optional)
+		compile ("org.apache.logging.log4j:log4j-jcl:$log4j2Version", optional)
```
artembilan: I found this as redundant.
Will remove on merge.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/log4j2/test/AmqpAppenderTests.java
```diff
@@ -0,0 +1,105 @@
...
+		Message received = template.receive(queue.getName());
+		assertNotNull(received);
+		assertEquals("testAppId.foo.INFO", received.getMessageProperties().getReceivedRoutingKey());
+	}
+
+	@Test
+	public void testProperties() {
+		Log logger = LogFactory.getLog("foo");
+		AmqpAppender appender = (AmqpAppender) TestUtils
+				.getPropertyValue(logger, "logger.context.configuration.appenders", Map.class).get("rabbitmq");
```
artembilan: This works the same way, but more cleaner and without the extra `jcl` dependency:
```
Logger logger = LogManager.getLogger("foo");
AmqpAppender appender = (AmqpAppender) TestUtils
		.getPropertyValue(logger, "context.configuration.appenders", Map.class).get("rabbitmq");
```

build.gradle
garyrussell: ? It's the bridge from commons-logging to log4j2.

How can the test work?

artembilan: For me that works just only basing on the Log4j 2 API: http://logging.apache.org/log4j/2.x/manual/api.html

garyrussell: So you changed the test as well; that's different from saying "it's redundant".

"I changed the tests to use log4j2 directly" would have been more accurate :smile:

artembilan: Yeah. Agree. But I thought that my next comment in the email stream should explain the reason.
Sorry for confuse.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/1150fe48997ab6f9a612d40a51b5c6051c5e7efe

PR #370, opened 2016-02-08
AMQP-568: Long LongString Header Handling

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-568

Fix `DataInputStream` mapping - return the `LongString` by default.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/DefaultMessagePropertiesConverter.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/DefaultMessagePropertiesConverter.java
@@ -13,6 +13,7 @@
 
 package org.springframework.amqp.rabbit.support;
 
+import java.io.DataInputStream;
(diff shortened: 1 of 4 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/DefaultMessagePropertiesConverter.java
```diff
@@ -217,17 +234,21 @@ else if (value instanceof Map<?, ?>) {
...
 	private Object convertLongString(LongString longString, String charset) {
 		try {
 			if (longString.length() <= this.longStringLimit) {
 				return new String(longString.getBytes(), charset);
-			} else {
-				return longString.getStream();
 			}
-		} catch (Exception e) {
+			else {
+				return this.convertLongLongStrings ? longString.getStream() : longString;
```
artembilan: Seems for me I understand your point here.
So, merging...

PR #373, opened 2016-02-11
AMQP-569: Support Delayed Message Exchange Plugin

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-569


artembilan:
I wonder if Travis failure relates somehow to your change...
```
org.springframework.amqp.core.build.MessageBuilderTests > fromBodyAndMessageRange FAILED

    java.lang.AssertionError at MessageBuilderTests.java:128

org.springframework.amqp.core.build.MessageBuilderTests > copyProperties FAILED

    java.lang.AssertionError at MessageBuilderTests.java:145

org.springframework.amqp.core.build.MessageBuilderTests > fromBodyAndMessage FAILED

    java.lang.AssertionError at MessageBuilderTests.java:62
```

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitUtils.java
```diff
@@ -250,4 +250,32 @@ public static boolean isMismatchedQueueArgs(Exception e) {
 		}
 	}
 
+	/**
+	 * Return true if there is a {@link ShutdownSignalException} in the cause tree and its
+	 * reason is "COMMAND_INVALID" and the operation being performed was exchangeDeclare.
```
artembilan: M-m-m. Maybe we should mention here a sample of the failure?
```
* For example there is Delayed Plugin on RabbitMQ but we requested Delayed Exchange to declare.
```

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitAdmin.java
```diff
@@ -523,8 +526,22 @@ private void declareExchanges(final Channel channel, final Exchange... exchanges
...
 				try {
-					channel.exchangeDeclare(exchange.getName(), exchange.getType(), exchange.isDurable(),
-						exchange.isAutoDelete(), exchange.getArguments());
+					if (exchange.isDelayed()) {
+						Map<String, Object> arguments = exchange.getArguments();
+						if (arguments == null) {
+							arguments = new HashMap<String, Object>();
+						}
+						else {
+							arguments = new HashMap<String, Object>(arguments);
```
artembilan: Why can't we place a new entry to the `exchange.getArguments()` directly?
What is the reason to copy them here to a new `Map` ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitAdmin.java
```diff
@@ -523,8 +526,22 @@ private void declareExchanges(final Channel channel, final Exchange... exchanges
...
+						}
+						arguments.put("x-delayed-type", exchange.getType());
+						channel.exchangeDeclare(exchange.getName(), ExchangeTypes.DELAYED, exchange.isDurable(),
+								exchange.isAutoDelete(), arguments);
+					}
+					else {
+						channel.exchangeDeclare(exchange.getName(), exchange.getType(), exchange.isDurable(),
+							exchange.isAutoDelete(), exchange.getArguments());
```
artembilan: I'd say this `else` is redundant.
We end up with the `channel.exchangeDeclare()` anyway, so I prefer to have it only once to demonstrate that we don't do any routing to different logic.

But yes: that's minor and IMHO.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitAdminIntegrationTests.java
```diff
@@ -336,6 +346,52 @@ public void testQueueDeclareBad() {
...
+		Queue queue = new Queue(UUID.randomUUID().toString(), true, false, false);
+		Binding binding = new Binding(queue.getName(), DestinationType.QUEUE, exchange.getName(), queue.getName(), null);
+
+		try {
+			this.rabbitAdmin.declareExchange(exchange);
+		}
+		catch (AmqpIOException e) {
+			if (RabbitUtils.isExchangeDeclarationFailure(e)
+					&& e.getCause().getCause().getMessage().contains("invalid exchange type 'x-delayed-message'")) {
+				Assume.assumeTrue("Broker does not have the delayed message exchange plugin installed", false);
```
artembilan: That's cool trick to stop the test and return with `green`  :smile:

spring-amqp/src/main/java/org/springframework/amqp/core/MessageProperties.java
```diff
@@ -331,6 +331,30 @@ public void setConsumerQueue(String consumerQueue) {
...
+			return null;
+		}
+	}
+
+	/**
+	 * Set the x-delay header.
+	 * @param xDelay the delay.
+	 * @since 1.6
+	 */
+	public void setXDelay(Integer xDelay) {
```
artembilan: M-m-m. Maybe just `setDelay()` ?
How about if the plugin will become the core functionality and the property will be just `delay`?

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitAdminIntegrationTests.java
```diff
@@ -336,6 +346,52 @@ public void testQueueDeclareBad() {
...
+		exchange.setDelayed(true);
+		Queue queue = new Queue(UUID.randomUUID().toString(), true, false, false);
+		Binding binding = new Binding(queue.getName(), DestinationType.QUEUE, exchange.getName(), queue.getName(), null);
+
+		try {
+			this.rabbitAdmin.declareExchange(exchange);
+		}
+		catch (AmqpIOException e) {
+			if (RabbitUtils.isExchangeDeclarationFailure(e)
+					&& e.getCause().getCause().getMessage().contains("invalid exchange type 'x-delayed-message'")) {
```
artembilan: This doesn't path for me:
```
connection error; protocol method: #method<connection.close>(reply-code=503, reply-text=COMMAND_INVALID - unknown exchange type 'x-delayed-message', class-id=40, method-id=10)
```
My RabbitMQ version is RabbitMQ 3.5.4, Erlang 18.0.
Or that is the Windows RabbitMQ variant...

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitAdmin.java
garyrussell: I was thinking it would be better to not change the user's arguments map. I added a boolean to represent this argument instead.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitAdmin.java
garyrussell: I don't understand the comment - the second argument (exchange type) is different in each call.

spring-amqp/src/main/java/org/springframework/amqp/core/MessageProperties.java
garyrussell: Even if the plugin becomes standard, it's still an extension to the spec but, yes, we can remove the X.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitAdmin.java
artembilan: Yes, of course.
As we extract `exchange.getArguments()` variable we can do the same for the exchange type.

As I said: it's IMO, so never mind and let's concentrate on other stuff!

garyrussell:
Re travis: those tests are failing for me too; investigating.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitAdmin.java
artembilan: Got it. 
Let's leave as is!

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitAdmin.java
garyrussell: Perhaps I can convince you that, since this is the most likely branch of the if statement, we don't need to extract these arguments to variables in 99.999% of the time :smile:

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitAdminIntegrationTests.java
garyrussell: So just drop `invalid` ??

artembilan: Looks like, if your RabbitMQ is 3.6.0

src/reference/asciidoc/amqp.adoc
```diff
@@ -2821,6 +2821,57 @@ The third creates names like `custom.gen-MRBv9sqISkuCiPfOYfpo4g`.
 
 Of course, you can provide your own naming strategy bean.
 
+[[delayed-message-exchange]]
+==== Delayed Message Exchange
+
+_Version 1.6_ introduces support for the
+https://github.com/rabbitmq/rabbitmq-delayed-message-exchange[Delayed Message Exchange Plugin].
```
artembilan: Maybe we should mention the blog as https://www.rabbitmq.com/blog/2015/04/16/scheduling-messages-with-rabbitmq/ ?

src/reference/asciidoc/amqp.adoc
```diff
@@ -2821,6 +2821,57 @@ The third creates names like `custom.gen-MRBv9sqISkuCiPfOYfpo4g`.
...
+----
+MessageProperties properties = new MessageProperties();
+properties.setXDelay(15000);
+template.send(exchange, routingKey,
+        MessageBuilder.withBody("foo".getBytes()).andProperties(properties).build());
+----
+
+or
+
+[source, xml]
```
artembilan: The bellow is Java code :smile:

artembilan:
That's all.
Pushing all other your changes to the master and waiting for your polishing here.

src/reference/asciidoc/amqp.adoc
garyrussell: I will change it to just the blog, since the community plugins page has a link to GH.

garyrussell:
Pushed

garyrussell:
Do not merge yet; I need a patch to `SimpleAmqpHeaderMapper`.

artembilan:
Doh... Of course!
And with this release we should fix SI as well `DefaultAmqpHeaderMapper` :disappointed:

garyrussell:
Yes; agreed - I think we can squeeze it in by Tuesday - I moved the JIRA to M1.

artembilan:
We still have time today.
When we press "Release" button here we can switch there!
:smile:

src/reference/asciidoc/index.adoc
```diff
@@ -41,3 +41,6 @@ include::si-amqp.adoc[]
 In addition to this reference documentation, there exist a number of other resources that may help you learn about AMQP.
 
 include::further-reading.adoc[]
+
+[appendix]
+include::appendix.adoc[]
```
artembilan: Not sure why but in the PDF we have this:
manual
Looks like in the `index.adoc` we should place new page breaker `<<<` before `include::whats-new.adoc[]`

garyrussell: ?? We never get page breaks on `n.x` sections.

artembilan: OK! 
Never mind.
But looks strange that the previous page with code isn't full because the next code block is so big that it is moved to the next page makes this What's New a big ugly.
I can play with this code block to move it to the previous page back.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/cad39a4c2bd758c5009070c8569f35085acd2859

PR #375, opened 2016-02-23
Decrement coundown latch after original is invoked.

ryl:
If the latch is counted down before the original method is invoked, assertion in the test case waiting on the latch may fire before the handler method is 100% complete. This can even lead to messages staying in queues if tests terminate quickly enough.


garyrussell:
Merged as 2c4985e80418128e9e6fd5bdb9f98332e80dc907 with shorter commit comment headline.

Thanks for the contribution @ryl

PR #380, opened 2016-02-29
[Review Only] AMQP-418: SimpleDirectMessageListenerContainer

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-418

Experimental

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleDirectMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleDirectMessageListenerContainer.java
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package org.springframework.amqp.rabbit.listener;
 
 import java.io.IOException;
@@ -21,12 +22,10 @@ import java.util.List;
 import java.util.concurrent.TimeoutException;
 
 import org.springframework.amqp.core.Message;
-import org.springframework.amqp.core.MessageListener;
 import org.springframework.amqp.core.MessageProperties;
 import org.springframework.amqp.rabbit.connection.Connection;
 import org.springframework.amqp.rabbit.connection.ConnectionFactory;
 import org.springframework.amqp.rabbit.connection.RabbitUtils;
-import org.springframework.amqp.rabbit.core.ChannelAwareMessageListener;
 import org.springframework.amqp.rabbit.support.DefaultMessagePropertiesConverter;
 import org.springframework.amqp.rabbit.support.MessagePropertiesConverter;
 import org.springframework.core.task.SimpleAsyncTaskExecutor;
@@ -44,8 +43,8 @@ import com.rabbitmq.client.Envelope;
  * listener directly on the rabbit client consumer thread. There is no txSize property -
  * each message is acked (or nacked) individually.
  * <p>
- * This should be considered experimental at this time; a fully-functional listener container
- * based on this is planned for the 2.0 release.
+ * This should be considered experimental at this time; a fully-functional listener
+ * container based on this is planned for the 2.0 release.
  * <p>
  * TODO: declare queues; advice chain; error handler; AmqpRejectAndDontRequeueException;
  * listener container factory; ...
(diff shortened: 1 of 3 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleDirectMessageListenerContainer.java
```diff
@@ -0,0 +1,218 @@
...
+					try {
+						for (int i = 0; i < consumersPerQueue; i++) {
+							for (String queue : getQueueNames()) {
+								Connection connection = getConnectionFactory().createConnection();
+								Channel channel = connection.createChannel(false);
+								channel.basicQos(prefetch);
+								RabbitUtils.setPhysicalCloseRequired(true);
+								SimpleConsumer consumer = new SimpleConsumer(channel, queue);
+								consumers.add(consumer);
+								channel.basicConsume(queue, consumer);
```
artembilan: I try to understand if we really need the separate Thread to initiate the `basicConsume()`.
Looks like you use that `this.taskExecutor` only to schedule this `while` task.
But why we can't do that directly from the `doStart()` Thread?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleDirectMessageListenerContainer.java
```diff
@@ -0,0 +1,218 @@
...
+					catch (Exception e) {
+						logger.error("Error creating consumer", e);
+						doShutdown();
+						try {
+							Thread.sleep(5000);
+						}
+						catch (InterruptedException e1) {
+							Thread.currentThread().interrupt();
+						}
+						continue;
```
artembilan: What is the point of this `continue`? Retry?
The simplified version of `restart()`? :smile:

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleDirectMessageListenerContainer.java
```diff
@@ -0,0 +1,218 @@
...
+		private void invokeListener(Message message, Channel channel) throws Exception {
+			Object messageListener = getMessageListener();
+			if (messageListener instanceof ChannelAwareMessageListener) {
+				((ChannelAwareMessageListener) messageListener).onMessage(message, channel);
+			}
+			else if (messageListener instanceof MessageListener) {
+					((MessageListener) messageListener).onMessage(message);
+			}
+			else {
+				throw new IllegalStateException();
```
artembilan: The exception should be more friendly.
Let me guess: `[Review Only]` :smile:

artembilan: From other side we should check the `messageListener` from the setter and/or `afterPropertiesSet`.
Or even better - `AbstractMessageListenerContainer.checkMessageListener()`.
Or... Why existing `invokeListener()` isn't good here ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleDirectMessageListenerContainer.java
```diff
@@ -0,0 +1,218 @@
...
+				BasicProperties properties, byte[] body) throws IOException {
+			MessageProperties messageProperties = messagePropertiesConverter.toMessageProperties(
+					properties, envelope, "UTF-8");
+			messageProperties.setMessageCount(0);
+			messageProperties.setConsumerTag(consumerTag);
+			messageProperties.setConsumerQueue(queue);
+			Message message = new Message(body, messageProperties);
+			try {
+				invokeListener(message, getChannel());
+				getChannel().basicAck(envelope.getDeliveryTag(), false);
```
artembilan: I wonder is it idempotent to `basicAck()` from my target listener (e.g. over `AmqpHeaders.CHANNEL` in the SI app) and then here one more time...

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleDirectMessageListenerContainer.java
garyrussell: In case the broker is not available; the task will loop attempting to reconnect. Currently hard-coded at 5 seconds.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleDirectMessageListenerContainer.java
garyrussell: See previous comment.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleDirectMessageListenerContainer.java
garyrussell: Yeah - this area needs more work for different ack modes, transactions etc.

garyrussell:
Closing; deferred to 2.0

PR #381, opened 2016-02-29
AMQP-582: Compatibility with SF-4.3

artembilan:
JIRA: https://jira.spring.io/browse/AMQP-582

Since the recent change in the SF-4.3 the `MessageMethodArgumentResolver` supports now the payload conversion option.
With that the conversion exception expectation has been changed.

Make `MethodRabbitListenerEndpointTests` compatible with SF-4.3, as well retain compatibility with previous versions.


artembilan:
Tested locally with normal build as well as with IO.

PR #382, opened 2016-03-03
AMQP-585: Add receiveTimeout to Async Template

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-585

- Complete Future with `AmqpReplyTimeoutException`
- Cancel all pending replies when the template is stopped

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
@@ -476,8 +476,8 @@ public class AsyncRabbitTemplate implements SmartLifecycle, MessageListener, Ret
 					}
 				}
 				else {
-					if (logger.isDebugEnabled()) {
-						logger.debug("No pending reply - perhaps timed out");
+					if (logger.isWarnEnabled()) {
+						logger.warn("No pending reply - perhaps timed out: " + message);
 					}
 				}
 			}
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
@@ -17,15 +17,18 @@
 package org.springframework.amqp.rabbit.core;
 
 import java.nio.charset.Charset;
+import java.util.Date;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ScheduledFuture;
 
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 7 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
```diff
@@ -421,6 +475,11 @@ public void onMessage(Message message) {
 						((RabbitMessageFuture) future).set(message);
 					}
 				}
+				else {
+					if (logger.isDebugEnabled()) {
+						logger.debug("No pending reply - perhaps timed out");
```
artembilan: Wouldn't it be better to `warn()` instead?
Together with the `message`, of course, to be able diagnose it from logs. 
Something similar what we have in the `TemporaryReplyChannel`:
```
if (errorDescription != null) {
	if (logger.isWarnEnabled()) {
		logger.warn(errorDescription + ":" + message);
	}
	if (GenericMessagingTemplate.this.throwExceptionOnLateReply) {
		throw new MessageDeliveryException(message, errorDescription);
	}
}
```

artembilan:
Not related to this PR, but can be fixed here, if I am right.
Look into `CorrelationMessagePostProcessor`:
```
if (this.correlationData != null && this.correlationData.getId() == null) {
	this.correlationData.setId(correlationId);
	future.setConfirm(new SettableListenableFuture<Boolean>());
}
```
So, `if (this.correlationData != null && this.correlationData.getId() != null)` we don't call `setConfirm()`.
Is that expected?

From other side that is an our internal code:
```
if (this.enableConfirms) {
	correlationData = new CorrelationData(null);
}
```
And never end up with the `this.correlationData.getId() != null` issue.
Jut in case I am missing anything...

artembilan:
That's all for now.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/AsyncRabbitTemplate.java
garyrussell: ok

garyrussell:
Regarding your last comment, it's just defensive coding. When using conversion, we set up a null correlationId object so we can pass it into the template; yes, it will always be null; we can change it to an assert if you prefer.

artembilan:
If I's seen the `future.setConfirm(new SettableListenableFuture<Boolean>());` in the `convertSendAndReceive()` like it is in the `sendAndReceive`, I'd didn't ask anything.

Actually that `if (this.correlationData != null && this.correlationData.getId() == null) {` is really just for the `this.correlationData.setId(correlationId);`, so I worried if there is a case when we don't add `future.setConfirm()`.

Right, it's fine after the closer look and your explanation.
Thanks.

garyrussell:
Pushed

artembilan:
Good merging...

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/491e0d8a76d93b7a9564a84368da115c2f47c4db after rebase and squash.

PR #383, opened 2016-03-04
Checkstyle Phase I - NewLines at EOF

garyrussell:


artembilan:
I'm not familiar with the `checkstyle` plugin, so all those commented out rules don't say to me anything specific.

If you decided to do that like it is here I don't have any objections and ready to merge.

Let me know if you still have anything else to do for this PR and what is our plan to utilize this plugin features?

Out of topic: how can you catch up all that job?! :smile:

garyrussell:
Yes, it can be merged; this is just phase I; it might take years before all rules are uncommented but we should have it as a goal.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/11292cd42833e96da4d2bcbc0ae5f4ee70353300 after rebase.

PR #385, opened 2016-03-08
Fix Tangles

garyrussell:
Package:

Async template references core and listener packages - needs to move up a level.

DeclareExchangeExceptionListener

AmqpReplyTimeoutException

Classes:

Admin Event

Container Event


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -1448,7 +1448,8 @@ private void publishConsumerFailedEvent(String reason, boolean fatal, Throwable
 		private void publishIdleContainerEvent(long idleTime) {
 			if (applicationEventPublisher != null) {
 				applicationEventPublisher.publishEvent(
-						new ListenerContainerIdleEvent(SimpleMessageListenerContainer.this, idleTime));
+						new ListenerContainerIdleEvent(SimpleMessageListenerContainer.this, idleTime, getListenerId(),
+								getQueueNames()));
```
artembilan: Yeah... That's good catch.
The even really represents a system snapshot, but `queues` in the container can be changed during the event parsing.

Looks from local perspective.
Merging...

PR #386, opened 2016-03-09
Enable 'this.' Rule

garyrussell:
Gradle task currently only fixes one violation per line; it's not
too hard to fix but it's easy enough to run multiple times.

Pass 1

Manually Fix Inner Classes

Pass 2

Pass 3

Pass 4

Pass 5

Manually fix up inner classes.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -178,7 +178,6 @@ subprojects { subproject ->
 
 	checkstyle {
 		configFile = new File(rootDir, "src/checkstyle/checkstyle.xml")
-		// ignoreFailures = true
 	}
 
 	artifacts {
--- a/build.gradle
+++ b/build.gradle
@@ -59,6 +59,7 @@ subprojects { subproject ->
 	apply plugin: 'checkstyle'
 
 	apply from:   "${rootDir}/src/checkstyle/fixHeaders.gradle"
+	apply from:   "${rootDir}/src/checkstyle/fixThis.gradle"
 
 	if (project.hasProperty('platformVersion')) {
 		apply plugin: 'spring-io'
@@ -177,6 +178,7 @@ subprojects { subproject ->
 
 	checkstyle {
 		configFile = new File(rootDir, "src/checkstyle/checkstyle.xml")
+		// ignoreFailures = true
 	}
 
 	artifacts {
--- a/src/checkstyle/fixThis.gradle
+++ b/src/checkstyle/fixThis.gradle
@@ -1,0 +1,66 @@
+task fixThis << {
+	fileTree("${buildDir}/reports/checkstyle").include('*.xml').each { report ->
+		println "processing $report"
+		def xml = new XmlParser(false, false).parse(report)
+		xml.file.each { f ->
+			// println "processing $f"
+			def errors = f.error
+			def hasThisError = false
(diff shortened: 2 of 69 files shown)


artembilan:
According to the Travis report, looks like you have enable not only `this.` rule :smile:

build.gradle
```diff
@@ -177,6 +178,7 @@ subprojects { subproject ->
 
 	checkstyle {
 		configFile = new File(rootDir, "src/checkstyle/checkstyle.xml")
+		// ignoreFailures = true
```
artembilan: I found that we don't need this option.
The `--continue` does the trick.

src/checkstyle/fixThis.gradle
```diff
@@ -0,0 +1,66 @@
...
+					// println line
+					ln++
+					def matcher = line =~ /Copyright (20\d\d)(?:-20\d\d)?/
+					if (matcher.count) {
+						years = matcher[0][1]
+						if (years != now) {
+							years = years + "-$now"
+						}
+						line.replaceFirst(/\{year}/, years)
+					}
```
artembilan: I think the `if` above is a redundant copy/paste artifact.

garyrussell: ?? Don't we want to change the copyright year if necessary?

garyrussell:
Rebased, pushed.

src/checkstyle/fixThis.gradle
artembilan: Yes! Of course!
OK. Does it work really then?
How is that `{year}` in the line?

garyrussell: No idea - I don't understand groovy :smile:

artembilan: Yeah... That was a placeholder in my `fixHeaders` script in the header template.
I think the correct replacement here should be like this:
```
.replaceFirst(/\\E20\\d\\d\(\\-20\\d\\d\)\?\\Q/, years)
```
See `fixHeaders.gradle` for more info:
```
headerTemplate = new File('src/checkstyle/checkstyle-header.txt')
			.text
			.replaceAll(/\^\\Q|\\E\$|\^\$|\^\.\*\$/, '')
			.replaceFirst(/\\E20\\d\\d\(\\-20\\d\\d\)\?\\Q/, '{year}')
			.trim() + System.lineSeparator()  + System.lineSeparator()
```

:wink:

artembilan: Doh.
This one:
```
.replaceFirst(/(20\d\d)(?:-20\d\d)?/, years)
```

garyrussell: I'll try that - thanks - I was surprised there were no changes - I just assumed your previous one to fix the licenses touched all the same files.

garyrussell: I ran with that and it made no changes, so let's just update the script during merge.

Thanks.

artembilan: Yep! 
Merging...

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/6d9afdfa0b15ca3ed68c26476143030be0123d25

PR #390, opened 2016-03-16
AMQP-549: @RabbitListener Inferred Conversion Type

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-549

When receiving messages with JSON content from foreign systems (without type
information in headers), the `@RabbitListener` is not convenient.

With this change, the type of the argument is inferred (if possible) from the
target method signature.

Methods with a single unannotated parameter or one with a `@Payload` annotation
are considered for this feature.

If the type cannot be inferred, a customized message converter can use the `Method`
object to determine the type itself.

This feature is not supported by `@RabbitHandler` methods (class-level `@RabbitListener`)
because the method selection is performed after the payload has been converted.

In addition to the inferred type and method, the target bean is also made available.
This might allow a custom converter to determine the type, even for a class-level
annotation.

In summary, when using a `@RabbitListener`, 3 new properties are added to `MessageProperties`:

- `inferredArgumentType` (if it can be determined)
- `targetMethod` (for method-level annotations)
- `targetObject` (in all cases)


garyrussell:
Review only - needs docs and more tests.

garyrussell:
Do not review/merge yet - I think a much simpler solution is to provide the `Type` directly in a header and let __any__ converter have access to it and make any decisions it wishes - this will also mean a compound converter, such as the `ContentTypeDelegatingMessageConverter` will work too.

The current solution only works for a simple JSON converter.

garyrussell:
Reworked with cleaner, more general approach.

artembilan:
A general comment after "at a glance" review:
Shouldn't the receiving `type` override expectation from the properties?
I mean that it is a consumer wished and responsibilities to convert an incoming JSON to its desired domain object.
In general the producer and consumer know nothing about each other. 
And I'd really would ignore that `__TypeId__`, if I can achieve the loosely coupled domains without any copy/paste hacks for those classes desired by the producer.

spring-amqp/src/main/java/org/springframework/amqp/support/converter/AbstractJavaTypeMapper.java
```diff
@@ -27,74 +27,81 @@
...
+		Map<String, Object> headers = properties.getHeaders();
+		Object classIdFieldNameValue = headers.get(headerName);
+		String classId = null;
+		if (classIdFieldNameValue != null) {
+			classId = classIdFieldNameValue.toString();
+		}
+		return classId;
+	}
+
+	private void validateIdTypeMapping() {
```
artembilan: I know it's not your code, but how about to make it better!

This method really doesn't do any *validation*, but just produces one more `Map` for `value <-> key` direction.

Looks like something like `BiMap` would be better, of course, and not sure if Spring has one for us, but we can do the same from the ctor anyway.

After that we can even get rid of the next `afterPropertiesSet()` altogether!

spring-amqp/src/test/java/org/springframework/amqp/support/converter/DefaultJackson2JavaTypeMapperTest.java
```diff
@@ -43,208 +44,209 @@
  * @author James Carr
  * @author Sam Nelson
  * @author Andreas Asplund
+ * @author Gary Russell
  */
 
 @RunWith(MockitoJUnitRunner.class)
 public class DefaultJackson2JavaTypeMapperTest {
```
artembilan: Let's rename it to the ...Test**s**, if we are already here!

spring-amqp/src/test/java/org/springframework/amqp/support/converter/Jackson2JsonMessageConverterTests.java
```diff
@@ -50,13 +54,14 @@
 public class Jackson2JsonMessageConverterTests {
```
artembilan: Not sure how this test has worked before, but looks like there is no any config for it.
And removing:
```
@ContextConfiguration
@RunWith(SpringJUnit4ClassRunner.class)
```
makes it worked, too, but much faster.

spring-amqp/src/test/java/org/springframework/amqp/support/converter/Jackson2JsonMessageConverterTests.java
```diff
@@ -69,222 +74,266 @@ public void before(){
...
+		assertThat(foo, instanceOf(Foo.class));
+	}
+
+	@Test
+	public void testInferredGenericTypeInfo() throws Exception {
+		byte[] bytes = "[ {\"name\" : \"foo\" } ]".getBytes();
+		MessageProperties messageProperties = new MessageProperties();
+		messageProperties.setContentType("application/json");
+		messageProperties
+				.setInferredArgumentType(Bar.class.getMethod("setFoos", List.class).getGenericParameterTypes()[0]);
```
artembilan: I haven't seen yet how you inferre the `Type` in the annotation processor, but how about this variant:
```
messageProperties.setInferredArgumentType((new ParameterizedTypeReference<List<Foo>>() {}).getType());
```
Although that produce the same result, so may even never mind...
Just for the general info then :smile:

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -164,15 +178,81 @@ private String createMessagingErrorMessage(String description, Object payload) {
...
 	 */
 	private class MessagingMessageConverterAdapter extends MessagingMessageConverter {
 
+		private final Object bean;
+
+		private final Method method;
+
+		private final Type inferredArgumentType;
+
+		public MessagingMessageConverterAdapter(Object bean, Method method) {
```
artembilan: The `RedundantModifierChecker` will say us that we are bad here: the class is `private`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -164,15 +178,81 @@ private String createMessagingErrorMessage(String description, Object payload) {
...
+				}
+			}
+			if (genericParameterType == null) {
+				for (int i = 0; i < parameterAnnotations.length; i++) {
+					MethodParameter methodParameter = new MethodParameter(this.method, i);
+					/*
+					 * We're looking for a single non-annotated parameter, or one annotated with
+					 * @Payload.
+					 * We ignore parameters with type Message because they are not involved with
+					 * conversion.
```
artembilan: OK. Why don't you ignore `Message` for the single parameter then?
And I guess that should be independently of the `@Payload` on it.

Right, you have that below, so, let's copy the logic to the single param!

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java
```diff
@@ -1005,6 +1057,36 @@ public Foo2 foo2(Foo2 foo2) {
...
+			return foo2 + queue;
+		}
+
+		@RabbitListener(queues="test.converted.args2")
+		public String foo2a(@Payload Foo2 foo2, @Header("amqp_consumerQueue") String queue) {
+			return foo2 + queue;
+		}
+
+		@RabbitListener(queues="test.converted.message")
+		public String foo2Message(@Payload Foo2 foo2, Message message) {
```
artembilan: OK. As I expected:
```
@RabbitListener(queues="test.notconverted.message")
public String justMessage(Message message) {
	return "foo";
}
```

ends up with the:
```
com.fasterxml.jackson.databind.JsonMappingException: No suitable constructor found for type [simple type, class org.springframework.amqp.core.Message]: can not instantiate from JSON object (missing default constructor or creator, or perhaps need to add/enable type information?)
 at [Source: { "bar" : "baz" }; line: 1, column: 3]
```
Fixing the single param inferring as I commented above makes me happy for the test like:
```
returned = template.convertSendAndReceive("", "test.notconverted.message", "{ \"bar\" : \"baz\" }",
				messagePostProcessor);
```
with the result:
```
"(Body:'{ \"bar\" : \"baz\" }' MessageProperties [headers={}, timestamp=null, messageId=null, userId=null, appId=null, clusterId=null, type=null, correlationId=null, replyTo=amq.rabbitmq.reply-to.g2dkAAtyYWJiaXRASE9NRQAACPEAAAAAAw==.DwnvYpsCuMKIeOumJkRO9w==, contentType=application/json, contentEncoding=UTF-8, contentLength=0, deliveryMode=null, expiration=null, priority=0, redelivered=false, receivedExchange=, receivedRoutingKey=test.notconverted.message, deliveryTag=1, messageCount=0])justMessageFoo2Service"
```

garyrussell:
Yes; I thought about that (inferred type should rule), but I wasn't sure; maybe it should be an option on the type mapper - by default, the inferred type wins. Especially because the inferred type header is only set in these narrow circumstances.

spring-amqp/src/main/java/org/springframework/amqp/support/converter/AbstractJavaTypeMapper.java
garyrussell: Agreed

spring-amqp/src/test/java/org/springframework/amqp/support/converter/Jackson2JsonMessageConverterTests.java
garyrussell: I knew there was a better way; and you would point it out :smile:

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java
garyrussell: :+1:

garyrussell:
One issue with the inferred type overriding the `__TypeId__` is subclasses - the type information might be more specific, so I think we need to make it an option.

spring-amqp/src/test/java/org/springframework/amqp/support/converter/Jackson2JsonMessageConverterTests.java
garyrussell: I don't understand - there's an autowired field - the config is in `src/test/resources` and it runs in 0.13 seconds for me.

garyrussell:
Pushed

garyrussell:
Also skip `Channel` and `Message<?>` parameters.

spring-amqp/src/main/java/org/springframework/amqp/support/converter/AbstractJavaTypeMapper.java
```diff
@@ -21,80 +21,101 @@
...
+	 */
+	@Deprecated
 	public void afterPropertiesSet() throws Exception {
-        validateIdTypeMapping();
-    }
+	}
+
+	protected boolean hasInferredTypeHeader(MessageProperties properties) {
+		boolean hasInferredTypeHeader = properties.getInferredArgumentType() != null;
+		return hasInferredTypeHeader;
```
artembilan: No reason to extract the variable.
The compilator is good here with optimization, but the source code is still this and maybe simplified.

spring-amqp/src/main/java/org/springframework/amqp/support/converter/DefaultJackson2JavaTypeMapper.java
```diff
@@ -29,49 +31,100 @@
...
+	 * <p>
+	 * If you wish to force the use of the  {@code __TypeId__} and associated headers
+	 * (such as when the actual type is a subclass of the method argument type),
+	 * set the precedence to {@link Jackson2JavaTypeMapper.TypePrecedence#TYPE_ID}.
+	 *
+	 * @param typePrecedence the precedence.
+	 * @since 1.6
+	 */
+	public void setTypePrecedence(TypePrecedence typePrecedence) {
+		this.typePrecedence = typePrecedence;
```
artembilan: `Assert.notNull(typePrecedence, "'typePrecedence' must not be null.");`
Otherwise NPE for the `equals` operation in the `toJavaType()`.

garyrussell:
You're right - lazy IDE refactoring - pushed.

artembilan:
The polishing for your confirmation: https://github.com/artembilan/spring-amqp/commit/1c3126cca04049de9eee891b2d27cdecbced23c1.
I hope the last one :smile:

garyrussell:
Thanks; pushed (plus one more commit :smile: ).

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/71c02959b1c20cce499df1eeb45acea67d56d73f

PR #391, opened 2016-03-19
AMQP-591: Fix Channel Cache (with Permits)

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-591

Permits are not released after a connection loss.

Cached channels held a reference to the old connection proxy; since this
was the key to the checkoutPermits map, the map entry was not found
when closing, and hence the permits were not increased.

The solution is to change the channel proxy to be permanent, and just refresh
its target connection; remove the hashCode method so cached channels can always
find their permits. Previously the hashcode included the hashcode of the delegate.

For `CacheMode.CONNECTION`, when a new proxy is created it is permanently stored
in `allocatedConnections` (previously called `openConnections`). When a connection
is retrieved from the cache, if the underlying physical connection is closed we
look at the next cached connection. If all cached connections are closed, we
attempt to refresh the underlying connection.

Also, the permits map entry for a connection was set up after the
listeners were called, so the admin, for example didn't consult the permits
at all.

Added an illegal state exception if the permits are not found when checking out as
well as error logging during return.

Add a test case.


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
```diff
@@ -369,12 +370,17 @@ private Channel getChannel(ChannelCachingConnectionProxy connection, boolean tra
 					if (!checkoutPermits.tryAcquire(this.channelCheckoutTimeout, TimeUnit.MILLISECONDS)) {
 						throw new AmqpTimeoutException("No available channels");
 					}
+					logger.error(
+							"Acquired permit for " + connection + ", remaining:" + checkoutPermits.availablePermits());
```
artembilan: This isn't `error` after the successful `tryAcquire()`.
INFO, or even DEBUG.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
```diff
@@ -582,6 +596,17 @@ else if (this.cacheMode == CacheMode.CONNECTION) {
 		return null;
 	}
 
+	private void refreshProxyConnnection(ChannelCachingConnectionProxy connection) {
```
artembilan: Typo: ternary `n` in the `Connection`.
Will fix on merge.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
```diff
@@ -529,47 +527,63 @@ public final Connection createConnection() throws AmqpException {
...
-							this.openConnectionNonTransactionalChannels.remove(connection);
-							this.openConnectionTransactionalChannels.remove(connection);
-							this.checkoutPermits.remove(connection);
+							this.idleConnections.addLast(connection);
+							if (connection.equals(lastIdle)) {
+								// all of the idle connections are closed.
+								connection = this.idleConnections.poll();
+								break;
+							}
 							connection = null;
```
artembilan: Looks like we have to add
```
this.checkoutPermits.remove(connection);
```
before this nulling.
Otherwise I see you re-new the `connection` in the next `if` and put it into the the `this.checkoutPermits`.

I can't catch that case with tests, so looks like I'm missing some very smart logic here around.

Would you mind elaborating more?

Thanks

artembilan: No, the bug has a place anyway.
When we exceed the `connectionCacheSize` (`CachingConnectionFactoryTests.testWithConnectionFactoryCachedConnection()`), we end up in the `ChannelCachingConnectionProxy.close()` had have:
```
CachingConnectionFactory.this.allocatedConnections.remove(this);
CachingConnectionFactory.this.allocatedConnectionNonTransactionalChannels.remove(this);
CachingConnectionFactory.this.allocatedConnectionTransactionalChannels.remove(this);
```
But `checkoutPermits` still has that already closed extra connection.

artembilan:
That's all.
Would like to find more vulnerabilities, but the `CachingConnectionFactory` is already so complex that I would wished to have separate factories for channels and for connections respectively. 

Plus it seems for me that `CachedChannelInvocationHandler` may be moved to the `item` abstraction of our `SimplePool` implementation in SI.

Those just thoughts for consideration in the future, but you feel free to answer to this as well.

garyrussell:
Pushed

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
```diff
@@ -1005,45 +1074,50 @@ public Channel createChannel(boolean transactional) {
...
-								logger.debug("Returning connection '" + this + "' to cache");
-							}
-							CachingConnectionFactory.this.idleConnections.add(this);
-							if (CachingConnectionFactory.this.connectionHighWaterMark
-									.get() < CachingConnectionFactory.this.idleConnections.size()) {
-								CachingConnectionFactory.this.connectionHighWaterMark
-										.set(CachingConnectionFactory.this.idleConnections.size());
+								logger.debug("Completely closing connection '" + this + "'");
 							}
+							destroy();
```
artembilan: Gary, is it fine that we go ahead in this method after the `destroy()` and say in logs like:
```
2016-03-22 15:42:29,792 DEBUG Test worker [org.springframework.amqp.rabbit.connection.CachingConnectionFactory] - <Completely closing connection 'Proxy@46204d1 Dedicated Rabbit Connection: SimpleConnection@261b168a [delegate=amqp://guest@127.0.0.1:5672/, localPort= 55310]'>
2016-03-22 15:42:29,794 DEBUG Test worker [org.springframework.amqp.rabbit.connection.CachingConnectionFactory] - <Returning connection 'Proxy@46204d1 Dedicated Rabbit Connection: null' to cache>
```
?
Shouldn't we `return` from here altogether?

artembilan: I would like to hear back from you on the matter.

For me it is just only blocker to merge.

Thanks.

garyrussell: Sorry, looks like my reply from last night was lost somehow.

No we shouldn't return, we have to return the proxy (closed) to the idle list so it can be reused. Remember, the complete goal of this change is to always reuse proxies and never destroy them.

However, I suppose we could put the following log in an `else` clause so we only log once.

artembilan: >  Remember, the complete goal of this change is to always reuse proxies and never destroy them.

Even if that is out of `connectionCacheSize`:
```
if (!this.target.isOpen() || countOpenIdleConnections() >= CachingConnectionFactory.this.connectionCacheSize) {
```
?

garyrussell: Yes, the "cacheSize" now represents the maximum number of currently open idle channels. The actual number of entries can be larger than that (up to the new limit) but we'll never have more than the cache size open and idle.

This is to preserve the proxy object so the permits are always correct - I considered continuing to remove them but it adds more complexity to an already complex class.

artembilan: OK. Got it!
Merging...

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/4e997815f1d1e513f4d4e2a57197c703336dbca3

artembilan:
Forgot to mention: for backport this requires the separate PR for `1.5.x`.
Thanks.

PR #392, opened 2016-03-21
AMQP-583: Fatal @RabbitListener Exceptions

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-583

The `ConditionalRejectingErrorHandler` now considers irrecoverable
`@RabbitListener` exceptions as fatal (rejecting the message so it is
discarded or routed to a DLX/DLQ).

Previously, only `MessageConversionException`s were fatal by default
and users had to configure a customized error handler.

`o.s.amqp.support.converter.MessageConversionException`
`o.s.messaging.converter.MessageConversionException`
`MethodArgumentNotValidException`
`MethodArgumentTypeMismatchException`

are all fatal exceptions.

Also, previously, the stack trace was logged twice; it is now only logged once.

__cherry-pick to 1.5.x__ What's new will need adjustment.

### DIFF
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/ErrorHandlerTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/ErrorHandlerTests.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.amqp.rabbit.listener;
+
+import static org.junit.Assert.fail;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
+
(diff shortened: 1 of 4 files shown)


artembilan:
Not sure if Travis failure is related to the PR, but I'm sharing its issue anyway:
```
org.springframework.amqp.rabbit.core.BatchingRabbitTemplateTests > testDebatchByContainerBadMessageRejected FAILED

    org.mockito.exceptions.verification.TooLittleActualInvocations at BatchingRabbitTemplateTests.java:338

572 tests completed, 1 failed, 46 skipped

:spring-rabbit:test FAILED
```

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/ErrorHandlerTests.java
```diff
@@ -0,0 +1,95 @@
...
+ * @since 1.6
+ *
+ */
+public class ErrorHandlerTests {
+
+	@Test
+	public void testFatalsAreRejected() throws Exception {
+		ConditionalRejectingErrorHandler handler = new ConditionalRejectingErrorHandler();
+		Log logger = spy(TestUtils.getPropertyValue(handler, "logger", Log.class));
+		doAnswer(new DoesNothing()).when(logger).warn(Matchers.anyString(), Matchers.any(Throwable.class));
```
artembilan: WDYT if we, at least, start using `BDDMockito` instead?
I don't tell already that it would be better to migrate like it is required by the Checkstyle.

artembilan:
Yes, `BatchingRabbitTemplateTests.testDebatchByContainerBadMessageRejected` fails for me as well.
Just because of this "the stack trace was logged twice; it is now only logged once" change.

Fixing during merger...

artembilan:
OK. IT is fine after fixing and rebasing.
So, merging after the aforementioned polishings.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/ee51a64f6abdbca459964f7b6c3d3cc54d2d95ab and cherry-picked with conflict fix for the "What's New".

PR #393, opened 2016-03-21
AMQP-579: @SendTo - Support Runtime SpEL

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-579

`@SendTo` currently supports initialization time SpEL (`#{...}`).

Add support for runtime SpEL evaluation against the request and reply objects.

This change also eliminates the `ThreadLocal` for `@SendTo`s in the
`DelegatingInvocableHandler` for multi-method listeners, in favor of
returning a `ResultHolder` containg the result and sendTo (when there
is no `replyTo` property.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
@@ -16,6 +16,9 @@
 
 package org.springframework.amqp.rabbit.listener.adapter;
 
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
@@ -32,6 +35,13 @@ import org.springframework.amqp.rabbit.support.RabbitExceptionTranslator;
 import org.springframework.amqp.support.converter.MessageConversionException;
 import org.springframework.amqp.support.converter.MessageConverter;
 import org.springframework.amqp.support.converter.SimpleMessageConverter;
+import org.springframework.context.expression.MapAccessor;
+import org.springframework.expression.BeanResolver;
+import org.springframework.expression.Expression;
+import org.springframework.expression.spel.standard.SpelExpressionParser;
+import org.springframework.expression.spel.support.StandardEvaluationContext;
+import org.springframework.expression.spel.support.StandardTypeConverter;
+import org.springframework.util.Assert;
 
 import com.rabbitmq.client.Channel;
 
@@ -51,13 +61,22 @@ public abstract class AbstractAdaptableMessageListener implements MessageListene
 
 	private static final String DEFAULT_ENCODING = "UTF-8";
 
+	private static final SpelExpressionParser PARSER = new SpelExpressionParser();
 
 	/** Logger available to subclasses */
 	protected final Log logger = LogFactory.getLog(getClass());
 
+	private final ConcurrentMap<String, Expression> responseExpressions = new ConcurrentHashMap<String, Expression>();
+
+	private final StandardEvaluationContext evalContext = new StandardEvaluationContext();
+
 	private String responseRoutingKey = DEFAULT_RESPONSE_ROUTING_KEY;
 
 	private String responseExchange = null;
+
+	private Address responseAddress = null;
+
+	private Expression responseExpression;
 
 	private volatile boolean mandatoryPublish;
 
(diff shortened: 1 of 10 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
```diff
@@ -104,6 +123,28 @@ public void setResponseExchange(String responseExchange) {
...
+	 * that return result objects, which will be wrapped in
+	 * a response message and sent to a response destination.
+	 * <p>
+	 * Can be a string starting with "SpEL:" in which case the expression is
+	 * evaluated at runtime; see the reference manual for more information.
+	 * @param defaultReplyTo The exchange.
+	 * @since 1.6
+	 */
+	public void setResponseAddress(String defaultReplyTo) {
+		if (defaultReplyTo.startsWith("SpEL:")) {
```
artembilan: That smells a bit.
How about to use:
```
PARSER.parseExpression(defaultReplyTo, new TemplateParserContext());
```
instead?
And rely on the standard SpEL syntax: `#{myBean.foo}` etc.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/DelegatingInvocableHandler.java
```diff
@@ -125,22 +125,32 @@ protected InvocableHandlerMethod getHandlerForPayload(Class<? extends Object> pa
...
-				}
-				Address replyTo = destinations.length == 1 ? new Address(resolve(destinations[0])) : null;
-				if (replyTo != null) {
-					this.defaultReplyToForHandler.put(handler, replyTo);
-				}
-			}
+			replyTo = extractSendTo(method.toString(), ann);
+		}
+		if (replyTo == null) {
+			SendTo ann = AnnotationUtils.getAnnotation(this.bean.getClass(), SendTo.class);
```
artembilan: `@SendTo` is only method level annotation:
```
@Target(ElementType.METHOD)
@Retention(RetentionPolicy.RUNTIME)
@Documented
public @interface SendTo {
```
Or have you raised an SF JIRA on the matter already ?

artembilan: Oh! I really have a bad memory...
https://jira.spring.io/browse/SPR-13578.
Never mind.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
artembilan: OK. I see your point around `resolve()` in the `MethodRabbitListenerEndpoint`, which with the `StandardBeanExpressionResolver` and its internal `beanExpressionParserContext` eats my case.
Anyway I guess we still can get a gain of the `TemplateParserContext` with the `@{` , `}` placeholder for this kind of expressions. 
At least that is what SpEL recommends to do for custom templates.
WDYT?

garyrussell: Yes, makes sense; but I'm not keen on `@`, I'll look for something else.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/DelegatingInvocableHandler.java
```diff
@@ -125,22 +125,32 @@ protected InvocableHandlerMethod getHandlerForPayload(Class<? extends Object> pa
...
+			replyTo = extractSendTo(this.getBean().getClass().getSimpleName(), ann);
 		}
+		if (replyTo != null) {
+			this.handlerSendTo.put(handler, replyTo);
```
artembilan: I'd say that parse into an `Expression` would be better here. More logical, than in the `AbstractAdaptableMessageListener`.

garyrussell:
Nice - much cleaner and the template context automatically gives us a literal expression if the prefix is not found.

Pushed

src/reference/asciidoc/amqp.adoc
```diff
@@ -1703,8 +1704,38 @@ public String spelReplyTo() {
...
+[source, java]
+----
+@RabbitListener(queues = "test.sendTo.spel")
+@SendTo("!{'some.reply.queue.with.' + result.queueName}")
+public Bar capitalizeWithSendToSpel(Foo foo) {
+    return processTheFooAndReturnABar(foo);
+}
+----
+
+The runtime nature of the SpEL expression is indicated with the `!{...}` prefix.
```
artembilan: This isn't `prefix` anymore.
Maybe `delimiters` like it is in the Core SpEL doc:
> Each evaluation block is delimited with prefix and suffix characters that you can define, a common choice is to use `#{ }` as the delimiters

Let me know your thought and I'll fix on merge.

garyrussell: Thanks yes, should be

    is indicated with `!{...}` delimiters.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/30cabc0ac64916cd5e2388f67f72e9a47fac7bcd

PR #399, opened 2016-03-24
AMQP-594: Fix for `@RabbitListener` when CGLIB

artembilan:
JIRA: https://jira.spring.io/browse/AMQP-594

Previously with the CGLIB proxy and generic interface implementation
the `RabbitListenerAnnotationBeanPostProcessor` registered two endpoints for "different"
methods for the same `@RabbitListener`.

* Fix the issue adding `ReflectionUtils.USER_DECLARED_METHODS` to the method processor
* Add test on the matter
* Also add a note to the Reference Manual regarding Proxy behavior

**Cherry-pick to 1.5.x**


garyrussell:
Merged as 1c4dd54e4c8c69424c40c2b1d33a21a5a5f03caa and cherry-picked after resolving conflicts.

PR #400, opened 2016-03-29
AMQP-591: Fix Permit Release on Factory Destroy

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-591

Don't release a permit if the channel being closed is
actually in the cache (e.g. during `destroy()` or `resetConnection()`).

Also drop ERROR log to DEBUG for permit release.

Also add Lars Hvile test (`@Ignored` because it needs user interaction).

__cherry-pick to 1.5.x__

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
@@ -886,7 +886,14 @@ public class CachingConnectionFactory extends AbstractConnectionFactory
 
 				// If we get here, we're supposed to shut down.
 				physicalClose();
-				releasePermit();
+				/*
+				 *  Only release a permit if this is a normal close; if the channel is
+				 *  in the list, it means we're closing a cached channel (for which a permit
+				 *  has already been released).
+				 */
+				if (!this.channelList.contains(proxy)) {
+					releasePermit();
+				}
 				return null;
 			}
 			else if (methodName.equals("getTargetChannel")) {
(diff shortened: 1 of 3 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
```diff
@@ -886,7 +886,14 @@ else if (methodName.equals("close")) {
...
 				physicalClose();
-				releasePermit();
+				/*
+				 *  Only release a permit if this is a normal close; if the channel is
+				 *  in the list, it means we're closing a cached channel (for which a permit
+				 *  has already been released).
+				 */
+				if (!this.channelList.contains(proxy)) {
+					releasePermit();
+				}
```
artembilan: I commented out this `if` altogether and `testAll` passes.
Even that new `RabbitReconnectProblemTests` passes for me after restarting RabbitMQ.

Or I am missing anything, or we really don't need this logic.
Just because we go to the `channelList` only if `this.channelList.size() < getChannelCacheSize()`.
See the previous `if` in this method:
```
if (!RabbitUtils.isPhysicalCloseRequired() &&
	(this.channelList.size() < getChannelCacheSize()
		|| this.channelList.contains(proxy))) {
	logicalClose((ChannelProxy) proxy);
	// Remain open in the channel list.
	releasePermit();
	return null;
}
```

garyrussell: Yes, it's needed - but as you point out, we didn't have a test case to prove it.

Added.

garyrussell: Pushed a bit more polish.

garyrussell: Actually --- hang on.

garyrussell:
Pushed

artembilan:
Good. 
I think I'm fine with the fix now.
Merging to let the reporter to test it in his environment against the latest our build.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/c4985b98305b83563381852aecf743d908db35df and cherry-picked.

PR #402, opened 2016-04-04
Checkstyle Rules

garyrussell:
Most rules are now enabled (not Javadoc, AssertJ, BDDMockito or leading whites).


artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/b89756eec7fd4cc22c0669fdd53992e1c474adf2
That was terrific! 
:bomb:

PR #403, opened 2016-04-07
Fix namespace mapping (META-INF/spring.schemas)

shuraa:
_I have signed and agree to the terms of the SpringSource Individual Contributor License Agreement._


garyrussell:
Thanks!

PR #405, opened 2016-04-09
AMQP-597: Document MetaAnnotations

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-597

Also clean up extra queues left due to `@DirtiesContext` on some tests.

### DIFF
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java
@@ -28,6 +28,10 @@ import static org.junit.Assert.assertTrue;
 import static org.mockito.Mockito.mock;
 
 import java.io.Serializable;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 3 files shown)


spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java
```diff
@@ -1061,6 +1080,30 @@ public String baz(Baz baz, String rk) {
 
 	}
 
+	@Target({ElementType.TYPE, ElementType.METHOD, ElementType.ANNOTATION_TYPE})
+	@Retention(RetentionPolicy.RUNTIME)
+	@RabbitListener(bindings = @QueueBinding(
+			value = @Queue,
+			exchange = @Exchange(value = "test.metaFanout", type = ExchangeTypes.FANOUT, autoDelete = "true")))
+	public @interface MyAnonFanoutListener {
```
artembilan: I wonder if they works well as really full meta-annotation.
I mean how about the `value` attribute in the custom annotation, which is visible from the annotation processor.
Do we use here in the SA the same mechanism as we do in the SI ?
Thanks

artembilan: M-m-m. Looks like no. We do only this:
```
private Collection<RabbitListener> findListenerAnnotations(Method method) {
```
but don't collect attributes or `AnnotationUtils.synthesizeAnnotation()` from the custom environment. 

OK. That's fine too.

More complex cases should be achieved with the regular `@Bean` definition for the AMQP entities (exchanges, queues and bidnings between them) and used as a `queus` from the normal `@RabbitListener` annotation.

How about to document that as well?

garyrussell:
Pushed

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/bfc854b42c1ef1c28d9444b03af68b5aac90734d after rebase.

PR #407, opened 2016-04-20
AMQP-599: Log Appenders - Add addresses Property

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-599

Also add docs for the appenders.

__cherry-pick to 1.5.x - except Log4j2 and docs__

### DIFF
--- a/src/reference/asciidoc/logging.adoc
+++ b/src/reference/asciidoc/logging.adoc
@@ -1,0 +1,183 @@
+[[logging]]
+=== Logging Subsystem AMQP Appenders
+
+The framework provides logging appenders for several popular logging subsystems:
+
+- log4j (since Spring AMQP _version 1.1_)
+- logback (since Spring AMQP _version 1.4_)
+- log4j2 (since Spring AMQP _version 1.6_)
+
+The appenders are configured using the normal mechanisms for the logging subsystem, available properties are specified
+in the following sections.
+
+==== Common properties
+
+The following properties are available with all appenders:
+
+.Common Appender Properties
+[cols="2l,2l,4", options="header"]
+|===
+| Property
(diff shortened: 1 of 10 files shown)


src/reference/asciidoc/logging.adoc
```diff
@@ -0,0 +1,183 @@
...
+| Name of the exchange to publish log events to.
+
+| exchangeType
+| topic
+| Type of the exchange to publish log events to - only needed if the appender declares the exchange.
+See `declareExchange`.
+
+| routingKeyPattern
+| %c.%p
+| Log4j-style pattern format to use to generate a routing key.
```
artembilan: You said ` Common properties`, but use you here `Log4j` word.
So, maybe really some general phrase would be better:
```
Logging subsystem pattern format to use to generate a routing key.
```
?

artembilan:
LGTM.
I can fix that minor concern on merge.

src/reference/asciidoc/logging.adoc
garyrussell: That's why I added "-style" - the other implementations use the same pattern.

Not sure how else it can be reworded without going into lots of details.

garyrussell: That's why I added "-style" - the other implementations use the same pattern.

But, yes, your suggestion is fine.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/d29a3a0a39b5861594c26e2d72028613694c1a7c
And cherry-picked as https://github.com/spring-projects/spring-amqp/commit/9db4aba23ab6e9a162975f1d59ad867b03c6b73c after resolving conflicts.

PR #409, opened 2016-04-26
AMQP-604: @QueueBinding: Support Arguments

### LINKED ISSUE #408: Unexpected exchange type: headers
Hi!

When I try to bind queue to headers exchange through annotation `@RabbitListener -> bindings -> @QueueBinding`, I receive exception: `Unexpected exchange type: headers`

I inspect class `RabbitListenerAnnotationBeanPostProcessor` and found next code:

```java
if (exchangeType.equals(ExchangeTypes.DIRECT)) {
    exchange = new DirectExchange(exchangeName,
            resolveExpressionAsBoolean(bindingExchange.durable()),
            resolveExpressionAsBoolean(bindingExchange.autoDelete()));
    actualBinding = new Binding(queueName, DestinationType.QUEUE, exchangeName, resolvedKey, null);
}
else if (exchangeType.equals(ExchangeTypes.FANOUT)) {
    exchange = new FanoutExchange(exchangeName,
            resolveExpressionAsBoolean(bindingExchange.durable()),
            resolveExpressionAsBoolean(bindingExchange.autoDelete()));
    actualBinding = new Binding(queueName, DestinationType.QUEUE, exchangeName, "", null);
}
else if (exchangeType.equals(ExchangeTypes.TOPIC)) {
    exchange = new TopicExchange(exchangeName,
            resolveExpressionAsBoolean(bindingExchange.durable()),
            resolveExpressionAsBoolean(bindingExchange.autoDelete()));
    actualBinding = new Binding(queueName, DestinationType.QUEUE, exchangeName, resolvedKey, null);
}
else {
    throw new BeanInitializationException("Unexpected exchange type: " + exchangeType);
}
```

i.e. binding to headers exchange not supported in annotation `@RabbitListener`.
Why?

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-604

Add support for queue, exchange and binding arguments to
`@QueueBinding`; support `ExchangeType.HEADERS`.

Resolves #408

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/Argument.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/Argument.java
@@ -35,7 +35,7 @@ public @interface Argument {
 	/**
 	 * @return the argument name.
 	 */
-	String name() default "";
+	String name();
 
 	/**
 	 * The argument value, an empty string is translated to {@code null} for example
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/Argument.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/Argument.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 the original author or authors.
+ * Copyright 2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,44 +20,33 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
-import org.springframework.amqp.core.ExchangeTypes;
-
 /**
- * An exchange to which to bind a {@code RabbitListener} queue.
+ * Represents an argument used when declaring queues etc within a
+ * {@code QueueBinding}.
  *
  * @author Gary Russell
- * @since 1.5
+ * @since 1.6
  *
  */
 @Target({})
 @Retention(RetentionPolicy.RUNTIME)
-public @interface Exchange {
+public @interface Argument {
 
 	/**
-	 * @return the exchange name.
+	 * @return the argument name.
 	 */
-	String value();
+	String name() default "";
 
 	/**
-	 * The exchange type - only DIRECT, FANOUT TOPIC, and HEADERS exchanges are supported.
-	 * @return the exchange type.
+	 * The argument value, an empty string is translated to {@code null} for example
+	 * to represent a present header test for a headers exchange.
+	 * @return the argument value.
 	 */
-	String type() default ExchangeTypes.DIRECT;
+	String value() default "";
 
 	/**
-	 * @return true if the exchange is to be declared as durable.
+	 * @return the type of the argument value.
 	 */
-	String durable() default "false";
-
-	/**
-	 * @return true if the exchange is to be declared as auto-delete.
-	 */
-	String autoDelete() default "false";
-
-	/**
-	 * @return the arguments to apply when declaring this exchange.
-	 * @since 1.6
-	 */
-	Argument[] arguments() default {};
+	String type() default "java.lang.String";
 
 }
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
@@ -20,6 +20,7 @@ import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -33,6 +34,7 @@ import org.springframework.amqp.core.DirectExchange;
 import org.springframework.amqp.core.Exchange;
 import org.springframework.amqp.core.ExchangeTypes;
 import org.springframework.amqp.core.FanoutExchange;
+import org.springframework.amqp.core.HeadersExchange;
 import org.springframework.amqp.core.Queue;
 import org.springframework.amqp.core.TopicExchange;
 import org.springframework.amqp.rabbit.config.RabbitListenerConfigUtils;
@@ -45,6 +47,7 @@ import org.springframework.amqp.rabbit.listener.RabbitListenerEndpointRegistry;
 import org.springframework.aop.framework.Advised;
 import org.springframework.aop.support.AopUtils;
 import org.springframework.beans.BeansException;
+import org.springframework.beans.factory.BeanClassLoaderAware;
 import org.springframework.beans.factory.BeanFactory;
 import org.springframework.beans.factory.BeanFactoryAware;
 import org.springframework.beans.factory.BeanInitializationException;
@@ -59,10 +62,12 @@ import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
 import org.springframework.context.expression.StandardBeanExpressionResolver;
 import org.springframework.core.Ordered;
 import org.springframework.core.annotation.AnnotationUtils;
+import org.springframework.expression.spel.standard.SpelExpressionParser;
 import org.springframework.messaging.handler.annotation.support.DefaultMessageHandlerMethodFactory;
 import org.springframework.messaging.handler.annotation.support.MessageHandlerMethodFactory;
 import org.springframework.messaging.handler.invocation.InvocableHandlerMethod;
 import org.springframework.util.Assert;
+import org.springframework.util.ClassUtils;
 import org.springframework.util.ReflectionUtils;
 import org.springframework.util.StringUtils;
 
@@ -96,19 +101,22 @@ import org.springframework.util.StringUtils;
  * @see MethodRabbitListenerEndpoint
  */
 public class RabbitListenerAnnotationBeanPostProcessor
-		implements BeanPostProcessor, Ordered, BeanFactoryAware, SmartInitializingSingleton {
+		implements BeanPostProcessor, Ordered, BeanFactoryAware, BeanClassLoaderAware, SmartInitializingSingleton {
 
 	/**
 	 * The bean name of the default {@link org.springframework.amqp.rabbit.listener.RabbitListenerContainerFactory}.
 	 */
 	static final String DEFAULT_RABBIT_LISTENER_CONTAINER_FACTORY_BEAN_NAME = "rabbitListenerContainerFactory";
 
+	private static final SpelExpressionParser PARSER = new SpelExpressionParser();
 
 	private RabbitListenerEndpointRegistry endpointRegistry;
 
 	private String containerFactoryBeanName = DEFAULT_RABBIT_LISTENER_CONTAINER_FACTORY_BEAN_NAME;
 
 	private BeanFactory beanFactory;
+
+	private ClassLoader beanClassLoader;
 
 	private final RabbitHandlerMethodFactoryAdapter messageHandlerMethodFactory =
 			new RabbitHandlerMethodFactoryAdapter();
@@ -174,6 +182,10 @@ public class RabbitListenerAnnotationBeanPostProcessor
 		}
 	}
 
+	@Override
+	public void setBeanClassLoader(ClassLoader classLoader) {
+		this.beanClassLoader = classLoader;
+	}
 
 	@Override
 	public void afterSingletonsInstantiated() {
@@ -487,7 +499,8 @@ public class RabbitListenerAnnotationBeanPostProcessor
 				Queue queue = new Queue(queueName,
 						resolveExpressionAsBoolean(bindingQueue.durable()),
 						exclusive,
-						autoDelete);
+						autoDelete,
+						resolveArguments(bindingQueue.arguments()));
 				((ConfigurableBeanFactory) this.beanFactory).registerSingleton(queueName + ++this.increment, queue);
 				queues.add(queueName);
 				Exchange exchange;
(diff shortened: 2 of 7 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/Argument.java
```diff
@@ -0,0 +1,52 @@
...
+ *
+ */
+@Target({})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Argument {
+
+	/**
+	 * @return the argument name.
+	 */
+	String name() default "";
```
artembilan: I think `name` must be required.
Otherwise this annotation does not make sense.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -497,37 +510,100 @@ private String resolve(String value) {
...
 			}
 		}
 		return queues.toArray(new String[queues.size()]);
 	}
 
+	private Map<String, Object> resolveArguments(Argument[] arguments) {
+		Map<String, Object> map = new HashMap<String, Object>();
+		for (Argument arg : arguments) {
+			String key = resolve(arg.name());
+			if (StringUtils.hasText(key)) {
```
artembilan: In pursuit of the `@Argument(name)` property, I think we shouldn't allow to resolve it to the empty `String`.
Although... it may make sense in some cases when we deploy application into different environment. 
WDYT?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -497,37 +510,100 @@ private String resolve(String value) {
...
+				Object value = resolve(arg.value());
+				Object type = resolve(arg.type());
```
artembilan: The `resolve()` always return `String`:
```
private String resolve(String value) {
	if (this.beanFactory != null && this.beanFactory instanceof ConfigurableBeanFactory) {
		return ((ConfigurableBeanFactory) this.beanFactory).resolveEmbeddedValue(value);
	}
	return value;
}
```
So, some code here around is redundant.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -497,37 +510,100 @@ private String resolve(String value) {
...
+					}
+					else {
+						map.put(key, value);
+					}
+				}
+				else {
+					Assert.isTrue(value instanceof String, "value must resolve to a String for type conversion");
+					String valueString = (String) value;
+					if (StringUtils.hasText(valueString)) {
+						map.put(key, PARSER.parseExpression(valueString).getValue(typeClass));
```
artembilan: Interesting trick :smile: 
How have you came up with that ?
Why don't use `ConversionService`?

artembilan:
That's all.
Really looks cool and pretty simple :smile: 

Wait for your feedback and ready to polish on merge.

garyrussell:
Pushed - good review - thanks!

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -497,37 +494,110 @@ private String resolve(String value) {
...
+						actualBinding);
 			}
 		}
 		return queues.toArray(new String[queues.size()]);
 	}
 
+	private Map<String, Object> resolveArguments(Argument[] arguments) {
+		Map<String, Object> map = new HashMap<String, Object>();
+		for (Argument arg : arguments) {
+			String key = resolveExpressionAsString(arg.name(), "name");
```
artembilan:
```
Caused by: java.lang.IllegalStateException: 'name' must resolve to a String
```
isn't so informative.
How about this:
```
resolveExpressionAsString(arg.name(), "@Argument.name");
...
throw new IllegalStateException("The [" + attribute + "] must resolve to a String. " +
					"Got [" + resolved + "] for [" + value + "]");
```
Where for the `@Argument(name = "#{new java.util.Date()}", value = "bax")`
We get an exception like:
```
Caused by: java.lang.IllegalStateException: The [@Argument.name()] must resolve to a String. Got [class java.util.Date] for [#{new java.util.Date()}]
```
What I love from Spring above all, is the very useful Stacktraces.
Let's don't break the traditions! :smile:

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -497,37 +494,110 @@ private String resolve(String value) {
...
+				Object value = resolveExpression(arg.value());
+				Object type = resolveExpression(arg.type());
+				Class<?> typeClass;
+				String typeName;
+				if (type instanceof Class) {
+					typeClass = (Class<?>) type;
+					typeName = typeClass.getName();
+				}
+				else {
+					Assert.isTrue(type instanceof String, "type must resolve to a Class or String");
```
artembilan: Similar message must be for the actual type, instead of just:
```
Caused by: java.lang.IllegalArgumentException: type must resolve to a Class or String
```
would be better to have this:
```
Assert.isTrue(type instanceof String, "type must resolve to a Class or String, but get [" +
							type.getClass().getName() + "]");
```
And therefore:
```
Caused by: java.lang.IllegalArgumentException: type must resolve to a Class or String, but get [java.util.Date]
```

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
garyrussell: 👍 thanks

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
garyrussell: `s/get/got/`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -497,37 +494,110 @@ private String resolve(String value) {
...
+					if (typeClass.equals(String.class) && !StringUtils.hasText((String) value)) {
+						map.put(key, null);
+					}
+					else {
+						map.put(key, value);
+					}
+				}
+				else {
+					Assert.isTrue(value instanceof String, "value must resolve to a String for type conversion");
+					String valueString = (String) value;
```
artembilan: M-m-m. Not sure why do we have to restrict conversion just only from `String`?
How about this variant:
```
@Argument(name = "file", value = "#{new java.io.File('/myFile')}", type = "java.lang.String")
```
Since we have `if (CONVERSION_SERVICE.canConvert(value.getClass(), typeClass)) {`, I think that would be fine just rely on the `ConversionService`.

OTOH does it still make sense to have the `type` option, if we allow to execute SpEL there?
My point is around this:
```
@Argument(name = "x-message-ttl", value = "#{10000}"))
```
If the `x-message-ttl` really can't resolve the actual value from the `String`...
And that is less verbose, rather than one more attribute for fully qualified class name.
WDYT?

garyrussell: Unfortunately, we can't do that - we don't know what type of argument value the argument name requires. In this case, yes, TTL is a number but we could match a header, for example `"foo", "10000"` where 10000 is a String.

But, you are right, we don't need the restriction since the CS can handle other conversions. The limit was there when I used SpEL to do the conversion.

artembilan: Good! That fine with me.

How about to add a not into the Reference Manual about SpEL support and how `CS` is involved?

Thanks for the patience to my nit-picking! :smile:

garyrussell:
Pushed

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -545,7 +545,7 @@ else if (exchangeType.equals(ExchangeTypes.HEADERS)) {
 	private Map<String, Object> resolveArguments(Argument[] arguments) {
 		Map<String, Object> map = new HashMap<String, Object>();
 		for (Argument arg : arguments) {
-			String key = resolveExpressionAsString(arg.name(), "name");
+			String key = resolveExpressionAsString(arg.name(), "@Annotation.name");
```
artembilan: Why `@Annotation`, but not `@Argument` directly?
We have context here, so why just don't reflect it?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -574,14 +575,13 @@ else if (exchangeType.equals(ExchangeTypes.HEADERS)) {
 					}
 				}
 				else {
-					Assert.isTrue(value instanceof String, "value must resolve to a String for type conversion");
-					String valueString = (String) value;
-					if (StringUtils.hasText(valueString)) {
-						if (CONVERSION_SERVICE.canConvert(String.class, typeClass)) {
-							map.put(key, CONVERSION_SERVICE.convert(valueString, typeClass));
+					if (value instanceof String && StringUtils.hasText((String) value)) {
```
artembilan: We still have this `String` restriction...
All other types are ignored for the `CS`.

If you worry here about the `String value() default "";`, how about do not `resolveExpression(arg.value())` if it is an empty? Therefore `value == null` and we are fine with the `map.put()`.
OTOH it is end-user problem that his SpEL evaluates to empty String, not `null`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -574,14 +575,13 @@ else if (exchangeType.equals(ExchangeTypes.HEADERS)) {
...
+						if (CONVERSION_SERVICE.canConvert(value.getClass(), typeClass)) {
+							map.put(key, CONVERSION_SERVICE.convert(value, typeClass));
 						}
 						else {
-							throw new IllegalStateException("Cannot convert from String to " + typeClass);
+							throw new IllegalStateException("Cannot convert from " + value.getClass()
+								+ " to " + typeClass);
```
artembilan: Yeah... Looks like we have to use `Class.getName()` here as well.
Although I can live with this extra `class` literal in the exception message.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
garyrussell: doh - brain fart - typo

garyrussell:
Pushed

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java
```diff
@@ -661,6 +686,35 @@ public void handleDifferent(Foo2 foo) {
...
+					exchange = @Exchange(value = "auto.headers", type = ExchangeTypes.HEADERS, autoDelete = "true"),
+					arguments = {
+							@Argument(name = "x-match", value = "any"),
+							@Argument(name = "foo", value = "bax"),
+							@Argument(name = "#{'baz'}", value = "#{'fiz'}")
+					}
+				)
+			}
+		)
+		public String handleWithHeadersExchange(String foo) {
```
artembilan: I think you will agree with me that this `@RabbitListener` declaration looks ugly. Like it would be better that we haven never introduced .
But the cat is out of the bag, plus there is some request from the community.
And yes: no one makes you to use this approach :smile:

artembilan:
LTGT, but one general comment:
Is that intentional that you don't mention SpEL and `CS` in the Docs?
I asked for that in one my comment.

Thanks

garyrussell:
Sorry; yes - missed that - let me add something (on call so it might be a little while).

garyrussell:
Pushed

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/54a613a08e2154fffc8784c7964a85975f4d8fce

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java
garyrussell: In some ways, yes, I agree; but the original implementation was also a user request.

Even though it might be a little ugly, it is really concise having everything in one place, especially when declaring auto-delete elements.

artembilan: Yeah... That's good argument.
Fair enough!

PR #411, opened 2016-04-29
Reduce Method Complexity

garyrussell:
https://sonar.spring.io/issues/search#componentRoots=org.springframework.amqp%3Aspring-amqp-dist|createdAt=2016-04-28T02%3A35%3A51%2B0000|sort=UPDATE_DATE|asc=false

Refactor `RabbitListenerAnnotationBeanPostProcessor.registerBeansForDeclaration` to
reduce its complexity.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
@@ -526,8 +526,8 @@ public class RabbitListenerAnnotationBeanPostProcessor
 		}
 		((ConfigurableBeanFactory) this.beanFactory).registerSingleton(exchangeName + ++this.increment,
 				exchange);
-		((ConfigurableBeanFactory) this.beanFactory).registerSingleton(exchangeName + ++this.increment,
-				actualBinding);
+		((ConfigurableBeanFactory) this.beanFactory).registerSingleton(
+				exchangeName + "." + queueName + ++this.increment, actualBinding);
 	}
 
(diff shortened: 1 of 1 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -487,61 +487,81 @@ else if (resolvedValueToUse instanceof Iterable) {
...
+			actualBinding = new Binding(queueName, DestinationType.QUEUE, exchangeName, routingKey,
+					resolveArguments(binding.arguments()));
+		}
+		else {
+			throw new BeanInitializationException("Unexpected exchange type: " + exchangeType);
+		}
+		((ConfigurableBeanFactory) this.beanFactory).registerSingleton(exchangeName + ++this.increment,
+				exchange);
+		((ConfigurableBeanFactory) this.beanFactory).registerSingleton(exchangeName + ++this.increment,
+				actualBinding);
```
artembilan: H-m. The same `exchangeName` for `Binding` bean, even if it is just a prefix.
How about to have `queueName/exchangeName` combination for the `Binding` bean?
I know it is an existing code, but I won't mind to discuss and fix it here.

garyrussell: good idea

garyrussell:
Pushed

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/b51c8fe71255865bf12debd2d1ed6ce5203be316 after squash.

PR #412, opened 2016-04-29
AMQP-605: Expose RabbitMQ ClientProperties

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-605

`AMQPAppender`s, `RabbitConnectionFactoryBean`; also add a getter
to the `AbstractConnectionFactory` to provide access to the underlying
factory.

__cherry-pick to 1.5.x (except docs and log4j2)__

Appenders tested with breakpoints and RabbitMQ Admin UI

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
@@ -77,6 +77,15 @@ public abstract class AbstractConnectionFactory implements ConnectionFactory, Di
 		this.rabbitConnectionFactory = rabbitConnectionFactory;
 	}
 
+	/**
+	 * Return a reference to the underlying Rabbit Connection factory.
+	 * @return the connection factory.
+	 */
+	public com.rabbitmq.client.ConnectionFactory getRabbitConnectionFactory() {
+		return this.rabbitConnectionFactory;
+	}
+
+
 	public void setUsername(String username) {
 		this.rabbitConnectionFactory.setUsername(username);
 	}
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
@@ -80,6 +80,7 @@ public abstract class AbstractConnectionFactory implements ConnectionFactory, Di
 	/**
 	 * Return a reference to the underlying Rabbit Connection factory.
 	 * @return the connection factory.
+	 * @since 1.5.6
 	 */
 	public com.rabbitmq.client.ConnectionFactory getRabbitConnectionFactory() {
 		return this.rabbitConnectionFactory;
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
@@ -375,6 +375,15 @@ public class RabbitConnectionFactoryBean extends AbstractFactoryBean<ConnectionF
 		this.connectionFactory.setExceptionHandler(exceptionHandler);
 	}
 
+	/**
+	 * Add custom properties to the connection factory's clientProperties.
+	 * @param properties the properties to add.
+	 * @since 1.5.6
+	 */
+	public void setCustomClientProperties(Map<String, Object> properties) {
+		this.connectionFactory.getClientProperties().putAll(properties);
+	}
+
 	@Override
 	public Class<?> getObjectType() {
 		return ConnectionFactory.class;
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
@@ -320,11 +320,12 @@ public class RabbitConnectionFactoryBean extends AbstractFactoryBean<ConnectionF
 	}
 
 	/**
+	 * Add custom client properties.
 	 * @param clientProperties the client properties.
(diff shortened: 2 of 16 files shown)


ShijunK:
this is awesome. Thank you. @garyrussell

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
```diff
@@ -77,6 +77,15 @@ public AbstractConnectionFactory(com.rabbitmq.client.ConnectionFactory rabbitCon
 		this.rabbitConnectionFactory = rabbitConnectionFactory;
 	}
 
+	/**
+	 * Return a reference to the underlying Rabbit Connection factory.
+	 * @return the connection factory.
+	 */
+	public com.rabbitmq.client.ConnectionFactory getRabbitConnectionFactory() {
```
artembilan: `@since 1.5.6`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
```diff
@@ -375,6 +375,15 @@ public void setExceptionHandler(ExceptionHandler exceptionHandler) {
...
 	}
 
+	/**
+	 * Add custom properties to the connection factory's clientProperties.
+	 * @param properties the properties to add.
+	 * @since 1.5.6
+	 */
+	public void setCustomClientProperties(Map<String, Object> properties) {
+		this.connectionFactory.getClientProperties().putAll(properties);
+	}
```
artembilan: I wonder why existing:
````java
/**
	 * @param clientProperties the client properties.
	 * @see com.rabbitmq.client.ConnectionFactory#setClientProperties(java.util.Map)
	 */
	public void setClientProperties(Map<String, Object> clientProperties) {
		this.connectionFactory.setClientProperties(clientProperties);
	}
````
isn't enough...
Right, I see they have a different code, but shouldn't an existing method be reworked to this new logic?

Just explain, please, your intentions for these two methods.
Thaks

File: src/reference/asciidoc/logging.adoc [hunk 3]
```diff
@@ -181,3 +181,41 @@ public class MyEnhancedAppender extends AmqpAppender {
...
+    }
+
+}
+----
+
+These properties appear on the RabbitMQ Admin UI when viewing the connection.
+
+With log4j2, subclasses are not supported, due to the way log4j2 uses static factory methods.
+
+Instead, you can specifiy custom client properties in the logback.xml like so:
```
artembilan: Typo: `log4j2.xml`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
```diff
@@ -470,10 +475,27 @@ private void activateOptions() {
...
 			this.connectionFactory.setVirtualHost(this.virtualHost);
+			if (this.clientConnectionProperties != null) {
+				updateClientConnectionProperties();
+			}
 			setUpExchangeDeclaration();
 			this.senderPool = Executors.newCachedThreadPool();
 		}
 
+		private void updateClientConnectionProperties() {
+			String[] props = this.clientConnectionProperties.split(",");
```
artembilan: And why we can't use the same `prop:name[,prop:name]...` approach for all other our appenders?
Yes, leave the extension good hook, but also allow to configure those appenders the same coma-separated value way.
WDYT?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
garyrussell: Doh - mistake I didn't see the existing method - however, that is bad (bug) because it replaces properties set up by the client itself; we need a getter to allow that; will fix to add.

garyrussell:
Pushed

artembilan:
> :spring-rabbit:checkstyleMain[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-amqp/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/LogAppenderUtils.java:30: Class LogAppenderUtils should be declared as final. [FinalClass]

Will fix on merge if only a concern.

src/reference/asciidoc/logging.adoc
```diff
@@ -181,3 +181,74 @@ public class MyEnhancedAppender extends AmqpAppender {
...
+==== Customizing the Client Properties
+
+===== Simple String Properties
+
+Each appender supports adding client properties to the RabbitMQ connection.
+
+.log4j
+[source, text]
+----
+log4j.appender.amqp.clientConnectionProperties=foo:bar,baz:qux
```
artembilan: How about to add a raw for this new property in the `Common Appender Properties` table above? 
(I can do that on merge. Just let me know!)

garyrussell: Good idea, thanks.

src/reference/asciidoc/amqp.adoc
```diff
@@ -385,6 +385,10 @@ Starting with Spring AMQP _version 1.5_, these properties can be set directly on
 If both discrete properties and `sslPropertiesLocation` is provided, properties in the latter will override the
 discrete values.
 
+Starting with _version 1.5.6_, the factory bean exposes a property `customClientProperties` which allows you to add
+properties to connections.
+These properties appear on the RabbitMQ Admin UI when viewing connections.
```
artembilan: These sentences don't make sense any more, since we don't introduce any new option.

garyrussell: agreed

artembilan:
That's all. 
Fixing and merging...

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/35b273311b15027cbf2dba6ddb9f2b727ded6329

artembilan:
Cherry-picked as https://github.com/spring-projects/spring-amqp/commit/df8249f843d4fde91cf5122315d2390f3acbef82

PR #416, opened 2016-05-05
AMQP-572: Always commit only in `afterCompletion`

artembilan:
JIRA: https://jira.spring.io/browse/AMQP-572

Since RabbitMQ doesn't support XA transactions and it really doesn't care about the transaction boundaries,
it doesn't matter when we call `channel.basicAck()` and `channel.txCommit()`.
Therefore any `TransactionManager` looks the same for RabbitMQ and we don't need to distinguish if we are within a local transaction or XA.

* Move `this.resourceHolder.commitAll()` in the `RabbitResourceSynchronization` to the `afterCompletion()` permanently, unconditionally.
* All test passes and now the ACK on the RabbitMQ will be done only if the commit to the DB is successful, for example.
* And it doesn't matter now if we use  XA transactions or Dave Syer's `Best Efforts 1PC pattern` with the `ChainedTransactionManager`

NOTE: There might be some breaking effect for end application if they do some "hard" work around TX synchronizations and rely on the previous behavior.
Since it is minor release, I think we can live with that and provide some workaround on demand.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/ConnectionFactoryUtils.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/ConnectionFactoryUtils.java
@@ -19,8 +19,6 @@ package org.springframework.amqp.rabbit.connection;
 import java.io.IOException;
 
 import org.springframework.amqp.AmqpIOException;
-import org.springframework.transaction.NoTransactionException;
-import org.springframework.transaction.interceptor.TransactionAspectSupport;
 import org.springframework.transaction.support.ResourceHolderSynchronization;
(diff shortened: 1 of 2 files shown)


garyrussell:
Should we add a `TransactionTemplate` test, as discussed in the JIRA?

artembilan:
??? See that modified tests-case in PR - `testSendInGlobalTransactionGuts`
It is fully based on the `TransactionTemplate`.
But right now we don't need to worry about `TransactionAspectSupport.currentTransactionStatus().isNewTransaction()`, because we don't care about that at all. We always just wait for the end of TX.

garyrussell:
Doh - I should have looked at the code, not just the PR - all I saw was a bunch of deleted stuff.

Sorry.

PR #417, opened 2016-05-05
AMQP-607: Don't Adjust Channel Cache Size

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-607

Listener container channels are not cached (since AMQP-275).

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
@@ -44,8 +44,6 @@ import org.springframework.amqp.ImmediateAcknowledgeAmqpException;
 import org.springframework.amqp.core.AcknowledgeMode;
 import org.springframework.amqp.core.Message;
 import org.springframework.amqp.core.Queue;
-import org.springframework.amqp.rabbit.connection.CachingConnectionFactory;
-import org.springframework.amqp.rabbit.connection.CachingConnectionFactory.CacheMode;
 import org.springframework.amqp.rabbit.connection.ConnectionFactory;
 import org.springframework.amqp.rabbit.connection.ConnectionFactoryUtils;
 import org.springframework.amqp.rabbit.connection.ConsumerChannelRegistry;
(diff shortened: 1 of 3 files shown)


artembilan:
M-m-m.
> Listener container channels are not cached

I see this code in the `BlockingQueueConsumer`:
```
this.resourceHolder = ConnectionFactoryUtils.getTransactionalResourceHolder(this.connectionFactory, this.transactional);
this.channel = this.resourceHolder.getChannel();
```
And further I jump through several code circles into the `CachedChannelInvocationHandler`:
```
else if (methodName.equals("close")) {
	// Handle close method: don't pass the call on.
	if (CachingConnectionFactory.this.active) {
			synchronized (this.channelList) {
				if (!RabbitUtils.isPhysicalCloseRequired() &&
						(this.channelList.size() < getChannelCacheSize()
								|| this.channelList.contains(proxy))) {
						releasePermitIfNecessary(proxy);
						logicalClose((ChannelProxy) proxy);
						return null;
					}
				}
			}

			// If we get here, we're supposed to shut down.
			physicalClose();
			releasePermitIfNecessary(proxy);
			return null;
}
```
Where I see that according to the  AMQP-275 fix we don't place `BlockingQueueConsumer`'s `channel` into the `channelList` but just `physicalClose();` close it.

Now pay attention to the next `releasePermitIfNecessary(proxy)`.

Since you have answered that container's channels are long-lived it looks like `checkoutPermits` together with the `channelCheckoutTimeout` is a bottleneck in case of `SMLC`.

Correct me if I am missing anything.

garyrussell:
I am not sure what you are saying - yes, the consumer channels count against the channel limit (when enabled), but they are never cached, just closed (and permit released).

This logic (adjusting the cache size) never made any sense to me, even before AMQP-275 - it doesn't take account of the fact that there might be multiple containers.

artembilan:
Yes-yes, that is my point here as well.
>  multiple containers

Regarding `channelCheckoutTimeout`. Do you find it is OK to protect against the limit even if we are containers? Typically the `channel` lives in the container together with the application. So, if I have enough containers with their `concurrency`, we may end up that one of them can't start because of channel limit on the `CF`. Which really make sense only for the short living `RabbitTemplate`'s channels.
Finally I'll end up only with the workaround with two distinct `CF`s - for long-living without limits for containers and short-living with cache limit for templates.

Is that more clear for you now ?

And yes, I'm fine with the fix and even ready to merge. 

We can move that discussion to separate JIRA, if that...

garyrussell:
I don't think we need to complicate the connection factory any more than it already is - the fact that the consumer channels are not actually put into the cache is not relevant (remember, with the recent change, the proxy __is always__ put in the cache, just the underlying channel is closed). A "user" checks out a channel and, via `RabbitUtils`, signals whether the channel should be physically closed. The connection factory does not need to be directly aware of the user type.

Feel free to open a JIRA against 2.0 and we can revisit - you've already expressed some interest in reworking the caching mechanism to use a pool of some sort and maybe separate out channel caching Vs. connection caching.

artembilan:
Done: https://jira.spring.io/browse/AMQP-608.
Merging after rebase...

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/965195cda9bd1d52b5f1ae089baef0e2069396f9

PR #418, opened 2016-05-06
AMQP-609: Add 'internal' to @Exhange

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-609

Also `ignoreDeclarationExceptions`.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/Exchange.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/Exchange.java
@@ -55,6 +55,17 @@ public @interface Exchange {
 	String autoDelete() default "false";
 
 	/**
+	 * @return true if the exchange is to be declared as internal.
+	 */
+	String internal() default "false";
+
+	/**
+	 * @return true if the declaration exceptions should be ignored.
+	 * @since 1.6
+	 */
+	String ignoreDeclarationExceptions() default "false";
+
+	/**
 	 * @return the arguments to apply when declaring this exchange.
 	 * @since 1.6
 	 */
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java
@@ -57,6 +57,7 @@ import org.mockito.stubbing.Answer;
 
 import org.springframework.amqp.AmqpException;
 import org.springframework.amqp.AmqpRejectAndDontRequeueException;
+import org.springframework.amqp.core.DirectExchange;
 import org.springframework.amqp.core.ExchangeTypes;
 import org.springframework.amqp.core.Message;
 import org.springframework.amqp.core.MessagePostProcessor;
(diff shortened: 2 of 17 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/Exchange.java
```diff
@@ -55,6 +55,17 @@
 	String autoDelete() default "false";
 
 	/**
+	 * @return true if the exchange is to be declared as internal.
+	 */
+	String internal() default "false";
```
artembilan: Ha! I almost added this option here when I was doing the original feature.
But than I decided to stay off it, just because an `internal` exchange does not make sense on the `@RabbitListener`, when we really would like to have some `binding` to the queue to retrieve messages.
Therefore we do allow to use an `internal` exchange by the client.

Am I missing anything in the `internal` exchange concept?

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java
```diff
@@ -715,6 +716,27 @@ public String handleWithHeadersExchange(String foo) {
...
+		@RabbitListener(bindings = {
+				@QueueBinding(
+					value = @Queue,
+					exchange = @Exchange(value = "auto.internal", autoDelete = "true",
+								ignoreDeclarationExceptions = "true"),
+					key = "auto.internal.rk")}
+		)
+		public String handleWithInternalExchangeIgnore(String foo) {
+			return foo.toUpperCase();
+		}
```
artembilan: M-m-m. I don't see tests for these listeners...

src/reference/asciidoc/amqp.adoc
```diff
@@ -2952,9 +2956,11 @@ For example, if a `queue` already exists with no `time to live` argument, and yo
 
 By default, the `RabbitAdmin` will immediately stop processing all declarations when any exception occurs; this could cause downstream issues - such as a *listener container* failing to initialize because another queue (defined after the one in error) is not declared.
 
-This behavior can be modified by setting the `ignore-declaration-failures` attribute to `true` on the `RabbitAdmin`.
+This behavior can be modified by setting the `ignore-declaration-exceptions` attribute to `true` on the `RabbitAdmin`.
```
artembilan: Ha! That was bug :smile:

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/Exchange.java
garyrussell: I thought so too, but then I thought some more. Let's say the admin has set up exchange to exchange binding

`foo -> bar`

where bar is an internal exchange.

The admin only allows users to bind queues to bar. Hence, in order to avoid mismatched configuration errors, we need to support `internal` on the `@Exchange`.

Now, thinking some more, from a user's perspective, I shouldn't have to care about knowing the configuration of the exchange I am binding to (except, perhaps, its type).

I was thinking of adding another option to the `@Exchange` - `existing = "true"` which means "bind to an existing exchange - don't declare it".

But in the end, I decided it was easier to simply support `ignoreDeclarationExceptions`.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java
garyrussell: I am just testing that the context loads ok - before the BPP changes, the test fails to load the context due to the mismatched exchange.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/Exchange.java
artembilan: OK. Merging as is

PR #421, opened 2016-05-17
GH-420: AcknowledgeMode JavaDocs

### LINKED ISSUE #420: quick question...hopefully
Maybe the method is poorly named or I maybe I don't fully understand the domain but....this logic looked suspicious to me:

https://github.com/spring-projects/spring-amqp/blob/master/spring-amqp/src/main/java/org/springframework/amqp/core/AcknowledgeMode.java#L32

Shouldn't the comparison be against AUTO instead of NONE?

garyrussell:
Fixes #420

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/core/AcknowledgeMode.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/core/AcknowledgeMode.java
@@ -17,21 +17,53 @@
 package org.springframework.amqp.core;
 
 /**
+ * Acknowledgment modes supported by the listener container.
+ *
  * @author Dave Syer
  *
  */
 public enum AcknowledgeMode {
 
-	NONE, MANUAL, AUTO;
+	/**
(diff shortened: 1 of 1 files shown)


spring-amqp/src/main/java/org/springframework/amqp/core/AcknowledgeMode.java
```diff
@@ -17,21 +17,53 @@
...
+ * Acknowledgment modes supported by the listener container.
+ *
  * @author Dave Syer
  *
  */
 public enum AcknowledgeMode {
 
-	NONE, MANUAL, AUTO;
+	/**
+	 * No acks - {@code autoAck=true} in {@code Channe.basicConsume()}.
```
artembilan: Typo: Channe**l**
Will fix on merge together with the `@author` effort respect :smile:

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/0ebb5529eb709752db4cf8bddb8d2081126fcd06

PR #425, opened 2016-05-25
AMQP-612: RabbitTemplate UserId Expression

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-612

https://www.rabbitmq.com/validated-user-id.html

Also, map the incoming user id to a received property to avoid unintended
propagation.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
@@ -71,6 +71,7 @@ import org.springframework.beans.factory.BeanFactoryAware;
 import org.springframework.context.expression.BeanFactoryResolver;
 import org.springframework.context.expression.MapAccessor;
 import org.springframework.expression.Expression;
+import org.springframework.expression.spel.standard.SpelExpressionParser;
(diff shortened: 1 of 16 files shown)


garyrussell:
Added more docs

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -552,6 +557,30 @@ public void setUseTemporaryReplyQueues(boolean value) {
...
+	public void setUserIdExpression(Expression userIdExpression) {
+		this.userIdExpression = userIdExpression;
+	}
+
+	/**
+	 * Set an expression to be evaluated to set the userId message property if it
+	 * evaluates to a non-null value and the property is not already set in the
+	 * message to be sent.
+	 * See https://www.rabbitmq.com/validated-user-id.html
+	 * @param userIdExpression the expression.
```
artembilan: There is no such a method parameter.
I think IDEA is good with this inspection :smile: 
Will rename during merge, if only concern.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/44a31245a4ee18df4971f19373574cbc1bcca39e.
See commit comment

PR #427, opened 2016-05-30
AMQP-614: Add Builder for Queue/Exchange

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-614

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/core/AbstractBuilder.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/core/AbstractBuilder.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.amqp.core;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/**
+ * Base class for builders supporting arguments.
+ *
+ * @author Gary Russell
+ * @since 1.6
+ *
+ */
+public abstract class AbstractBuilder {
+
+	private Map<String, Object> arguments;
+
+	/**
+	 * Return the arguments map, after creating one if necessary.
+	 * @return the arguments.
+	 */
+	protected Map<String, Object> getOrCreateArguments() {
+		if (this.arguments == null) {
+			this.arguments = new LinkedHashMap<String, Object>();
+		}
+		return this.arguments;
+	}
+
+	protected Map<String, Object> getArguments() {
+		return this.arguments;
+	}
+
+}
--- a/spring-amqp/src/main/java/org/springframework/amqp/core/ExchangeBuilder.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/core/ExchangeBuilder.java
@@ -1,0 +1,162 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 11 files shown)


artembilan:
Well, I'd prefer to have two top level builders, `QueueBuilder` and `ExchangeBuilder`. Even if they inherit from one parent with common options.
Another concern is generics issue using the inheritance. I think you would like to let to chain all those option methods with any order. So, the parent `Builder` class should have one more generic argument for the target implementor type. All the super chain methods should return that generic type, so in the typing time the IDE will know for which type is the next method to suggest.
You can find such an approach if the Spring Integraiton Java DSL. For example: https://github.com/spring-projects/spring-integration-java-dsl/blob/master/src/main/java/org/springframework/integration/dsl/core/MessageProducerSpec.java.

Another concern is the `.queue()`  factory method if we expose `QueueBuilder` as a top-level API.
Yes, I agree that it will looks a bit bizarre - `QueueBuilder.queue()`. So, maybe we can end up just with `new QueueBuilder()/new QueueBuilder()`? Or `QueueBuilder.anonymous()`, `.forName()`?
Any way would be fine with me. I just don't like this so generic builder name as top-level `Builder` API...

garyrussell:
Yes, that's why I narrowed the superclass methods to the specific builder type.

@maciejwalkowiak solved the `QueuBuilder.queue()` weirdness with `.durable()`, `nonDurable()` which seems like a good approach to me (but in that case I think we should remove the instance `durable()` method.

I will rebase on top of his commit and polish the exchange builder to be a stand-alone class.

maciejwalkowiak:
Thanks! 👍

garyrussell:
Pushed; @artembilan please don't squash when merging so that @maciejwalkowiak gets credit for his work.

spring-amqp/src/main/java/org/springframework/amqp/core/AbstractBuilder.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.Map;
+
+/**
+ * Base class for builders supporting arguments.
+ *
+ * @author Gary Russell
+ * @since 1.6
+ *
+ */
+public abstract class AbstractBuilder {
```
artembilan: I wonder if this all new API should be moved into `support` package.

spring-amqp/src/main/java/org/springframework/amqp/core/ExchangeBuilder.java
```diff
@@ -0,0 +1,162 @@
...
+	 * @return the builder.
+	 */
+	public ExchangeBuilder withArgument(String key, Object value) {
+		getOrCreateArguments().put(key, value);
+		return this;
+	}
+
+	/**
+	 * Add arguments.
+	 * @param the arguments arguments map.
```
artembilan: Missed `arguments` param.
Will fix on merge.

spring-amqp/src/main/java/org/springframework/amqp/core/QueueBuilder.java
```diff
@@ -0,0 +1,143 @@
...
+
+	private boolean exclusive;
+
+	private boolean autoDelete;
+
+	/**
+	 * Creates a builder for a durable queue with a generated
+	 * unique name - spring.gen-<random>.
+	 *
+	 * @param name the name of the queue.
```
artembilan: No param `name`

spring-amqp/src/main/java/org/springframework/amqp/core/ExchangeBuilder.java
garyrussell: Oops - inserted `the` in the wrong place 😄

spring-amqp/src/main/java/org/springframework/amqp/core/QueueBuilder.java
```diff
@@ -0,0 +1,143 @@
...
+	 * @return The Builder.
+	 */
+	public static QueueBuilder durable() {
+		return durable(namingStrategy.generateName());
+	}
+
+	/**
+	 * Creates a builder for a non-durable (transient) queue.
+	 *
+	 * @param name the name of the queue.
```
artembilan: DITO

spring-amqp/src/main/java/org/springframework/amqp/core/AbstractBuilder.java
garyrussell: I thought about that but a) the MessageBuilder is here and b) it will cause a package tangle (template references support) - and support will reference core - we generally don't have support classes reference other packages.

artembilan: Good. I see `import org.springframework.amqp.support.converter.MessageConverter;` in the `org.springframework.amqp.core.AmqpTemplate`.
So, fine now with me. 
Merging with all those minor polishings.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/225e72f89e976fd3422bcc68ff2964ce5e4a8833 and https://github.com/spring-projects/spring-amqp/commit/8f0a4a88c026c9c904dfb6639de5fae87afee0a2.

@maciejwalkowiak , thank you for contribution!
Just in time for `1.6.0.RELEASE` :smile: 

Looking forward for more!

PR #429, opened 2016-06-22
AMQP-619: Populate Bean CL into JSON converters

artembilan:
JIRA: https://jira.spring.io/browse/AMQP-619

Previously `DefaultJackson2JavaTypeMapper` used `getClass().getClassLoader()` to load class by its name from `MessageProperties`.
Sometimes it causes problems if we are based on the reloadable context like Spring Boot DevTools.

* Implement `BeanClassLoaderAware` in the `AbstractJavaTypeMapper` and `AbstractJsonMessageConverter`
* Delegate `classLoader` from the `AbstractJsonMessageConverter` bean to its internal `AbstractJavaTypeMapper` instance
* Use populated `classLoader` to instantiate classes for `Type` `MessageProperties`
* Add mock test to verify that BeanClassLoader is populated into internal `AbstractJavaTypeMapper`

The solution verified with provided in the JIRA application via local artifacts `install`

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/support/converter/AbstractJavaTypeMapper.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/support/converter/AbstractJavaTypeMapper.java
@@ -21,6 +21,8 @@ import java.util.HashMap;
 import java.util.Map;
 
 import org.springframework.amqp.core.MessageProperties;
+import org.springframework.beans.factory.BeanClassLoaderAware;
+import org.springframework.util.ClassUtils;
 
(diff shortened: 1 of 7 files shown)


artembilan:
Note, the fix isn't compatible with `1.5.x`: the `Jackson2JsonMessageConverter.typeMapperSet` has been introduced in `1.6`.

So, maybe we can revise it on demand?

garyrussell:
Merged as 0e042b0cc232125b2ee889de98c93d8922aa4fca with [this polish](https://github.com/garyrussell/spring-amqp/commit/ad9633312fe40c66b19ae2ac45eddc9192fd28c5).

PR #431, opened 2016-07-06
Extend toString method of MessageProperties

dchernyshov:
Prior to this commit, MessageProperties#toString printed correlationId and did not print correlationIdString. In this commit, correlationIdString is added to toString in order to make better logs readability.

I have signed and agree to the terms of the Spring Individual Contributor
License Agreement.


artembilan:
Good catch!
Merging...
Thank you for contribution.

Looking forward for more!

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/185fe59f7dbec1763642c3411698a9aebb8ca02d.
Pay attention that I have found more missed properties.

PR #432, opened 2016-07-13
AMQP-621: Fix Consumer.basicCancel race condition

artembilan:
JIRA: https://jira.spring.io/browse/AMQP-621

The `SimpleMessageListenerContainer` has logic to initiate `BlockingQueueConsumer.basicCancel()` and mark `BlockingQueueConsumer` as inactive immediately.
But the `handleCancelOk` answer can come back from the Broker a bit late, in async manner.
During this time window we can still receive messages via `handleDelivery`.
With `prefetchCount = 1` we may end up with the deadlock on `handleDelivery` thread during that time window, when main consumer loop identifies consumer as inactive and exits for `stop` process
and can't do that because the `BlockingQueueConsumer.this.queue.put(new Delivery(consumerTag, envelope, properties, body))` operation blocks the consumer from any other operations.

* Fix the race condition via additional `BlockingQueueConsumer.cancelled()` state to let main loop to spin until `handleCancelOk` honoring any in-flight messages
* The `cancelled` state is based on the `cancelReceived` flag which is changed from the `handleCancelOk` as a Broker response for the `basicCancel`
and in the `handleCancel` event from the Broker as a reason of some unexpected state on Broker, e.g. consuming queue has been removed
* Protect message lost via `basicReject` in case of late arrival into `handleDelivery`, when the consumer has been cancelled already

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
@@ -97,7 +97,11 @@ public class BlockingQueueConsumer {
 
 	private InternalConsumer consumer;
 
-	private final AtomicBoolean cancelReceived = new AtomicBoolean(false);
+	/**
+	 * The flag indicating that consumer has been cancelled from all queues
+	 * via {@code handleCancelOk} callback replies.
+	 */
+	private final AtomicBoolean cancelled = new AtomicBoolean(false);
 
 	private final AcknowledgeMode acknowledgeMode;
 
@@ -323,7 +327,7 @@ public class BlockingQueueConsumer {
 	}
 
 	protected boolean cancelled() {
-		return this.cancelReceived.get();
+		return this.cancelled.get();
 	}
 
 	/**
@@ -390,7 +394,7 @@ public class BlockingQueueConsumer {
 			checkMissingQueues();
 		}
 		Message message = handle(this.queue.poll(timeout, TimeUnit.MILLISECONDS));
-		if (message == null && this.cancelReceived.get()) {
+		if (message == null && this.cancelled.get()) {
 			throw new ConsumerCancelledException();
 		}
 		return message;
@@ -582,7 +586,7 @@ public class BlockingQueueConsumer {
 
 	public void stop() {
 		if (this.consumer != null && this.consumer.getChannel() != null && this.consumerTags.size() > 0
-				&& !this.cancelReceived.get()) {
+				&& !this.cancelled.get()) {
 			try {
 				RabbitUtils.closeMessageConsumer(this.consumer.getChannel(), this.consumerTags.keySet(),
 						this.transactional);
@@ -739,7 +743,7 @@ public class BlockingQueueConsumer {
 				logger.warn("Cancel received for " + consumerTag + "; " + BlockingQueueConsumer.this);
 			}
 			BlockingQueueConsumer.this.consumerTags.remove(consumerTag);
-			BlockingQueueConsumer.this.cancelReceived.set(true);
+			basicCancel();
 		}
 
 		@Override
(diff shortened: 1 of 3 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
```diff
@@ -97,8 +97,6 @@
 
 	private InternalConsumer consumer;
 
-	private final AtomicBoolean cancelled = new AtomicBoolean(false);
-
 	private final AtomicBoolean cancelReceived = new AtomicBoolean(false);
```
garyrussell: Since this is now overloaded to mean either a `cancel` or `cancelOk` has been received, we should probably rename it to something like `consumerCancelled` instead and add javadoc `cancel or canceOk received from broker`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
```diff
@@ -749,22 +747,30 @@ public void handleCancelOk(String consumerTag) {
...
 		@Override
 		public void handleDelivery(String consumerTag, Envelope envelope, AMQP.BasicProperties properties, byte[] body)
 				throws IOException {
-			if (logger.isDebugEnabled()) {
-				logger.debug("Storing delivery for " + BlockingQueueConsumer.this);
-			}
-			try {
-				BlockingQueueConsumer.this.queue.put(new Delivery(consumerTag, envelope, properties, body));
+			long deliveryTag = envelope.getDeliveryTag();
+			if (cancelled()) {
```
garyrussell: I don't believe this is necessary - we'll never get a delivery after a `cancel` or `cancelOk`, but I suppose it doesn't hurt, except for the (small) overhead.

garyrussell:
I think it's a bit more complicated than this...

When we receive a `basicCancel` for one of the consumer tags, we need to abort the consumer - including cancelling all the other consumer tags (call `basicCancel()` like we do in `queuesChanged()` - in fact, deleting a queue from the broker is quite similar to changing the containers queues manually.

When we issue `basicCancel`s for each consumer in `basicCancel()`  , we must return `false` from `cancelled()` while `this.consumerTags.size() > 0` - i.e. we need to keep the loop running until all the consumer tags have have been successfully cancelled.

artembilan:
Gary, please, take a look into change. I am totally agree with you and that was my original intention of changes.
I need some confirmation from you because we have some test failed, which looks like fully rely on the current state, but they are not so easy to fix:
```
org.springframework.amqp.rabbit.connection.LocalizedQueueConnectionFactoryTests > testFailOver FAILED
    java.lang.AssertionError at LocalizedQueueConnectionFactoryTests.java:132
                                                                                  
org.springframework.amqp.rabbit.listener.ContainerInitializationTests > testMismatchedQueueDuringRestart FAILED
    java.lang.AssertionError at ContainerInitializationTests.java:94
                                                                                  
org.springframework.amqp.rabbit.listener.MessageListenerRecoveryCachingConnectionIntegrationTests > testSingleListenerDoesRecoverFromMissingQueueWhenNotFatalGlobalProps FAILED
    java.lang.AssertionError at MessageListenerRecoveryCachingConnectionIntegrationTests.java:440
                                                                                  
org.springframework.amqp.rabbit.listener.MessageListenerRecoveryCachingConnectionIntegrationTests > testSingleListenerDoesRecoverFromMissingQueueWhenNotFatal FAILED
    java.lang.AssertionError at MessageListenerRecoveryCachingConnectionIntegrationTests.java:440
                                                                                  
org.springframework.amqp.rabbit.listener.SimpleMessageListenerContainerTests > testConsumerCancel FAILED
    java.lang.AssertionError at SimpleMessageListenerContainerTests.java:573                                                                                
```

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
```diff
@@ -323,7 +327,7 @@ protected boolean hasDelivery() {
 	}
 
 	protected boolean cancelled() {
-		return this.cancelReceived.get();
+		return this.cancelled.get();
```
garyrussell: Must return `false` until `consumerTags.size() == 0`.

garyrussell:
Polishing here https://github.com/garyrussell/spring-amqp/commit/5891211bf08d43aeaded67384f415bc7ac6655f1

The problem was you never set `cancelled` for the single queue case.

I also added some more protection.

artembilan:
> The problem was you never set cancelled for the single queue case.

Doh... :disappointed: 

Thank you for catching that and for the polishing!

Now I confirm that all the tests pass with your fix.

Feel free to merge.

garyrussell:
I added another commit (test)... https://github.com/garyrussell/spring-amqp/commit/881b54b66eda1d08fc6aec2e90f773e3393c79bf

artembilan:
> I added another commit

Fine with me.
Thank you!

garyrussell:
Merged as 8a520b34258d83f771bdaf9efde19ec005c53ce0

PR #433, opened 2016-07-15
AMQP-618: Add `contentType` note into JSON Docs

artembilan:
JIRA: https://jira.spring.io/browse/AMQP-618

Also, in the `Jackson2JsonMessageConverter`, wrap `log.warn()` with String concatenation into `log.isWarnEnabled()`


garyrussell:
Merged as 999090f63df91883fd3c5090176d7a67a6879f7c with minor polish.

PR #434, opened 2016-07-15
AMQP-623: Don't eat errors in AbstractAdaptableML

artembilan:
JIRA: https://jira.spring.io/browse/AMQP-623

Previously the `AbstractAdaptableMessageListener.onMessage()` just logged target listener errors without any re-throws to the container.
That makes any default listeners (`MessagingMessageListenerAdapter`) not so useful for MAQP protocol handling.

* Do not use `handleListenerException(Throwable ex)` in the `AbstractAdaptableMessageListener#onMessage(Message)` for consistency with `onMessage(Message, Channel)`.
Any custom implementation may decide to override those method and use `handleListenerException(Throwable ex)` at their own risk.


garyrussell:
Rebased, merged as b338a01ebd4c3a2e99770b63f7aa0550405269a0

PR #438, opened 2016-07-28
AMQP-628: Master to 2.0, SF 5.0

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-628

Remove log4j Appender (deprecated in 1.6.2).


src/reference/asciidoc/whats-new.adoc
```diff
@@ -1,229 +1,13 @@
 [[whats-new]]
 === What's New
 
-==== Changes in 1.6 Since 1.5
+==== Changes in 2.6 Since 1.5
```
artembilan: Typo: in *2.0* :smile: 

Also I wonder if you are going to address XSD moving to `2.0` as a part of this PR as well .

garyrussell: Oops; yes, of course.

garyrussell:
Pushed

build.gradle
```diff
@@ -79,8 +79,8 @@ subprojects { subproject ->
 	}
 
 	// ensure JDK 6 compatibility
```
artembilan: This isn't true any more :smile: 
Or change it to `8` or just remove altogether!

garyrussell: Remove - thanks

artembilan:
Since we don't have `log4j` code any more, we don't need its dependency. Even if it is `optional`.

I can remove on merge, of course, but would like to hear from you about a general our logging system in project. Right now it is:
```
testCompile "log4j:log4j:$log4jVersion" (1.2.17)
```

SF-5.0 has moved to log4j 2 already.

Different JIRA or we can consider it here as well?

garyrussell:
Yeah - separate JIRA I guess - I was going to suggest going to logback to be consistent with boot but if SF uses log4j2 ...

artembilan:
Done: https://jira.spring.io/browse/AMQP-630

Merging this one with those minor concerns in the `build.gradle`.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/2ee8c34a8cea2c7f6706943d8092e7d9996ba4fa

PR #441, opened 2016-08-07
AMQP-624: Add SSL Support to Logging Appenders

steve-oakey:
JIRA: https://jira.spring.io/browse/AMQP-624

Adds the ability to configure the log4j2 and Logback
AMQP appenders to use SSL. As it uses a RabbitConnectionFactoryBean
to configure create the ConnectionFactory, the same options
when configuring SSL for the factory bean are available and
the same default values are used for the SSL configuration.


artembilan:
@steve-oakey , thank you for contribution!
I think we can live with your fix, although it's enough difficult to review because there are a lot of code format changing. 
Let us know if that won't be hurt for your to revert that. Especially line length to `120`.
Also keep in mind that we decided to move to `JKS` as a default key(trust)Store type in the current version `2.0`.
Thanks

steve-oakey:
I should be able to fix up the line length. I will also change the default
to JKS. I'll try to push that update this evening.

-Steve

steve-oakey:
@artembilan - OK this diff should be better to look at, and I changed the default keystore types in both appenders to `JKS`. I had to do a bit of manual editing to fix the line wrapping... I could never quite seem to get it correct just using just the preferences from the Spring Framework repository.

garyrussell:
@steve-oakey Sorry for the pain - I use STS/eclipse and will export my current code formatting settings and post them in the project.

Thanks for your contribution.

I don't recall why I put those tests in the `test` sub-package; but I must have originally had some reason to do so, perhaps for some long-gone test case.

artembilan:
> I don't recall why I put those tests in the test sub-package; 

I think that was done for reason to be sure about methods visibility. Seems for me there were some setters in the previous `AmqpAppender` without `public` modifier :smile: .

garyrussell:
Yes; I am sure that was it.

Updated eclipse settings:

https://github.com/spring-projects/spring-amqp/commit/eed2851e60be14a175b84a5ef3f3979ed85833b3

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
```diff
@@ -484,6 +547,58 @@ private void activateOptions() {
...
+		 */
+		protected ConnectionFactory createRabbitConnectionFactory() {
+			RabbitConnectionFactoryBean factoryBean = new RabbitConnectionFactoryBean();
+			configureRabbitConnectionFactory(factoryBean);
+			try {
+				factoryBean.afterPropertiesSet();
+				return factoryBean.getObject();
+			}
+			catch (Exception e) {
+				LogFactory.getLog("log4j2AppenderErrors").error("Failed to create customized Rabbit ConnectionFactory.",
```
artembilan: Would you mind explaining this code a bit?
I guess it is something global for appenders/loggers problems.
Aren't there some public API or constant do not remember this String literal?

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/logback/AmqpAppenderTest.java
```diff
@@ -0,0 +1,182 @@
...
+import org.springframework.amqp.rabbit.connection.RabbitConnectionFactoryBean;
+import org.springframework.core.io.ClassPathResource;
+
+/**
+ *
+ * @author Stephen Oakey
+ *
+ * @since 2.0
+ */
+public class AmqpAppenderTest {
```
artembilan: Our code style is like `*Tests`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/logback/AmqpAppender.java
```diff
@@ -435,8 +556,60 @@ public void start() {
...
+	protected ConnectionFactory createRabbitConnectionFactory() {
+		RabbitConnectionFactoryBean factoryBean = new RabbitConnectionFactoryBean();
+		configureRabbitConnectionFactory(factoryBean);
+		try {
+			factoryBean.afterPropertiesSet();
+			return factoryBean.getObject();
+		}
+		catch (Exception e) {
+			addError("Failed to create customized Rabbit ConnectionFactory.", e);
+			return null;
```
artembilan: I'm not good in all those logging tricks, but I would like to have some test for the wrong case, e.g.:
````java
@Test
public void testNotFullSslConfiguration() {
	AmqpAppender appender = new AmqpAppender();
	appender.setUseSsl(true);
	appender.setTrustStore("foo");

	appender.start();
}
````
Where I end up with:
```
LOGBACK: No context given for ch.qos.logback.core.pattern.parser.Compiler@3ac42916
LOGBACK: No context given for ch.qos.logback.core.pattern.parser.Compiler@629f0666
LOGBACK: No context given for org.springframework.amqp.rabbit.logback.AmqpAppender[null]

java.lang.IllegalArgumentException: Target ConnectionFactory must not be null

	at org.springframework.util.Assert.notNull(Assert.java:165)
	at org.springframework.amqp.rabbit.connection.AbstractConnectionFactory.<init>(AbstractConnectionFactory.java:76)
	at org.springframework.amqp.rabbit.connection.CachingConnectionFactory.<init>(CachingConnectionFactory.java:230)
	at org.springframework.amqp.rabbit.logback.AmqpAppender.start(AmqpAppender.java:545)
	at org.springframework.amqp.rabbit.logback.AmqpAppenderTest.testNotFullSslConfiguration(AmqpAppenderTest.java:180)
```
I understand that the first three `LOGBACK` warnings are about not full logback configuration.
But the next `must not be null` may cause some confuse in the target application.

With the real config I have something like this:
```
10:47:12,626 |-INFO in ch.qos.logback.core.joran.action.NestedComplexPropertyIA - Assuming default type [ch.qos.logback.classic.PatternLayout] for [layout] property
10:47:12,664 |-ERROR in org.springframework.amqp.rabbit.logback.AmqpAppenderIntegrationTests$EnhancedAppender[AMQP] - Failed to create customized Rabbit ConnectionFactory. java.lang.IllegalStateException: keyStore property required
	at java.lang.IllegalStateException: keyStore property required
	at 	at org.springframework.util.Assert.state(Assert.java:640)
	at 	at org.springframework.amqp.rabbit.connection.RabbitConnectionFactoryBean.setUpSSL(RabbitConnectionFactoryBean.java:484)
...
```
but anyway somewhere down the logs I see:
```
10:47:12,671 |-ERROR in ch.qos.logback.core.joran.spi.Interpreter@27:13 - RuntimeException in Action for tag [appender] java.lang.IllegalArgumentException: Target ConnectionFactory must not be null
	at java.lang.IllegalArgumentException: Target ConnectionFactory must not be null
	at 	at org.springframework.util.Assert.notNull(Assert.java:165)
	at 	at org.springframework.amqp.rabbit.connection.AbstractConnectionFactory.<init>(AbstractConnectionFactory.java:76)
	at 	at org.springframework.amqp.rabbit.connection.CachingConnectionFactory.<init>(CachingConnectionFactory.java:230)
	at 	at org.springframework.amqp.rabbit.logback.AmqpAppender.start(AmqpAppender.java:545)
```
The first one is good, but the second should be avoided.
Seems for me simple `if...else` in the `start()` should be fine.

WDYT?

File: spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java (hunk 1 above)
Comment (steve-oakey) [inline spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java]: I did not take the time to familiarize myself entirely with the log4j2 public API, but I used the same method this class already used to log appender errors as seen [here](https://github.com/spring-projects/spring-amqp/blob/master/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java#L144).

File: spring-rabbit/src/test/java/org/springframework/amqp/rabbit/logback/AmqpAppenderTest.java (hunk 2 above)
Comment (steve-oakey) [inline spring-rabbit/src/test/java/org/springframework/amqp/rabbit/logback/AmqpAppenderTest.java]: Will fix.

File: spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java (hunk 1 above)
Comment (artembilan) [inline spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java]: Makes sense. Thank you!

Comment (garyrussell) [inline spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java]: Not sure where I got that from; it looks like we should be using the `StatusAppender` - available as `LOGGER` in `AbstractLifecycle` (a super class of the appender) - I just took a look at the JmsAppender and it uses LOGGER.

File: spring-rabbit/src/main/java/org/springframework/amqp/rabbit/logback/AmqpAppender.java (hunk 3 above)
Comment (steve-oakey) [inline spring-rabbit/src/main/java/org/springframework/amqp/rabbit/logback/AmqpAppender.java]: The reason why I chose to return `null` and hence passing `null` to the constructor of the `CachingConnectionFactory` was to simply cause a `RuntimeException` that would stop the application from starting. I agree it would be better to avoid the second error message in this case (the `keyStore property required` message is enough). 

What would you propose the logic if the `if..else` statement in the `start()` method do? Would it just not proceed with setting up the appender while allowing the application to start? Or throw a variant of `RuntimeException` to cause the application to fail?

File: spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java (hunk 1 above)
Comment (steve-oakey) [inline spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java]: @garyrussell - Thanks.  I will make the change in both places when I make the required updates for this PR.

**Comment by artembilan** [comment]
> What would you propose the logic if the if..else statement in the start() method do? Would it just not proceed with setting up the appender while allowing the application to start? Or throw a variant of RuntimeException to cause the application to fail?

Well, I don't know the proper answer because not familiar withe whole logging system picture.

Right, now I'm looking into `JdbcAppender` and its logic is like:
```
 final JdbcDatabaseManager manager = JdbcDatabaseManager.getJDBCDatabaseManager(
                managerName.toString(), bufferSizeInt, connectionSource, tableName, columnConfigs
        );
        if (manager == null) {
            return null;
        }

        return new JdbcAppender(name, filter, ignoreExceptions, manager);
```
So, if `JdbcDatabaseManager`  hasn't been initialized, we just return `null` to the logger configuration.
Where a manager does this on the connection error:
```
  public final synchronized void startup() {
        if (!this.isRunning()) {
            try {
                this.startupInternal();
                this.running = true;
            } catch (final Exception e) {
                logError("could not perform database startup operations", e);
            }
        }
}
The `JmsAppender` does just this:
```
try {
                return new JmsAppender(name, filter, layout, ignoreExceptions, jmsManager);
            } catch (final JMSException e) {
                LOGGER.error("Error creating JmsAppender [{}].", name, e);
                return null;
            }
```
but it is in its `Builder implements org.apache.logging.log4j.core.util.Builder<JmsAppender>`.

Which might fail just with this in the end:
```
 LOGGER.error("Unable to create Appender of type " + node.getName());
```

So, would you mind to investigate how does it work if we don't throw RuntimeException?
Seems for me any logging system problems should not impact to the target application.
Consider the case when you don't use AMQP appender in the production, etc, etc.

steve-oakey:
This sounds like the correct solution. I will make the updates tonight.
Thanks for the help looking into the other appenders. I did not think to do
so.

steve-oakey:
OK I made the following changes:

- The Logback appender will not start if the ConnectionFactory is invalid (i.e. returns null)
- The log4j2 appender factory method will return null if the ConnectionFactory is invalid.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
```diff
@@ -463,25 +530,77 @@ protected AmqpManager(String name) {
...
+				this.senderPool = Executors.newCachedThreadPool();
+				return true;
 			}
-			this.connectionFactory.setUsername(this.username);
-			this.connectionFactory.setPassword(this.password);
-			this.connectionFactory.setVirtualHost(this.virtualHost);
-			if (this.clientConnectionProperties != null) {
-				LogAppenderUtils.updateClientConnectionProperties(this.connectionFactory,
-						this.clientConnectionProperties);
+			return false;
```
artembilan: This `false` isn't consulted in the `public static AmqpAppender createAppender()`:
```
AmqpAppender appender = new AmqpAppender(name, filter, theLayout, ignoreExceptions, manager);
manager.activateOptions();
appender.startSenders();
```
There we end up with the:
```
Caused by: java.lang.NullPointerException
	at org.springframework.amqp.rabbit.log4j2.AmqpAppender.startSenders(AmqpAppender.java:206)
	at org.springframework.amqp.rabbit.log4j2.AmqpAppender.createAppender(AmqpAppender.java:197)
```
Although I think we really have to return `null` from that `@PluginFactory` instead of uncompleted `AmqpAppender` .

artembilan: I added
```
useSsl="true"
trustStore="true"
```
to the `log4j2-amqp-appender.xml` and ran `AmqpAppenderTests`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/logback/AmqpAppender.java
```diff
@@ -409,34 +535,84 @@ public void setClientConnectionProperties(String clientConnectionProperties) {
...
-		this.locationLayout.setContext(getContext());
-		this.locationLayout.start();
-		this.connectionFactory = new CachingConnectionFactory();
-		this.connectionFactory.setHost(this.host);
-		this.connectionFactory.setPort(this.port);
-		if (this.addresses != null) {
-			this.connectionFactory.setAddresses(this.addresses);
+		ConnectionFactory rabbitConnectionFactory = createRabbitConnectionFactory();
+		if (rabbitConnectionFactory != null) {
+			super.start();
```
artembilan: Logback looks good:
```
15:36:01,343 |-ERROR in org.springframework.amqp.rabbit.logback.AmqpAppenderIntegrationTests$EnhancedAppender[AMQP] - Failed to create customized Rabbit ConnectionFactory. java.lang.IllegalStateException: keyStore property required
	at java.lang.IllegalStateException: keyStore property required
	at 	at org.springframework.util.Assert.state(Assert.java:640)
	at 	at org.springframework.amqp.rabbit.connection.RabbitConnectionFactoryBean.setUpSSL(RabbitConnectionFactoryBean.java:484)
	at 	at org.springframework.amqp.rabbit.connection.RabbitConnectionFactoryBean.createInstance(RabbitConnectionFactoryBean.java:458)
	at 	at org.springframework.amqp.rabbit.connection.RabbitConnectionFactoryBean.createInstance(RabbitConnectionFactoryBean.java:64)
	at 	at org.springframework.beans.factory.config.AbstractFactoryBean.afterPropertiesSet(AbstractFactoryBean.java:134)
	at 	at org.springframework.amqp.rabbit.logback.AmqpAppender.createRabbitConnectionFactory(AmqpAppender.java:570)
	at 	at org.springframework.amqp.rabbit.logback.AmqpAppender.start(AmqpAppender.java:538)
...
15:36:01,343 |-INFO in ch.qos.logback.classic.joran.action.LoggerAction - Setting level of logger [org.springframework.amqp.rabbit.logback] to DEBUG
15:36:01,343 |-INFO in ch.qos.logback.classic.joran.action.LoggerAction - Setting additivity of logger [org.springframework.amqp.rabbit.logback] to false
15:36:01,343 |-INFO in ch.qos.logback.core.joran.action.AppenderRefAction - Attaching appender named [AMQP] to Logger[org.springframework.amqp.rabbit.logback]
15:36:01,344 |-INFO in ch.qos.logback.classic.joran.action.RootLoggerAction - Setting level of ROOT logger to INFO
15:36:01,344 |-INFO in ch.qos.logback.core.joran.action.AppenderRefAction - Attaching appender named [CONSOLE] to Logger[ROOT]
15:36:01,344 |-INFO in ch.qos.logback.classic.joran.action.ConfigurationAction - End of configuration.
15:36:01,345 |-INFO in ch.qos.logback.classic.joran.JoranConfigurator@4e268090 - Registering current configuration as safe fallback point
```
At least I see analogy with `JMSQueueAppender`:
```
} catch (Exception e) {
      addError("Error while activating options for appender named [" + name
          + "].", e);
    }

    if (this.queueConnection != null && this.queueSession != null
        && this.queueSender != null) {
      super.start();
}
```
and `SMTPAppenderBase`:
```
if (session == null) {
      addError("Failed to obtain javax.mail.Session. Cannot start.");
      return;
    }
```
in the Logback out-of-the-box.

artembilan:
@steve-oakey , great work any way!

Would you mind in addition to those my desired fixes in the Log4J2 Appender to consider adding some note to the Reference Manual (`logging.adoc`). At least some short sentence about SSL support and a link to the `<<_rabbitconnectionfactorybean_and_configuring_ssl>>` for understanding appropriate options.

In addition I see that there is need some polishing in the `logging.adoc` because we don't support log4j1 any more. So, no need to mention it there. But, eh, that might not be your task already. Although feel free to contribute such an improvement as well! :smile: 

Note: please, stop to squash commits. We lose a history of your changes a bit. In addition it is much easier to review just actual changes, since we already know everything rest. We do squash before merging to `master`.

Thank you for your help and patience!

steve-oakey:
I should thank you guys for your continued patience! Sorry about the commit squashing, force of habit from our work flow at work. 

I'll try to fix everything up by the end of the weekend sometime. Stay tuned!

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
steve-oakey: I feel silly. I must have made a when re-basing the branch on master before pushing the changes back up. I most definitely had a check against `manager.activateOptions` and was returning `null` if false. Coding and watching the Olympics at the same time was probably not the best idea...

steve-oakey:
OK guys, hopefully I've got everything cleaned up here. Let me know if there is anything else.

steve-oakey:
OK guys, hopefully I've got everything cleaned up here. Let me know if there is anything else.

artembilan:
LGTM.
I'm ready to merge with simple overflow fix for the long `clientConnectionProperties` literal.

@garyrussell ?

garyrussell:
Yes; fine with me - @steve-oakey many thanks for the contribution.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/09f03c74e9f569f6917aea24abf581e0b9d4be2d

@steve-oakey,  thank you for the contribution.
Looking forward for more!

PR #443, opened 2016-08-12
Pending Confirms Count

Giri-Chintala:
Resolves https://jira.spring.io/browse/AMQP-632

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
@@ -622,6 +622,23 @@ public class RabbitTemplate extends RabbitAccessor implements BeanFactoryAware, 
 		return unconfirmed.size() > 0 ? unconfirmed : null;
 	}
 
+    /**
+     * Gets unconfirmed messages count.
+     *
+     * @return The count of the messages that are not confirmed yet by RabbitMQ.
+     */
+    public int getUnconfirmedCount() {
+        int unconfirmedCount = 0;
+
+        synchronized (this.publisherConfirmChannels) {
+            for (Channel channel : this.publisherConfirmChannels.keySet()) {
+                unconfirmedCount = unconfirmedCount
+                        + ((PublisherCallbackChannel) channel).getPendingConfirmsCount(this);
+            }
+        }
+        return unconfirmedCount;
+    }
+
 	private void evaluateFastReplyTo() {
 		this.usingFastReplyTo = useDirectReplyTo();
 		this.evaluatedFastReplyTo = true;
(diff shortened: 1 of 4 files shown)


pivotal-issuemaster:
@Giri-Chintala Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-amqp&pullRequestId=443)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-amqp&pullRequestId=443) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/faq) for frequently asked questions.

artembilan:
Hello @Giri-Chintala !

Sorry for delay and thank for contribution.
We don't forget about about and really appreciate the help.

From my point of view I don't see so much benefit of this functionality. Just because `PublisherCallbackChannelImpl.processAck` is done in separate Thread by the Broker initiative. So, any `number` as a result of this new method invocation might not reflect the reality, when during our logic new acks are arriving on background.

I may agree with more general solution like:
```
public Collection<CorrelationData> getUnconfirmed() {
...
}
```
with proper JavaDoc to explain that this method returns only a snapshot at an invocation time.
And the real confirmation picture may be different over seconds or so.

WDYT?

Giri-Chintala:
Sorry about the delayed response. 
i am trying to find a way to pause producer from sending messages or send messages to  alternative RabbitMQ server when number of pending confirms exceeds some limit. 

i agree, number as a result of this new method invocation might not reflect the reality, But same argument applies to `getUnconfirmed(long age)`.  
periodically checking `getUnconfirmedCount()` gives somes idea about how is the RabbitMQ server behaving. 

i agree with `Collection<CorrelationData> getUnconfirmed()` as long as unconfirms are not removed from the collection.

garyrussell:
@Giri-Chintala 

> But same argument applies to getUnconfirmed(long age).

Not really because of the `age` argument; but I agree, it's ok as long as the user realizes it's only approximate.

@artembilan - while I think it might be useful for the user to be able to determine exactly which confirms are pending, I am not sure we want to expose the actual `CorrelationData` object because the user could make changes to it, even by mistake.

I therefore think that, for these purposes, just exposing a count should be enough.

artembilan:
@garyrussell ,

>  I am not sure we want to expose the actual CorrelationData object because the user could make changes to it, even by mistake.

Actually we have similar API already - `RabbitTemplate`:
```
/**
* Gets unconfirmed correlation data older than age and removes them.
* @param age in milliseconds
* @return the collection of correlation data for which confirms have
* not been received or null if no such confirms exist.
*/
public Collection<CorrelationData> getUnconfirmed(long age) {
```
unless a difference that end-user mutation won't have effect already because they are removed already.

So, if you insist that it must be just `int` value, let's consider it as a solution and I'll come back here for the final review!

garyrussell:
>Actually we have similar API already - RabbitTemplate:

>unless a difference that end-user mutation won't have effect already because they are removed already.

Correct - that's a big difference; once they are removed they are no longer relevant so it's fine to give the user access. I am just wary of doing that when they are still active objects.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/a3f6b1746eebea8294b8adece418b12deb102432.

Pay attention to beautiful polishing for `RabbitTemplate.getUnconfirmedCount()`:
````java
return this.publisherConfirmChannels.keySet()
			        .stream()
			        .mapToInt(channel -> ((PublisherCallbackChannel) channel).getPendingConfirmsCount(this))
			        .sum();
````
Java 8 in the power :smile: 

Thank you for contribution and looking forward for more!

PR #444, opened 2016-08-18
AMQP-633: Fix Non-Transactional Template Issue

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-633

Previously, the `RabbitTemplate` running on a transactional container thread always
used the consumer channel, regardless of whether the template was transactional.

A non-transactional template should not use the container's transactional channel.

Also some polishing around passing the connection factory into `doExecute()`.

__cherry-pick to 1.6.x__

Also tested with a SCSt app that publishes before throwing an exception.


artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/b25a3d22e72065b28e8c773fe03c1bd8bed061b3 and cherry-picked.

PR #446, opened 2016-08-31
AMQP-634: DirectMessageListenerContainer

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-634

Initial commit for `DirectMessageListenerContainer` that runs the listener on
the rabbitmq client thread.

Replaces #380

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplate.java
@@ -207,7 +207,7 @@ public class RabbitManagementTemplate implements AmqpManagementOperations {
 
 	private List<Queue> converteQueueList(List<QueueInfo> queues) {
 		return queues.stream()
-				.map(qi -> convert(qi))
+				.map(this::convert)
 				.collect(Collectors.toList());
 	}
 
@@ -218,7 +218,7 @@ public class RabbitManagementTemplate implements AmqpManagementOperations {
 
 	private List<Exchange> convertExchangeList(List<ExchangeInfo> exchanges) {
 		return exchanges.stream()
-				.map(ei -> convert(ei))
+				.map(this::convert)
 				.collect(Collectors.toList());
 	}
 
@@ -243,7 +243,7 @@ public class RabbitManagementTemplate implements AmqpManagementOperations {
 
 	private List<Binding> convertBindingList(List<BindingInfo> bindings) {
 		return bindings.stream()
-				.map(bi -> convert(bi))
+				.map(this::convert)
 				.collect(Collectors.toList());
 	}
 
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplate.java
@@ -16,8 +16,8 @@
 
 package org.springframework.amqp.rabbit.core;
 
-import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import org.springframework.amqp.AmqpException;
 import org.springframework.amqp.core.AmqpManagementOperations;
@@ -206,11 +206,9 @@ public class RabbitManagementTemplate implements AmqpManagementOperations {
 	}
 
 	private List<Queue> converteQueueList(List<QueueInfo> queues) {
-		List<Queue> convertedQueues = new ArrayList<Queue>();
-		for (QueueInfo qi : queues) {
-			convertedQueues.add(convert(qi));
-		}
-		return convertedQueues;
+		return queues.stream()
+				.map(qi -> convert(qi))
+				.collect(Collectors.toList());
 	}
 
 	private Queue convert(QueueInfo qi) {
@@ -219,14 +217,9 @@ public class RabbitManagementTemplate implements AmqpManagementOperations {
 	}
 
 	private List<Exchange> convertExchangeList(List<ExchangeInfo> exchanges) {
-		List<Exchange> convertedExchanges = new ArrayList<Exchange>();
-		for (ExchangeInfo ei : exchanges) {
-			Exchange converted = convert(ei);
-			if (converted != null) {
-				convertedExchanges.add(converted);
-			}
-		}
-		return convertedExchanges;
+		return exchanges.stream()
+				.map(ei -> convert(ei))
+				.collect(Collectors.toList());
 	}
 
 	private Exchange convert(ExchangeInfo ei) {
@@ -249,11 +242,9 @@ public class RabbitManagementTemplate implements AmqpManagementOperations {
 	}
 
 	private List<Binding> convertBindingList(List<BindingInfo> bindings) {
-		List<Binding> convertedBindings = new ArrayList<Binding>();
-		for (BindingInfo bi : bindings) {
-			convertedBindings.add(convert(bi));
-		}
-		return convertedBindings;
+		return bindings.stream()
+				.map(bi -> convert(bi))
+				.collect(Collectors.toList());
 	}
 
 	private Binding convert(BindingInfo bi) {
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
@@ -261,17 +261,39 @@ public abstract class AbstractMessageListenerContainer extends RabbitAccessor
 	}
 
 	/**
-	 * Set the message listener implementation to register. This can be either a Spring {@link MessageListener} object
-	 * or a Spring {@link ChannelAwareMessageListener} object.
+	 * Set the message listener implementation to register. This can be either a Spring
+	 * {@link MessageListener} object or a Spring {@link ChannelAwareMessageListener}
+	 * object. Using the strongly typed
+	 * {@link #setChannelAwareMessageListener(ChannelAwareMessageListener)} is preferred.
 	 *
 	 * @param messageListener The listener.
-	 * @throws IllegalArgumentException if the supplied listener is not a {@link MessageListener} or a
-	 * {@link ChannelAwareMessageListener}
+	 * @throws IllegalArgumentException if the supplied listener is not a
+	 * {@link MessageListener} or a {@link ChannelAwareMessageListener}
 	 * @see MessageListener
 	 * @see ChannelAwareMessageListener
 	 */
 	public void setMessageListener(Object messageListener) {
 		checkMessageListener(messageListener);
+		this.messageListener = messageListener;
+	}
+
+	/**
+	 * Set the {@link MessageListener}; strongly typed version of
+	 * {@link #setMessageListener(Object)}.
+	 * @param messageListener the listener.
+	 * @since 2.0
+	 */
+	public void setMessageListener(MessageListener messageListener) {
+		this.messageListener = messageListener;
+	}
+
+	/**
+	 * Set the {@link ChannelAwareMessageListener}; strongly typed version of
+	 * {@link #setMessageListener(Object)}.
+	 * @param messageListener the listener.
+	 * @since 2.0
+	 */
+	public void setChannelAwareMessageListener(ChannelAwareMessageListener messageListener) {
 		this.messageListener = messageListener;
 	}
 
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
@@ -46,7 +46,6 @@ import org.springframework.amqp.rabbit.listener.exception.ListenerExecutionFaile
 import org.springframework.amqp.support.converter.MessageConversionException;
 import org.springframework.amqp.support.converter.MessageConverter;
 import org.springframework.amqp.support.postprocessor.MessagePostProcessorUtils;
-import org.springframework.aop.Pointcut;
 import org.springframework.aop.framework.ProxyFactory;
 import org.springframework.aop.support.DefaultPointcutAdvisor;
 import org.springframework.beans.factory.BeanNameAware;
@@ -362,7 +361,7 @@ public abstract class AbstractMessageListenerContainer extends RabbitAccessor
 	 * provided here should not contain a transaction interceptor (otherwise two transactions would be be applied).
 	 * @param adviceChain the advice chain to set
 	 */
-	public void setAdviceChain(Advice[] adviceChain) {
+	public void setAdviceChain(Advice... adviceChain) {
 		this.adviceChain = Arrays.copyOf(adviceChain, adviceChain.length);
 	}
 
@@ -501,9 +500,8 @@ public abstract class AbstractMessageListenerContainer extends RabbitAccessor
 		}
 		ProxyFactory factory = new ProxyFactory();
 		for (Advice advice : getAdviceChain()) {
-			factory.addAdvisor(new DefaultPointcutAdvisor(Pointcut.TRUE, advice));
+			factory.addAdvisor(new DefaultPointcutAdvisor(advice));
 		}
-		factory.setProxyTargetClass(false);
 		factory.addInterface(ContainerDelegate.class);
 		factory.setTarget(delegate);
 		this.proxy = (ContainerDelegate) factory.getProxy(ContainerDelegate.class.getClassLoader());
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
@@ -71,6 +71,10 @@ public abstract class AbstractMessageListenerContainer extends RabbitAccessor
 
 	public static final boolean DEFAULT_DEBATCHING_ENABLED = true;
 
+	private final ContainerDelegate delegate = this::actualInvokeListener;
+
+	private ContainerDelegate proxy = this.delegate;
+
 	private volatile String beanName;
 
 	private volatile boolean autoStartup = true;
@@ -469,9 +473,9 @@ public abstract class AbstractMessageListenerContainer extends RabbitAccessor
 	protected void validateConfiguration() {
 	}
 
-	protected ContainerDelegate initializeProxy(Object delegate) {
+	protected void initializeProxy(Object delegate) {
 		if (this.getAdviceChain().length == 0) {
-			return null;
+			return;
 		}
 		ProxyFactory factory = new ProxyFactory();
 		for (Advice advice : getAdviceChain()) {
@@ -480,7 +484,7 @@ public abstract class AbstractMessageListenerContainer extends RabbitAccessor
 		factory.setProxyTargetClass(false);
 		factory.addInterface(ContainerDelegate.class);
 		factory.setTarget(delegate);
-		return (ContainerDelegate) factory.getProxy(ContainerDelegate.class.getClassLoader());
+		this.proxy = (ContainerDelegate) factory.getProxy(ContainerDelegate.class.getClassLoader());
 	}
 
 	/**
@@ -506,6 +510,7 @@ public abstract class AbstractMessageListenerContainer extends RabbitAccessor
 			synchronized (this.lifecycleMonitor) {
 				this.lifecycleMonitor.notifyAll();
 			}
+			initializeProxy(this.delegate);
 			doInitialize();
 		}
 		catch (Exception ex) {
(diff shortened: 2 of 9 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -446,6 +469,20 @@ public void setupMessageListener(Object messageListener) {
 	protected void validateConfiguration() {
 	}
 
+	protected ContainerDelegate initializeProxy(Object delegate) {
+		if (this.getAdviceChain().length == 0) {
+			return null;
+		}
+		ProxyFactory factory = new ProxyFactory();
+		for (Advice advice : getAdviceChain()) {
+			factory.addAdvisor(new DefaultPointcutAdvisor(Pointcut.TRUE, advice));
```
artembilan: This is redundant. Another ctor does exactly that:
```
public DefaultPointcutAdvisor(Advice advice) {
	this(Pointcut.TRUE, advice);
}
```

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -872,6 +909,12 @@ protected Exception wrapToListenerExecutionFailedExceptionIfNeeded(Exception e,
 		return e;
 	}
 
+	public interface ContainerDelegate {
```
artembilan: `@FunctionalInterface`? :smile: 
Although who uses it outside of the container...
Maybe remove that `public` modifier? Or even make it `protected` ...

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -0,0 +1,244 @@
...
+ * @since 2.0
+ *
+ */
+public class DirectMessageListenerContainer extends AbstractMessageListenerContainer {
+
+	private final List<SimpleConsumer> consumers = new ArrayList<SimpleConsumer>();
+
+	private final MessagePropertiesConverter messagePropertiesConverter = new DefaultMessagePropertiesConverter();
+
+	private final ContainerDelegate delegate = new ContainerDelegate() {
```
artembilan: Method reference:
```
private final ContainerDelegate delegate = super::invokeListener;
```
?
And looks like we can move it to the `AbstractMessageListenerContainer` and really make `ContainerDelegate` as `private`!

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -0,0 +1,244 @@
...
+		ContainerDelegate proxy = initializeProxy(this.delegate);
+		if (proxy != null) {
+			this.proxy = proxy;
+		}
+	}
+
+	@Override
+	protected void invokeListener(Channel channel, Message message) throws Exception {
+		this.proxy.invokeListener(channel, message);
+	}
```
artembilan: Looks like this method is fully equal to the same in the `SimpleMessageListenerContainer`.
We really can move it to the `AbstractMessageListenerContainer`.
Great work!

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/ReplyingMessageListener.java
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * A Message Listener that returns a reply - intended for lambda use in a
+ * {@link MessageListenerAdapter}.
+ *
+ * @author Gary Russell
+ * @since 2.0
+ *
+ */
+@FunctionalInterface
+public interface ReplyingMessageListener<T, R> {
```
artembilan: We can rebase now to the `master` and avoid this here.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -322,6 +332,23 @@ public void setDeBatchingEnabled(boolean deBatchingEnabled) {
...
 	/**
+	 * Public setter for the {@link Advice} to apply to listener executions. If {@link #setTxSize(int) txSize>1} then
+	 * multiple listener executions will all be wrapped in the same advice up to that limit.
+	 * <p>
+	 * If a {code #setTransactionManager(PlatformTransactionManager) transactionManager} is provided as well, then
+	 * separate advice is created for the transaction and applied first in the chain. In that case the advice chain
+	 * provided here should not contain a transaction interceptor (otherwise two transactions would be be applied).
+	 * @param adviceChain the advice chain to set
+	 */
+	public void setAdviceChain(Advice[] adviceChain) {
```
artembilan: Can't we make this method as `varargs` instead ?
Just looking to your extra code in the test :smile:

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -0,0 +1,226 @@
...
+	/**
+	 * Create an instance with the provided connection factory.
+	 * @param connectionFactory the connection factory.
+	 */
+	public DirectMessageListenerContainer(ConnectionFactory connectionFactory) {
+		setConnectionFactory(connectionFactory);
+	}
+
+	@Override
+	public final void setConnectionFactory(ConnectionFactory connectionFactory) {
```
artembilan: Why do we need override this method? To add `final`? What is the reason?
Thanks

artembilan: Ah! Called from ctor. 
can't we then just make `RabbitAccessor.setConnectionFactory()` as `final` already in the version `2.0` ?

garyrussell: I have no idea where that came from 😄

garyrussell: Yes; I guess we can do that.

garyrussell: Yes; I guess we can do that - remember, I first was going to put this in 1.6 😄

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -0,0 +1,226 @@
...
+			this.queue = queue;
+			this.ackRequired = !getAcknowledgeMode().isAutoAck() && !getAcknowledgeMode().isManual();
+		}
+
+		@Override
+		public void handleDelivery(String consumerTag, Envelope envelope,
+				BasicProperties properties, byte[] body) throws IOException {
+			MessageProperties messageProperties = DirectMessageListenerContainer.this.messagePropertiesConverter
+					.toMessageProperties(properties, envelope, "UTF-8");
+			messageProperties.setMessageCount(0);
```
artembilan: What does this `messageCount` property mean?
We have delivery but mark it as `0`. No JavaDocs as well.
Thanks

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainerTests.java
```diff
@@ -0,0 +1,109 @@
...
+	public void testSimple() {
+		CachingConnectionFactory cf = new CachingConnectionFactory("localhost");
+		ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
+		executor.setThreadNamePrefix("client-");
+		executor.afterPropertiesSet();
+		cf.setExecutor(executor);
+		DirectMessageListenerContainer container = new DirectMessageListenerContainer(cf);
+		container.setQueueNames(Q1, Q2);
+		container.setConsumersPerQueue(2);
+		container.setMessageListener(new MessageListenerAdapter((ReplyingMessageListener<String, String>) in -> {
```
artembilan: Can't we relax it a bit and let pass a new `ReplyingMessageListener` directly and only internally wrap it into `MessageListenerAdapter`?

OTOH I really would like to see:
```
public <I, O> void setMessageListener(ReplyingMessageListener<I, O> messageListener) {
	setMessageListener(new MessageListenerAdapter(messageListener));
}
```
WDYT?

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainerTests.java
```diff
@@ -0,0 +1,109 @@
...
+
+				LogFactory.getLog(this.getClass()).info(in);
```
artembilan: Why do we need this `info`?
Just to be sure that we really are in the consumer thread?

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainerTests.java
garyrussell: You mean automatically wrap the lambda in an adapter? People usually want to configure the adapter with things like JSON message converters etc.

artembilan: Doh. Missed that.
Also I see your next test with the `MessageListener` lambda.
So, ignore that my comment.

Although seems for me a couple overloaded methods for the `MessageListener` and `ChannelAwareMessageListener` should be useful.
No? 
What am I missing when we have one method for generic `Object`?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplate.java
```diff
@@ -206,11 +206,9 @@ public Exchange getExchange(String vhost, String name) {
 	}
 
 	private List<Queue> converteQueueList(List<QueueInfo> queues) {
-		List<Queue> convertedQueues = new ArrayList<Queue>();
-		for (QueueInfo qi : queues) {
-			convertedQueues.add(convert(qi));
-		}
-		return convertedQueues;
+		return queues.stream()
+				.map(qi -> convert(qi))
```
kartoffelsup: I'm fairly certain that you could replace all of these with method references. e.g.: `.map(this::convert))`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
garyrussell: When you use `basicGet` the broker includes an estimate for the current number of messages on the queue - this does not apply for `basicConsume` so we zero it here.

This predates my involvement on the project 😄 

I will add some javadocs to `MessafgeProperties`.

From `GetResponse`:

     * @param messageCount the server's most recent estimate of the number of messages remaining on the queue

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainerTests.java
garyrussell: Yeah - simply inherited from the existing code; will add more setters.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplate.java
garyrussell: Yes, of course; thanks.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
artembilan: OK. If we have those JavaDocs already, how about to leave this property as `null`?
E.g. for meaning `undefined`...

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
artembilan: Are you going to address this here or in separate commit?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
garyrussell: I think I'd rather change it to an int and let it default to 0 - probably a "less-likely" breaking change - compile time Vs. runtime.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
garyrussell: I will address here.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
artembilan: Hm...
Someone has permissions to delete my comments.

OK. Will be serious.
You missed at least to answer me.
Thanks

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -446,6 +495,20 @@ public void setupMessageListener(Object messageListener) {
...
 
+	protected void initializeProxy(Object delegate) {
+		if (this.getAdviceChain().length == 0) {
+			return;
+		}
+		ProxyFactory factory = new ProxyFactory();
+		for (Advice advice : getAdviceChain()) {
+			factory.addAdvisor(new DefaultPointcutAdvisor(Pointcut.TRUE, advice));
+		}
+		factory.setProxyTargetClass(false);
```
artembilan: It is `false` by default.

And I pointed previously that we don't need to specify `Pointcut.TRUE` explicitly. It is implied by the second `DefaultPointcutAdvisor` ctor.

Right, minor, but still hurts my eyes :smile:

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -0,0 +1,220 @@
...
+	 * Set to false to requeue failed deliveries.
+	 * @param requeueRejected the requeue rejected.
+	 */
+	public void setDefaultRequeueRejected(boolean requeueRejected) {
+		this.defaultRequeueRejected = requeueRejected;
+	}
+
+	@Override
+	protected void doInitialize() throws Exception {
+		Assert.notNull(getQueueNames(), "queue(s) are required");
```
artembilan: I think according https://jira.spring.io/browse/AMQP-637 requirement we even can start container without queues.
Am I missing anything?

garyrussell: Yes, but that's not implemented yet 😄

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -322,6 +353,23 @@ public void setDeBatchingEnabled(boolean deBatchingEnabled) {
 	}
 
 	/**
+	 * Public setter for the {@link Advice} to apply to listener executions. If {@link #setTxSize(int) txSize>1} then
```
artembilan: Looks like we don't have `setTxSize()` here. That is an option for the simple container.

(And yes, GH has come back to life :smile: )

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -0,0 +1,220 @@
...
+	 * @param connectionFactory the connection factory.
+	 */
+	public DirectMessageListenerContainer(ConnectionFactory connectionFactory) {
+		setConnectionFactory(connectionFactory);
+	}
+
+	/**
+	 * Set a task executor for the container - used to create the consumers not at
+	 * runtime.
+	 * @param taskExecutor the task excutor.
```
artembilan: typo: ex**e**cutor

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
garyrussell: See the DMLC class javadoc:

    There is no txSize property - each message is acked (or nacked) individually.

garyrussell: See the DMLC class javadoc:

    There is no txSize property - each message is acked (or nacked) individually.

I guess we can just delete this sentence from this common class, and handle it in the reference docs.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -0,0 +1,220 @@
...
+			super(channel);
+			this.queue = queue;
+			this.ackRequired = !getAcknowledgeMode().isAutoAck() && !getAcknowledgeMode().isManual();
+		}
+
+		@Override
+		public void handleDelivery(String consumerTag, Envelope envelope,
+				BasicProperties properties, byte[] body) throws IOException {
+			MessageProperties messageProperties = DirectMessageListenerContainer.this.messagePropertiesConverter
+					.toMessageProperties(properties, envelope, "UTF-8");
```
artembilan: Is it normal that we use hard-codded charset?
I see the same issue in the `BlockingQueueConsumer`. Meanwhile `RabbitTemplate` uses configurable property.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
artembilan: :+1:

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainerTests.java
```diff
@@ -0,0 +1,104 @@
...
+import org.springframework.amqp.rabbit.connection.CachingConnectionFactory;
+import org.springframework.amqp.rabbit.core.RabbitTemplate;
+import org.springframework.amqp.rabbit.listener.adapter.MessageListenerAdapter;
+import org.springframework.amqp.rabbit.listener.adapter.ReplyingMessageListener;
+import org.springframework.amqp.rabbit.test.BrokerRunning;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
+
+/**
+ * @author Gary Russell
+ * @since 1.6
```
artembilan: `@since 2.0`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
garyrussell: This is original code (before my time on the project). Yes, it looks like it should be configurable - however, I seem to recall that the rabbit client unconditionally uses UTF-8 for all its String conversion (such as to `LongString`). The only field we provide conversion for is the correlation id; so it wouldn't be correct to use the same charset for both (unless it's UTF-8). Let's ignore for now (or open a new JIRA to review).

artembilan: Eh... Agree - `LongStringHelper`:
```
return new String(bytes, "utf-8");
```
So, right: ignoring for now.

garyrussell: By the way, in 1.6 we added a [String version of correlationId](https://jira.spring.io/browse/AMQP-566).

Perhaps we could deprecate the `byte[]` version in 1.6 and remove it in 2.0 - that way we can get rid of this configurable charset altogether and always use `UTF-8` to match the rabbit client.

artembilan: Makes sense. 
And free `RabbitTemple` from that property as well.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/f369fb617fa736c81849079cf0ab00f6774be81d

PR #447, opened 2016-08-31
AMQP-643: Java 8 and Lambdas

garyrussell:
Mark functional interfaces as such.

Convert many anonymous classes (mainly in tests) to lambdas.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/ReplyingMessageListener.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/ReplyingMessageListener.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 the original author or authors.
+ * Copyright 2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -14,29 +14,24 @@
  * limitations under the License.
  */
 
-package org.springframework.amqp.rabbit.support;
-
-import java.util.Collection;
-
-import org.springframework.amqp.core.MessageListener;
-import org.springframework.amqp.rabbit.core.ChannelAwareMessageListener;
+package org.springframework.amqp.rabbit.listener.adapter;
 
 /**
- * {@link MessageListener}s and {@link ChannelAwareMessageListener}s that also implement this
- * interface can have configuration verified during initialization.
+ * A Message Listener that returns a reply - intended for lambda use in a
+ * {@link MessageListenerAdapter}.
  *
  * @author Gary Russell
- * @since 1.5
+ * @since 2.0
  *
  */
 @FunctionalInterface
-public interface ListenerContainerAware {
+public interface ReplyingMessageListener<T, R> {
 
 	/**
-	 * Return the queue names that the listener expects to listen to.
-	 *
-	 * @return the queue names.
+	 * Handle the message and return a reply.
+	 * @param t the request.
+	 * @return the reply.
 	 */
-	Collection<String> expectedQueueNames();
+	R handleMessage(T t);
 
 }
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitTemplatePublisherCallbacksIntegrationTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitTemplatePublisherCallbacksIntegrationTests.java
@@ -56,18 +56,15 @@ import org.junit.After;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
-import org.mockito.invocation.InvocationOnMock;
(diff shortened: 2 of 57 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/ReplyingMessageListener.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+@FunctionalInterface
+public interface ReplyingMessageListener<T, R> {
+
+	/**
+	 * Handle the message and return a reply.
+	 * @param t the request.
+	 * @return the reply.
+	 */
+	R handleMessage(T t);
```
artembilan: Why do we need this if it is fully copying the `java.util.function.Function<T, R>`  ?

garyrussell: Good point, generally - although, since the MLA takes an `Object` ctor arg, we always have to cast the lambda so I thought this 

    new MessageListenerAdapter((ReplyingMessageListener<String, String>) in -> in.toUpperCase()));

would add some semantic sugar over

    new MessageListenerAdapter((Function<String, String>) in -> in.toUpperCase()));

Also, the MLA uses the method name `handleMessage` by default.

I suppose we could add code to call `apply` if it's a `Function` but then what if someone has an existing listener with a `handleMessage` method that also implements `Function` ? We'd start calling the "wrong" method.

artembilan: Ah! Sure! `handleMessage` default method does the trick :smile: 

Although it is really dilemma, because we have many similar interfaces which could be replaced by the `java.util.function` stuff...
Not sure that SF has removed some of their in favor of those generics, but right, we should still keep some backward compatibility, if case of implementations. 
So, yeah, let's leave it as is for now!

I think the PR is cool and I'll merge it over a time.

Busy with `ScopedProxyFactoryBean` hell :angry:

garyrussell: FYI, most of the lambda conversions were automatic with STS so you shouldn't need to spend too much time reviewing them.

artembilan: :+1: 
Right, I don't look into test.
The greed light from Travis is enough for me :smile:

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/8cc1caf1128fa0d828457ed140ad693ec3ab65a7

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitTemplatePublisherCallbacksIntegrationTests.java
```diff
@@ -205,26 +190,18 @@ public Void doInRabbit(Channel channel) throws Exception {
...
 		container.setQueueNames(ROUTE);
-		container.setMessageListener(new MessageListenerAdapter(new Object() {
-
-			@SuppressWarnings("unused")
-			public String handleMessage(String in) {
-				return in.toUpperCase();
-			}
-		}));
+		container.setMessageListener(
+				new MessageListenerAdapter((ReplyingMessageListener<String, String>) in -> in.toUpperCase()));
```
artembilan: I've just merged this, but I think it won't hurt if we discussion it a bit more.
1. My IDEA says me that this can be simplified as method reference `String::toUpperCase`. But, eh, it is just this case and that's all.
2. How about to improve it to something like this?
```
MessageListenerAdapter messageListener = new MessageListenerAdapter();
messageListener.<String, String>setDelegate(String::toUpperCase);
```
where the target code is:
```
public <I, O> void setDelegate(ReplyingMessageListener<I, O> delegate) {
```
Eh... One more ctor with generics doesn't work.

And also I think this Lambda deserves some Docs in favor of Java 8 universe :smile: 

WDYT?

garyrussell: >And also I think this Lambda deserves some Docs in favor of Java 8 universe 

Not sure what you mean. Do you mean convert any example code in the docs to lambdas?

artembilan: No, I mean that we have to mention it in the docs somewhere as a build-in variant to achieve a Lambda configuration.
Otherwise end-user may create its own variant of this interface with the same method to meet `MessageListenerAdapter` requirements. And once he will be surprised that we have one already but it isn't documented. 
This is end-user high level API without any usage in the Framework. So, should be mentioned in Docs. 
No? :smile:

PR #449, opened 2016-09-02
AMQP-644: Non-Fatal Exception Stopped Container

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-644

Previously, any `NoSuchMethodException` was container-fatal (the container was stopped)
on the assumption that the listener was mis-configured.

However, this could be a transient, data-related, problem so should be considered
message-fatal, not container-fatal.

Also added `ClassCastException` because when the `MessageListenerAdapter` delegate is a
lambda, we get this exception instead of a `NoSuchMethodException`.

__cherry-pick to 1.6.x__

### DIFF
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerIntegration2Tests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerIntegration2Tests.java
@@ -34,6 +34,7 @@ import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 import java.io.IOException;
+import java.io.Serializable;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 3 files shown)


spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerIntegration2Tests.java
```diff
@@ -475,6 +468,34 @@ public void onMessage(Message message) {
...
+		this.template.convertAndSend(this.queue.getName(), "foo");
+		assertTrue(latch.await(10, TimeUnit.SECONDS));
+		assertTrue(this.container.isRunning());
+		this.container.stop();
+	}
+
+	@Test
+	public void testTransientBadMessageDoesntStopContainerLambda() throws Exception {
+		final CountDownLatch latch = new CountDownLatch(2);
+		this.container = createContainer(new MessageListenerAdapter((ReplyingMessageListener<String, Void>) m -> {
```
artembilan: I guess this code can't be cherry-picked.
NP. Will fix during backport 
:smile:

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/78c7a364cffe499a4629a841c00fe8aba826ef13 and cherry-picked as https://github.com/spring-projects/spring-amqp/commit/384cd1e194d4e7b1a39fe5a093db652922ff120c.

PR #450, opened 2016-09-02
Clean Broker After Running Tests

garyrussell:
Also update S-F and spring-retry to snapshots.

### DIFF
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/config/QueueParserIntegrationTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/config/QueueParserIntegrationTests.java
@@ -74,7 +74,7 @@ public final class QueueParserIntegrationTests {
 		assertEquals(null, result);
 
 		connectionFactory.destroy();
-
+		brokerIsRunning.getAdmin().deleteQueue("arguments");
 	}
 
 }
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/LocalizedQueueConnectionFactoryIntegrationTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/LocalizedQueueConnectionFactoryIntegrationTests.java
@@ -63,6 +63,7 @@ public class LocalizedQueueConnectionFactoryIntegrationTests {
 		RabbitTemplate template = new RabbitTemplate(targetConnectionFactory);
 		template.convertAndSend("", queue.getName(), "foo");
 		assertEquals("foo", template.receiveAndConvert(queue.getName()));
+		admin.deleteQueue(queue.getName());
 	}
 
 }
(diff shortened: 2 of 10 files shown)


spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/LocalizedQueueConnectionFactoryIntegrationTests.java
```diff
@@ -63,6 +63,7 @@ public void testConnect() {
 		RabbitTemplate template = new RabbitTemplate(targetConnectionFactory);
 		template.convertAndSend("", queue.getName(), "foo");
 		assertEquals("foo", template.receiveAndConvert(queue.getName()));
+		admin.deleteQueue(queue.getName());
```
artembilan: Why do we need this if the `queue` is `autoDeleted` ?

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/config/QueueParserIntegrationTests.java
```diff
@@ -74,7 +74,7 @@ public void testArgumentsQueue() throws Exception {
 		assertEquals(null, result);
 
 		connectionFactory.destroy();
-
+		brokerIsRunning.getAdmin().deleteQueue("arguments");
```
artembilan: Can't we mark this `arguments` with the `auto-delete="true"`?
OTOH the `QueueParserIntegrationTests-context.xml` has one more `foo` declaration...

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/LocalizedQueueConnectionFactoryIntegrationTests.java
garyrussell: Because they previously were not auto-delete so running against a broker that has previously had the tests run will fail.

We can remove these in a couple of years or so 😄

garyrussell: Sorry - I was talking about something else.

Here, we never consume from the queue so it will not be deleted - auto-delete queues are deleted only when the last consumer disconnects.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/config/QueueParserIntegrationTests.java
garyrussell: The foo queue is never used; it can be removed; as I mentioned in the other comments, it doesn't really help to set auto-delete because it causes tests to fail against a broker that has the older tests run against it.

artembilan: Yep! Makes sense.
Leaving as is.
Doh... :smile:

artembilan:
OK. Good.
Merging with removal that bogus `foo` queue in the config.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/b34f5027a5e0355ae8e85f6da45237ab32051a0d

PR #451, opened 2016-09-07
AMQP-637: DirectMLContainer - Add/Remove Queues

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-637

Also pull up some more common properties to abstract container class.

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/AmqpIOException.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/AmqpIOException.java
@@ -28,10 +28,20 @@ import java.io.IOException;
 @SuppressWarnings("serial")
 public class AmqpIOException extends AmqpException {
 
+	/**
+	 * Construct an instance with the provided cause.
+	 * @param cause the cause.
+	 */
 	public AmqpIOException(IOException cause) {
 		super(cause);
 	}
 
+	/**
+	 * Construct an instance with the provided message and cause.
+	 * @param message the message.
+	 * @param cause the cause.
+	 * @since 2.0
+	 */
 	public AmqpIOException(String message, Throwable cause) {
 		super(message, cause);
 	}
--- a/spring-amqp/src/main/java/org/springframework/amqp/AmqpIOException.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/AmqpIOException.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2010 the original author or authors.
+ * Copyright 2002-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -23,12 +23,17 @@ import java.io.IOException;
  * can be commonly thrown from AMQP operations.
  *
  * @author Mark Pollack
+ * @author Gary Russell
  */
 @SuppressWarnings("serial")
 public class AmqpIOException extends AmqpException {
 
 	public AmqpIOException(IOException cause) {
 		super(cause);
+	}
+
+	public AmqpIOException(String message, Throwable cause) {
+		super(message, cause);
 	}
 
 }
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
@@ -19,13 +19,11 @@ package org.springframework.amqp.rabbit.listener;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.Map.Entry;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.apache.commons.logging.Log;
 
@@ -63,7 +61,7 @@ import com.rabbitmq.client.Envelope;
  */
 public class DirectMessageListenerContainer extends AbstractMessageListenerContainer {
 
-	private final List<SimpleConsumer> consumers = new LinkedList<SimpleConsumer>();
+	private final List<SimpleConsumer> consumers = new LinkedList<>();
 
 	private final MultiValueMap<String, SimpleConsumer> consumersByQueue = new LinkedMultiValueMap<>();
 
@@ -73,7 +71,7 @@ public class DirectMessageListenerContainer extends AbstractMessageListenerConta
 
 	private volatile int consumersPerQueue = 1;
 
-	private volatile long recoveryInterval;
+	private volatile long recoveryInterval = DEFAULT_RECOVERY_INTERVAL;
 
 	/**
 	 * Create an instance with the provided connection factory.
@@ -95,9 +93,14 @@ public class DirectMessageListenerContainer extends AbstractMessageListenerConta
 	/**
 	 * Each queue runs in its own consumer; set this property to create multiple
 	 * consumers for each queue.
+	 * If the container is already running, the number of consumers per queue will
+	 * be adjusted up or down as necessary.
+	 * If the adjustment fails for any reason, you may have mismatched number of
+	 * consumers on each queue. A subsequent (successful) call will fix the mismatch.
 	 * @param consumersPerQueue the consumers per queue.
 	 */
 	public void setConsumersPerQueue(int consumersPerQueue) {
+		Assert.isTrue(consumersPerQueue >= 1, "'consumersPerQueue' must be 1 or greater");
 		if (isRunning()) {
 			adjustConsumers(consumersPerQueue);
 		}
@@ -116,6 +119,14 @@ public class DirectMessageListenerContainer extends AbstractMessageListenerConta
 		super.setExclusive(exclusive);
 	}
 
+	/**
+	 * Set the interval before retrying when starting the container; default 5000.
+	 * @param recoveryInterval the interval.
+	 */
+	public void setRecoveryInterval(long recoveryInterval) {
+		this.recoveryInterval = recoveryInterval;
+	}
+
 	@Override
 	public void setQueueNames(String... queueName) {
 		Assert.state(!isRunning(), "Cannot set queue names while running, use add/remove");
@@ -130,48 +141,56 @@ public class DirectMessageListenerContainer extends AbstractMessageListenerConta
 
 	@Override
 	public void addQueueNames(String... queueNames) {
-		addQueues(Arrays.asList(queueNames));
+		try {
+			addQueues(Arrays.stream(queueNames));
+		}
+		catch (AmqpIOException e) {
+			throw new AmqpIOException("Failed to add " + queueNames, e.getCause());
+		}
 		super.addQueueNames(queueNames);
 	}
 
 	@Override
 	public void addQueues(Queue... queues) {
-		addQueues(Arrays.asList(queues)
-				.stream()
-				.map(Queue::getName)
-				.collect(Collectors.toList()));
+		try {
+			addQueues(Arrays.stream(queues)
+					.map(Queue::getName));
+		}
+		catch (AmqpIOException e) {
+			throw new AmqpIOException("Failed to add " + queues, e.getCause());
+		}
 		super.addQueues(queues);
 	}
 
-	private void addQueues(Collection<String> queues) {
+	private void addQueues(Stream<String> queueNameStream) {
 		if (isRunning()) {
 			synchronized (this.consumersMonitor) {
 				Set<String> current = getQueueNamesAsSet();
-				for (String queue : queues) {
-					if (current.contains(queue)) {
-						throw new IllegalStateException("Queue " + queue + " is already configured for this container: "
-								+ this + ", no queues added");
-					}
-				}
 				List<String> added = new ArrayList<>();
-				for (String queue : queues) {
-					try {
-						consumeFromQueue(queue);
-						added.add(queue);
+				queueNameStream.forEach(queue -> {
+					if (current.contains(queue)) {
+						this.logger.warn("Queue " + queue + " is already configured for this container: "
+								+ this + ", ignoring add");
 					}
-					catch (IOException e) {
-						logger.error("Failed to add queue: " + queue + " undoing adds (if any)");
-						removeQueues(added);
-						throw new AmqpIOException("Failed to add " + queues, e);
+					else {
+						try {
+							consumeFromQueue(queue);
+							added.add(queue);
+						}
+						catch (IOException e) {
+							logger.error("Failed to add queue: " + queue + " undoing adds (if any)");
+							removeQueues(added.stream());
+							throw new AmqpIOException(e);
+						}
 					}
-				}
+				});
 			}
 		}
 	}
 
 	@Override
 	public boolean removeQueueNames(String... queueNames) {
-		removeQueues(Arrays.asList(queueNames));
+		removeQueues(Arrays.stream(queueNames));
 		return super.removeQueueNames(queueNames);
 	}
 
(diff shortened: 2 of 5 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -95,59 +100,155 @@ public void setConsumersPerQueue(int consumersPerQueue) {
...
+					}
+				}
+				List<String> added = new ArrayList<>();
+				for (String queue : queues) {
+					try {
+						consumeFromQueue(queue);
+						added.add(queue);
+					}
+					catch (IOException e) {
+						logger.error("Failed to add queue: " + queue + " undoing adds (if any)");
```
garyrussell: @artembilan Any ideas why checkstyle is not flagging `this.` here?

artembilan: Let me pull your changes locally.
On GH it isn't obvious.

artembilan: Ha! Because `logger` is a property in super class.
I guess that broader area is just not supported.

Not sure how that would be easy to trace properties from super classes without compilation...

I guess we don't have choice unless live with this limitation.

garyrussell: Ah; thanks - makes sense.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -91,79 +98,243 @@ public void setTaskExecutor(TaskExecutor taskExecutor) {
...
+
+	@Override
+	public void addQueueNames(String... queueNames) {
+		addQueues(Arrays.asList(queueNames));
+		super.addQueueNames(queueNames);
+	}
+
+	@Override
+	public void addQueues(Queue... queues) {
+		addQueues(Arrays.asList(queues)
```
kartoffelsup: There should also be a method called `Arrays.stream(array...)`

spring-amqp/src/main/java/org/springframework/amqp/AmqpIOException.java
```diff
@@ -31,4 +32,8 @@ public AmqpIOException(IOException cause) {
 		super(cause);
 	}
 
+	public AmqpIOException(String message, Throwable cause) {
```
artembilan: `@since 2.0`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -54,18 +63,16 @@
  */
 public class DirectMessageListenerContainer extends AbstractMessageListenerContainer {
 
-	private final List<SimpleConsumer> consumers = new ArrayList<SimpleConsumer>();
+	private final List<SimpleConsumer> consumers = new LinkedList<SimpleConsumer>();
```
artembilan: We can use diamond here as well.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -54,18 +63,16 @@
...
 
 	private volatile TaskExecutor taskExecutor;
 
 	private volatile int consumersPerQueue = 1;
 
-	private volatile int prefetch = 1;
-
-	private volatile boolean defaultRequeueRejected = false;
-
 	private volatile long recoveryInterval;
```
artembilan: recoveryinterval_usage

I wonder if we can move `recoveryBackOff` option from the `SimpleMessageListenerContainer` to the super `AbstractMessageListenerContainer`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
artembilan: Thank you, @kartoffelsup , for pointing that out!

Looking at this further I even think that we don't need those `.collect(Collectors.toList())`.
We can just get `Stream` here and in the previous `addQueueNames()` method and just `.map()` each item.
Although need to understand @garyrussell 's logic in the `addQueues(Collection<String> queues)` regarding double iteration over the same collection....

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -91,79 +98,243 @@ public void setTaskExecutor(TaskExecutor taskExecutor) {
...
+	}
+
+	private void addQueues(Collection<String> queues) {
+		if (isRunning()) {
+			synchronized (this.consumersMonitor) {
+				Set<String> current = getQueueNamesAsSet();
+				for (String queue : queues) {
+					if (current.contains(queue)) {
+						throw new IllegalStateException("Queue " + queue + " is already configured for this container: "
+								+ this + ", no queues added");
```
artembilan: Why can't we just ignore existing queue and WARN this fact instead of full rejection?
Thanks

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -91,79 +98,243 @@ public void setTaskExecutor(TaskExecutor taskExecutor) {
...
+				.stream()
+				.map(Queue::getName)
+				.collect(Collectors.toList()));
+		return super.removeQueues(queues);
+	}
+
+	private void removeQueues(Collection<String> queues) {
+		if (isRunning()) {
+			synchronized (this.consumersMonitor) {
+				for (String queue : queues) {
```
artembilan: The same `Stream` processing we can apply here as well:
````java
@Override
public boolean removeQueueNames(String... queueNames) {
	removeQueues(Arrays.stream(queueNames));
	return super.removeQueueNames(queueNames);
}

@Override
public boolean removeQueues(Queue... queues) {
	removeQueues(Arrays.stream(queues)
		.map(Queue::getName));
	return super.removeQueues(queues);
}

private void removeQueues(Stream<String> queues) {
	if (isRunning()) {
		synchronized (this.consumersMonitor) {
			queues.map(this.consumersByQueue::remove)
					.flatMap(Collection::stream)
					.forEach(consumer -> {
						try {
							consumer.getChannel().basicCancel(consumer.getConsumerTag());
							this.consumers.remove(consumer);
						}
						catch (IOException e) {
							logger.error("Failed to cancel consumer: " + consumer, e);
						}
					});
		}
	}
}
````
WDYT?

File: spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java [hunk 8]
```diff
@@ -91,79 +98,243 @@ public void setTaskExecutor(TaskExecutor taskExecutor) {
...
+										List<SimpleConsumer> list = this.consumersByQueue.get(queue);
+										if (list != null) {
+											list.remove(consumer);
+										}
+									}
+									catch (IOException e1) {
+										logger.error("Failed to cancel consumer during revert", e);
+									}
+								}
+								throw new AmqpIOException("Failed to increase consumers", e);
```
artembilan: Why this operation has to be atomic?
Why we can't adjust as much as we can and don't let error to break just added new consumers?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
garyrussell: That is mentioned in https://jira.spring.io/browse/AMQP-636

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
garyrussell: I was thinking because we'd have a container in a strange state (different number of consumers on each queue), but I suppose I can rework it so the counts will be corrected on the next call.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
garyrussell: It doesn't handle `null` returned from `remove()`, but I've done something similar.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -130,77 +141,84 @@ public void setQueues(Queue... queues) {
...
 	@Override
 	public boolean removeQueueNames(String... queueNames) {
-		removeQueues(Arrays.asList(queueNames));
+		removeQueues(Arrays.stream(queueNames));
 		return super.removeQueueNames(queueNames);
 	}
 
 	@Override
 	public boolean removeQueues(Queue... queues) {
 		removeQueues(Arrays.asList(queues)
```
artembilan: The same `Arrays.stream()` can be applied here as well.

garyrussell: doh - missed that one.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -88,82 +93,247 @@ public void setTaskExecutor(TaskExecutor taskExecutor) {
...
+				.map(Queue::getName));
+		return super.removeQueues(queues);
+	}
+
+	private void removeQueues(Stream<String> queueNames) {
+		if (isRunning()) {
+			synchronized (this.consumersMonitor) {
+				queueNames.forEach(queue -> {
+					List<SimpleConsumer> consumersForQueue = this.consumersByQueue.remove(queue);
+					if (consumersForQueue != null) {
```
artembilan: Why can't we use `.filter(consumers -> consumers != null)`?
And come back to the streams again :smile:

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -88,82 +93,247 @@ public void setTaskExecutor(TaskExecutor taskExecutor) {
...
+		super.setQueues(queues);
+	}
+
+	@Override
+	public void addQueueNames(String... queueNames) {
+		try {
+			addQueues(Arrays.stream(queueNames));
+		}
+		catch (AmqpIOException e) {
+			throw new AmqpIOException("Failed to add " + queueNames, e.getCause());
```
artembilan: The `queueNames` must be wrapped to `Arrays.asList()`:
array_in_sout

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -88,82 +93,247 @@ public void setTaskExecutor(TaskExecutor taskExecutor) {
...
 	}
 
 	@Override
+	public void addQueues(Queue... queues) {
+		try {
+			addQueues(Arrays.stream(queues)
+					.map(Queue::getName));
+		}
+		catch (AmqpIOException e) {
+			throw new AmqpIOException("Failed to add " + queues, e.getCause());
```
artembilan: DITO for `Arrays.asList()`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -88,82 +93,247 @@ public void setTaskExecutor(TaskExecutor taskExecutor) {
...
+		return super.removeQueues(queues);
+	}
+
+	private void removeQueues(Stream<String> queueNames) {
+		if (isRunning()) {
+			synchronized (this.consumersMonitor) {
+				queueNames.forEach(queue -> {
+					List<SimpleConsumer> consumersForQueue = this.consumersByQueue.remove(queue);
+					if (consumersForQueue != null) {
+						consumersForQueue.stream().forEach(consumer -> {
```
artembilan: IDEA says me that `.stream()` is redundant here and `.forEach()` is a part of `Iterable` API.

garyrussell: Yeah - spotted that, but reworked anyway for comment above.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
kartoffelsup: Or `Arrays.toString`

artembilan: Ha! I think this would be even better. 
We definitely need only a string representation and there is no reason in that intermediate `ArrayList`  object.

Good catch, @kartoffelsup !
Thank you!

garyrussell: I am not too worried about performance; this should not happen often; feel free to change during merge if you wish (already pushed the other).

artembilan:
```
org.springframework.amqp.rabbit.listener.DirectMessageListenerContainerTests > testQueueManagement FAILED

    java.lang.AssertionError at DirectMessageListenerContainerTests.java:135
```
I guess it is for this line:
```
		assertEquals(2, admin.getQueueProperties(Q1).get(RabbitAdmin.QUEUE_CONSUMER_COUNT));
```
Looks like a race condition after `container.addQueueNames(Q1, Q2);`.
Although it isn't clear how, because `channel.basicConsume()` is based on the `BlockingRpcContinuation`.

OTOH, the original `start()` and, therefore consume, is done in the separate thread:
```
this.taskExecutor.execute(() -> {

			synchronized (this.consumersMonitor) {
				while (!initialized.get() && isRunning()) {
					try {
						for (String queue : getQueueNames()) {
							consumeFromQueue(queue);
						}
					}
...		
```
Why does `addQueueNames()` performs `channel.basicConsume()` at the caller thread directly?

garyrussell:
I want to return success/failure to the caller; I guess we can add a while loop in the test.

artembilan:
>  I guess we can add a while loop in the test.

OK.
I thought about the message consumption in that case, but your compromise should be enough.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainerTests.java
```diff
@@ -167,20 +166,29 @@ public void testAddRemoveConsumers() {
...
 	}
 
+	private boolean consumersOnQueue(String queue, int count) throws Exception {
+		int n = 0;
+		RabbitAdmin admin = this.brokerRunning.getAdmin();
+		while (n++ < 100 && !admin.getQueueProperties(queue).get(RabbitAdmin.QUEUE_CONSUMER_COUNT)
+				.equals(Integer.valueOf(count))) {
+			Thread.sleep(100);
+		}
+		return admin.getQueueProperties(queue).get(RabbitAdmin.QUEUE_CONSUMER_COUNT).equals(Integer.valueOf(count));
```
artembilan: IDEA says me that boxing is unnecessary.
Fixing on merge...

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/9d668288c42d9c6af499a65de47e51d870294ea4

PR #452, opened 2016-09-09
AMQP-637: Fix Shutdown

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-637

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
@@ -431,7 +431,9 @@ public class DirectMessageListenerContainer extends AbstractMessageListenerConta
 			this.logger.error("Consumer canceled - queue deleted? " + this);
 			synchronized (DirectMessageListenerContainer.this.consumersMonitor) {
 				List<SimpleConsumer> list = DirectMessageListenerContainer.this.consumersByQueue.get(this.queue);
-				list.remove(this);
+				if (list != null) {
+					list.remove(this);
+				}
 				DirectMessageListenerContainer.this.cancellationLock.release(this);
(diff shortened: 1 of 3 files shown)


garyrussell:
Don't merge; needs polish

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -387,16 +407,36 @@ public void handleDelivery(String consumerTag, Envelope envelope,
...
+		}
+
+		@Override
 		public void handleCancelOk(String consumerTag) {
+			if (this.logger.isDebugEnabled()) {
+				this.logger.debug("CancelOk " + this);
+			}
+			synchronized (DirectMessageListenerContainer.this.consumersMonitor) {
+				DirectMessageListenerContainer.this.cancellationLock.release(this);
+				DirectMessageListenerContainer.this.consumers.remove(this);
```
artembilan: Why don't we remove `Consumer` from the `consumersByQueue` as well, like we do that in the `handleCancel()`?

garyrussell: I guess we could for consistency - right now it's removed in adjustConsumers when we cancel (or cleared in doShutdown).

artembilan: OK. Merging with that fix.

garyrussell: No; I pushed

artembilan:
OK. Merging...

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/1834b6c9f4fafdec5038316f2deea9e163b99863

PR #454, opened 2016-09-09
AMQP-638 DirectMessageListenerContainer Events

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-638

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
@@ -142,7 +142,7 @@ public abstract class AbstractMessageListenerContainer extends RabbitAccessor
 
 	private volatile int prefetchCount = DEFAULT_PREFETCH_COUNT;
 
-	private Long idleEventInterval;
+	private long idleEventInterval;
 
 	private volatile long lastReceive = System.currentTimeMillis();
 
@@ -633,11 +633,11 @@ public abstract class AbstractMessageListenerContainer extends RabbitAccessor
 	 * How often to emit {@link ListenerContainerIdleEvent}s in milliseconds.
 	 * @param idleEventInterval the interval.
 	 */
-	public void setIdleEventInterval(Long idleEventInterval) {
+	public void setIdleEventInterval(long idleEventInterval) {
 		this.idleEventInterval = idleEventInterval;
 	}
 
-	protected Long getIdleEventInterval() {
+	protected long getIdleEventInterval() {
 		return this.idleEventInterval;
 	}
 
@@ -1140,7 +1140,7 @@ public abstract class AbstractMessageListenerContainer extends RabbitAccessor
 	}
 
 	protected void updateLastReceive() {
-		if (this.idleEventInterval != null) {
+		if (this.idleEventInterval > 0) {
 			this.lastReceive = System.currentTimeMillis();
 		}
 	}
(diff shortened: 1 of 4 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -608,6 +630,27 @@ protected long getShutdownTimeout() {
 	}
 
 	/**
+	 * How often to emit {@link ListenerContainerIdleEvent}s in milliseconds.
+	 * @param idleEventInterval the interval.
+	 */
+	public void setIdleEventInterval(Long idleEventInterval) {
```
artembilan: I think we can require here an valid value already and really don't let to provide pointless `null`.
Therefore primitive `long`.
But, eh, minor, of course :smile:

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -267,6 +282,19 @@ protected void doStart() throws Exception {
...
+			if (this.taskScheduler != null) {
+				afterPropertiesSet();
+			}
+			this.idleTask = this.taskScheduler.scheduleAtFixedRate(() -> {
+				long now = System.currentTimeMillis();
+				if (now - getLastReceive() > idleEventInterval && now - this.lastAlertAt > idleEventInterval) {
+					publishIdleContainerEvent(now - getLastReceive());
+					this.lastAlertAt = now;
+				}
+			}, this.getIdleEventInterval() / 2);
```
artembilan: Can't we use here an extracted `idleEventInterval` variable already ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -1191,16 +1165,16 @@ public void run() {
...
 							}
 							else {
 								long now = System.currentTimeMillis();
 								long lastAlertAt = SimpleMessageListenerContainer.this.lastNoMessageAlert.get();
-								long lastReceive = SimpleMessageListenerContainer.this.lastReceive;
-								if (now > lastReceive + SimpleMessageListenerContainer.this.idleEventInterval
-										&& now > lastAlertAt + SimpleMessageListenerContainer.this.idleEventInterval
+								long lastReceive = getLastReceive();
+								if (now > lastReceive + getIdleEventInterval()
+										&& now > lastAlertAt + getIdleEventInterval()
```
artembilan: Why don't extract variable and don't call the method several times?
More over with the current `setIdleEventInterval(Long)` signature we can end up here with `NPE`...

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -267,6 +282,19 @@ protected void doStart() throws Exception {
 			this.taskExecutor = new SimpleAsyncTaskExecutor(
 					(getBeanName() == null ? "container" : getBeanName()) + "-");
 		}
+		long idleEventInterval = getIdleEventInterval();
+		if (idleEventInterval > 0) {
+			if (this.taskScheduler != null) {
+				afterPropertiesSet();
```
artembilan: I think you mean here `if (this.taskScheduler == null)`.
Otherwise we end up with an `NPE` in the next `this.taskScheduler.scheduleAtFixedRate()`.

garyrussell: Doh

artembilan: No, you aren't. Just omission.
Fixing on merge...

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/666b39c7c65830061871b11c8a09fb05beca4fe7

PR #457, opened 2016-09-13
AMQP-642: DirecMLContainer Transaction Support

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-642

Also switch the SMLC to use `basicNack` instead of `basicReject`.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitUtils.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitUtils.java
@@ -291,13 +291,14 @@ public abstract class RabbitUtils {
 	}
 
 	/**
-	 * Determine whether a message should be requeued; returns true if ex is a
+	 * Determine whether a message should be requeued; returns true if the throwable is a
 	 * {@link MessageRejectedWhileStoppingException} or defaultRequeueRejected is true and
 	 * there is not an {@link AmqpRejectAndDontRequeueException} in the cause chain.
 	 * @param defaultRequeueRejected the default requeue rejected.
 	 * @param throwable the throwable.
 	 * @param logger the logger to use for debug.
 	 * @return true to requeue.
+	 * @since 2.0
 	 */
 	public static boolean shouldRequeue(boolean defaultRequeueRejected, Throwable throwable, Log logger) {
 		boolean shouldRequeue = defaultRequeueRejected ||
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitUtils.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitUtils.java
@@ -23,6 +23,8 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
 import org.springframework.amqp.AmqpIOException;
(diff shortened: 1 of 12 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitUtils.java
```diff
@@ -288,4 +290,29 @@ public static boolean isExchangeDeclarationFailure(Exception e) {
 		}
 	}
 
+	/**
+	 * Determine whether a message should be requeued; returns true if ex is a
```
artembilan: If `exception`.
I don't think that it is a good practice to abbreviate in public OpenSource Docs.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitUtils.java
```diff
@@ -288,4 +290,29 @@ public static boolean isExchangeDeclarationFailure(Exception e) {
...
+	 * {@link MessageRejectedWhileStoppingException} or defaultRequeueRejected is true and
+	 * there is not an {@link AmqpRejectAndDontRequeueException} in the cause chain.
+	 * @param defaultRequeueRejected the default requeue rejected.
+	 * @param throwable the throwable.
+	 * @param logger the logger to use for debug.
+	 * @return true to requeue.
```
artembilan: `@since 2.0` ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitUtils.java
garyrussell: `ex` was the old variable name 😄

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -449,30 +449,75 @@ public void handleDelivery(String consumerTag, Envelope envelope,
...
+							RabbitUtils.shouldRequeue(isDefaultRequeueRejected(), e, this.logger));
 				}
-				if (this.ackRequired) {
-					getChannel().basicNack(envelope.getDeliveryTag(), false, shouldRequeue);
+				catch (IOException e1) {
+					this.logger.error("Failed to nack message", e1);
 				}
 			}
+			if (isChannelTransacted()) {
+				RabbitUtils.commitIfNecessary(getChannel());
```
artembilan: Not sure what `commit` is doing in the `rallback` function...

garyrussell: We've already rolled back the "real" work (any publishes that were done on the channel). We do, however, have to commit the `basicReject` so the message is requeued. There's a comment (line 641 in BQC) that we should copy here.

artembilan: Got it.
Thank you.
Pretty odd, but works for me.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/a39d29d0c7621b58f9ac3dc6952ad6b01990ea7d

PR #461, opened 2016-09-28
AMQP-648: Fix Race in handleCancel (SMLC)

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-648

The `cancelled` boolean is set when there are no more consumers in `handleCancelOk`.

Consider a case where a container is configured for multiple queues; one queue is deleted
from the broker. In `handleCancel`, we attempt to cancel the rest of the consumers so the
container can recover.

However, if those cancels fail (e.g. if the broker goes down), we'll never finalize the
consumer.

Move the consumer removal (and setting `cancelled`) to `basicCancel` instead. We don't really
care whether or not we get the `cancelOk`s.

Add mock test - also tested with the SI sample modified for 2 queues
- set a breakpoint in `handleCancel`
- delete a queue
- stop/start rabbitmq
- release the breakpoint
- container recovers

__backport to 1.6.x__

### DIFF
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumerTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumerTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2015 the original author or authors.
+ * Copyright 2002-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumerTests.java
```diff
@@ -137,6 +154,53 @@ public void testPrefetchIsSetOnFailedPassiveDeclaration() throws IOException {
...
+				catch (ShutdownSignalException e) {
+				}
+				catch (InterruptedException e) {
+					Thread.currentThread().interrupt();
+				}
+			}
+		});
+		assertTrue(consumerLatch.await(10, TimeUnit.SECONDS));
+		Consumer consumer = consumerCaptor.getValue();
+		consumer.handleCancel("consumer1");
```
artembilan: Not sure where is recover and queue deletion, but, eh, at least this test confirms that we do:
```
if (message == null && this.cancelled.get()) {
	throw new ConsumerCancelledException();
}
```
automatically on only one canceled `Consumer`.
Plus we are now independent of the `IOException` during cancellation process.

OK. Merging since the core code looks stable.

PR #462, opened 2016-09-28
AMQP-645: Deprecate byte[] Correlation Id

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-645

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/DefaultMessagePropertiesConverter.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/DefaultMessagePropertiesConverter.java
@@ -91,17 +91,30 @@ public class DefaultMessagePropertiesConverter implements MessagePropertiesConve
 	}
 
 	/**
+	 * @param correlationIPolicy true to use.
+	 * @see #setCorrelationIdPolicy(CorrelationIdPolicy)
+	 * @deprecated - the byte[] version of correlation id will be removed in 2.0
+	 */
+	@Deprecated
+	public void setCorrelationIdAsString(CorrelationIdPolicy correlationIPolicy) {
+		this.correlationIdPolicy = correlationIPolicy;
+	}
+
+	/**
 	 * For inbound, determine whether correlationId, correlationIdString or
 	 * both are populated. For outbound, determine whether correlationIdString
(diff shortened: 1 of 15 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/DefaultMessagePropertiesConverter.java
```diff
@@ -91,17 +91,30 @@ public DefaultMessagePropertiesConverter(int longStringLimit, boolean convertLon
...
 	 * or correlationId is used when mapping; if {@code CorrelationIdPolicy.BOTH}
 	 * is set for outbound, String takes priority and we fallback to bytes.
 	 * Default {@code CorrelationIdPolicy.BYTES}.
 	 * @param correlationIPolicy true to use.
+	 * @deprecated - the byte[] version of correlation id will be removed in 2.0
 	 */
-	public void setCorrelationIdAsString(CorrelationIdPolicy correlationIPolicy) {
-		this.correlationIdPolicy = correlationIPolicy;
+	@Deprecated
+	public void setCorrelationIdPolicy(CorrelationIdPolicy correlationIPolicy) {
```
artembilan: Why do we need this new method if we deprecate it anyway?
I am even sure that we should deprecate `CorrelationIdPolicy` altogether and, right, remove in `2.0`.

Am I missing something?

garyrussell: I just noticed that it was mis-named `setCorrelationIdAsString`, so I added one with the proper name and deprecated them both.

artembilan: OK. Let it be. Makes sense, although doesn't provide much value if it deprecated anyway.
I see you remove `CorrelationIdPolicy`, so I mark it with `@Deprecated` and merge.
Anything else to pay attention?

garyrussell: ok; thanks. It has to be deprecated because the policy is going away in 2.0 - see the 2.0 PR.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/dbe3b374a837c1aa92ebaba8fa9ba6efba1ff503

PR #463, opened 2016-09-28
AMQP-645: Remove byte[] Correlation Id

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-645


spring-amqp/src/main/java/org/springframework/amqp/core/MessageProperties.java
```diff
@@ -208,20 +205,40 @@ public String getType() {
...
+	 * Get the correlation id.
+	 * @return the id.
+	 */
+	public String getCorrelationId() {
+		return this.correlationId;
 	}
 
+	/**
+	 * Set the correlation id.
+	 * @param correlationId the id.
```
artembilan: There is no such a param.
Will fix on merge.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/7f719b654788ae6383a34b6064c186f9733b82d1

PR #468, opened 2016-10-05
AMQP-631: Add `ConnectionNameStrategy` for `CF`

artembilan:
JIRA: https://jira.spring.io/browse/AMQP-631

To allow to identify application connections in the Broker or other monitoring and tracing tools, the Rabbit Connection can be supplied with `connection_name` property.

* Introduce `ConnectionNameStrategy` for injection into the `AbstractConnectionFactory` which is used to generate name for the underlying `rabbitConnectionFactory.newConnection()` operation

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
@@ -71,7 +71,7 @@ public abstract class AbstractConnectionFactory implements ConnectionFactory, Di
 	private volatile int closeTimeout = DEFAULT_CLOSE_TIMEOUT;
 
 	private ConnectionNameStrategy connectionNameStrategy =
-			connectionFactory -> (this.beanName != null ? this.beanName : getClass()) +
+			connectionFactory -> (this.beanName != null ? this.beanName : "SpringAMQP") +
 					"#" + this.defaultConnectionNameStrategyCounter.getAndIncrement();
 
 	private volatile String beanName;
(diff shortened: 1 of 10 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
```diff
@@ -66,6 +70,10 @@
 
 	private volatile int closeTimeout = DEFAULT_CLOSE_TIMEOUT;
 
+	private ConnectionNameStrategy connectionNameStrategy =
+			connectionFactory -> (this.beanName != null ? this.beanName : getClass()) +
+					"#" + this.defaultConnectionNameStrategyCounter.getAndIncrement();
```
garyrussell: I don't think we should use `toString()` here - it adds an unnecessary `class ` prefix:

    class org.springframework.amqp.rabbit.connection.CachingConnectionFactory#0

Also, the FQCN is too much - maybe just simple name? Or `springAMQP#0` ??

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
```diff
@@ -281,36 +282,47 @@ public int getCloseTimeout() {
 		return this.closeTimeout;
 	}
 
+	/**
+	 * Provide a {@link ConnectionNameStrategy} to build the name for target RabbitMQ connection.
+	 * The {@link #beanName} together with the counter is used by default.
+	 * @param connectionNameStrategy the {@link ConnectionNameStrategy} to use.
```
garyrussell: for __the__ target

together with __a__ counter

garyrussell: And checkstyle of course 😄

garyrussell:
Merged as b59371edf8879f638d1e0d662562e03072280551

PR #470, opened 2016-10-06
AMQP-650: AsyncRabbitTemplate Refactoring

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-650

- Extract `AsyncAmqpTemplate` interface (see AMQP-616)
- Accept any `AbstractMessageListenerContainer` for replies
- group all `sendAndReceive` methods together

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/core/AsyncAmqpTemplate.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/core/AsyncAmqpTemplate.java
@@ -1,0 +1,143 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.amqp.core;
+
+import org.springframework.amqp.AmqpException;
+import org.springframework.util.concurrent.ListenableFuture;
+
+/**
+ * Classes implementing this interface can perform asynchronous send and
+ * receive operations.
+ *
+ * @author Gary Russell
+ * @since 2.0
+ *
+ */
+public interface AsyncAmqpTemplate {
+
+	/**
+	 * Send a message to the default exchange with the default routing key. If the message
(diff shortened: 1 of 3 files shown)


spring-amqp/src/main/java/org/springframework/amqp/core/AsyncAmqpTemplate.java
```diff
@@ -0,0 +1,143 @@
...
+
+	/**
+	 * Convert the object to a message and send it to the default exchange with the
+	 * provided routing key.
+	 * @param routingKey the routing key.
+	 * @param message the message.
+	 * @param <C> the expected result type.
+	 * @return the {@link AmqpConverterFuture}.
+	 */
+	<C> AmqpConverterFuture<C> convertSendAndReceive(String routingKey, Object message) throws AmqpException;
```
artembilan: I wonder what is a difference of this method from others that only this one is throwing `AmqpException`, although the last one is `RuntimeException` any way...

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/AsyncRabbitTemplate.java
```diff
@@ -71,15 +74,15 @@
  * @since 1.6
  */
 public class AsyncRabbitTemplate implements SmartLifecycle, MessageListener, ReturnCallback, ConfirmCallback,
-		BeanNameAware {
+		BeanNameAware, AsyncAmqpTemplate {
```
artembilan: Nit-pick, but `AsyncAmqpTemplate` is the main contract of this class. So, it should be first in the `implements` declaration. :smile:

spring-amqp/src/main/java/org/springframework/amqp/core/AsyncAmqpTemplate.java
garyrussell: Looks like it was a copy/paste from `RabbitTemplate`; I don't remember but I must have removed them from the other methods, I am not sure the value of `throws RuntimeException` except, perhaps documentation. But they should be consistent. I can go either way - WDYT?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/AsyncRabbitTemplate.java
garyrussell: True.

spring-amqp/src/main/java/org/springframework/amqp/core/AsyncAmqpTemplate.java
artembilan: I would have it removed.
Thanks

artembilan:
LGTM.
It is ready for merge, but having my problem with tests it would be better to merge #471 first.
Thanks

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/73fcb2c816683d4b0aadb3d54608d6ea67c6cac9 after rebase.

PR #472, opened 2016-10-07
AMQP-616: Support Direct ReplyTo in Async Template

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-616

As an alternative to using a SMLC reply-container for the async template,
instead, use a specially configured DMLC to listen on the channels amq.rabbitmq.reply-to
pseudo queue.

Cache these containers for reuse.

When retrieving a container from the cache, check that it's still valid before using.

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/core/AsyncAmqpTemplate.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/core/AsyncAmqpTemplate.java
@@ -34,7 +34,7 @@ public interface AsyncAmqpTemplate {
 	 * @param message the message.
 	 * @return the {@link AmqpMessageFuture}.
 	 */
-	AmqpMessageFuture sendAndReceive(Message message);
+	ListenableFuture<Message> sendAndReceive(Message message);
 
 	/**
 	 * Send a message to the default exchange with the supplied routing key. If the message
@@ -43,7 +43,7 @@ public interface AsyncAmqpTemplate {
 	 * @param message the message.
 	 * @return the {@link AmqpMessageFuture}.
 	 */
-	AmqpMessageFuture sendAndReceive(String routingKey, Message message);
+	ListenableFuture<Message> sendAndReceive(String routingKey, Message message);
 
 	/**
 	 * Send a message to the supplied exchange and routing key. If the message
@@ -53,7 +53,7 @@ public interface AsyncAmqpTemplate {
 	 * @param message the message.
 	 * @return the {@link AmqpMessageFuture}.
 	 */
-	AmqpMessageFuture sendAndReceive(String exchange, String routingKey, Message message);
+	ListenableFuture<Message> sendAndReceive(String exchange, String routingKey, Message message);
 
 	/**
 	 * Convert the object to a message and send it to the default exchange with the
@@ -62,7 +62,7 @@ public interface AsyncAmqpTemplate {
 	 * @param <C> the expected result type.
 	 * @return the {@link AmqpConverterFuture}.
 	 */
-	<C> AmqpConverterFuture<C> convertSendAndReceive(Object object);
+	<C> ListenableFuture<C> convertSendAndReceive(Object object);
 
 	/**
 	 * Convert the object to a message and send it to the default exchange with the
@@ -72,7 +72,7 @@ public interface AsyncAmqpTemplate {
 	 * @param <C> the expected result type.
 	 * @return the {@link AmqpConverterFuture}.
 	 */
-	<C> AmqpConverterFuture<C> convertSendAndReceive(String routingKey, Object object);
+	<C> ListenableFuture<C> convertSendAndReceive(String routingKey, Object object);
 
 	/**
 	 * Convert the object to a message and send it to the provided exchange and
@@ -83,7 +83,7 @@ public interface AsyncAmqpTemplate {
 	 * @param <C> the expected result type.
 	 * @return the {@link AmqpConverterFuture}.
 	 */
-	<C> AmqpConverterFuture<C> convertSendAndReceive(String exchange, String routingKey, Object object);
+	<C> ListenableFuture<C> convertSendAndReceive(String exchange, String routingKey, Object object);
 
 	/**
 	 * Convert the object to a message and send it to the default exchange with the
@@ -94,7 +94,7 @@ public interface AsyncAmqpTemplate {
 	 * @param <C> the expected result type.
 	 * @return the {@link AmqpConverterFuture}.
 	 */
-	<C> AmqpConverterFuture<C> convertSendAndReceive(Object object, MessagePostProcessor messagePostProcessor);
+	<C> ListenableFuture<C> convertSendAndReceive(Object object, MessagePostProcessor messagePostProcessor);
 
 	/**
 	 * Convert the object to a message and send it to the default exchange with the
@@ -106,7 +106,7 @@ public interface AsyncAmqpTemplate {
 	 * @param <C> the expected result type.
 	 * @return the {@link AmqpConverterFuture}.
 	 */
-	<C> AmqpConverterFuture<C> convertSendAndReceive(String routingKey, Object object,
+	<C> ListenableFuture<C> convertSendAndReceive(String routingKey, Object object,
 			MessagePostProcessor messagePostProcessor);
 
 	/**
@@ -120,23 +120,7 @@ public interface AsyncAmqpTemplate {
 	 * @param <C> the expected result type.
 	 * @return the {@link AmqpConverterFuture}.
 	 */
-	<C> AmqpConverterFuture<C> convertSendAndReceive(String exchange, String routingKey, Object object,
+	<C> ListenableFuture<C> convertSendAndReceive(String exchange, String routingKey, Object object,
 			MessagePostProcessor messagePostProcessor);
-
-	/**
-	 * Return type from {@code sendAndReceive()} methods.
-	 *
-	 */
-	interface AmqpMessageFuture extends ListenableFuture<Message> {
-
-	}
-
-	/**
-	 * Return type from {@code convertSendAndReceive()} methods.
-	 *
-	 */
-	interface AmqpConverterFuture<C> extends ListenableFuture<C> {
-
-	}
 
 }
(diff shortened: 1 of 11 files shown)


spring-amqp/src/main/java/org/springframework/amqp/core/AsyncAmqpTemplate.java
```diff
@@ -115,12 +115,12 @@
...
 	 * @param exchange the exchange
 	 * @param routingKey the routing key.
-	 * @param message the message.
+	 * @param object the object to convert.
 	 * @param messagePostProcessor the post processor.
 	 * @param <C> the expected result type.
 	 * @return the {@link AmqpConverterFuture}.
 	 */
-	<C> AmqpConverterFuture<C> convertSendAndReceive(String exchange, String routingKey, Object message,
+	<C> AmqpConverterFuture<C> convertSendAndReceive(String exchange, String routingKey, Object object,
```
artembilan: Recall me, please, why do we need these two `ListenableFuture` extensions without any customization and so on.
I mean `AmqpConverterFuture`  and `AmqpMessageFuture`
Thanks

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/AsyncRabbitTemplate.java
```diff
@@ -186,6 +200,35 @@ public AsyncRabbitTemplate(RabbitTemplate template, AbstractMessageListenerConta
 	}
 
 	/**
+	 * Construct an instance using the provided arguments. "Direct replyTo" is used for
+	 * replies.
+	 * @param connectionFactory the connection factory.
+	 * @param exchange the default exchange to which requests will be sent.
+	 * @param routingKey the default routing key.
```
artembilan: `@since 2.0`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/AsyncRabbitTemplate.java
```diff
@@ -186,6 +200,35 @@ public AsyncRabbitTemplate(RabbitTemplate template, AbstractMessageListenerConta
...
+		this.template.setExchange(exchange == null ? "" : exchange);
+		this.template.setRoutingKey(routingKey);
+		this.container = null;
+		this.replyAddress = null;
+	}
+
+	/**
+	 * Construct an instance using the provided arguments. "Direct replyTo" is used for
+	 * replies.
+	 * @param template a {@link RabbitTemplate}
```
artembilan: DITO

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/AsyncRabbitTemplate.java
```diff
@@ -312,105 +352,132 @@ public RabbitMessageFuture sendAndReceive(String exchange, String routingKey, Me
...
+			}
+			channel = container.getChannel();
+		}
+		final DirectReplyToMessageListenerContainer theContainer = container;
+		container.setMessageListener(message -> {
+			onMessage(message);
+			try {
+				synchronized (this.directConsumersMonitor) {
+					if (this.running) {
+						this.directConsumers.put(theContainer);
```
artembilan: Why can't we live with he single `DirectReplyToMessageListenerContainer` instance?
Why do we need pool of them?
Thanks

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/AsyncRabbitTemplateTests.java
```diff
@@ -336,12 +420,12 @@ public ConnectionFactory connectionFactory() {
...
 		@Bean
 		public Queue requests() {
-			return new AnonymousQueue();
+			return new AnonymousQueue(new AnonymousQueue.Base64UrlNamingStrategy());
 		}
 
 		@Bean
 		public Queue replies() {
-			return new AnonymousQueue();
+			return new AnonymousQueue(new AnonymousQueue.Base64UrlNamingStrategy());
```
artembilan: I think we can revert this change already since `AnonymousQueue` is now Base64 by default.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/AsyncRabbitTemplate.java
garyrussell: Because we need a pool of channels - a channel needs to be dedicated for a send/receive operation. I suppose we could implement the pool within the DMLC and use `consumersPerQueue` to increase the pool size I'll take a look at it.

artembilan: M-m-m. Let me guess: with the exclusive `channel` we pursue correlation.
One send and only after that one receive.

artembilan: Wow! Good to know.
Thanks

artembilan:
Travis: 
```
:spring-rabbit:javadoc/home/travis/build/spring-projects/spring-amqp/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java:591: error: @param name not found

	 * @param the consumer.
	          ^

/home/travis/build/spring-projects/spring-amqp/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java:593: warning: no @param for consumer

	protected void consumerRemoved(SimpleConsumer consumer) {
	               ^
```

Will fix on merge if only concern.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectReplyToMessageListenerContainer.java
```diff
@@ -0,0 +1,132 @@
...
+			Channel channel = null;
+			while (channel == null) {
+				for (SimpleConsumer consumer : getConsumers()) {
+					if (this.inUseConsumerChannels.putIfAbsent(consumer.getChannel(), Boolean.TRUE) == null) {
+						channel = consumer.getChannel();
+						break;
+					}
+				}
+				if (channel == null) {
+					this.consumerCount++;
```
artembilan: I think we are missing to decrease this variable when we remove consumers

garyrussell: The count is never decremented; we only remove them from the in use map - the monitor thread will recover any consumers that die.

garyrussell:
All good comments; thanks; working on a new commit - almost ready but will finish tomorrow.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectReplyToMessageListenerContainer.java
artembilan: Well, but do we need extra consumers, if the pace from produces is calming down?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -447,8 +454,12 @@ private void checkMissingQueues(String[] queueNames) {
 	}
 
 	private void consumeFromQueue(String queue) {
-		for (int i = 0; i < this.consumersPerQueue; i++) {
-			doConsumeFromQueue(queue);
+		List<SimpleConsumer> list = this.consumersByQueue.get(queue);
+		// Possible race with setConsumersPerQueue and the task launched by start()
+		if (list == null || list.size() == 0) {
```
artembilan: `CollectionUtils.isEmpty()`, but, eh..., minor.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectReplyToMessageListenerContainer.java
```diff
@@ -0,0 +1,198 @@
...
+	 * Set cancelConsumer to true if the client is not prepared to handle/discard a
+	 * late arriving reply.
+	 * @param channel the channel.
+	 * @param cancelConsumer true to cancel the consumer.
+	 * @param message a message to be included in the cancel event if cancelConsumer is true.
+	 */
+	public void releaseConsumerFor(Channel channel, boolean cancelConsumer, String message) {
+		SimpleConsumer consumer = this.inUseConsumerChannels.remove(channel);
+		if (consumer != null && cancelConsumer) {
+			this.logger.error("Consumer canceled by client" + consumer);
```
artembilan: What is the purpose of this `error()`? Why simple `debug()` is not enough?
and at least whitespace after "client" word.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectReplyToMessageListenerContainer.java
```diff
@@ -0,0 +1,198 @@
...
+			consumer.cancelConsumer("Consumer " + this + " canceled due to " + message);
```
artembilan: Or `message` must never be `null`, or we should reconsider this sentence to honor `null` variant.

garyrussell: I don't understand these comments - the `message` is only used when `cancelConsumer` is true. What am I missing? Please elaborate.

artembilan: Ah! Indeed.
So, we should ensure that `message` isn't `null`, when `cancelConsumer == true`.
I don't say that only one `cancelMessage` and check it for `null` would be good solution (although for computer it is really enough :smile: ), but at least we will provide the proper message for human being. Otherwise we can end up with the:
```
... canceled due to null
```

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/8cdbd4cd170a620ec23f7901ad570859c5011b22

PR #474, opened 2016-10-10
AMQP-654: DMLC: Recover from Broker Interruption

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-654


kartoffelsup:
You can terminate the stream with a forEach, the collect operation is redundant. You are basically iterating twice.

garyrussell:
@kartoffelsup Thanks for reviewing.

Unfortunately not; the `cancelConsumer()` method modifies `this.consumers` so we will get a `ConcurrentModificationException`.

I anticipated such a suggestion and considered adding a comment to the `collect()`. I guess I should have done so.

kartoffelsup:
Makes sense, sorry - didn't catch that! A comment really sounds like a good idea here though :)

artembilan:
LGTM.
Merging...

@kartoffelsup , thank you very much for review!

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/7aa924e9ec502361cf6c287750eaf3eeac473b03

PR #478, opened 2016-10-12
AMQP-196: Use RabbitOperations in Gateway Support

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-196

Deprecate old methods.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitGatewaySupport.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitGatewaySupport.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2015 the original author or authors.
+ * Copyright 2002-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,6 +22,7 @@ import org.apache.commons.logging.LogFactory;
 import org.springframework.amqp.rabbit.connection.ConnectionFactory;
 import org.springframework.beans.factory.BeanInitializationException;
 import org.springframework.beans.factory.InitializingBean;
+import org.springframework.util.Assert;
 
 /**
  * Convenient super class for application classes that need RabbitMQ access.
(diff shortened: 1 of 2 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitGatewaySupport.java
```diff
@@ -79,18 +80,39 @@ public final ConnectionFactory getConnectionFactory() {
...
 	 */
+	@Deprecated
 	public final void setRabbitTemplate(RabbitTemplate rabbitTemplate) {
 		this.rabbitTemplate = rabbitTemplate;
 	}
 
 	/**
+	 * Set the RabbitTemplate for the gateway.
+	 * @param rabbitTemplate The Rabbit template.
+	 * @see #setConnectionFactory(org.springframework.amqp.rabbit.connection.ConnectionFactory)
```
artembilan: `@since 1.6`  ?
Will fix on merge.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitGatewaySupport.java
```diff
@@ -79,18 +80,39 @@ public final ConnectionFactory getConnectionFactory() {
...
+	 */
+	public final void setRabbitOperations(RabbitOperations rabbitTemplate) {
+		Assert.isInstanceOf(RabbitTemplate.class, rabbitTemplate);
```
artembilan: Seems for me the original concern that `RabbitTemplate` maybe AOP proxy, therefore with assert is gonna fail.
Isn't it ?

garyrussell: Yes, but then the cast will fail anyway - this commit is just to introduce the new methods and deprecate the old ones.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/0cae572a185a0eacc528b88b961767c68ba64360

PR #480, opened 2016-10-14
AMQP-616: Direct ReplyTo Container Race Condition

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-616

Consider the case where a reply is received at the same time as the future times
out or is canceled. At the same time, another request gets the same channel.

It is possible that the "late" future timeout/cancel will remove the consumer from
the in-use map even though the "next" user is waiting for the reply.

Add an epoch to the consumer which is incremented on each use. When releasing
the consumer, only do so if the epoch matches.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -97,7 +97,7 @@ subprojects { subproject ->
 
 		springVersion = project.hasProperty('springVersion') ? project.springVersion : '5.0.0.BUILD-SNAPSHOT'
 
-		springRetryVersion = '1.2.0.BUILD-SNAPSHOT'
+		springRetryVersion = '1.2.0.RC1'
 	}
 
 	eclipse {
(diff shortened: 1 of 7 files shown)


artembilan:
Hm. Travis doesn't like something here...:
```
org.springframework.amqp.rabbit.AsyncRabbitTemplateTests > testConvert1ArgDirect FAILED

    java.lang.AssertionError at AsyncRabbitTemplateTests.java:104
```

garyrussell:
Yes - I introduced a race by moving the remove until after `onMessage()` - will fix.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/d2bc2df562d3e498f706a339051a35993dce8e22

PR #485, opened 2016-10-17
Update to amqp-client Version 4.0.0.M2

garyrussell:


artembilan:
Does Travis failure say anything to you?
```
org.springframework.amqp.rabbit.connection.CachingConnectionFactoryIntegrationTests > testReceiveFromNonExistentVirtualHost FAILED

    java.lang.AssertionError
```
Does this `4.0` Client still work with Broker `3.x`, at least `3.6` ?

garyrussell:
Strange - the tests run on my machine - that was one of the tests I had to change (different exception). Yes, it's compatible, the client version is no longer tied to the server version - see [here](https://groups.google.com/forum/#!topic/rabbitmq-users/KpLgNNKyxic).

garyrussell:
Failed on my linux machine with 3.6.2 - investigating

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactoryIntegrationTests.java
```diff
@@ -255,14 +256,16 @@ public void testSendAndReceiveFromVolatileQueue() throws Exception {
...
-		exception.expect(AmqpIOException.class);
-		String result = (String) template.receiveAndConvert("foo");
-		assertEquals("message", result);
-
+		try {
+			template.receiveAndConvert("foo");
+			fail("Exception expected");
+		}
+		catch (AmqpIOException | AmqpAuthenticationException e) {
+			// expected
```
artembilan: I understand that it doesn't have much sense, but seems for me we should finish this test with some assertion. 
```
exception.expect(Matchers.anyOf(Matchers.instanceOf(AmqpIOException.class),
				Matchers.instanceOf(AmqpAuthenticationException.class)));
```
?

Just let me know and I'll fix on merge!

artembilan: Well, if there is no assertion, we fail with just a raw exception which isn't caught by this `catch` block.
With the test report it isn't clear what's wrong. Otherwise, with an assertion any our future changes make us to think twice to do them, since we have particular test for an appropriate expectation.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/a1bec868c88d1ea342b91cd5159490be091980c9

PR #487, opened 2016-10-19
AMQP-655: RabbitTemplate and DirectReply Container

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-655

Use a `DirectReplyToMessageListenerContainer` for Direct reply-to by default.

Add an option to revert to previous behavior of creating a new consumer for each
request.

Fix a problem in the `AmqpInvokerServiceExporter` where the correlation id was not
returned.

Fix a test `testSendToMissingExchange` - sometimes we can get a connectio already closed
exception.

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/remoting/service/AmqpInvokerServiceExporter.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/remoting/service/AmqpInvokerServiceExporter.java
@@ -79,11 +79,12 @@ public class AmqpInvokerServiceExporter extends RemoteInvocationBasedExporter im
 			RemoteInvocation invocation = (RemoteInvocation) invocationRaw;
 			remoteInvocationResult = invokeAndCreateResult(invocation, getService());
 		}
-		send(remoteInvocationResult, replyToAddress);
+		send(remoteInvocationResult, replyToAddress, message);
 	}
 
-	private void send(Object object, Address replyToAddress) {
+	private void send(Object object, Address replyToAddress, Message requestMessage) {
 		Message message = this.messageConverter.toMessage(object, new MessageProperties());
+		message.getMessageProperties().setCorrelationId(requestMessage.getMessageProperties().getCorrelationId());
 
 		getAmqpTemplate().send(replyToAddress.getExchangeName(), replyToAddress.getRoutingKey(), message);
 	}
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
@@ -601,17 +601,18 @@ public class RabbitTemplate extends RabbitAccessor implements BeanFactoryAware, 
 	public Collection<String> expectedQueueNames() {
 		this.isListener = true;
 		Collection<String> replyQueue = null;
-		if (this.replyAddress != null) {
+		if (this.replyAddress == null || this.replyAddress.equals(Address.AMQ_RABBITMQ_REPLY_TO)) {
+			throw new IllegalStateException("A listener container must not be provided when using direct reply-to");
+		}
+		else if (this.replyAddress != null) {
 			Address address = new Address(this.replyAddress);
 			if ("".equals(address.getExchangeName())) {
 				replyQueue = Collections.singletonList(address.getRoutingKey());
 			}
 			else {
-				logger.debug("Cannot verify reply queue because it has the form 'exchange/routingKey'");
+				logger.info("Cannot verify reply queue because 'replyAddress' is not a simple queue name: "
+						+ this.replyAddress);
 			}
-		}
-		else if (this.replyAddress == null || this.replyAddress.equals(Address.AMQ_RABBITMQ_REPLY_TO)) {
-			throw new IllegalStateException("A listener container must not be provided when using direct reply-to");
 		}
 		return replyQueue;
 	}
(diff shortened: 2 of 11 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -592,6 +610,9 @@ public void setUserIdExpressionString(String userIdExpression) {
 				logger.debug("Cannot verify reply queue because it has the form 'exchange/routingKey'");
 			}
 		}
+		else if (this.replyAddress == null || this.replyAddress.equals(Address.AMQ_RABBITMQ_REPLY_TO)) {
```
artembilan: Looks like this `else if` must be first.
Because in case of ` this.replyAddress = Address.AMQ_RABBITMQ_REPLY_TO` the `this.replyAddress` is really `not null`.
Need a test ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -1252,50 +1276,86 @@ protected Message doSendAndReceiveWithFixed(final String exchange, final String
...
-			}
-			pendingReply.setSavedCorrelation(savedCorrelation);
-			if (RabbitTemplate.this.correlationKey == null) { // using standard correlationId property
-				message.getMessageProperties().setCorrelationId(messageTag);
-			}
-			else {
-				message.getMessageProperties().setHeader(RabbitTemplate.this.correlationKey, messageTag);
-			}
+		}
+		ChannelHolder channelHolder = container.getChannelHolder();
```
artembilan: container_npe

And I see how. You don't initialize `container` variable in the `synchronized`.
If `this.directReplyToContainers.get(connectionFactory) != null` we go out from the `synchronized` block, but `container` can be still `null` because that was the reason to step into `synchronized` block.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -1252,50 +1276,86 @@ protected Message doSendAndReceiveWithFixed(final String exchange, final String
...
+		return reply;
+	}
+
+	private void saveAndSetHeaders(final Message message, final PendingReply pendingReply, String messageTag) {
+		// Save any existing replyTo and correlation data
+		String savedReplyTo = message.getMessageProperties().getReplyTo();
+		pendingReply.setSavedReplyTo(savedReplyTo);
+		if (StringUtils.hasLength(savedReplyTo) && logger.isDebugEnabled()) {
+			logger.debug("Replacing replyTo header:" + savedReplyTo
+					+ " in favor of template's configured reply-queue:"
```
artembilan: Whitespaces after colons.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -1252,50 +1276,86 @@ protected Message doSendAndReceiveWithFixed(final String exchange, final String
...
+					+ RabbitTemplate.this.replyAddress);
+		}
+		message.getMessageProperties().setReplyTo(RabbitTemplate.this.replyAddress);
```
artembilan: `RabbitTemplate.` prefix for is redundant.
But, eh..., minor.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -1566,7 +1627,8 @@ public void addListener(Channel channel) {
 			if (this.publisherConfirmChannels.putIfAbsent(key, this) == null) {
 				publisherCallbackChannel.addListener(this);
 				if (logger.isDebugEnabled()) {
-					logger.debug("Added pubsub channel: " + channel + " to map, size now " + this.publisherConfirmChannels.size());
+					logger.debug("Added confirmable channel: " + channel + " to map, size now "
```
artembilan: I don't know this `confirmable` word :smile: 

Maybe `publisher-confirm` would be better? 
At least that is how `PublisherCallbackChannelImpl.toString()` represents it...

spring-amqp/src/main/java/org/springframework/amqp/remoting/service/AmqpInvokerServiceExporter.java
```diff
@@ -79,11 +79,12 @@ public void onMessage(Message message) {
...
 			remoteInvocationResult = invokeAndCreateResult(invocation, getService());
 		}
-		send(remoteInvocationResult, replyToAddress);
+		send(remoteInvocationResult, replyToAddress, message);
 	}
 
-	private void send(Object object, Address replyToAddress) {
+	private void send(Object object, Address replyToAddress, Message requestMessage) {
 		Message message = this.messageConverter.toMessage(object, new MessageProperties());
+		message.getMessageProperties().setCorrelationId(requestMessage.getMessageProperties().getCorrelationId());
```
garyrussell: Well, really, the client has to be Spring AMQP - it's using Spring Remoting over AMQP.

> do we have...

Yes; the `RemotingTests` failed with this (using a DRTMLC requires the correlationId).

Yes, it's a bug and should be back-ported.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/9c0aac63caca9d5358ea6bd31122e6277b46e2d1

PR #488, opened 2016-10-20
AMQP-663: Fix Race in Container Test

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-663

The test is waiting for the consumer to be refreshed after deleting a queue by
inspecting the consumers map. There is a short period where the map is empty and
the test does not protect for that when iterating. Adding hasNext() won't help -
it will reduce the probability but won't eliminate it.

Catch the NoSuchElementException in the while loop and continue to wait.

__cherry-pick to 1.6.x, 1.5.x__


artembilan:
... and cherry-picked.

PR #489, opened 2016-10-21
fix javadoc formatting in class EnableRabbit

tbcs:


artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/9df8c13ee782529142ededffbdbfb8d6afc1a646 with some minor fix for similar formatting problem in the XML sample.

Thank you very much for the contribution.
Looking forward for more!

PR #490, opened 2016-10-21
RabbitManagementTemplate for CloudFoundry

garyrussell:
Requires: https://github.com/rabbitmq/hop/pull/87

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplate.java
@@ -70,6 +70,20 @@ public class RabbitManagementTemplate implements AmqpManagementOperations {
 
 	/**
 	 * Construct a template using the supplied uri.
+	 * @param uri the uri - must include user info, e.g.
+	 * "http://guest:guest@localhost:15672/api/".
+	 */
+	public RabbitManagementTemplate(String uri) {
+		try {
+			this.rabbitClient = new Client(uri, null, null);
+		}
+		catch (Exception e) {
+			throw new AmqpException(e);
+		}
+	}
+
+	/**
+	 * Construct a template using the supplied uri.
 	 * @param uri the uri.
 	 * @param username the user.
 	 * @param password the password.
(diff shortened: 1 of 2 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplate.java
```diff
@@ -70,6 +70,20 @@ public RabbitManagementTemplate(Client rabbitClient) {
 
 	/**
 	 * Construct a template using the supplied uri.
+	 * @param uri the uri - must include user info, e.g.
+	 * "http://guest:guest@localhost:15672/api/".
```
artembilan: `@since 2.0`? 
And can we use a new `Client(String uri)` ctor already ?

Just let me know and I'll fix on merge

garyrussell: Yes, we can use the new ctor - however, I'd like to cherry-pick this to 1.6.x  - we need to discuss with Michael when 1.0.1.RELEASE will be available.

artembilan: Well, I think for `1.6` we can do all that `UserInfo` work manually, if release isn't soon enough.

@michaelklishin , your thoughts? 
Thanks

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/bc7190e808934c6bac011fa471b8888c81eb9e3f

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitManagementTemplate.java
garyrussell: True; that's what I did for my [CloudFoundry PoC](https://github.com/garyrussell/sandbox/blob/master/repartition/src/main/java/org/springframework/amqp/protonj/RepartitionApplication.java#L120).

artembilan: Right, I also agree about `1.1`: there are so much changes already...

PR #493, opened 2016-10-24
AMQP-665: DMLC: Shut Down Scheduler

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-665

Spring Boot app with a DMLC, context.close(); app never terminates because the non-daemon threads hold the JVM.
    
Shut down the TS if an external one has not been provided.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
@@ -89,6 +89,8 @@ public class DirectMessageListenerContainer extends AbstractMessageListenerConta
 
 	private TaskScheduler taskScheduler;
 
+	private boolean taskSchedulerSet;
+
 	private long monitorInterval = DEFAULT_MONITOR_INTERVAL;
 
 	private volatile boolean started;
@@ -157,6 +159,7 @@ public class DirectMessageListenerContainer extends AbstractMessageListenerConta
 	 */
 	public void setTaskScheduler(TaskScheduler taskScheduler) {
 		this.taskScheduler = taskScheduler;
+		this.taskSchedulerSet = true;
 	}
 
 	/**
@@ -302,6 +305,7 @@ public class DirectMessageListenerContainer extends AbstractMessageListenerConta
 		if (this.taskScheduler == null) {
 			ThreadPoolTaskScheduler threadPoolTaskScheduler = new ThreadPoolTaskScheduler();
 			threadPoolTaskScheduler.setThreadNamePrefix(getBeanName() + "-consumerMonitor-");
+			threadPoolTaskScheduler.setDaemon(true);
 			threadPoolTaskScheduler.afterPropertiesSet();
 			this.taskScheduler = threadPoolTaskScheduler;
 		}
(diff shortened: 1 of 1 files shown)


artembilan:
I fully understand that your change isn't related to the Travis failure, but at least, I'll pay your attention :smile: : 
```
org.springframework.amqp.rabbit.core.RabbitTemplateDirectReplyToContainerIntegrationTests > testSendToMissingExchange FAILED

    org.springframework.amqp.AmqpException
```
Also I would like to see some explanation in the commit message. 
What is the reason of this kind of changes.
The JIRA ins't clear as well.
Plus that might be helpful for me in other places :smile: 
Thanks

garyrussell:
I updated the commit comment; we will probably have to wait for a Bamboo failure to diagnose the test failure - travis does not provide information.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -302,6 +302,7 @@ protected void doInitialize() throws Exception {
 		if (this.taskScheduler == null) {
 			ThreadPoolTaskScheduler threadPoolTaskScheduler = new ThreadPoolTaskScheduler();
 			threadPoolTaskScheduler.setThreadNamePrefix(getBeanName() + "-consumerMonitor-");
+			threadPoolTaskScheduler.setDaemon(true);
```
artembilan: H-m. Why don't call `ThreadPoolTaskScheduler.shutdown()` in the `actualShutDown()` instead?

garyrussell: OTOH - if someone creates these manually, we'll end up with a memory leak - will change.

artembilan: Right, something like `boolean taskSchedulerExplicitlySet`. We use such an approach in other places.

Otherwise we should revise the `shutdown` logic, because of just:
```
public void destroy() {
	shutdown();
}
```
In the `AbstractMessageListenerContainer`. 
I see that we call `shutdown();` in the `actualStart()`. Might be dangerous at some point.

OTOT re. `daemon`. I would like to understand how it finishes to schedule then, if we don't call any shutdown at all...

garyrussell: The monitor task is cancelled but the TE thread remains; if it's not daemon it holds up the JVM. We must make it daemon or shut down the TE when appropriate.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -552,6 +556,10 @@ protected void doShutdown() {
 				this.hasStopped = true;
 			}
 		}
+		if (!this.taskSchedulerSet && this.taskScheduler != null) {
+			((ThreadPoolTaskScheduler) this.taskScheduler).shutdown();
+			this.taskScheduler = null;
```
artembilan: Well, with the `squash` GH doesn't see changes and don't notify via email.
I have missed this change therefore :cry:

artembilan: I'm afraid that we can't do this unconditionally.
Look at the logic in the `actualStart()` and usage for `shutdown()` and this `doShutdown()`.
So, we retry until we are able to consume, but already on the first failure we destroy `this.consumerMonitorTask`.

Am I missing something?

OTOH do we still need `setDaemon(true)` if we close internal `taskScheduler` gracefully ?

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/e34f7711a6596ba8a366d618e673c74306b37249

PR #496, opened 2016-11-01
AMQP-667: Support 'delayed' on `@Exchange`

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-667

Add `delayed` property to `@Exchange` to support delayed message exchanges with
`@QueueBinding`.

Note: Ran the test locally with `"${no.prop:true}"` - Bamboo and travis don't have
the plugin.

__cherry-pick to 1.6.x__

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/core/ExchangeTypes.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/core/ExchangeTypes.java
@@ -34,6 +34,4 @@ public abstract class ExchangeTypes {
 
 	public static final String SYSTEM = "system";
 
-	public static final String DELAYED = "x-delayed-message";
-
 }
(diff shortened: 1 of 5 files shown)


artembilan:
Gary, how about to remove `ExchangeTypes.DELAYED` constant?
Looks like we have to use only other types and consult `"x-delayed-message"` only during the exchange declaration:
```
arguments.put("x-delayed-type", exchange.getType());
channel.exchangeDeclare(exchange.getName(), ExchangeTypes.DELAYED, exchange.isDurable(),
				exchange.isAutoDelete(), exchange.isInternal(), arguments);
```
I mean that config like:
```
exchange = @Exchange(value = "auto.start", type = ExchangeTypes.DELAYED, autoDelete = "true", delayed = "true")
```
confuses a bit...

spring-amqp/src/main/java/org/springframework/amqp/core/ExchangeTypes.java
```diff
@@ -34,6 +34,4 @@
 
 	public static final String SYSTEM = "system";
 
-	public static final String DELAYED = "x-delayed-message";
```
artembilan: I will deprecate it for `1.6.x`
Merging...

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/75a4a8dd7d55b843f745ba4a5f789f1007599fe4 and cherry-picked as https://github.com/spring-projects/spring-amqp/commit/943d45f8ed8d5aa86fd1067b117a1731aeffe516

PR #497, opened 2016-11-01
Avoid Synthetic ctors

garyrussell:
Private inner classes with a private ctor (real or implied) cause the compiler
to generate synthetic package-visibility ctors, with a synthetic class parameter if needed.

Also a few more lambdas.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/RetryInterceptorBuilder.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/RetryInterceptorBuilder.java
@@ -200,9 +200,6 @@ public abstract class RetryInterceptorBuilder<T extends MethodInterceptor> {
 	public abstract T build();
 
 
-	private RetryInterceptorBuilder() {
-	}
-
 	public static final class StatefulRetryInterceptorBuilder extends RetryInterceptorBuilder<StatefulRetryOperationsInterceptor> {
 
 		private final StatefulRetryOperationsInterceptorFactoryBean factoryBean =
@@ -211,6 +208,9 @@ public abstract class RetryInterceptorBuilder<T extends MethodInterceptor> {
 		private MessageKeyGenerator messageKeyGenerator;
 
 		private NewMessageIdentifier newMessageIdentifier;
+
+		StatefulRetryInterceptorBuilder() {
+		}
 
 		/**
 		 * Stateful retry requires messages to be identifiable. Default is to use the message id header; use a custom
@@ -283,9 +283,6 @@ public abstract class RetryInterceptorBuilder<T extends MethodInterceptor> {
 			return this.factoryBean.getObject();
 		}
 
-		private StatefulRetryInterceptorBuilder() {
-		}
-
 	}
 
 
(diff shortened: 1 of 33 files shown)


spring-rabbit/src/test/java/org/springframework/amqp/rabbit/test/RabbitMatchers.java
```diff
@@ -39,7 +39,7 @@ private RabbitMatchers() {
 
 		private final String pattern;
 
-		private RegexMatcher(String pattern) {
+		RegexMatcher(String pattern) {
```
artembilan: Enclosing `RabbitMatchers ` class has `private` ctor.
Omission?
I can fix on merge.

Were you accurate with inner classes, but `public`? :wink: 
https://github.com/spring-projects/spring-integration/commit/9ae8f329d7a4aed25d9b207a1628878f4a0989f3

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/RetryInterceptorBuilder.java
```diff
@@ -200,9 +200,6 @@ protected void applyCommonSettings(AbstractRetryOperationsInterceptorFactoryBean
 	public abstract T build();
 
 
-	private RetryInterceptorBuilder() {
-	}
```
artembilan: ??? I don't see package protected replacement.
Intentional? 
`protected` class is compiled with `protected` ctor?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/RetryInterceptorBuilder.java
```diff
@@ -212,6 +209,9 @@ private RetryInterceptorBuilder() {
 
 		private NewMessageIdentifier newMessageIdentifier;
 
+		StatefulRetryInterceptorBuilder() {
+		}
```
artembilan: Shouldn't we add `super();`?
Why then we do that in other places?

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/test/RabbitMatchers.java
garyrussell: No need - we never instantiate `RabbitMatchers` - it has a private ctor because it's just a utility class with static methods.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/RetryInterceptorBuilder.java
garyrussell: Again; never used.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/RetryInterceptorBuilder.java
garyrussell: Yes; just to keep sonar quiet.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/RetryInterceptorBuilder.java
artembilan: Got it. Thank you!
Merging with minor `super()` for that mentioned ctor....

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/87e2fd4041e9280904c2e1ab8fb5989f1473419b

PR #498, opened 2016-11-14
AMQP-671: Upgrade to Log4j2-2.7 and others

artembilan:
JIRA: https://jira.spring.io/browse/AMQP-671

* Upgrade to Log4j2-2.7 and fix compatibility
* Upgrade to Spring Retry 1.2 and remove reflection manipulation in the `SimpleMessageListenerContainer`, when `StatefulRetryOperationsInterceptor` is detected to set `UseRawKey` to `true`
* Deprecate `MissingMessageIdAdvice`
* Upgrade to Jackson `2.8.4` and fix usage of deprecated API
* Upgrade to SF-4.3.4 since it is minimum in Spring Boot 1.5
* The `1.7` XSD
* `What's New` for `1.7`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -11,7 +11,7 @@ buildscript {
 }
 
 plugins {
-	id "org.sonarqube" version "1.2"
+	id "org.sonarqube" version "2.1"
 }
 
 description = 'Spring AMQP'
@@ -93,18 +93,18 @@ subprojects { subproject ->
 		commonsIoVersion = '2.4'
 		hamcrestVersion = '1.3'
 		jacksonVersion = '1.9.13'
-		jackson2Version = '2.3.2'
+		jackson2Version = '2.8.4'
 		junitVersion = '4.12'
 		log4jVersion = '1.2.17'
-		log4j2Version = '2.5'
-		logbackVersion = '1.1.2'
-		mockitoVersion = '1.9.5'
+		log4j2Version = '2.7'
+		logbackVersion = '1.1.7'
+		mockitoVersion = '1.10.19'
 		rabbitmqVersion = project.hasProperty('rabbitmqVersion') ? project.rabbitmqVersion : '3.6.5'
 		rabbitmqHttpClientVersion = '1.0.0.RELEASE'
 
-		springVersion = project.hasProperty('springVersion') ? project.springVersion : '4.2.8.RELEASE'
+		springVersion = project.hasProperty('springVersion') ? project.springVersion : '4.3.4.RELEASE'
 
-		springRetryVersion = '1.1.3.RELEASE'
+		springRetryVersion = '1.2.0.RC1'
 	}
 
 	eclipse {
@@ -114,7 +114,7 @@ subprojects { subproject ->
 	}
 
 	jacoco {
-		toolVersion = "0.7.2.201409121644"
+		toolVersion = "0.7.7.201606060606"
 	}
 
 	// dependencies that are common across all java projects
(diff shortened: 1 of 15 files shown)


build.gradle
```diff
@@ -114,7 +114,7 @@ subprojects { subproject ->
 	}
 
 	jacoco {
-		toolVersion = "0.7.2.201409121644"
+		toolVersion = "0.7.7.201606060606"
```
garyrussell: Is this compatible with Sonar? I believe we tried to update this some time ago and the sonar builds failed.

build.gradle
```diff
@@ -242,7 +242,7 @@ project('spring-rabbit') {
 
 		compile ("org.apache.logging.log4j:log4j-core:$log4j2Version", optional)
 
-		testCompile "commons-cli:commons-cli:$commonsCliVersion"
+//		testCompile "commons-cli:commons-cli:$commonsCliVersion"
```
garyrussell: ??

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/retry/MissingMessageIdAdvice.java
```diff
@@ -40,8 +40,11 @@
  * Must be placed before the retry interceptor in the advice chain.
  * @author Gary Russell
  * @since 1.1.2
+ * @deprecated since 1.7 in favor of non-retryable messages logic in the {@code StatefulRetryOperationsInterceptor}
  *
  */
+@Deprecated
+@SuppressWarnings("deprecation")
```
garyrussell: We can't just deprecate it; it simply won't work any more with spring-retry 1.2.

build.gradle
garyrussell: Thanks - the SI coverage on sonar looks ok, so it seems ok.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/retry/MissingMessageIdAdvice.java
garyrussell: I think we have no choice.

garyrussell: Or, revert the container changes and just unconditionally `setRawKey` to true.

build.gradle
artembilan: I wondered why do we need this dependency...

And looks like there is nothing on the matter.

Although I have a problem with the `MessageListenerContainerRetryIntegrationTests.testStatefulRetryWithNoMessageIds()`, but not sure that it is related... I might have just missed something about Retry.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/retry/MissingMessageIdAdvice.java
artembilan: Eh... As for me the last variant sounds better as a smooth movement from 1.6 to 2.0 over 1.7.

Although I admit that I don't understand this new Retry logic yet...

build.gradle
artembilan: Well, there is no comments why we need that `commons-cli`: https://github.com/spring-projects/spring-amqp/commit/1150fe48997ab6f9a612d40a51b5c6051c5e7efe#diff-c197962302397baf3a4cc36463dce5eaR230.

Any ideas ?

garyrussell: I have no recollection as to why I added it - it's only a test dependency; let's just remove it, not just comment it out.

garyrussell:
Merged as df261f24918bf76ba4e9e2b6ded8fb6e179aeb0e with minor what's new polish.

PR #499, opened 2016-11-15
AMQP-672: SMLC: Support Start with No Queues

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-672

Also add events for consumer start/stop/restart.

Stop consumer immediately if no queues.

__I will issue another PR against master once reviewed/merged__

### DIFF
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerIntegration2Tests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerIntegration2Tests.java
@@ -18,23 +18,26 @@ package org.springframework.amqp.rabbit.listener;
 
 import static org.hamcrest.Matchers.contains;
 import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.equalTo;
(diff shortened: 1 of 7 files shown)


spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerIntegration2Tests.java
```diff
@@ -153,12 +158,66 @@ public void testChangeQueues() throws Exception {
...
+		container.setApplicationEventPublisher(new ApplicationEventPublisher() {
+
+			@Override
+			public void publishEvent(Object event) {
+				// NOSONAR
+			}
+
+			@Override
+			public void publishEvent(ApplicationEvent event) {
+				System.out.println(event);
```
artembilan: Overdid. 
Will remove on merge

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/5626f6ccfa9afb80d2e0037ab34ec12852e13c98.
Although I was a donkey to override commit message :cry: .
Didn't notice that it is a single commit in the PR...

Can we live with that? 
Or would be better to override history?

PR #500, opened 2016-11-16
chore: primitive long in ActiveObjectCounter.await()

panchenko:
There is no need to pass timeout as object Long here, looks like a typo.


artembilan:
Thank you, @panchenko, for the contribution!

PR #501, opened 2016-11-16
AMQP-673: 4.0.0.RC2 Client Phase 1, 2 (of 2)

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-673

Remove Use of Deprecated QueueingConsumer


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -1789,15 +1826,15 @@ private void restoreProperties(Message message, PendingReply pendingReply) {
...
 		channel.basicQos(1);
 		final CountDownLatch latch = new CountDownLatch(1);
-		QueueingConsumer consumer = new QueueingConsumer(channel) {
+		DefaultConsumer consumer = new DefaultConsumer(channel) {
 
 			@Override
 			public void handleCancel(String consumerTag) throws IOException {
-				super.handleCancel(consumerTag);
-				latch.countDown();
+				future.completeExceptionally(new ConsumerCancelledException());
```
artembilan: Hm... I think we should introduce some context to the exception to determine which consumer was canceled.
At least mention that `consumerTag`.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactoryIntegrationTests.java
```diff
@@ -312,8 +319,8 @@ public void testMixTransactionalAndNonTransactional() throws Exception {
 	}
 
 	@Test
-	public void testHardErrorAndReconnect() throws Exception {
-
+	public void testHardErrorAndReconnectNoAoto() throws Exception {
```
artembilan: Typo? `NoAuto` ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -1874,6 +1917,48 @@ public void returned(AmqpMessageReturnedException e) {
 	}
 
 	/**
+	 * A wrapper for the contents of a message delivery.
+	 */
+	public static class Delivery {
```
artembilan: This looks similar to the `BlockingQueueConsumer.Delivery`.
Can't we extract it as a top-level class now ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -1021,6 +1012,52 @@ else if (channelTransacted) {
 		return receiveMessage;
 	}
 
+	public Delivery getDeliveryAndCancelConsumer(Channel channel, CompletableFuture<Delivery> future,
```
artembilan: Is there critical reason to have this method as `public` ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -869,15 +871,9 @@ public Message receive(long timeoutMillis) throws AmqpException {
...
-			if (timeoutMillis < 0) {
-				delivery = consumer.nextDelivery();
-			}
-			else {
-				delivery = consumer.nextDelivery(timeoutMillis);
-			}
-			channel.basicCancel(consumer.getConsumerTag());
+			CompletableFuture<Delivery> future = new CompletableFuture<>();
+			Delivery delivery = getDeliveryAndCancelConsumer(channel, future,
+					createConsumer(queueName, channel, future), timeoutMillis);
```
artembilan: Looks like `future` variable and `createConsumer()` logic can be hide in the `getDeliveryAndCancelConsumer()`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -995,15 +991,10 @@ else if (channelTransacted) {
...
-			Delivery delivery;
-			if (RabbitTemplate.this.receiveTimeout < 0) {
-				delivery = consumer.nextDelivery();
-			}
-			else {
-				delivery = consumer.nextDelivery(RabbitTemplate.this.receiveTimeout);
-			}
-			channel.basicCancel(consumer.getConsumerTag());
+			CompletableFuture<Delivery> future = new CompletableFuture<>();
+			long timeoutMillis = this.receiveTimeout;
```
artembilan: What is the reason in this variable?
Thanks

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -1806,13 +1843,19 @@ public void handleConsumeOk(String consumerTag) {
...
 
+			@Override
+			public void handleDelivery(String consumerTag, Envelope envelope, BasicProperties properties, byte[] body)
+					throws IOException {
+				future.complete(new Delivery(envelope, properties, body));
+			}
+
 		};
 		channel.basicConsume(queueName, consumer);
 		if (!latch.await(10, TimeUnit.SECONDS)) {
```
artembilan: Do we really still need this blocking operation?
We are waiting on the `CompletableFuture<Delivery>` anyway.
And that looks more natural to wait not more than `this.receiveTimeout`.
WDYT?

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactoryIntegrationTests.java
```diff
@@ -349,6 +356,71 @@ public void testHardErrorAndReconnect() throws Exception {
...
+					latch.countDown();
+					// This will be thrown on the Connection thread just before it dies, so basically ignored
+					throw new RuntimeException(cause);
+				});
+				Channel targetChannel = ((ChannelProxy) channel).getTargetChannel();
+				if (targetChannel instanceof AutorecoveringChannel) {
+					((AutorecoveringChannel) targetChannel).addRecoveryListener(listener);
+				}
+				else {
+					recoveryLatch.countDown();
```
artembilan: Is there some case when we jump to this `else` block?
Thanks

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/ClientRecoveryCompatibilityTests.java
```diff
@@ -44,6 +49,10 @@
 	public void testDefeatRecovery() throws Exception {
 		final Channel channel1 = mock(Channel.class);
 		when(channel1.isOpen()).thenReturn(true);
+		willAnswer(i -> {
+			return null;
```
artembilan: Minor. 
Can be simplified: `i -> null`

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/ClientRecoveryCompatibilityTests.java
```diff
@@ -68,6 +77,12 @@ public void testDefeatRecovery() throws Exception {
 		when(rabbitConn.isOpen()).thenReturn(false).thenReturn(true);
 		when(channel1.isOpen()).thenReturn(false);
 		conn2 = ccf.createConnection();
+		try {
+			channel = conn2.createChannel(false);
```
artembilan: Assignment isn't used.
Plus looks like we need `fail()`  after this operator.
Otherwise in case of successful execution we don't step into the `catch` below.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -1021,6 +1012,52 @@ else if (channelTransacted) {
...
+			// no result in time
+		}
+		try {
+			if (exception == null || !(exception instanceof ConsumerCancelledException)) {
+				channel.basicCancel(consumer.getConsumerTag());
+			}
+		}
+		catch (Exception e) {
+			if (this.logger.isDebugEnabled()) {
+				this.logger.debug("Failed to cancel consumer", e);
```
artembilan: Minor. 
We don't need `logger.isDebugEnabled()` if our log string is constant.
Although that might be better to include to the log at least `consumerTag`...

OTOH is that really intentional to `DEBUG` error problem?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/SimpleConnection.java
```diff
@@ -71,15 +74,41 @@ public void close() {
 
 	@Override
 	public boolean isOpen() {
-		return this.delegate != null
-				&& (this.delegate.isOpen() || this.delegate.getClass().getSimpleName().contains("AutorecoveringConnection"));
+		if (this.delegate instanceof AutorecoveringConnection && !this.delegate.isOpen()) {
+			throw new AmqpException("Auto recovery connection is not currently open");
```
garyrussell: If a connection is in recovery, we get "connection already closed exception" on channel creation, which is misleading - previously we always returned true for these connections. So I thought this is a more useful error.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
garyrussell: Remained after refactoring - will fix with previous comment.

artembilan:
BTW, the test from Travis really fails for me, too:
```
org.springframework.amqp.AmqpException: Auto recovery connection is not currently open
	at org.springframework.amqp.rabbit.connection.SimpleConnection.isOpen(SimpleConnection.java:78)
	at org.springframework.amqp.rabbit.connection.CachingConnectionFactory$ChannelCachingConnectionProxy.isOpen(CachingConnectionFactory.java:1153)
	at org.springframework.amqp.rabbit.connection.CachingConnectionFactory.createBareChannel(CachingConnectionFactory.java:496)
	at org.springframework.amqp.rabbit.connection.CachingConnectionFactory.access$500(CachingConnectionFactory.java:97)
	at org.springframework.amqp.rabbit.connection.CachingConnectionFactory$CachedChannelInvocationHandler.invoke(CachingConnectionFactory.java:932)
	at com.sun.proxy.$Proxy44.exchangeDeclare(Unknown Source)
	at org.springframework.amqp.rabbit.core.RabbitAdmin.declareExchanges(RabbitAdmin.java:504)
	at org.springframework.amqp.rabbit.core.RabbitAdmin.lambda$declareExchange$0(RabbitAdmin.java:157)
	at org.springframework.amqp.rabbit.core.RabbitTemplate.doExecute(RabbitTemplate.java:1503)
	at org.springframework.amqp.rabbit.core.RabbitTemplate.execute(RabbitTemplate.java:1458)
	at org.springframework.amqp.rabbit.core.RabbitTemplate.execute(RabbitTemplate.java:1439)
	at org.springframework.amqp.rabbit.core.RabbitAdmin.declareExchange(RabbitAdmin.java:156)
	at org.springframework.amqp.rabbit.core.RabbitAdminIntegrationTests.testDeclareDelayedExchange(RabbitAdminIntegrationTests.java:376)
```

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
garyrussell: Debug is enough here - most likely failure to cancel is closed connection.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
garyrussell: 👍

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
garyrussell: We're blocking downstream anyway on the future. This makes life a bit easier because we don't need to deal with races.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
garyrussell: 👍

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactoryIntegrationTests.java
garyrussell: 😄

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactoryIntegrationTests.java
garyrussell: Only if we run with a client that does not have autoRecovery true.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/ClientRecoveryCompatibilityTests.java
garyrussell: This is left behind from debugging - I needed to figure out where the close() was called from; can be removed.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/ClientRecoveryCompatibilityTests.java
garyrussell: 👍

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/SimpleConnection.java
artembilan: Ok. Why just not return `this.delegate.isOpen()`?
At least the exception is thrown only if it isn't open.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
artembilan: But looks like we are blocking for 10 secs plus that `receiveTimeout`. At most...
Also I can live with such a distinguishes between warming and real process.

If you see that there also can be some races, let's just leave as is and go ahead!

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactoryIntegrationTests.java
artembilan: And do we run such one?
I just want to be sure that all the code in test-case makes sense.

garyrussell: No we don't (I decided to retain the old test above) so I guess this can be removed.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/SimpleConnection.java
artembilan: My general concern here that we never `try...catch` this `isOpen()` and many our algorithm on that may end up unexpectedly. Like that `RabbitAdminIntegrationTests` on the `this.rabbitAdmin.declareExchange(exchange);`

artembilan:
Hm. The PR still fails on the : 
```
org.springframework.amqp.rabbit.core.RabbitAdminIntegrationTests > testDeclareDelayedExchange FAILED

    org.springframework.amqp.AmqpException at RabbitAdminIntegrationTests.java:377
```

garyrussell:
Interesting; I didn't think that would fix it but I'm struggling to see how it happens - when we destroy the connection in init(), the target is set to null so we should always get a new connection. Investigating...

garyrussell:
Ah - Linux (and presumable windows) only.

garyrussell:
Problem was with the missing delayed message exchange plugin.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/SimpleConnection.java
```diff
@@ -69,17 +74,55 @@ public void close() {
...
+	 * True if the connection is open.
+	 * @return true if the connection is open
+	 * @throws AutoRecoverConnectionNotCurrentlyOpenException if the connection is an
+	 * {@link AutorecoveringConnection} and is currently closed; this is required to
+	 * prevent the CCF from discarding this connection and opening a new one, in which
+	 * case the "old" connection would eventually be recovered and orphaned - also any
+	 * consumers belonging to it might be recovered too and the broker will deliver
+	 * messages to them when there is no code actually running to deal with those messages
+	 * (when using the SMLC). If we have actually closed the connection (e.g. via
+	 * CCF.resetConnection()) this will return false.
```
artembilan: I don't like reductions in public JavaDocs.
You can use `{@code CachingConnectionFactory}`.
Otherwise this is frustrating read for beginners.

artembilan:
Polishing for your confirmation: https://github.com/artembilan/spring-amqp/commit/58a4872e6749aee7c6dacc0b794cc1471951e099

garyrussell:
Polishing LGTM; thanks!

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/cf8584f366f37e38c051c4ce6b55d0ab6c607d89

PR #508, opened 2016-11-28
AMQP-676: @RabbitListener Improvements

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-676

1. Add the ability to propagate exceptions from POJO `@RabbitListener` methods to senders
(when using `replyTo/@SendTo`. Exceptions are propagated in a `RemoteInvocationResult`.
2. Add `RemoteInvocationAwareMessageConverterAdapter` to support unwrapping these exceptions
on the client side.
3. Note: this feature generally requires the use of the `SimpleMessageConverter` since exceptions
are `Serializable` and generally, are not JSON friendly.
4. Add `errorHandler` to `@RabbitListener`, allowing users to catch exceptions, send a modified result,
or throw some other exception.
5. Remove the (now) unnecessary wrapping of spring-messaging `MessageConversionException`s.

__review only - needs docs__

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/support/converter/RemoteInvocationAwareMessageConverterAdapter.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/support/converter/RemoteInvocationAwareMessageConverterAdapter.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.amqp.support.converter;
+
+import org.springframework.amqp.AmqpException;
+import org.springframework.amqp.core.Message;
+import org.springframework.amqp.core.MessageProperties;
+import org.springframework.beans.factory.BeanClassLoaderAware;
+import org.springframework.remoting.support.RemoteInvocationResult;
+import org.springframework.util.Assert;
+
+/**
+ * A delegating adapter that unwraps {@link RemoteInvocationResult} after invoking
+ * the delegate to convert from a message.
+ *
+ * @author Gary Russell
+ * @since 2.0
+ *
+ */
+public class RemoteInvocationAwareMessageConverterAdapter implements MessageConverter, BeanClassLoaderAware {
+
+	private final SimpleMessageConverter defaultMessageConverter = new SimpleMessageConverter();
+
+	private final MessageConverter delegate;
+
+	public RemoteInvocationAwareMessageConverterAdapter() {
+		this.delegate = this.defaultMessageConverter;
+	}
+
+	public RemoteInvocationAwareMessageConverterAdapter(MessageConverter delegate) {
+		Assert.notNull(delegate, "'delegate' converter cannot be null");
+		this.delegate = delegate;
+	}
+
+	@Override
+	public void setBeanClassLoader(ClassLoader classLoader) {
+		this.defaultMessageConverter.setBeanClassLoader(classLoader);
+	}
+
+	@Override
+	public Message toMessage(Object object, MessageProperties messageProperties) throws MessageConversionException {
+		return this.delegate.toMessage(object, messageProperties);
+	}
+
+	@Override
+	public Object fromMessage(Message message) throws MessageConversionException {
+		Object result = this.delegate.fromMessage(message);
+		if (result instanceof RemoteInvocationResult) {
+			try {
+				result = ((RemoteInvocationResult) result).recreate();
+			}
+			catch (Throwable e) {
+				throw new AmqpException(e);
+			}
+		}
+		return result;
+	}
+
+}
--- a/spring-amqp/src/main/java/org/springframework/amqp/support/converter/RemoteInvocationAwareMessageConverterAdapter.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/support/converter/RemoteInvocationAwareMessageConverterAdapter.java
@@ -26,6 +26,7 @@ import org.springframework.util.Assert;
 /**
  * A delegating adapter that unwraps {@link RemoteInvocationResult} after invoking
  * the delegate to convert from a message.
+ * Delegates to a {@link SimpleMessageConverter} by default.
  *
  * @author Gary Russell
  * @since 2.0
--- a/spring-amqp/src/main/java/org/springframework/amqp/support/converter/RemoteInvocationAwareMessageConverterAdapter.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/support/converter/RemoteInvocationAwareMessageConverterAdapter.java
@@ -16,7 +16,7 @@
 
 package org.springframework.amqp.support.converter;
 
-import org.springframework.amqp.AmqpException;
+import org.springframework.amqp.AmqpRemoteException;
 import org.springframework.amqp.core.Message;
 import org.springframework.amqp.core.MessageProperties;
 import org.springframework.beans.factory.BeanClassLoaderAware;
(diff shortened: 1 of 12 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerErrorHandler.java
```diff
@@ -0,0 +1,47 @@
...
+	 * the sender using normal {@code replyTo/@SendTo} semantics.
+	 * @param amqpMessage the raw message received.
+	 * @param message the converted spring-messaging message.
+	 * @param exception the exception the listener threw, wrapped in a
+	 * {@link ListenerExecutionFailedException}.
+	 * @return the return value to be sent to the sender.
+	 * @throws an exception which may be the original or different.
+	 */
+	Object handleError(Message amqpMessage, org.springframework.messaging.Message<?> message,
+			ListenerExecutionFailedException exception) throws Exception;
```
artembilan: Any reason to keep it wrapped to the `ListenerExecutionFailedException`?
Isn't that `failedMessage` is exactly our `amqpMessage`?
Thanks

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerErrorHandler.java
```diff
@@ -0,0 +1,47 @@
...
+
+	/**
+	 * Handle the error. If an exception is not thrown, the return value is returned to
+	 * the sender using normal {@code replyTo/@SendTo} semantics.
+	 * @param amqpMessage the raw message received.
+	 * @param message the converted spring-messaging message.
+	 * @param exception the exception the listener threw, wrapped in a
+	 * {@link ListenerExecutionFailedException}.
+	 * @return the return value to be sent to the sender.
+	 * @throws an exception which may be the original or different.
```
artembilan: I think this must be with `Exception` after `@throws`.
See Travis report.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -102,12 +115,39 @@ public void onMessage(org.springframework.amqp.core.Message amqpMessage, Channel
...
+					}
+					else {
+						logger.trace("Error handler returned no result");
+					}
+				}
+				catch (Exception ex) {
+					if (!this.propagateExceptions) {
+						throw ex;
+					}
+					handleResult(new RemoteInvocationResult(ex), amqpMessage, channel, message);
```
artembilan: Would you mind explaining the reason of the `RemoteInvocationResult` layer?
To get a gain from the `recreate()` ability to rethrow?
How about just to equip `SimpleMessageConverter` with a flag `rethrowResult`, since it looks like with this feature we always can deserialize `Exception` as is?

spring-amqp/src/main/java/org/springframework/amqp/support/converter/RemoteInvocationAwareMessageConverterAdapter.java
```diff
@@ -0,0 +1,73 @@
...
+
+	@Override
+	public Object fromMessage(Message message) throws MessageConversionException {
+		Object result = this.delegate.fromMessage(message);
+		if (result instanceof RemoteInvocationResult) {
+			try {
+				result = ((RemoteInvocationResult) result).recreate();
+			}
+			catch (Throwable e) {
+				throw new AmqpException(e);
```
garyrussell: `fromMessage()` would need to throw `Exception`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerErrorHandler.java
garyrussell: Good point; I guess we can unwrap.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapter.java
garyrussell: Because `RemoteInvocationResult.recreate()` merges the remote and local stack trace.

spring-amqp/src/main/java/org/springframework/amqp/support/converter/RemoteInvocationAwareMessageConverterAdapter.java
panchenko: Might be introduce something like AmqpRemoteException?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -102,12 +115,39 @@ public void onMessage(org.springframework.amqp.core.Message amqpMessage, Channel
...
+					Object result = this.errorHandler.handleError(amqpMessage, message, e);
+					if (result != null) {
+						handleResult(result, amqpMessage, channel, message);
+					}
+					else {
+						logger.trace("Error handler returned no result");
+					}
+				}
+				catch (Exception ex) {
+					if (!this.propagateExceptions) {
```
panchenko: Might be rename `propagateExceptions` everywhere to something more explicit, e.g. `returnExceptionsToSender`?

To me _if not propagate then throw_ looks confusing, as quite often it's written the other way around.
After rename it can be written as _if returnExceptionsToSender then handleResult(...) else throw_ and it's clear what it does.

spring-amqp/src/main/java/org/springframework/amqp/support/converter/RemoteInvocationAwareMessageConverterAdapter.java
```diff
@@ -0,0 +1,73 @@
...
+
+	private final MessageConverter delegate;
+
+	public RemoteInvocationAwareMessageConverterAdapter() {
+		this.delegate = this.defaultMessageConverter;
+	}
+
+	public RemoteInvocationAwareMessageConverterAdapter(MessageConverter delegate) {
+		Assert.notNull(delegate, "'delegate' converter cannot be null");
+		this.delegate = delegate;
```
panchenko: It seems like defaultMessageConverter is not needed when constructing this with the provided converter. Might be use a boolean field instead, so setBeanClassLoader works only if default converter?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerErrorHandler.java
garyrussell: I decided to leave this as-is - otherwise we have to get into the business of dealing with `Throwable`.

spring-amqp/src/main/java/org/springframework/amqp/support/converter/RemoteInvocationAwareMessageConverterAdapter.java
artembilan: I think that won't hurt, @panchenko .

There is `RemoteInvocationFailureException`, which has a `SoapFaultException` extension. So, our `AmqpRemoteInvocationException` would be fine in that family. Although that can't be `AmqpException` any more then...

spring-amqp/src/main/java/org/springframework/amqp/support/converter/RemoteInvocationAwareMessageConverterAdapter.java
artembilan: That has crossed my mind before as well. But I decided do not worry about this `defaultMessageConverter`, because in most cases the `RemoteInvocationAwareMessageConverterAdapter` will be used by default.
Although for better memory consumption we really can consider `boolean` flag instead of extra `new SimpleMessageConverter()` in each `RemoteInvocationAwareMessageConverterAdapter` instance.

Works for me either way. 
But yeah, good catch, @panchenko ! 😄

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/MethodRabbitListenerEndpoint.java
```diff
@@ -80,6 +84,25 @@ public void setMessageHandlerMethodFactory(MessageHandlerMethodFactory messageHa
...
+	 * @since 2.0
+	 */
+	public void setReturnExceptions(boolean returnExceptions) {
+		this.returnExceptions = returnExceptions;
+	}
+
+	/**
+	 * Set the {@link RabbitListenerErrorHandler} to invoke if the listener method
+	 * throws an exception.
+	 * @param errorHandler the error handler.
```
artembilan: `@since 2.0`

Will fix on merge

src/reference/asciidoc/amqp.adoc
```diff
@@ -2126,6 +2126,42 @@ static class TxServiceImpl implements TxService<Foo> {
 }
 ----
 
+[[annotation-error-handling]]
+====== Handling Exceptions
+
+By default, if an annotated listener method throws an exception, the exeption is thrown to the container and the message will be requeued and redelivered, discarded, or routed to a Dead Letter Exchange, depending on the container and broker configuration.
```
artembilan: Typo: ex**c**eption

Will fix on merge.

src/reference/asciidoc/amqp.adoc
```diff
@@ -2126,6 +2126,42 @@ static class TxServiceImpl implements TxService<Foo> {
...
+Use the `errorHandler` to provide the bean name of a `RabbitListenerErrorHandler` implementation.
+This functional interface has one method:
+
+[source, java]
+----
+@FunctionalInterface
+public interface RabbitListenerErrorHandler {
+
+	Object handleError(Message amqpMessage, org.springframework.messaging.Message<?> message,
+			ListenerExecutionFailedException exception) throws Exception;
```
artembilan: Whitespaces for indents instead of tabs.
Yes, in docs only 😄 

Fixing on merge...

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/04c7f851e789da189eb7756e6d1456cd8f0593c6

PR #513, opened 2016-11-30
Backport AMQP-673 and Earlier 4.0.0 Commits

garyrussell:


artembilan:
Does Travis report ring a bell for you?
```
org.springframework.amqp.rabbit.core.RabbitTemplatePublisherCallbacksIntegrationTests2 > classMethod FAILED

    org.springframework.beans.NotWritablePropertyException at RabbitTemplatePublisherCallbacksIntegrationTests2.java:64
```

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AutoRecoverConnectionNotCurrentlyOpenException.java
```diff
@@ -0,0 +1,36 @@
...
+package org.springframework.amqp.rabbit.connection;
+
+import org.springframework.amqp.AmqpException;
+
+/**
+ * An exception thrown if the connection is an auto recover connection
+ * that is not currently open.
+ *
+ * @author Gary Russell
+ * @since 2.0
```
artembilan: `@since 1.7`

Will fix on merge

build.gradle
```diff
@@ -99,8 +99,8 @@ subprojects { subproject ->
 		log4j2Version = '2.7'
 		logbackVersion = '1.1.7'
 		mockitoVersion = '1.10.19'
-		rabbitmqVersion = project.hasProperty('rabbitmqVersion') ? project.rabbitmqVersion : '3.6.5'
-		rabbitmqHttpClientVersion = '1.0.0.RELEASE'
+		rabbitmqVersion = project.hasProperty('rabbitmqVersion') ? project.rabbitmqVersion : '4.0.0'
```
artembilan: Does it mean that we don't support client `3.6.x` any more?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/Delivery.java
```diff
@@ -0,0 +1,78 @@
...
+
+import com.rabbitmq.client.AMQP;
+import com.rabbitmq.client.AMQP.BasicProperties;
+import com.rabbitmq.client.Envelope;
+
+/**
+ * Encapsulates an arbitrary message - simple "bean" holder structure.
+ *
+ * @author Gary Russell
+ * @since 2.0
```
artembilan: `@since 1.7`

src/reference/asciidoc/amqp.adoc
```diff
@@ -644,6 +644,17 @@ The `cacheMode` property (`CHANNEL` or `CONNECTION` is also included).
 .JVisualVM Example
 image::images/cacheStats.png[align="center"]
 
+[[auto-recovery]]
+===== RabbitMQ Automatic Connection/Topology recovery
+
+Since the first version of Spring AMQP, the framework has provided its own connection and channel recovery in the event of a broker failure.
+Also, as discussed in <<broker-configuration>>, the `RabbitAdmin` will re-declare any infrastructure beans (queues etc) when the connection is re-established.
+It therefore does not rely on the https://www.rabbitmq.com/api-guide.html#recovery[Auto Recovery] that is now provided by the `amqp-client` library.
+Spring AMQP now uses the `4.0.x` version of `amqp-client`, which has auto recovery enabled by default.
```
artembilan: If we assume to still support pre-4.0 clients, this sentence should be adjusted a bit:

Spring AMQP now **supports** the `4.0.x`...

build.gradle
garyrussell: I don't see any reason to - I guess we'll have to talk to @wilkinsona about the amqp-client version for boot 1.5, though.

artembilan: Eh... Yes, since client 4.0 is compatible with Broker 3.6 and this amqp-client is transitive dependency of Spring AMQP and I'm pretty sure there is no any other Spring project depended of amqp-client, therefore we are safe to upgrade to this major version.
Yes, it breaks a bit our IO Platform rules, but still arguments to upgrade and forget are valuable enough.
WDYT, @wilkinsona ?

src/reference/asciidoc/amqp.adoc
garyrussell: The 4.0.0. client is compatible with earlier brokers; I don't think we need to "support" older clients in this new minor release (1.7.0).

We should also cherry-pick @acogoluegnes commit after this is merged bd4397a015d51bc5b7a5d954bf940396aa138eea

build.gradle
wilkinsona: The AMQP client version's defined in the Platform. Brussels (based on Boot 1.5) has 3.6.5 in it at the moment. We could jump to 4.0 if necessary, but it does strike me as a little odd for AMQP 1.7 to require a new major version of the AMQP client. You'd know much better than me if that's a reasonable upgrade to be making, though.

garyrussell: Although the version number looks like a major release; I don't believe it's much different to the 3.6.x client - the numbering reflects that it has been detached from server versioning.

http://www.rabbitmq.com/blog/2016/11/24/rabbitmq-java-client-4-0-is-released/

artembilan: I guess the main reason of this upgrade is a robust auto-recovery function which is enabled by default now.
Seems for me there was even some community request to provide this upgrade earlier than Spring AMPQ 2.0...

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/4efba7d5ee552b22409a592c9c4ac4a3cf979b8e

artembilan:
> We should also cherry-pick @acogoluegnes commit after this is merged bd4397a

Done.

PR #518, opened 2016-12-03
Fix Sonar Detected Bad instanceof Check

garyrussell:
https://sonar.spring.io/component_issues?id=org.springframework.amqp%3Aspring-amqp-dist%3Amaster#resolved=false|types=BUG

Checking wrong connection object to attach recovery listener.

After merge, needs backport of this default listener.


garyrussell:
I mean when I backported support for the 4.0.0 client, I missed adding the default (logging) connection recovery listener. I wanted to wait for this to be merged (or approved) before doing that.

PR #521, opened 2016-12-08
AMQP-687: Fix Conditional Rollback

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-687

When using a rule-based transaction attribute with a rollback/no-rollback rule,
the rule is ignored when providing the transaction attribute directly to the container
rather than via a transaction advice,

Also, user exceptions are always wrapped in a `ListenerExecutionFailedException` which
limits the ability to configure rules, even when using an advice.

- Honor the attribute `rollBackOn()` mechanism within the container
- Add `ListenerFailedRuleBasedTransactionAttribute` to allow the user to configure rules based on the actual exception.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
@@ -695,7 +695,9 @@ public class DirectMessageListenerContainer extends AbstractMessageListenerConta
 								new TransactionTemplate(this.transactionManager, this.transactionAttribute);
 					}
 					this.transactionTemplate.execute(s -> {
-						ConnectionFactoryUtils.bindResourceToTransaction(new RabbitResourceHolder(getChannel(), false),
+						RabbitResourceHolder resourceHolder = new RabbitResourceHolder(getChannel(), false);
+						resourceHolder.addDeliveryTag(getChannel(), deliveryTag);
+						ConnectionFactoryUtils.bindResourceToTransaction(resourceHolder,
 								this.connectionFactory, true);
 						// unbound in ResourceHolderSynchronization.beforeCompletion()
 						try {
@@ -741,22 +743,22 @@ public class DirectMessageListenerContainer extends AbstractMessageListenerConta
 			}
 			catch (ImmediateAcknowledgeAmqpException e) {
 				if (this.logger.isDebugEnabled()) {
-					this.logger.debug("User requested ack for failed delivery");
+					this.logger.debug("User requested ack for failed delivery: " + deliveryTag);
 				}
 				handleAck(deliveryTag, channelLocallyTransacted);
 			}
 			catch (Exception e) {
 				if (causeChainHasImmediateAcknowledgeAmqpException(e)) {
 					if (this.logger.isDebugEnabled()) {
-						this.logger.debug("User requested ack for failed delivery");
+						this.logger.debug("User requested ack for failed delivery: " + deliveryTag);
 					}
+					handleAck(deliveryTag, channelLocallyTransacted);
 				}
 				else {
 					this.logger.error("Failed to invoke listener", e);
 					if (this.transactionManager != null) {
 						if (this.transactionAttribute.rollbackOn(e)) {
-							rollback(deliveryTag, e);
-							throw e;
+							throw e; // encompassing transaction will handle the rollback.
 						}
 						else {
 							if (this.logger.isDebugEnabled()) {
@@ -766,6 +768,7 @@ public class DirectMessageListenerContainer extends AbstractMessageListenerConta
 					}
 					else {
 						rollback(deliveryTag, e);
+						// no need to rethrow e - we'd ignore it anyway, not throw to client
 					}
 				}
 			}
@@ -774,14 +777,7 @@ public class DirectMessageListenerContainer extends AbstractMessageListenerConta
 		private void handleAck(long deliveryTag, boolean channelLocallyTransacted) throws IOException {
 			try {
 				if (this.ackRequired) {
-					if (isChannelTransacted() && !channelLocallyTransacted) {
-
-						// Not locally transacted but it is transacted so it
-						// could be synchronized with an external transaction
-						ConnectionFactoryUtils.registerDeliveryTag(this.connectionFactory, getChannel(), deliveryTag);
-
-					}
-					else {
+					if (!isChannelTransacted() || channelLocallyTransacted) {
 						getChannel().basicAck(deliveryTag, false);
 					}
 				}
@@ -800,7 +796,7 @@ public class DirectMessageListenerContainer extends AbstractMessageListenerConta
 			}
 			if (this.ackRequired) {
 				try {
-					getChannel().basicReject(deliveryTag,
+					getChannel().basicNack(deliveryTag, true,
 							RabbitUtils.shouldRequeue(isDefaultRequeueRejected(), e, this.logger));
 				}
 				catch (IOException e1) {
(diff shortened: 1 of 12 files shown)


garyrussell:
Needs docs - I will backport to 1.7/1.6 after review/merge.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -687,28 +690,89 @@ public void handleDelivery(String consumerTag, Envelope envelope,
...
+		private void callExecuteListener(Message message, long deliveryTag) throws Exception {
+			boolean channelLocallyTransacted = isChannelLocallyTransacted();
 			try {
 				executeListener(getChannel(), message);
-				boolean channelLocallyTransacted = isChannelLocallyTransacted();
+				handleAck(deliveryTag, channelLocallyTransacted);
+			}
+			catch (ImmediateAcknowledgeAmqpException e) {
+				if (this.logger.isDebugEnabled()) {
+					this.logger.debug("User requested ack for failed delivery");
```
artembilan: Minor: We don't need extra `isDebugEnabled()` if we have only static log message.
The same in the `causeChainHasImmediateAcknowledgeAmqpException(e)` block of code.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -687,28 +690,89 @@ public void handleDelivery(String consumerTag, Envelope envelope,
...
+				}
+				handleAck(deliveryTag, channelLocallyTransacted);
+			}
+			catch (Exception e) {
+				if (causeChainHasImmediateAcknowledgeAmqpException(e)) {
+					if (this.logger.isDebugEnabled()) {
+						this.logger.debug("User requested ack for failed delivery");
+					}
```
artembilan: Don't you miss here `handleAck(deliveryTag, channelLocallyTransacted);` then ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -687,28 +690,89 @@ public void handleDelivery(String consumerTag, Envelope envelope,
...
+				}
+				else {
+					this.logger.error("Failed to invoke listener", e);
+					if (this.transactionManager != null) {
+						if (this.transactionAttribute.rollbackOn(e)) {
```
artembilan: The `setTransactionAttribute()` must be supplied with the `Assert.notNull()`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -687,28 +690,89 @@ public void handleDelivery(String consumerTag, Envelope envelope,
...
+							rollback(deliveryTag, e);
+							throw e;
+						}
+						else {
+							if (this.logger.isDebugEnabled()) {
+								this.logger.debug("No rollback for " + e);
```
artembilan: And? What else?
Do we just leave message unacked in the queue? 
What happens afterwards?

Thanks

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -687,28 +690,89 @@ public void handleDelivery(String consumerTag, Envelope envelope,
...
+							}
+						}
+					}
+					else {
+						rollback(deliveryTag, e);
+					}
```
artembilan: Why don't we re-throw `e` in this `else`?
The previous one does that.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -770,13 +777,35 @@ private boolean doReceiveAndExecute(BlockingQueueConsumer consumer) throws Throw
...
+				if (getTransactionManager() != null) {
+					if (getTransactionAttribute().rollbackOn(ex)) {
+						consumer.rollbackOnExceptionIfNecessary(ex);
+						throw ex;
+					}
+					else {
+						if (this.logger.isDebugEnabled()) {
+							this.logger.debug("No rollback for " + ex);
+						}
+						break;
```
artembilan: Well, I see this `SimpleMessageListenerContainer` implementation is a bit different.
You have here fallback to the `commitIfNecessary()` for all `ImmediateAcknowledgeAmqpException`s and where no `rollbackOn()`.
Also you re-throw `e` in the `else` below. 

That is exactly what I expect in the `DirectMessageListenerContainer`.
See my concerns there.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
garyrussell: No; in this case we're running on the amqp-client thread - the previous one should not throw e.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
garyrussell: Here we're running on the async consumer thread so we need to throw. See my comment for the DMLC case.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
garyrussell: My last comment was incorrect.

garyrussell: In this else block, we're not running in a transaction template, so there is no reason to throw the exception since it is ignored anyway (because we're running on the amqp-client thread). Will add a comment.

garyrussell:
Please see commit comment on latest push for info.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -770,13 +777,35 @@ private boolean doReceiveAndExecute(BlockingQueueConsumer consumer) throws Throw
 				executeListener(channel, message);
 			}
 			catch (ImmediateAcknowledgeAmqpException e) {
+				if (this.logger.isDebugEnabled()) {
+					this.logger.debug("User requested ack for failed delivery");
```
artembilan: I guess we can't extract `deliveryTag` here to achieve the the same logs like in the `DirectMessageListenerContainer`.
Maybe we can just use the whole `BlockingQueueConsumer.toString()` here?
```
this.logger.debug("User requested ack for failed delivery in: " + consumer);
```

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -770,13 +777,35 @@ private boolean doReceiveAndExecute(BlockingQueueConsumer consumer) throws Throw
...
+				}
 				break;
 			}
 			catch (Throwable ex) { //NOSONAR
-				consumer.rollbackOnExceptionIfNecessary(ex);
-				throw ex;
+				if (causeChainHasImmediateAcknowledgeAmqpException(ex)) {
+					if (this.logger.isDebugEnabled()) {
+						this.logger.debug("User requested ack for failed delivery");
```
artembilan: The same here.
My minor concern that we are wrapping constant to extra `isDebugEnabled()` call.

artembilan:
Hm. I missed your latest commit somehow.
GH notification doesn't work properly all the time.

LGTM, merging...

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/7434ae6cd259b198fff9418efbf3199dced2ac24

Issue #522, opened 2016-12-13
Rabbit SSL connection error RabbitConnectionFactoryBean 


harrydevnull:
In case if we use  RabbitConnectionFactoryBean with 
_useSSL=true
&&sslPropertiesLocation == null && keyStore == null &&trustStore == null
&& keyStorePassphrase == null && this.trustStorePassphrase == null_

it makes a call to amqp-rabbit-connectionfactory 
```
if (this.sslAlgorithmSet) {
				this.connectionFactory.useSslProtocol(this.sslAlgorithm);
			}
			else {
				this.connectionFactory.useSslProtocol();
			}
```
but that code internally uses a NullTrustManager; meaning it doesn't do the server certificate validation. The spring documentation also mentions this.
I did raise an issue with rabbit-mq maintainers [Issue](https://github.com/rabbitmq/rabbitmq-java-client/issues/226) but it didn't go anywhere.

My suggestion is that we should avoid using 
`this.connectionFactory.useSslProtocol(this.sslAlgorithm);`
or
`this.connectionFactory.useSslProtocol();`

instead delegate to Java's default trustStore mechanism
 the rabbit-amqp library is susceptible for man-in-middle attacks; I believe  we should provide more safer defaults.

should I make a PR? I think this should work !!
```
public void useDefaultTrustStoreMechanism(){
  		SSLContext c = SSLContext.getInstance(this.sslAlgorithm);
        c.init(null, new TrustManager[] {  getDefaultTrustManager(this.sslAlgorithm, (KeyStore)null }, null);
        useSslProtocol(c);
	
}


public static X509TrustManager getDefaultTrustManager(String algorithm, KeyStore keystore) {

      TrustManagerFactory factory;
      X509TrustManager x509TrustManager = null;

      try {
          factory = TrustManagerFactory.getInstance(algorithm);
          factory.init(keystore);
          for (TrustManager trustManager : factory.getTrustManagers()) {

              if (trustManager instanceof X509TrustManager) {
                   x509TrustManager = (X509TrustManager)trustManager;
                  break;
              }
          }

      } catch (NoSuchAlgorithmException  e) {
          e.printStackTrace();
      }catch ( KeyStoreException e){
          e.printStackTrace();
      }

      return x509TrustManager;

}
```

garyrussell:
We can't make it the default in 1.6.x (or earlier) because it's a breaking change. We could add it as an option there, though.

[...]

PR #523, opened 2016-12-13
Rabbit SSL connection error RabbitConnectionFactoryBean

harrydevnull:
#522

In case if we use RabbitConnectionFactoryBean with
useSSL=true
&&sslPropertiesLocation == null && keyStore == null &&trustStore == null
&& keyStorePassphrase == null && this.trustStorePassphrase == null

it makes a call to amqp-rabbit-connectionfactory which does nothing which is a security flaw as such. It opens a possibility of man-in-middle attack.
this fix would would delegate to Java's trust mechanism if truststore or ssllocationpath is null.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
@@ -128,17 +128,20 @@ public class RabbitConnectionFactoryBean extends AbstractFactoryBean<ConnectionF
 
 	/**
 	 *
-	 * @return whether or not Server Side certificate has to be validated or not
+	 * @return true if Server Side certificate has to be skipped
+	 * @since 1.6.6
      */
 	public boolean isSkipServerCertificateValidation() {
 		return this.skipServerCertificateValidation;
 	}
 
 	/**
-	 * whether or not Server Side certificate has to be validated or not
-	 * this would be used if useSSL is set to true and should only be used on dev or Qa regions
-	 * skipServerCertificateValidation should never be set to true in production
-	 *
+	 * whether or not Server Side certificate has to be validated or not.
+	 * This would be used if useSSL is set to true and should only be used on dev or Qa regions
+	 * skipServerCertificateValidation should <b> never be set to true in production</b>
+	 * @param skipServerCertificateValidation Flag to override Server side certificate checks; if set to true com.rabbitmq.client.NullTrustManager would be used
+	 * @since 1.6.6
+	 * @see com.rabbitmq.client.NullTrustManager
 	 */
 	public void setSkipServerCertificateValidation(boolean skipServerCertificateValidation) {
 		this.skipServerCertificateValidation = skipServerCertificateValidation;
@@ -661,11 +664,11 @@ public class RabbitConnectionFactoryBean extends AbstractFactoryBean<ConnectionF
 	 * @throws KeyStoreException
      */
 	private void useDefaultTrustStoreMechanism() throws NoSuchAlgorithmException, KeyManagementException, KeyStoreException {
-		SSLContext c = SSLContext.getInstance(this.sslAlgorithm);
-		TrustManagerFactory factory = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
-		factory.init((KeyStore) null);
-		c.init(null, factory.getTrustManagers(), null);
-		this.connectionFactory.useSslProtocol(c);
+		SSLContext sslContext = SSLContext.getInstance(this.sslAlgorithm);
+		TrustManagerFactory trustManagerFactory = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
+		trustManagerFactory.init((KeyStore) null);
+		sslContext.init(null, trustManagerFactory.getTrustManagers(), null);
+		this.connectionFactory.useSslProtocol(sslContext);
 	}
 
 }
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
@@ -68,6 +68,7 @@ import com.rabbitmq.client.impl.nio.NioParams;
  * @author Gary Russell
  * @author Heath Abelson
  * @author Arnaud Cogoluègnes
+ * @author Hareendran
  *
  * @since 1.4
  */
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
@@ -20,6 +20,7 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.security.KeyManagementException;
 import java.security.KeyStore;
+import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
 import java.security.SecureRandom;
 import java.util.Arrays;
@@ -34,6 +35,7 @@ import javax.net.ssl.KeyManagerFactory;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.TrustManager;
 import javax.net.ssl.TrustManagerFactory;
+import javax.net.ssl.X509TrustManager;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -559,12 +561,7 @@ public class RabbitConnectionFactoryBean extends AbstractFactoryBean<ConnectionF
 	protected void setUpSSL() throws Exception {
 		if (this.sslPropertiesLocation == null && this.keyStore == null && this.trustStore == null
 				&& this.keyStoreResource == null && this.trustStoreResource == null) {
-			if (this.sslAlgorithmSet) {
-				this.connectionFactory.useSslProtocol(this.sslAlgorithm);
-			}
-			else {
-				this.connectionFactory.useSslProtocol();
-			}
+			useDefaultTrustStoreMechanism();
 		}
 		else {
 			if (this.sslPropertiesLocation != null) {
@@ -627,6 +624,50 @@ public class RabbitConnectionFactoryBean extends AbstractFactoryBean<ConnectionF
 	 */
 	protected SSLContext createSSLContext() throws NoSuchAlgorithmException {
 		return SSLContext.getInstance(this.sslAlgorithm);
+	}
+
+
+	/**
+	 * In case the client doesn't provide the sslProperties location or the truststore is null delegate to Java's default
+	 * trust mechanism
+	 * @throws NoSuchAlgorithmException
+	 * @throws KeyManagementException
+	 * @throws KeyStoreException
+     */
+	public void useDefaultTrustStoreMechanism() throws NoSuchAlgorithmException, KeyManagementException,
+			KeyStoreException {
+		SSLContext c = SSLContext.getInstance(this.sslAlgorithm);
+		c.init(null, new TrustManager[] {  getDefaultTrustManager(TrustManagerFactory.getDefaultAlgorithm(), null) }, null);
+		this.connectionFactory.useSslProtocol(c);
+	}
+
+
+	public static X509TrustManager getDefaultTrustManager(String algorithm, KeyStore keystore) throws NoSuchAlgorithmException, KeyStoreException {
+
+		TrustManagerFactory factory;
+		X509TrustManager x509TrustManager = null;
+
+		try {
+			factory = TrustManagerFactory.getInstance(algorithm);
+			factory.init(keystore);
+			for (TrustManager trustManager : factory.getTrustManagers()) {
+
+				if (trustManager instanceof X509TrustManager) {
+					x509TrustManager = (X509TrustManager) trustManager;
+					break;
+				}
+			}
+
+		}
+		catch (NoSuchAlgorithmException e) {
+			throw  e;
+		}
+		catch (KeyStoreException e) {
+			throw  e;
+		}
+
+		return x509TrustManager;
+
 	}
 
 }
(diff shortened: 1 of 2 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
```diff
@@ -629,4 +626,48 @@ protected SSLContext createSSLContext() throws NoSuchAlgorithmException {
...
+	 * In case the client doesn't provide the sslProperties location or the truststore is null delegate to Java's default
+	 * trust mechanism
+	 * @throws NoSuchAlgorithmException
+	 * @throws KeyManagementException
+	 * @throws KeyStoreException
+     */
+	public void useDefaultTrustStoreMechanism() throws NoSuchAlgorithmException, KeyManagementException,
+			KeyStoreException {
+		SSLContext c = SSLContext.getInstance(this.sslAlgorithm);
+		c.init(null, new TrustManager[] {  getDefaultTrustManager(TrustManagerFactory.getDefaultAlgorithm(), null) }, null);
```
artembilan: Why can't we use here something like this?
```
TrustManagerFactory tmf = TrustManagerFactory.getInstance("SunX509");
tmf.init(null);
c.init(null, tmf.getTrustManagers(), null);
```

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/SSLConnectionTests.java
```diff
@@ -74,7 +74,9 @@ public void testAlgNoProps() throws Exception {
 		fb.setSslAlgorithm("TLSv1.2");
 		fb.afterPropertiesSet();
 		fb.getObject();
-		verify(rabbitCf).useSslProtocol("TLSv1.2");
+		//since useSslProtocol is not called directly; commenting off the verification ; and not sure on what to verify
+		// as well may be perhaps checking whether  useSslProtocol (SSLContext) method is invoked at least 1 time
+		//verify(rabbitCf).useSslProtocol("TLSv1.2");
```
artembilan: I think we still can at least verify something like this:
```
verify(rabbitCf).useSslProtocol(any(SSLContext.class));
```
Or even `ArgumentCaptor<SSLContext>` to verify that `protocol` is propagates properly ans so on.
The same is below.

Anyway that isn't good to leave that comment without any particular solution.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
```diff
@@ -629,4 +626,48 @@ protected SSLContext createSSLContext() throws NoSuchAlgorithmException {
...
+					break;
+				}
+			}
+
+		}
+		catch (NoSuchAlgorithmException e) {
+			throw  e;
+		}
+		catch (KeyStoreException e) {
+			throw  e;
```
artembilan: These catches look redundant.
Although, if you agree with my comment above about `tmf.getTrustManagers()`, we won't need this `getDefaultTrustManager()` method at all.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
harrydevnull: if we were to work with IBM Jdk ( think websphere) it would be `IbmX509`

harrydevnull: and on my mac it is ssl.TrustManagerFactory.algorithm=PKIX

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
```diff
@@ -629,4 +626,36 @@ protected SSLContext createSSLContext() throws NoSuchAlgorithmException {
...
+     */
+	public void useDefaultTrustStoreMechanism() throws NoSuchAlgorithmException, KeyManagementException,
+			KeyStoreException {
+		SSLContext c = SSLContext.getInstance(this.sslAlgorithm);
+		c.init(null, new TrustManager[] {  getDefaultTrustManager(TrustManagerFactory.getDefaultAlgorithm(), null) }, null);
+		this.connectionFactory.useSslProtocol(c);
+	}
+
+
+	public static X509TrustManager getDefaultTrustManager(String algorithm, KeyStore keystore) throws NoSuchAlgorithmException, KeyStoreException {
```
artembilan: Even if it is fine to have it as a `static` method, I'm not sure that it is good to `public` it anyway.
The same `public` concern for the `useDefaultTrustStoreMechanism()` method above.

harrydevnull: update the visibility modifiers 37f55a3

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
```diff
@@ -629,4 +627,36 @@ protected SSLContext createSSLContext() throws NoSuchAlgorithmException {
...
+		c.init(null, new TrustManager[] {  getDefaultTrustManager(TrustManagerFactory.getDefaultAlgorithm(), null) }, null);
+		this.connectionFactory.useSslProtocol(c);
+	}
+
+
+	private static X509TrustManager getDefaultTrustManager(String algorithm, KeyStore keystore) throws NoSuchAlgorithmException, KeyStoreException {
+
+		TrustManagerFactory factory = TrustManagerFactory.getInstance(algorithm);
+		X509TrustManager x509TrustManager = null;
+        factory.init(keystore);
```
artembilan: Some formatting issue.
And I still don't understand why we can't just use `getTrustManagers()` without any filtering?
Thanks

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/SSLConnectionTests.java
```diff
@@ -85,9 +93,34 @@ public void testNoAlgNoProps() throws Exception {
 		fb.setUseSSL(true);
 		fb.afterPropertiesSet();
 		fb.getObject();
-		verify(rabbitCf).useSslProtocol();
+		verify(rabbitCf).useSslProtocol(Mockito.any(SSLContext.class));
+	}
+
+	@Test
+	public void testuseSslProtocolShouldNotbecalled() throws Exception {
```
artembilan: Minor, but would be better to follow with camel case formatting for naming.
I mean this is preferable: `testUseSslProtocolShouldNotBeCalled`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
harrydevnull: I don't know the answer persumably it might be a glue code for different  JDK implementations.

while searching; I did come across one article which showed on how to create sslcontext for IBM  jdks.
the doc says  `IBM TrustManagerFactory is called IbmX509 or IbmPKIX.` 
I presume spring classes has to work on any JDK.

http://www.ibm.com/support/knowledgecenter/SSYKE2_6.0.0/com.ibm.java.security.component.60.doc/security-component/jsse2Docs/knowndiffsun.html
http://www.ibm.com/support/knowledgecenter/SSYKE2_6.0.0/com.ibm.java.security.component.60.doc/security-component/jsse2Docs/pkixtrustmgrsupport.html

artembilan: Thank you for some info, @harrydevnull .

I don't mind that `TrustManagerFactory.getDefaultAlgorithm()`, I just worry why we don't use it like this:
```
ctx.init(null, factory.getTrustManagers(), null);
```
Why do you need to extract exactly one `x509TrustManager` and don't use all of them for the `SSLContext`?

artembilan:
Some general concern, BTW: https://github.com/spring-projects/spring-amqp/issues/522#issuecomment-266783886

I really don't see any options to revert to the old behavior.
Any ideas?

And I don't think that we should backport it, because people may always choose to use `com.rabbitmq.client.ConnectionFactory` directly, where you can specify any desired behavior with the `useSslProtocol(SSLContext context)`.

harrydevnull:
my thought process says other way round. we should definitely back-port
these changes.
we should never have used NullTrustManager in the source Tree ( but that is
in rabbit-mq  source)
if are not doing it the correct way we should fail fast.

attaching a article which google flagged as a ssl vulnerability.

http://www.appcelerator.com/blog/2016/02/google-security-alert-unsafe-implementation-of-the-interface-x509trustmanager/

In the mean while I'll try to convince to rabbit-mq maintainers to get rid
of NullTrustManger from source Tree. the patch on spring layer actually
shields away the user from the underlying vulnerability.
(which according to me as developer is really awesome)

artembilan:
Indeed we can back-port it since it isn't so big change, but more over useful.

Well, even if it looks like a vulnerability I definitely sure that there might be some use-cases when `NullTrustManger` is useful:
```
 * Convenience class providing a default implementation of javax.net.ssl.X509TrustManager.
 * Trusts every single certificate presented to it.
```
I remember as I looked for such one several year ago to achieve distributed testing on several JVMs without a requirement to transfer trust store everywhere or update certificates expiration dates and so on.

Therefore our concern to you still provide an option to revert to the `NullTrustManger` on the current `2.0` and make an option to jump to your solution in the `1.6` and `1.7`, meanwhile still leave `NullTrustManger` as default one.

@garyrussell , need your vote, please 😄

harrydevnull:
Agreed. As a matter I do have the copy of NullTrustManager in source repo and it being used only in dev regions as we use self signed certificates.

the way I see is to add a new skipServerCertificateValidation flag default to false
and in setUpSSL method 
`
if(!skipServerCertificateValidation){
 useDefaultTrustStoreMechanism()
}else{
this.connectionFactory.useSslProtocol()
}
`
what do you think?

artembilan:
`skipServerCertificateValidation` - sounds perfect!
👍

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
```diff
@@ -121,6 +123,27 @@
 
 	private volatile SecureRandom secureRandom;
 
+	private boolean skipServerCertificateValidation;
```
artembilan: This must be `true` by default in back-port commits.
Just for the record

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
```diff
@@ -121,6 +123,27 @@
...
+
+
+	/**
+	 *
+	 * @return whether or not Server Side certificate has to be validated or not
```
artembilan: Would be better to have some general Java Doc description for `public` method. 
The `@return` not enough.
Plus `@since 1.6.6`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
```diff
@@ -121,6 +123,27 @@
...
+     */
+	public boolean isSkipServerCertificateValidation() {
+		return this.skipServerCertificateValidation;
+	}
+
+	/**
+	 * whether or not Server Side certificate has to be validated or not
+	 * this would be used if useSSL is set to true and should only be used on dev or Qa regions
+	 * skipServerCertificateValidation should never be set to true in production
+	 *
```
artembilan: Need `@param`:
```
RabbitConnectionFactoryBean.java:143: warning: no @param for skipServerCertificateValidation

	public void setSkipServerCertificateValidation(boolean skipServerCertificateValidation) {
```
Also `@since 1.6.6`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
```diff
@@ -559,11 +582,11 @@ protected ConnectionFactory createInstance() throws Exception {
 	protected void setUpSSL() throws Exception {
 		if (this.sslPropertiesLocation == null && this.keyStore == null && this.trustStore == null
 				&& this.keyStoreResource == null && this.trustStoreResource == null) {
-			if (this.sslAlgorithmSet) {
-				this.connectionFactory.useSslProtocol(this.sslAlgorithm);
+			if (this.skipServerCertificateValidation) {
+				this.connectionFactory.useSslProtocol(this.getSslAlgorithm());
```
artembilan: No `this.` for methods, although we can use property directly here any way.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
```diff
@@ -629,4 +652,20 @@ protected SSLContext createSSLContext() throws NoSuchAlgorithmException {
...
+	/**
+	 * In case the client doesn't provide the sslProperties location or the truststore is null delegate to Java's default
+	 * trust mechanism
+	 * @throws NoSuchAlgorithmException
+	 * @throws KeyManagementException
+	 * @throws KeyStoreException
+     */
+	private void useDefaultTrustStoreMechanism() throws NoSuchAlgorithmException, KeyManagementException, KeyStoreException {
+		SSLContext c = SSLContext.getInstance(this.sslAlgorithm);
+		TrustManagerFactory factory = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
```
artembilan: Spring is an Open Source. Would be better to use expressive names even for variable in `private` methods.
Just because all this code is visible for everyone 😄 
I mean both these `c` and `factory`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
harrydevnull: checkstyle did complain if i don't use this.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/f8e7732ce69e5f3e591700bebbf00682ce7ab231
and cherry-picked as https://github.com/spring-projects/spring-amqp/commit/2262590ed55da5db44290cd45796204d8ed39d24

@harrydevnull , thank you for contribution.
Looking forward for more!

harrydevnull:
Thank you for merging my PR.
Really happy to have contributed back to spring framework..

PR #526, opened 2016-12-14
AMQP-687: Fix Conditional Rollback [1.6.x]

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-687

When using a rule-based transaction attribute with a rollback/no-rollback rule,
the rule is ignored when providing the transaction attribute directly to the container
rather than via a transaction advice,

Also, user exceptions are always wrapped in a `ListenerExecutionFailedException` which
limits the ability to configure rules, even when using an advice.

- Honor the attribute `rollBackOn()` mechanism within the container
- Add `ListenerFailedRuleBasedTransactionAttribute` to allow the user to configure rules based on the actual exception.

Add docs.

AMQP-686: Handle ImmediateAcknowledgeAmqpException

JIRA: https://jira.spring.io/browse/AMQP-686

Previously, `ImmediateAcknowledgeAmqpException` only worked if thrown by
an error handler, not a listener.

Examine the cause chain to see if this exception appears, and handle accordingly.

An `AmqpRejectAndDontRequeueException` found first stops the chain traversal.

Also clarify the javadocs to make it clear the current transaction is ended
immediately, not all message delivery.

checkstyle

Polishing and Fixes

While polishing for PR comments, I found some more problems (existing code).

When using an external transaction manager, txRollback() was called twice, once
within the transaction and once by the after completion in the transaction synchronization.

This was benign because the delivery tags were not registered, but incorrect.

Now, the delivery tag is always registered and the rollback done by the synchronization.

Added more tests and added tests for the local transaction scenario.

Add missing .this

Polishing

Conflicts:
	spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
	spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
	spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
	spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/ExternalTxManagerSMLCTests.java
	spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/ExternalTxManagerTests.java
	spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/LocallyTransactedTests.java
	src/reference/asciidoc/whats-new.adoc

Resolved.

Conflicts:
	spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
	spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java

Resolved.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -98,7 +98,7 @@ subprojects { subproject ->
 		log4jVersion = '1.2.17'
 		log4j2Version = '2.5'
 		logbackVersion = '1.1.2'
-		mockitoVersion = '1.9.5'
+		mockitoVersion = '1.10.19'
 		rabbitmqVersion = project.hasProperty('rabbitmqVersion') ? project.rabbitmqVersion : '3.6.5'
 		rabbitmqHttpClientVersion = '1.0.0.RELEASE'
 
--- a/spring-amqp/src/main/java/org/springframework/amqp/ImmediateAcknowledgeAmqpException.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/ImmediateAcknowledgeAmqpException.java
@@ -17,10 +17,12 @@
 package org.springframework.amqp;
 
 /**
- * Special exception for listener implementations that want to signal that the current batch of messages should be
- * acknowledged immediately (i.e. as soon as possible) without rollback, and without consuming any more messages.
+ * Special exception for listener implementations that want to signal that the current
+ * batch of messages should be acknowledged immediately (i.e. as soon as possible) without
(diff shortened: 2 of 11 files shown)


artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/21dd858d41b64f1f9b2c626af11e45290506183d after rebase.

PR #529, opened 2016-12-18
Fixing a race condition in SimpleMessageListenerContainer

zaharidichev:
JIRA: [https://jira.spring.io/browse/AMQP-690](https://jira.spring.io/browse/AMQP-690)

Fixing race condition causing SimpleMessageListenerContainer to drop consumers to zero. The introduced method should take into consideration only the count of the consumers that have not started their shutdown procedure.

This quite the simple fix to a pretty elusive problem. I have been thinking of a way to actually write a test for this but to be frank, I have not come up with a good idea. 

Additionally I must say that to me the set of states that a consumer can be in (active,not active) is not complete. I mean for example: 

```
@ManagedMetric(metricType = MetricType.GAUGE)
public int getActiveConsumerCount() {
        return this.cancellationLock.getCount();
}
```

This piece of code if called gets the count of the consumers that have still not unregistered themselves from the ActiveObjectCounter. However does that mean that they are "active". I mean  surely they can still be processing messages. However if they have been subject to the considerStoppingAConsumer method, they are not going to be active for a lot longer. That being said, somebody can mistakenly use this method to do things such as what caused this original bug.

If I have to define the states of a consumer they would be: 

- active
- shutting down
- shut down

Its just my two cents.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
@@ -102,8 +102,7 @@ public class SimpleMessageListenerContainer extends AbstractMessageListenerConta
 
 	private long receiveTimeout = DEFAULT_RECEIVE_TIMEOUT;
 
-	// Map entry value, when false, signals the consumer to terminate
-	private Set<BlockingQueueConsumer> activeConsumers;
+	private Set<BlockingQueueConsumer> consumers;
 
 	private final ActiveObjectCounter<BlockingQueueConsumer> cancellationLock = new ActiveObjectCounter<>();
 
@@ -145,13 +144,13 @@ public class SimpleMessageListenerContainer extends AbstractMessageListenerConta
 		}
 		synchronized (this.consumersMonitor) {
 			if (logger.isDebugEnabled()) {
-				logger.debug("Changing activeConsumers from " + this.concurrentConsumers + " to " + concurrentConsumers);
+				logger.debug("Changing consumers from " + this.concurrentConsumers + " to " + concurrentConsumers);
 			}
 			int delta = this.concurrentConsumers - concurrentConsumers;
 			this.concurrentConsumers = concurrentConsumers;
-			if (isActive() && this.activeConsumers != null) {
+			if (isActive() && this.consumers != null) {
 				if (delta > 0) {
-					Iterator<BlockingQueueConsumer> consumerIterator = this.activeConsumers.iterator();
+					Iterator<BlockingQueueConsumer> consumerIterator = this.consumers.iterator();
 					while (consumerIterator.hasNext() && delta > 0) {
 						BlockingQueueConsumer consumer = consumerIterator.next();
 						consumer.basicCancel();
@@ -461,7 +460,7 @@ public class SimpleMessageListenerContainer extends AbstractMessageListenerConta
 		super.doStart();
 		synchronized (this.consumersMonitor) {
 			int newConsumers = initializeConsumers();
-			if (this.activeConsumers == null) {
+			if (this.consumers == null) {
 				if (logger.isInfoEnabled()) {
 					logger.info("Consumers were initialized and then cleared " +
 							"(presumably the container was stopped concurrently)");
@@ -475,7 +474,7 @@ public class SimpleMessageListenerContainer extends AbstractMessageListenerConta
 				return;
 			}
 			Set<AsyncMessageProcessingConsumer> processors = new HashSet<AsyncMessageProcessingConsumer>();
-			for (BlockingQueueConsumer consumer : this.activeConsumers) {
+			for (BlockingQueueConsumer consumer : this.consumers) {
 				AsyncMessageProcessingConsumer processor = new AsyncMessageProcessingConsumer(consumer);
 				processors.add(processor);
 				getTaskExecutor().execute(processor);
@@ -501,8 +500,8 @@ public class SimpleMessageListenerContainer extends AbstractMessageListenerConta
 
 		try {
 			synchronized (this.consumersMonitor) {
-				if (this.activeConsumers != null) {
-					Iterator<BlockingQueueConsumer> consumerIterator = this.activeConsumers.iterator();
+				if (this.consumers != null) {
+					Iterator<BlockingQueueConsumer> consumerIterator = this.consumers.iterator();
 					while (consumerIterator.hasNext()) {
 						BlockingQueueConsumer consumer = consumerIterator.next();
 						consumer.basicCancel();
@@ -525,7 +524,7 @@ public class SimpleMessageListenerContainer extends AbstractMessageListenerConta
 		}
 
 		synchronized (this.consumersMonitor) {
-			this.activeConsumers = null;
+			this.consumers = null;
 		}
 
 	}
@@ -533,7 +532,7 @@ public class SimpleMessageListenerContainer extends AbstractMessageListenerConta
 	private boolean isActive(BlockingQueueConsumer consumer) {
 		boolean consumerActive;
 		synchronized (this.consumersMonitor) {
-			consumerActive = this.activeConsumers != null && this.activeConsumers.contains(consumer);
+			consumerActive = this.consumers != null && this.consumers.contains(consumer);
 		}
 		return consumerActive && this.isActive();
 	}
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@zaharidichev Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-amqp&pullRequestId=529)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-amqp&pullRequestId=529) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/faq) for frequently asked questions.

pivotal-issuemaster:
@zaharidichev Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-amqp&pullRequestId=529)!

zaharidichev:
This quite the simple fix to a pretty elusive problem. I have been thinking of a way to actually write a test for this but to be frank, I have not come up with a good idea. 

Additionally I must say that to me the set of states that a consumer can be in (active,not active) is not complete. I mean for example: 

```
@ManagedMetric(metricType = MetricType.GAUGE)
public int getActiveConsumerCount() {
        return this.cancellationLock.getCount();
}
```

This piece of code if called gets the count of the consumers that have still not unregistered themselves from the ActiveObjectCounter. However does that mean that they are "active". I mean  surely they can still be processing messages. However if they have been subject to the considerStoppingAConsumer method, they are not going to be active for a lot longer. That being said, somebody can mistakenly use this method to do things such as what caused this original bug.

If I have to define the states of a consumer they would be: 

- active
- shutting down
- shut down

Its just my two cents.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -633,9 +633,15 @@ private void considerAddingAConsumer() {
 		}
 	}
 
+	private int getConsumersThatHaveNotInitiatedStopping() {
+		synchronized (this.consumersMonitor) {
+			return (int) this.consumers.values().stream().filter(x -> x = Boolean.TRUE).count();
```
artembilan: Hello, @zaharidichev !

Thank you for contribution first of all!

Well, let me review your PR and share my vision.

Your `filter()` looks wrong. It always assigns `Boolean.TRUE` to the `x` and returns that `true` 😄 .
OTOH the issue is a bug and we must back port it to versions which aren't support Java 8 yet.

One more thought that it looks like we don't need this method at all:
- leave `this.consumers.size() > this.concurrentConsumers` condition as is;
- instead of `this.consumers.put(consumer, false)` use `this.consumers.remove(consumer)`;
- the `isActive(BlockingQueueConsumer)` logic is based on the fact that a `value` for the `consumer` can be `null`, so we are safe to just remove `consumer` from the cache and don't care about its state;
- the `AsyncMessageProcessingConsumer.run()` should be free from of the `synchronized (SimpleMessageListenerContainer.this.consumersMonitor) {` block as well, since we don't need it here any more.

That way we will be able to avoid such kind of hell iteration in the `Map` on each `considerStoppingAConsumer()` call.
And at the same time we still reach atomicity for the stopping consumers.

Looks like we have to change `setConcurrentConsumers()`  to use `this.consumers.remove(consumer)` as well, because `SimpleMessageListenerContainerLongTests` fails for me locally with my logic. 

>  I mean surely they can still be processing messages.

Yes, indeed. That is a main point why even being shutting down we still treat it as an active because any "in-flight" message might be in process yet. Even direct `stop()` doesn't prevent that behavior. But at least it is some middleground between nothing and real state. Anyway externally you can't be sure how an internal state may be changed in between your actions. That's why we should ensure the proper state management internally to prevent any unexpected involvements. 

So, let's fix the current problem and leave `activeCount` as is or feel free to start a new JIRA for some argument to reconsider it.

At the same time we still can ask @garyrussell for his thoughts on the matter.

Thanks

zaharidichev: Hi @artembilan !

First of all, sorry for the lousy filter. I have put in the requested changes and actually I agree with them. I think it looks decent now. After these changes however, the question arises - Why do we even need a key value pair here?  

I mean just a Set<BlockingQueueConsumer> activeConsumers would be good enough to keep track of the number of active consumers as well as whether a particular consumer is still active (if its a member of the set). That way we also simplify the logic of `isActive(BlockingQueueConsumer)` as it will be pretty much an `activeConsumers.contains()` call.

artembilan: Please, move your comment to the top level. Now it isn't visible because of your changes.

zaharidichev:
Hi @artembilan !

First of all, sorry for the lousy filter. I have put in the requested changes and actually I agree with them. I think it looks decent now. After these changes however, the question arises - Why do we even need a key value pair here?

I mean just a` Set<BlockingQueueConsumer> activeConsumers` would be good enough to keep track of the number of active consumers as well as whether a particular consumer is still active (if its a member of the set). That way we also simplify the logic of `isActive(BlockingQueueConsumer)` as it will be pretty much an `activeConsumers.contains()` call.

artembilan:
@garyrussell , 

Need your input on this PR.

I think I wouldn't hesitate if the discussion was only about `master`, but since we are going to back port the fix it would be better to determine the real reason of that `Map<BlockingQueueConsumer, Boolean>` and why we can't really switch to the `Set<BlockingQueueConsumer>` like we have already in the `DirectMessageListenerContainer`.

Thanks

garyrussell:
I haven't really been following this PR to-date. 

It looks to me that `AMQP-621` - adding the additional test for `consumer.cancelled()` - removed the need for `consumers` to be a map any more - the `isActive()` call is now redundant since whenever we reduce consumers (automatically or manually), we always cancel the consumer.

`consumer.cancelled()` will eventually return true, even if we never receive the `cancelOk` (because of a network glitch or similar).

So, yes, I believe it can be changed to a simple set now.

artembilan:
Great!

Thank you for confirmation, @garyrussell !

@zaharidichev , you can proceed as you think about that.
We'll correct you afterward if that.

artembilan:
@zaharidichev ,

We appreciate your contribution and valuable research around that issue.
Only concern that we are going to release tomorrow, so you have time only until this evening (our EST).
Otherwise we accepts your PR as is and will do all the appropriate changes on own own on merge.

Thanks

zaharidichev:
@artembilan Sure, I am just thinking it over once again.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -106,7 +103,7 @@
 	private long receiveTimeout = DEFAULT_RECEIVE_TIMEOUT;
 
 	// Map entry value, when false, signals the consumer to terminate
-	private Map<BlockingQueueConsumer, Boolean> consumers;
+	private Set<BlockingQueueConsumer> activeConsumers;
```
artembilan: Great!
Although the comment on the property is wrong 😀. Plus I don't see reason to rename it. Just consumers is enough.

garyrussell:
The field name in the test needs to be reverted too; Artem can do that on merge 😄

zaharidichev:
@garyrussell, If  we are talking about SimpleMessageListenerContainerTests, I already did that.

artembilan:
@zaharidichev ,

Does this say anything to you?
```
org.springframework.amqp.rabbit.listener.MessageListenerContainerErrorHandlerIntegrationTests > testErrorHandlerThrowsARADRE FAILED

    org.springframework.beans.NotReadablePropertyException at MessageListenerContainerErrorHandlerIntegrationTests.java:134
```

zaharidichev:
@artembilan 
Actually it does indeed say something. What did not say anything however was gradle testAll which I ran right before committing. Are these integration tests not included in the task ?

artembilan:
Well, I think your problem is here:
```
@Rule
public BrokerRunning brokerIsRunning = BrokerRunning.isRunningWithEmptyQueues(queue.getName());
```
with meaning that you don't have working RabbitMQ.
It is indeed `integration test`, but as you see it works well only when RabbitMQ Broker is on the `5672` port.

zaharidichev:
That clears things up quite a lot. Thanks for pointing that out. And thanks for all the help :)

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/4ad793811b3140d6dde4002603c16e0a300398b0
and cherry-picked as  https://github.com/spring-projects/spring-amqp/commit/f0f05db9451aeef83e6d8c20c452155782521d7c after fixing conflicts.

@zaharidichev , 
thank you for the contribution.
Looking forward for more!

garyrussell:
Looks like we didn't do a `testAll` build: https://build.spring.io/artifact/AMQP-MEIGHT/JOB1/build-926/Test-Results/spring-rabbit/build/reports/spring-io-jdk8-tests/index.html

garyrussell:
(I will fix)

PR #532, opened 2016-12-27
AMQP-675: Fix RabbitConverterFuture race condition

artembilan:
JIRA: https://jira.spring.io/browse/AMQP-675

https://build.spring.io/browse/AMQP-MEIGHT-932

The AMQP reply to the `Future<?>` may come back very quickly and we may not have yet chance to sent `returnType`.
Therefore no any conversion for the body will happens and we receive just raw `Map`.

* Move `returnType` population for the `RabbitConverterFuture` to the `CorrelationMessagePostProcessor` directly, before any send execution
* Remove `boolean start` condition since it doesn't make sense and `startTimer()` relates to the "late reply task", not to the send operation


garyrussell:
Merged as 27cf0a39aab1527809b097b435efb597540d2f7e with checkstyle fix and `final` returnType field in MPP.

PR #533, opened 2016-12-29
Various Test Performance Fixes

garyrussell:
Including:

- add `Lifecycle` to `RabbitTemplate` and `stop()` it
in non-spring-managed tests where send and receive operations
have been done - prevent the replyTo container from
reopening the connection
- use a shared admin for several tests
- close connections - a few remain which we can track down later

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/LocalizedQueueConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/LocalizedQueueConnectionFactory.java
@@ -362,7 +362,9 @@ public class LocalizedQueueConnectionFactory implements ConnectionFactory, Routi
 	@Override
 	public void destroy() throws Exception {
 		for (ConnectionFactory connectionFactory : this.nodeFactories.values()) {
-			((DisposableBean) connectionFactory).destroy();
+			if (connectionFactory instanceof DisposableBean) {
+				((DisposableBean) connectionFactory).destroy();
+			}
 		}
 		if (this.defaultConnectionFactory instanceof DisposableBean) {
 			((DisposableBean) this.defaultConnectionFactory).destroy();
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/LocalizedQueueConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/LocalizedQueueConnectionFactory.java
@@ -364,6 +364,9 @@ public class LocalizedQueueConnectionFactory implements ConnectionFactory, Routi
 		for (ConnectionFactory connectionFactory : this.nodeFactories.values()) {
 			((DisposableBean) connectionFactory).destroy();
 		}
+		if (this.defaultConnectionFactory instanceof DisposableBean) {
+			((DisposableBean) this.defaultConnectionFactory).destroy();
+		}
 	}
 
 }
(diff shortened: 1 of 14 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/LocalizedQueueConnectionFactory.java
```diff
@@ -364,6 +364,9 @@ public void destroy() throws Exception {
 		for (ConnectionFactory connectionFactory : this.nodeFactories.values()) {
 			((DisposableBean) connectionFactory).destroy();
```
artembilan: I think for safety and clear code we have to use ` instanceof DisposableBean` here as well

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -661,6 +664,41 @@ public int getUnconfirmedCount() {
 		}
 	}
 
+	@Override
+	public void start() {
+	}
+
+	@Override
+	public void stop() {
```
artembilan: `BatchingRabbitTemplate extends RabbitTemplate` but doesn't call `super.stop()` in its `stop()` impl.
Omission? 
Although for safety I'm sure we have to call `super` any way. Or comment somehow why we don't do that...

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -661,6 +664,41 @@ public int getUnconfirmedCount() {
...
+		synchronized (this.directReplyToContainers) {
+			if (this.directReplyToContainers.size() > 0) {
+				Iterator<Entry<ConnectionFactory, DirectReplyToMessageListenerContainer>> iterator =
+						this.directReplyToContainers.entrySet().iterator();
+				while (iterator.hasNext()) {
+					Entry<ConnectionFactory, DirectReplyToMessageListenerContainer> entry = iterator.next();
+					if (entry.getValue().isRunning()) {
+						entry.getValue().stop();
+					}
+					iterator.remove();
```
artembilan: Why `this.directReplyToContainers.clear()` isn't good in the end instead ?
OTOH:
```
	@Override
	public void stop() {
		synchronized (this.directReplyToContainers) {
				this.directReplyToContainers.values()
						.stream()
						.filter(AbstractMessageListenerContainer::isRunning)
						.forEach(AbstractMessageListenerContainer::stop);

				this.directReplyToContainers.clear();
		}
	}
```

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -661,6 +664,41 @@ public int getUnconfirmedCount() {
...
+				}
+			}
+		}
+	}
+
+	@Override
+	public boolean isRunning() {
+		synchronized (this.directReplyToContainers) {
+			if (this.directReplyToContainers.size() > 0) {
+				for (DirectReplyToMessageListenerContainer container : this.directReplyToContainers.values()) {
```
artembilan: Something similar with Streams here as well:
```
public boolean isRunning() {
		synchronized (this.directReplyToContainers) {
			return this.directReplyToContainers.values()
					.stream()
					.anyMatch(AbstractMessageListenerContainer::isRunning);
		}
	}
```

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/b5f530d80cec0b9851a2a82a9e96d8e17c594e20 after rebase

PR #540, opened 2017-01-10
AMQP-693: AsyncRabbitTemplate TaskScheduler Fixes

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-693

- Destroy the internal scheduler (if used) in stop() to allow the JVM to terminate
- Fix synchronization around `this.taskScheduler`
- Assert that the template is started to avoid NPEs

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/AsyncRabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/AsyncRabbitTemplate.java
@@ -331,6 +331,7 @@ public class AsyncRabbitTemplate implements AsyncAmqpTemplate, ChannelAwareMessa
 	 * @see #setReceiveTimeout(long)
 	 */
 	public synchronized void setTaskScheduler(TaskScheduler taskScheduler) {
+		Assert.notNull(taskScheduler, "'taskScheduler' cannot be null");
 		this.internalTaskScheduler = false;
 		this.taskScheduler = taskScheduler;
 	}
@@ -722,7 +723,7 @@ public class AsyncRabbitTemplate implements AsyncAmqpTemplate, ChannelAwareMessa
 
 		void startTimer() {
 			if (AsyncRabbitTemplate.this.receiveTimeout > 0) {
-				synchronized (AsyncRabbitTemplate.this.taskScheduler) {
+				synchronized (AsyncRabbitTemplate.this) {
 					this.timeoutTask = AsyncRabbitTemplate.this.taskScheduler.schedule(new TimeoutTask(),
 							new Date(System.currentTimeMillis() + AsyncRabbitTemplate.this.receiveTimeout));
 				}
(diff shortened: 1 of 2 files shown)


garyrussell:
Repushed to fix typo in commit comment.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/AsyncRabbitTemplate.java
```diff
@@ -713,8 +722,10 @@ void setNackCause(String nackCause) {
 
 		void startTimer() {
 			if (AsyncRabbitTemplate.this.receiveTimeout > 0) {
-				this.timeoutTask = AsyncRabbitTemplate.this.taskScheduler.schedule(new TimeoutTask(), // NOSONAR sync
-						new Date(System.currentTimeMillis() + AsyncRabbitTemplate.this.receiveTimeout));
+				synchronized (AsyncRabbitTemplate.this.taskScheduler) {
```
artembilan: image

Not sure if that is so big deal, maybe something like `final Object taskSchedulerMonitor` would be better any way ?
At least, let's add some comment here to avoid similar discussions in the future

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/AsyncRabbitTemplate.java
```diff
@@ -328,8 +330,9 @@ public void setReceiveTimeout(long receiveTimeout) {
 	 * @param taskScheduler the task scheduler
 	 * @see #setReceiveTimeout(long)
 	 */
-	public void setTaskScheduler(TaskScheduler taskScheduler) {
-		this.taskScheduler = taskScheduler; // NOSONAR synchronization
+	public synchronized void setTaskScheduler(TaskScheduler taskScheduler) {
+		this.internalTaskScheduler = false;
+		this.taskScheduler = taskScheduler;
```
artembilan: `Assert.notNull()` is missed.
Otherwise `this.internalTaskScheduler == false` and `this.taskScheduler == null` - and, therefore, NPE in the target code.
Test-case to be sure what we do ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/AsyncRabbitTemplate.java
garyrussell: That was a mistake - should be `synchronized (ART.this)` - to match the other synchronizations. Will push in the morning (or you can fix during merge if this is the only problem).

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/ea42002a04c83c161f068fb9e38aee2cc2475063 and cherry-picked as https://github.com/spring-projects/spring-amqp/commit/62cb3eba1ad0b10f4bf5bbafbefbf836ae74754f

PR #541, opened 2017-01-11
AMQP-693: Polishing

garyrussell:
Move assert - avoid the race which would cause an NPE if the template is stopped while a send is in process.
Remove the pending future in this case.


artembilan:
... and cherry-picked as https://github.com/spring-projects/spring-amqp/commit/0b370fa71534ad5542f555737bea7fc9cb61050c

PR #544, opened 2017-01-13
AMQP-697: CorrelationAwareMessagePostProcessor

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-697

Add support for modifying publisher confirms `CorrelationData` during message post processing.

Change `AsyncRabbitTemplate` to use a stateless MPP, move all state to the `CorrelationData`.

__master only - I will issue a new 1.x backport PR after review/merge__

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/core/MessagePostProcessor.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/core/MessagePostProcessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


spring-amqp/src/main/java/org/springframework/amqp/core/MessagePostProcessor.java
```diff
@@ -31,6 +32,23 @@
...
+	 * @throws AmqpException an exception.
+	 */
 	Message postProcessMessage(Message message) throws AmqpException;
 
+	/**
+	 * Change (or replace) the message and/or change its correlation data.
+	 * @param message the message.
+	 * @param correlation the correlation data.
+	 * @return the message.
+	 * @since 1.6.7
```
artembilan: `@since 2.0.0` already.
Because it can't be `default` for Java < 8 😄

garyrussell: Right; that's why my comment says I'll create a separate PR for the backport - see `CorrelationAwareMessagePostProcessor`.

artembilan: OK, Fixing this `@since` and merging

garyrussell: No - it's just a "pull-up" so the since is correct

artembilan: Doh...
Got it now.
Thanks

PR #547, opened 2017-01-17
AMQP-699: Suppress Stack Trace in RabbitAdmin

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-699

When `ignoreDeclarationExceptions` is true, only log exception stack traces under DEBUG logging.
Just log the exception and its message under WARN.

For `IOException`, extract the cause to get a useful message in the WARN log.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitAdmin.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitAdmin.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitAdmin.java
```diff
@@ -589,11 +589,20 @@ private void declareBindings(final Channel channel, final Binding... bindings) t
...
 						+ (element == null ? "broker-generated" : ": " + element)
 						+ ", continuing...", t);
 			}
+			else if (this.logger.isWarnEnabled()) {
+				Throwable cause = t;
+				if (t instanceof IOException && t.getCause() != null) {
+					cause = t.getCause();
+				}
+				this.logger.warn("Failed to declare " + elementType
+						+ (element == null ? "broker-generated" : ": " + element)
```
artembilan: M-m-m.
Shouldn't we place `: ` before `broker-generated`? Otherwise it looks like:
```
"Failed to declare queuebroker-generated, continuing... 
```

artembilan: Just let me know and I'll fix on merge!

garyrussell: Good catch.

artembilan:
The polishing for your confirmation: https://github.com/artembilan/spring-amqp/commit/b546da8ebe92f00547c96be69ebe1cfafa720572

garyrussell:
👍 thanks

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/1027fa3a05d8aa4ec19458f3dadcc3c91dcabb33 and cherry-picked.

PR #550, opened 2017-01-20
AMQP-701: @RabbitListener BPP Performance

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-701

`RabbitListenerAnnotationBeanPostProcessor` introspects the class each
time a bean of that type is instantiated.

When complex beans with many methods are scoped as prototype, the process
can be inordinately expensive.

Cache the results of the introspection.

Since we have no need for the cache of non-prototype beans after context
initialization, clear the cache.

The cache will be initialized for the prototype bean on first access.

__cherry-pick to 1.7.x and master__

It might need some rework on master; let me know if its non-trivial and I'll issue a separate PR.


artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/ac0dc5cae5eb69f3ebe308fdd7dfa1dcc1ebda5f and chery-picked to `1.7.x` as is.

Cherry-picked to `master` as https://github.com/spring-projects/spring-amqp/commit/27241383738c0ff1e6c459cdc1de729d2fd4d6ca after resolving conflicts.

PR #552, opened 2017-01-24
AMQP-703: Add Connection Name to Cache Properties

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-703

__cherry-pick to 1.7.x__


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/SimpleConnection.java
```diff
@@ -124,6 +124,10 @@ public int getPort() {
 		return this.delegate.getPort();
 	}
 
+	com.rabbitmq.client.Connection getDelegate() {
```
artembilan: Is it really intentional do not be `public` ?

garyrussell: Yes, it was intended; I don't think we need to allow end-users to get to the delegate.

artembilan:
... and cherry-picked.

PR #555, opened 2017-01-31
AMQP-706: Improve DefaultClassMapper Documentation

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-706

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/support/converter/DefaultClassMapper.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/support/converter/DefaultClassMapper.java
@@ -17,7 +17,7 @@
 package org.springframework.amqp.support.converter;
 
 import java.util.HashMap;
-import java.util.Hashtable;
+import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Map.Entry;
 
@@ -26,11 +26,15 @@ import org.springframework.beans.factory.InitializingBean;
 import org.springframework.util.ClassUtils;
 
 /**
- * Maps to/from JSON using type information in the {@link MessageProperties};
- * the default name of the message property containing the type is '__TypeId__'.
- * An optional property {@link #setDefaultType(Class)}
- * is provided that allows mapping to a statically defined type, if no message property is
- * found in the message properties.
+ * Maps to/from JSON using type information in the {@link MessageProperties}; the default
+ * name of the message property containing the type is
+ * {@value #DEFAULT_CLASSID_FIELD_NAME}. An optional property
+ * {@link #setDefaultType(Class)} is provided that allows mapping to a statically defined
+ * type, if no message property is found in the message properties.
+ * {@link #setIdClassMapping(Map)} can be used to map tokens in the
+ * {@value #DEFAULT_CLASSID_FIELD_NAME} header to classes. If this class is not a
+ * Spring-managed bean, call {@link #afterPropertiesSet()} to set up the class to id
+ * mapping.
  * @author Mark Pollack
  * @author Gary Russell
  *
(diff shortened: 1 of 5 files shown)


artembilan:
I wonder how Travis failure is related to your changes:
```
org.springframework.amqp.support.converter.DefaultClassMapperTests > shouldReturnHashtableForFieldWithHashtable FAILED

    java.lang.AssertionError at DefaultClassMapperTests.java:93
```

spring-amqp/src/main/java/org/springframework/amqp/support/converter/DefaultClassMapper.java
```diff
@@ -45,27 +49,61 @@
...
 	public void setDefaultType(Class<?> defaultType) {
 		this.defaultType = defaultType;
 	}
 
-	public void setDefaultHashtableClass(Class<?> defaultHashtableClass) {
-		this.defaultHashtableClass = defaultHashtableClass;
+	/**
+	 * Set the type of {@link Map} to use. For outbound messages, set the
+	 * {@value #DEFAULT_CLASSID_FIELD_NAME} header to {@code HashTable}. For inbound messages,
+	 * if the {@value #DEFAULT_CLASSID_FIELD_NAME} header is {@code HashTable} convert to thi
```
artembilan: Typo: thi**s**

garyrussell:
Most likely because I didn't run a build 😄 

Oops...

Fixing

spring-amqp/src/main/java/org/springframework/amqp/support/converter/DefaultClassMapper.java
```diff
@@ -45,27 +49,61 @@
...
+	 */
+	@Deprecated
+	public void setDefaultHashtableClass(Class<?> defaultMapClass) {
+		this.defaultMapClass = defaultMapClass;
+	}
+
+	/**
+	 * Set the type of {@link Map} to use. For outbound messages, set the
+	 * {@value #DEFAULT_CLASSID_FIELD_NAME} header to {@code HashTable}. For inbound messages,
+	 * if the {@value #DEFAULT_CLASSID_FIELD_NAME} header is {@code HashTable} convert to this
```
artembilan: What was the reason to introduce a new method if you still tell about `HashTable`?
Mostly looks like both methods are for the same reason...

garyrussell: I deprecated the old setter `setDefaultHashtableClass`.

HashMap has been preferred to `Hashtable` since JDK 1.2 😄 

In this context, `Hashtable` is simply the header name - we can't change that without adding more code. Otherwise we'd have interoperability issues and I didn't think it worth it just for a header name.

artembilan: Looking to the code I see:
```
if (Map.class.isAssignableFrom(classOfObjectToConvert)) {
	return DEFAULT_HASHTABLE_TYPE_ID;
}
```
So, when we send a `Map`, we always populated a `__TypeId__` as a `Hashtable`.
When we receive the header like `Hashtable` we convert to the `LinkedHashMap` by default.

I may understand the reason to accept `Hashtable` for legacy applications, but that isn't clear for me why we have to still produce data with the old protocol? To assist the same legacy apps? Postel's Law ?

Thanks 

You have a thi**s** typo in the `setDefaultHashtableClass` JavaDoc, BTW

Pulling locally for merge...

garyrussell: Right `DEFAULT_HASHTABLE_TYPE_ID` is `Hashtable` - that's just the header name.

Previously, if the sending system published any kind of map, the receiving system would build a `HashTable`. All I have done is changed that default to `LinkedHashMap`.

The header name remains the same so a 1.7.x app can talk to a 2.0 app without problems.

garyrussell: Right `DEFAULT_HASHTABLE_TYPE_ID` is `Hashtable` - that's just the `__TypeId__` header value (special token for maps).

Previously, if the sending system published any kind of map, the receiving system would build a `HashTable`. All I have done is changed that default to `LinkedHashMap`.

The header name remains the same so a 1.7.x app can talk to a 2.0 app without problems.

garyrussell: Further, if no class mapper is configured, Jackson will create a `LinkedHashMap` by default.

garyrussell: I also think we could set the `defaultType` to `LinkedHashMap` - to prevent the...

	throw new MessageConversionException(
			"failed to convert Message content. Could not resolve "
					+ getClassIdFieldName() + " in header " +
					"and no defaultType provided");

again, to be consistent with Jackson

artembilan: Thank you for explanation!
BTW I see this is tagged for `1.7.x` as well.
So, not sure in a story about  1.7.x to/from 2.0 communication issue...
But, yeah... Let it live as you described!

artembilan: >defaultType to LinkedHashMap  

OK. Will do on merge.

garyrussell: Oh - yes; let's drop the 1.7.x now that I made behavior changes.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/8fc9e4e91cadccca1c87729aa173e6710e29a0ea

PR #556, opened 2017-02-04
Fix asciidoc markup formatting for inlined code

kryger:


src/reference/asciidoc/appendix.adoc
```diff
@@ -659,4 +659,4 @@ Adds support for Dead Letter Exchanges/Dead Letter Queues.
 
 Adds an option to support adding a message id to logged messages.
 
-Adds an option to allow the specification of a `Charset` name to be used when converting `String`s to `byte[]`.
+Adds an option to allow the specification of a `Charset` name to be used when converting ``String``s to `byte[]`.
```
artembilan: This is how it looks in the PDF variant:
pdf

and HTML:
html



So, I'm afraid we can't accept this change.

In other places we use workaround like "`String` s".
In other words a whitespace before that `s` for plural.

I will merge it when you fix that way.
Thanks

kryger:
Hi @artembilan , amended the commit as per the request, I haven't tried rendering into pdf but the markdown preview looks OK.

PR #557, opened 2017-02-06
AMQP-617: Don't alias Queue and Exchange name attr

artembilan:
JIRA: https://jira.spring.io/browse/AMQP-617

If `id` attribute is present don't register a bean alias for the `name` attribute

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/AbstractExchangeParser.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/AbstractExchangeParser.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2017 the original author or authors.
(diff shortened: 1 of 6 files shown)


garyrussell:
Cool beans! Doc Polish: If you wish to declare a queue and exchange with the same `name`, you must provide an `id`.

artembilan:
Gary,  looks like you have missed the link to the docs.

garyrussell:
Sorry - updated

garyrussell:
I've changed the last sentence...


>There is no change if the element has just a `name` attribute, the bean can still be referenced by the `name`, for example in binding declarations, but you still can't reference it if the name contains SpEL - provide an `id` for reference purposes.

artembilan:
Cool! I'm fine with your polishing, Gary
Thank you!

garyrussell:
Merged as 72ce807fd3f095bf8408fb09e2840896838c580e

PR #558, opened 2017-02-06
AMQP-708: Fix Leaked AutoRecovered Channel

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-708

When publisher confirms or returns are enabled, and a container is stopped
while the broker is down, an auto-recover channel is leaked. It is later
recovered by the client library but has no code actually consuming from it.

While the channel is flagged to physically close it (to prevent the client
from recovering it), that close is avoided because `isOpen()` returns false
until the connection/channel is recovered.

Unconditionaly close the channel if physical closure is required.

Add a test with mocks; also tested with the SI sample and a real broker.
Before the patch, observe the leaked consumer on the admin console.
After the patch, the consumer is no longer retained after recovery.

Also, ignore `AlreadyClosedException` s in `RabbitUtils`.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
@@ -1072,9 +1072,7 @@ public class CachingConnectionFactory extends AbstractConnectionFactory
 				catch (Exception e2) { }
 				finally {
 					try {
-						if (channel.isOpen()) {
-							channel.close();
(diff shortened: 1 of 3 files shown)


artembilan:
... and cherry-picked to `1.6.x` as https://github.com/spring-projects/spring-amqp/commit/6fad1a2568aae8a234c30b9fd13b1ec1ab55bed9 and to `1.7.x` https://github.com/spring-projects/spring-amqp/commit/917ac4bf2aac9f4577371311ed91f50ad9dad971 after resolving conflicts.

PR #560, opened 2017-02-14
AMQP-710: Fix Exception Return From Void Method

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-710

Throw the original exception if `returnException` is set for a
method with void return, with no return address.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/DelegatingInvocableHandler.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/DelegatingInvocableHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 the original author or authors.
+ * Copyright 2015-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/DelegatingInvocableHandler.java
```diff
@@ -225,4 +225,15 @@ public String getMethodNameFor(Object payload) {
 		return handlerForPayload == null ? "no match" : handlerForPayload.getMethod().toGenericString(); //NOSONAR
 	}
 
+	/**
+	 * Return a the method that will be invoked for this payload.
```
artembilan: Typo with articles.
I think just `a` should be enough

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -137,17 +137,29 @@ public void onMessage(org.springframework.amqp.core.Message amqpMessage, Channel
...
-					throw e;
-				}
-				handleResult(new RemoteInvocationResult(e.getCause()), amqpMessage, channel, message);
+				returnOrThrow(amqpMessage, channel, message, e.getCause(), e);
+			}
+		}
+	}
+
+	private void returnOrThrow(org.springframework.amqp.core.Message amqpMessage, Channel channel, Message<?> message,
+			Throwable exToReturn, Exception exToThrow) throws Exception {
```
artembilan: Is that a new codding style to abbreviate variables names?

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/e1d1dd29c08bfbe2051f6e9c5019544157abfe41

PR #561, opened 2017-02-15
AMQP-711: Make Transaction Rollback Consistent

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-711

Previously, messsages were unconditionally requeued when an external
transaction manager is used; this was not consistent with local
transactions, where the normal requeue logic was honored.

__backport to 1.7.x will require rework due to the container changes - I will issue another PR after this is reviewed/merged__

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/ConnectionFactoryUtils.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/ConnectionFactoryUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


garyrussell:
This needs more work - it doesn't work if the TxMgr is a `RabbitTransactionManager`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/ConnectionFactoryUtils.java
```diff
@@ -167,18 +167,19 @@ public static void releaseResources(RabbitResourceHolder resourceHolder) {
 		RabbitUtils.closeConnection(resourceHolder.getConnection());
 	}
 
-	public static void bindResourceToTransaction(RabbitResourceHolder resourceHolder,
+	public static RabbitResourceHolder bindResourceToTransaction(RabbitResourceHolder resourceHolder,
 			ConnectionFactory connectionFactory, boolean synched) {
 		if (TransactionSynchronizationManager.hasResource(connectionFactory)
 				|| !TransactionSynchronizationManager.isActualTransactionActive() || !synched) {
-			return;
+			return (RabbitResourceHolder) TransactionSynchronizationManager.getResource(connectionFactory);
```
artembilan: When `|| !TransactionSynchronizationManager.isActualTransactionActive() || !synched`, the `.getResource(connectionFactory)` is `null`.
Please, ensure in target users an NPE protection.

garyrussell:
>cherry-picked...

That's why the PR says:

> __backport to 1.7.x will require rework due to the container changes - I will issue another PR after this is reviewed/merged__

😄

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/ConnectionFactoryUtils.java
garyrussell: I think it's only possible with weird configuration - such as...

    container.setTransactionAttribute(new DefaultTransactionAttribute(TransactionDefinition.PROPAGATION_NEVER));

...but possible - good catch.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/5d500b3685c8ea15623ff8626a4a4616299dc150

PR #565, opened 2017-02-20
Polish BrokerRunning @Rule

garyrussell:
- Improve Javadocs
- Fail fast when broker required and don't log the full stack trace

__cherry-pick to 1.7.x__


artembilan:
... and cherry-picked.

PR #567, opened 2017-02-25
AMQP-716: ExchangeBuilder and Durable

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-716

The builder was not consistent with the Abstract class regarding
the default for the `durable` flag.

__cherry-pick to 1.6.x__

I will issue a separate PR for master after merge - it will need `what's new`, and code changes including tests.

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/core/ExchangeBuilder.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/core/ExchangeBuilder.java
@@ -97,11 +97,24 @@ public final class ExchangeBuilder extends AbstractBuilder {
 	}
 
 	/**
-	 * Set the durable flag.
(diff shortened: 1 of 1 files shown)


garyrussell:
In 2.0 it will be true by default.

garyrussell:
>I will issue a separate PR for master after merge - it will need what's new, and code changes including tests.

garyrussell:
This PR is for the deprecation.

artembilan:
OK! Thanks

Any comments doesn't explain why we don't fix inconsistency...

Merging...

garyrussell:
We will fix the inconsistency in 2.0 😄

artembilan:
... and cherry-picked to `1.6.x`

garyrussell:
Thanks - master PR soon...

PR #568, opened 2017-02-26
AMQP-717: RabbitTemplate Invoke Method

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-717

Provide a mechanism to run `RabbitTemplate` opertions in a dedicated channel
instead of returning the channel to the cache on each operation.

Facilitates low-level channel operations like `waitForConfirmsOrDie()`.

For this reason, we need to subvert the channel proxy refresh logic and use the
bare channel in this mode.


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/TemplateCallback.java
```diff
@@ -0,0 +1,42 @@
...
+
+	/**
+	 * Execute any number of operations using a dedicated {@link Channel} as long as those
+	 * operations are performed on the template argument and on the calling thread. The
+	 * channel will be physically closed when the callback exits.
+	 *
+	 * @param template The template.
+	 * @return The result.
+	 */
+	T doInRabbit(RabbitTemplate template);
```
artembilan: Can't it be just `RabbitOperations`?
And more over this interface should move to the `RabbitOperations` directly.
Like we have `MessageGroupStore.MessageGroupCallback` relationship.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -1703,33 +1710,42 @@ public Boolean isMandatoryFor(final Message message) {
 
 	private <T> T doExecute(ChannelCallback<T> action, ConnectionFactory connectionFactory) {
 		Assert.notNull(action, "Callback object must not be null");
-		Channel channel;
+		Channel channel = null;
+		boolean invokeScope = false;
+		if (this.activeTemplateCallbacks.get() > 0) {
+			channel = this.dedicatedChannels.get();
```
artembilan: Looks like we don't need this `this.activeTemplateCallbacks`, the simple `this.dedicatedChannels.get()` works here the same way. See your logic there in the `invoke()`:
```
Assert.state(currentChannel == null, () -> "Nested invoke() calls are not supported; channel '" + currentChannel
				+ "' is already associated with this thread");
```

I won't mind if I am missing something on the matter...

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -1750,6 +1766,59 @@ public Boolean isMandatoryFor(final Message message) {
...
+	public <T> T invoke(TemplateCallback<T> action) {
+		final Channel currentChannel = this.dedicatedChannels.get();
+		Assert.state(currentChannel == null, () -> "Nested invoke() calls are not supported; channel '" + currentChannel
+				+ "' is already associated with this thread");
+		this.activeTemplateCallbacks.incrementAndGet();
+		RabbitResourceHolder resourceHolder = null;
+		Connection connection = null; // NOSONAR (close)
+		Channel channel;
+		if (isChannelTransacted()) {
+			resourceHolder = ConnectionFactoryUtils.getTransactionalResourceHolder(getConnectionFactory(), true);
```
artembilan: Doesn't this mean that we reuse `channel` in case of outer transaction? But with the *dedicated* channel like this PR claims...
Looks like  `ConnectionFactoryUtils.getTransactionalResourceHolder()` returns exactly:
```
RabbitResourceHolder resourceHolder = (RabbitResourceHolder) TransactionSynchronizationManager
				.getResource(connectionFactory);
```

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
garyrussell: This was just an attempt at optimization - only access the `ThreadLocal` if we know for sure it might contain a value. In most cases, so I wanted an optimized code path.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
garyrussell: Yes, that crossed my mind - I considered throwing an exception (or not store in the TL) if they attempted to use this technique while in a global transaction but it seemed to me there is no real issue here (except we'd just document it's unnecessary to use `invoke` if there is already a global channel.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/TemplateCallback.java
garyrussell: True

garyrussell:
I am not a fan of the multi/exec approach - plus that would not be appropriate if we have a global channel bound to the thread. With my technique it would work either way.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
artembilan: Yeah... I see now. 
I was worried about `ConnectionFactoryUtils.releaseResources(resourceHolder);` in the end, but now I see `resourceHolder.isSynchronizedWithTransaction()` there and we just move on idempotently.
So, never mind.

Thanks for your patience!

garyrussell:
Another problem with the multi/exec is you have to rely on the user calling the exec; with this scoped approach, we can reliably release the resources when the `invoke` exits.

artembilan:
> we can reliably release the resources when the invoke exits.

yeah-yeah! I see.

OK, never mind. Just wanted to share a bit other angle of vision to the problem...

Fine with me any way.

You can go ahead with Docs and we will merge it today to be ready for release tomorrow/after tomorrow.

garyrussell:
👍

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -155,6 +155,13 @@
 
 	private static final SpelExpressionParser PARSER = new SpelExpressionParser();
 
+	/*
+	 * Not static as normal since we want this TL to be scoped within the template instance.
```
artembilan: ???
Please, elaborate more.
The `ThreadLocal` is stateless and really just a delegate to the `Thread.currentThread()`

garyrussell: I think your understanding is incorrect, while we normally use static TLs, it's not required if you want a TL per instance...

	public static class Foo {

		static ThreadLocal<String> foo = new ThreadLocal<>();

		ThreadLocal<String> bar = new ThreadLocal<>();

	}

        ...

	Foo one = new Foo();
	Foo.foo.set("foo");
	one.bar.set("bar");
	Foo two = new Foo();
	System.out.println(two.foo.get() + "/" + two.bar.get());

    foo/null

In this case, we need an instance per template because they might invoke a template that's configured to use a different connection factory.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/7793edaba06fb40d14c0d99615f6c0f8224bcc88

PR #572, opened 2017-02-27
CheckStyle: JavaDoc

garyrussell:
@artembilan - you might not want to bother looking at all the changes but you should look at the checkstyle config file changes.

I had to suppress `JavaDocStyle` for two classes because `<bean/>` tags within `<pre/>` confused checkstyle.


artembilan:
I don't understand what's wrong with Travis: works well for me locally.
Merging...

PR #575, opened 2017-03-04
support anonymous queues in @RabbitListener queuesToDeclare

panchenko:
This code seems to be inconsistent:
* declareQueue() already calls resolveExpression(), no need to resolve it another time
* if queue name is empty a random name is generated and returned - which is ignored

That's why I suggest this change, it allows doing code like this (without this change it fails as listener container is trying to use an empty queue name).

```java
@RabbitListener(queuesToDeclare = @Queue)
public void onMessage(@Payload String message) {}
```

WDYT?

Btw, a good question for this setup would be obtaining this queue name, the following works, but is not that convenient:
```java
@RabbitListener(... group = "foo")
...
final List<?> foo = applicationContext.getBean("foo", List.class);
final String fooQueue = ((AbstractMessageListenerContainer) foo.get(0)).getQueueNames()[0];
```

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
@@ -495,8 +495,7 @@ public class RabbitListenerAnnotationBeanPostProcessor
 						"@RabbitListener can have only one of 'queues', 'queuesToDeclare', or 'bindings'");
 			}
 			for (int i = 0; i < queuesToDeclare.length; i++) {
-				declareQueue(queuesToDeclare[i]);
-				resolveAsString(resolveExpression(queuesToDeclare[i].value()), result);
+				result.add(declareQueue(queuesToDeclare[i]));
 			}
 		}
 		if (bindings.length > 0) {
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java
@@ -16,6 +16,7 @@
 
 package org.springframework.amqp.rabbit.annotation;
 
+import static org.hamcrest.Matchers.arrayWithSize;
 import static org.hamcrest.Matchers.contains;
 import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.equalTo;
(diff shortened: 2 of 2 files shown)


garyrussell:
Good catch, thanks!! - although we don't need the intermediate `queueName` variable.

While we're in there, I'd prefer that we switch to the new default anonymous queue naming strategy.

```
@@ -128,6 +128,8 @@ public class RabbitListenerAnnotationBeanPostProcessor
 
        private final Set<String> emptyStringArguments = new HashSet<>();
 
+       private final AnonymousQueue.Base64UrlNamingStrategy anonymousNamer = new AnonymousQueue.Base64UrlNamingStrategy();
+
        private RabbitListenerEndpointRegistry endpointRegistry;
 
        private String containerFactoryBeanName = DEFAULT_RABBIT_LISTENER_CONTAINER_FACTORY_BEAN_NAME;
@@ -495,8 +497,7 @@ public class RabbitListenerAnnotationBeanPostProcessor
                                                "@RabbitListener can have only one of 'queues', 'queuesToDeclare', or 'bindings'")
                        }
                        for (int i = 0; i < queuesToDeclare.length; i++) {
-                               declareQueue(queuesToDeclare[i]);
-                               resolveAsString(resolveExpression(queuesToDeclare[i].value()), result);
+                               result.add(declareQueue(queuesToDeclare[i]));
                        }
                }
                if (bindings.length > 0) {
@@ -550,7 +551,7 @@ public class RabbitListenerAnnotationBeanPostProcessor
                boolean exclusive = false;
                boolean autoDelete = false;
                if (!StringUtils.hasText(queueName)) {
-                       queueName = UUID.randomUUID().toString();
+                       queueName = this.anonymousNamer.generateName();
                        // default exclusive/autodelete to true when anonymous
                        if (!StringUtils.hasText(bindingQueue.exclusive())
                                        || resolveExpressionAsBoolean(bindingQueue.exclusive())) {

```

Here's a (slightly) better way to get the queue name, avoiding the need to put it in a group - good trick, though 😄 

```java
@SpringBootApplication
public class Agh575Application implements CommandLineRunner {

	public static void main(String[] args) {
		SpringApplication.run(Agh575Application.class, args);
	}

	@Autowired
	private RabbitListenerEndpointRegistry registry;

	@RabbitListener(id="foo", queuesToDeclare = @Queue)
	public void listen(byte[] in) {
		System.out.println(new String(in));
	}

	@Override
	public void run(String... arg0) throws Exception {
		SimpleMessageListenerContainer container = (SimpleMessageListenerContainer) this.registry
				.getListenerContainer("foo");
		System.out.println(container.getQueueNames()[0]);
	}

}
```

We should probably add a note to the reference manual somewhere to tell people how to obtain the anonymous queue name.

Let me know if you want to polish up this PR or if we should take it over.

M3 is due for release next Tuesday, so it would be good to get this in.

panchenko:
@garyrussell Surely I would like to polish this. However not sure why the test fails in travis, locally it works.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/77445943d62add5f5b31231344246fe15d36ad07

garyrussell:
Yes, it's a bug [AMQP-721](https://jira.spring.io/browse/AMQP-721). I will do the backports.

artembilan:
Thanks!

garyrussell:
Forget that; no; it's a 2.0 feature.

PR #576, opened 2017-03-05
use AnonymousQueue.Base64UrlNamingStrategy in RabbitListenerAnnotationBeanPostProcessor

panchenko:
As suggested by @garyrussell in #575 

It's logically a separate change, so IMHO another PR would make sense.


artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/29a70c7d74374b958f24191a7feaf1c550b246ad.

Fully unclear why `checkTestConfigs` fails on the Travis...

PR #577, opened 2017-03-05
drop UniquelyNamedQueue

panchenko:
Just noticed this class for creating _durable_ queues with _random_ names. To be honest I don't see a good use case for it - using it would result in having lots of queues in the broker.
There is AnonymousQueue for creating non-durable queues which makes more sense.
2.0 is a major release, so is a good chance to delete it.


garyrussell:
Thanks, I agree; merged as 8cc3a76d587dc16367393707824343dfb87fce55 (with what's new) and deprecated in 1.7.x 3122446e966c35e1d7b992404a210c77a5d2da74

garyrussell:
Also, please note that we prefer to track changes like this via JIRA [AMQP-720](https://jira.spring.io/browse/AMQP-720) so they show up in the JIRA release notes for a version.

PR #578, opened 2017-03-07
AMQP-722 Support multiple routing keys in @QueueBinding

panchenko:
I described the idea in https://jira.spring.io/browse/AMQP-722

If we are going forward I could add javadocs to the affected privates methods and probably change Collection to Predicate, etc.

WDYT?

### DIFF
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessorTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessorTests.java
@@ -17,7 +17,9 @@
 package org.springframework.amqp.rabbit.annotation;
 
 import static org.hamcrest.Matchers.allOf;
+import static org.hamcrest.Matchers.contains;
 import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.Matchers.instanceOf;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
@@ -28,10 +30,16 @@ import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
 import java.util.Iterator;
+import java.util.List;
+import java.util.stream.Collectors;
 
 import org.junit.Test;
 
+import org.springframework.amqp.core.Binding;
 import org.springframework.amqp.core.Queue;
 import org.springframework.amqp.rabbit.config.MessageListenerTestContainer;
 import org.springframework.amqp.rabbit.config.RabbitListenerContainerTestFactory;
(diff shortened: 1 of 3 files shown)


spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessorTests.java
```diff
@@ -195,6 +203,33 @@ public void invalidValueInAnnotationTestBean() {
...
+
+		RabbitListenerContainerTestFactory factory = context.getBean(RabbitListenerContainerTestFactory.class);
+		assertThat("one container should have been registered", factory.getListenerContainers(), hasSize(1));
+		RabbitListenerEndpoint endpoint = factory.getListenerContainers().get(0).getEndpoint();
+		assertEquals(Collections.singletonList("my_queue"),
+				((AbstractRabbitListenerEndpoint) endpoint).getQueueNames());
+		final List<Queue> queues = new ArrayList<>(context.getBeansOfType(Queue.class).values());
+		queues.sort(Comparator.comparing(Queue::getName));
+		assertThat(queues.stream().map(Queue::getName).collect(Collectors.toList()),
+				contains("my_queue", "secondQueue", "testQueue"));
```
panchenko: This check can be simplified by moving `testQueue`/`secondQueue` from `Config` into another class, which is included only when needed. Can be a separate PR. Should I?

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessorTests.java
```diff
@@ -269,6 +304,16 @@ public void handleIt(String body) {
 		}
 	}
 
+	@Component
+	static class MultipleBindingsTestBean {
+
+		@RabbitListener(bindings = {
+				@QueueBinding(exchange = @Exchange("my_exchange"), value = @org.springframework.amqp.rabbit.annotation.Queue(value = "my_queue", arguments = @Argument(name = "foo", value = "bar")), key = "red"),
+				@QueueBinding(exchange = @Exchange("my_exchange"), value = @org.springframework.amqp.rabbit.annotation.Queue("my_queue"), key = "yellow") })
```
panchenko: Eclipse formatter insist on having it this way :) I can wrap it manually.

garyrussell: Wouldn't it be a lot easier to change the key property to be an array...

	String[] key() default {};

... and then

		@RabbitListener(bindings = {
				@QueueBinding(exchange = @Exchange("my_exchange"), 
						value = @org.springframework.amqp.rabbit.annotation.Queue("my_queue"), 
							key = { "red", "yellow" }) })
		public void handleIt(String body) {
		}

?

panchenko:
@garyrussell Your suggestion totally makes sense, I've updated this PR.

garyrussell:
Merged as ddb42da8605bfcc287156da0b621383d391291db with minor polish and docs.

Thanks @panchenko !!

(Please limit commit comment headlines to 50 chars).

PR #579, opened 2017-03-07
test: restrict memory usage, so tests don't fail on travis

panchenko:
The changes from https://discuss.gradle.org/t/gradle-travis-sudo-false-leads-to-an-error/17034/3

With these numbers it works, it's a good question if it would be better to have numbers higher or lower :-)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,6 +24,7 @@ apply plugin: 'idea'
 def docsDir = 'src/reference/asciidoc' // Will be default with newer asciidoctor plugin
 
 ext {
+	isTravisBuild = System.getenv().get("TRAVIS") == 'true'
 	linkHomepage = 'https://projects.spring.io/spring-amqp'
 	linkCi       = 'https://build.spring.io/browse/AMQP'
 	linkIssue    = 'https://jira.spring.io/browse/AMQP'
@@ -149,6 +150,10 @@ subprojects { subproject ->
 		jacoco {
 			append = false
 			destinationFile = file("$buildDir/jacoco.exec")
+		}
+		if (isTravisBuild) {
+			// restrict memory usage, so tests don't fail with exit code 137 on travis
+			maxHeapSize = '256m'
 		}
 	}
 


artembilan:
@panchenko ,

It is always pleasure to have contributions from you, but would you mind elaborating more what has made you to propose this change? 
And it is really good practice to include "Why?" to the commit message: https://chris.beams.io/posts/git-commit/

Thanks

panchenko:
I've simplified changes here.

Lately tests in travis fail with exit code 137, in various internet topics that's explained as processed was killed by the container environment most likely because of high memory usage.

Currently [default max heap](https://docs.oracle.com/javase/8/docs/technotes/guides/vm/gc-ergonomics.html) is "Smaller of 1/4th of the physical memory or 1GB" and seems like that's too much for travis.

The only question if this limit should be used always or only on travis.

garyrussell:
Thanks @panchenko - my inclination is to make it travis only - I (personally) have never had problems on my laptop.

artembilan:
We use locally daemon and `--parallel` to build as fast as possible.
The `fork` is expensive operation because it starts a new JVM, so, that is really should be just for Travis.
Thanks

build.gradle
```diff
@@ -150,6 +151,10 @@ subprojects { subproject ->
 			append = false
 			destinationFile = file("$buildDir/jacoco.exec")
 		}
+		if (isTravisBuild) {
+			// restrict memory usage, so tests don't fail with exit code 137 on travis
+			maxHeapSize = '256m'
```
artembilan: @panchenko ,

Looks like now we adjust only `Xmx` .
Isn't that Travis option really enough:
```
env:
  global:
  - MALLOC_ARENA_MAX=1
  - GRADLE_OPTS="-Xmx512m"
```
?

panchenko: There are 2 different things: the main gradle process which does the build, it's affected by GRADLE_OPTS and another forked process which performs the tests (currently it's forked once and performs all the tests, if specifying the "forkEvery" parameter then fork terminates after that number of tests and a new one is forked).

Currently only the forked "Gradle Test Executor" experience problems, so this change is enough.

artembilan: Fair enough, @panchenko !

Let me "put in on board" and merge!

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/4c130b02f02dda003a055d61ffdd5b12a89b963e

Thank you again, @panchenko !

PR #581, opened 2017-03-11
AMQP-724 fix NPE parsing rabbit:listener-container

panchenko:


spring-rabbit/src/test/java/org/springframework/amqp/rabbit/config/ListenerContainerParserTests.java
```diff
@@ -225,14 +225,22 @@ public void testAnonParent() throws Exception {
...
 				"Configuration problem: Listener Container - cannot set channel-transacted with acknowledge='NONE'"));
 		}
 	}
 
+	@Test
+	public void AMQP_724_potentialNPE() {
+		final DefaultListableBeanFactory bf = new DefaultListableBeanFactory();
+		final XmlBeanDefinitionReader reader = new XmlBeanDefinitionReader(bf);
+		reader.setValidating(false);
+		reader.loadBeanDefinitions(new ClassPathResource(getClass().getSimpleName() + "-context.xml", getClass()));
```
panchenko: That file contains
`<rabbit:listener-container connection-factory="connectionFactory" channel-transacted="true" transaction-size="5" >`
and if running without validation acknowledgeMode is null, but with validation is has default value from xsd.

Or a separate file here?

panchenko:
Reproducible in 1.6 / 1.7 too.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/RabbitNamespaceUtils.java
```diff
@@ -192,7 +192,7 @@ public static BeanDefinition parseContainer(Element containerEle, ParserContext
 		String channelTransacted = containerEle.getAttribute(CHANNEL_TRANSACTED_ATTRIBUTE);
 		if (StringUtils.hasText(channelTransacted)) {
 			// Note: a placeholder will pass this test, but if it resolves to true, it will be caught during container initialization
-			if (acknowledgeMode.isAutoAck() && channelTransacted.equalsIgnoreCase("true")) {
+			if (acknowledgeMode == AcknowledgeMode.NONE && channelTransacted.equalsIgnoreCase("true")) {
```
panchenko: `acknowledgeMode != null && acknowledgeMode.isAutoAck()` is also possible, but looks longer :)

artembilan:
Thank you, @panchenko , for sharing the concern and attempt to fix the issue!

To be honest that logic in the XML parser is fully wrong. 
We might just don't have actual value for the property the because of property-placeholder and SpEL configuration in the XML, e.g.:
```
acknowledge="${acknowledge.mode:manual}"  channel-transacted="#{true}"
```
That logic must go to the target class, because for the good Spring design we must have XML and Java config consistency. Therefore an XML parser must be as lightweight as possible.

Yes, I see that `acknowledge` doesn't support arbitrary strings, only appropriate tokens:
```
<xsd:simpleType>
	<xsd:restriction base="xsd:NMTOKEN">
			<xsd:enumeration value="auto" />
			<xsd:enumeration value="manual" />
			<xsd:enumeration value="none" />
	</xsd:restriction>
</xsd:simpleType>
```
But exactly this is a problem because we can't use property-placeholder here.
Plus that hard-coded logic for the target values in the `RabbitNamespaceUtils` makes it useless as well.

Let's consider to fix the problem that way, since we are already here!

We don't care so much about XML support these days because Java & Annotation configuration, especially together with Spring Boot, makes life easier 😄

garyrussell:
My guess is that property wasn't made property-placedholder-friendly because it's not likely something you would want to change via properties - it's something you decide at design-time, not deployment-time.

While we must continue to support/extend XML configuration, I am not sure it's worth spending too much energy on things like this.

artembilan:
OK. Then let's merge the fix as is ! 😄 
Right?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/RabbitNamespaceUtils.java
artembilan: We can flip condition: `AcknowledgeMode.NONE == acknowledgeMode`. 
This way we will protect ourselves from `NPE` as well.

Well, I'd say that we accept the PR, but I don't like that test-case. Especially the name mentioning `AMQP-274`.

@garyrussell , your opinion?

garyrussell: OK with me - we generally don't need issue numbers in test names anyway - `git blame` is our friend.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/df3cb446df56dacb2617f97d5528cd8f7c8a6b41

PR #582, opened 2017-03-13
minor optimization Message.toString()

panchenko:
* StringBuffer -> StringBuilder in Message.toString()
* in tests: correlationId is a string, `new String(String)` can be avoided

Just to have less warnings in IDE :-)


artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/3a43611d46c37d311fdb22465547dcb07b5534ad.
Thank you!

PR #583, opened 2017-03-13
AMQP-726: Always Cancel Auto Recovering Consumers

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-726

When using auto recovering channels, `RabbitUtils.closeMessageConsumer()` skipped canceling
the consumers if the channel reports as being closed.

This should not cause a problem since the channel is physically closed, which should prevent
the consumers from being recovered.

Change the utility method to unconditionally cancel such consumers (which removes them as
candidates for recovery).

Also add the following defensive code:

1. If stop() is called without first calling `basicCancel()` (so the utility does the canceling),
set the `abortStarted` timer to signal the `handleDelivery` to use `offer()` instead of `put()`.
This happens from `run()` in `SMLC.restart()`.
2. If we time out the `offer()`, unconditionally drain the queue, reject, cancel the consumer and close the channel.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
@@ -812,18 +812,16 @@ public class BlockingQueueConsumer {
 				if (BlockingQueueConsumer.this.abortStarted > 0) {
 					if (!BlockingQueueConsumer.this.queue.offer(new Delivery(consumerTag, envelope, properties, body),
 							BlockingQueueConsumer.this.shutdownTimeout, TimeUnit.MILLISECONDS)) {
+						RabbitUtils.setPhysicalCloseRequired(true);
+						// Defensive - should never happen
+						BlockingQueueConsumer.this.queue.clear();
+						getChannel().basicNack(envelope.getDeliveryTag(), true, true);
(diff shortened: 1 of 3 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
```diff
@@ -808,8 +811,20 @@ public void handleDelivery(String consumerTag, Envelope envelope, AMQP.BasicProp
...
+							BlockingQueueConsumer.this.queue.clear();
+							getChannel().basicNack(envelope.getDeliveryTag(), true, true);
+							getChannel().basicCancel(consumerTag);
+							RabbitUtils.setPhysicalCloseRequired(true);
+							getChannel().close();
+						}
+						catch (Exception e) {
+							if (logger.isDebugEnabled()) {
+								logger.debug("Error performing 'basicCancel'", e);
+							}
```
artembilan: This is an exception. Why just don't let to bubble as is?
More over we may have an exception not only on the `basicCancel` here.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/2d8c8113e2377b9cfcb5952bbb829b3395c0b40c, cherry-picked to `1.7.x` and `1.6.x` as https://github.com/spring-projects/spring-amqp/commit/52909a707b69479f37bef514e17ceaa1dcae1de6

PR #584, opened 2017-03-13
AMQP-726: Fix for ChannelProxy; Polishing

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-726

- `instanceof` in RabbitUtils didn't take account of `ChannelProxy`
- fix partial stubbing problem with channel close - use a single stubbing with atomic boolean

__cherry-pick (first commit only) to 1.7.x, 1.6.x; second commit is for master only__

### DIFF
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumerTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumerTests.java
@@ -26,6 +26,7 @@ import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.BDDMockito.given;
 import static org.mockito.BDDMockito.willAnswer;
 import static org.mockito.BDDMockito.willThrow;
+import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
@@ -38,6 +39,7 @@ import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.logging.log4j.Level;
@@ -48,6 +50,7 @@ import org.mockito.Mockito;
 
 import org.springframework.amqp.AmqpRejectAndDontRequeueException;
 import org.springframework.amqp.core.AcknowledgeMode;
+import org.springframework.amqp.rabbit.connection.ChannelProxy;
 import org.springframework.amqp.rabbit.connection.Connection;
 import org.springframework.amqp.rabbit.connection.ConnectionFactory;
 import org.springframework.amqp.rabbit.support.ConsumerCancelledException;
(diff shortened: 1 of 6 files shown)


garyrussell:
Note that the second commit tests look like large changes; in fact it's a simple rename and one new test case but GitHub doesn't detect that.

    DirectMessageListenerContainerTests -> DirectMessageListenerContainerIntegrationTests

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumerTests.java
```diff
@@ -240,11 +243,14 @@ private void testRequeueOrNotGuts(Exception ex, boolean expectedRequeue,
...
 		Connection connection = mock(Connection.class);
-		Channel channel = mock(AutorecoveringChannel.class);
+		ChannelProxy channel = mock(ChannelProxy.class);
+		Channel rabbitChannel = mock(AutorecoveringChannel.class);
+		when(channel.getTargetChannel()).thenReturn(rabbitChannel);
 
 		when(connectionFactory.createConnection()).thenReturn(connection);
 		when(connection.createChannel(anyBoolean())).thenReturn(channel);
-		when(channel.isOpen()).thenReturn(true);
+		final AtomicBoolean isOpen = new AtomicBoolean(true);
```
artembilan: I think this 
```
when(channel.isOpen()).thenReturn(true, false);
```
would be much better.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainerMockTests.java
```diff
@@ -0,0 +1,100 @@
...
+		ConnectionFactory connectionFactory = mock(ConnectionFactory.class);
+		Connection connection = mock(Connection.class);
+		ChannelProxy channel = mock(ChannelProxy.class);
+		Channel rabbitChannel = mock(AutorecoveringChannel.class);
+		given(channel.getTargetChannel()).willReturn(rabbitChannel);
+
+		given(connectionFactory.createConnection()).willReturn(connection);
+		given(connection.createChannel(anyBoolean())).willReturn(channel);
+		final AtomicBoolean isOpen = new AtomicBoolean(true);
+		willAnswer(i -> isOpen.get()).given(channel).isOpen();
```
artembilan: The same multi `return()` could be done here as well.

Will fix during merge...

artembilan:
Cherry-picked first commit as https://github.com/spring-projects/spring-amqp/commit/68176de5795d8f8e1210ef703f76ffc3c00d998f and merged as https://github.com/spring-projects/spring-amqp/commit/b8470e095a0bc28639e4af184a95b592a3fa6163

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumerTests.java
garyrussell: Well, we don't know how often `isOpen()` will be called.

artembilan:
```
 /**
     * Sets consecutive return values to be returned when the method is called. E.g:
     * <pre class="code"><code class="java">
     * when(mock.someMethod()).thenReturn(1, 2, 3);
     * </code></pre>
     *
     * Last return value in the sequence (in example: 3) determines the behavior of further consecutive calls.
...
    OngoingStubbing<T> thenReturn(T value, T... values);
```

Does it make sense to you?

garyrussell: I understand, but `isOpen()` is called quite often; I specifically coded it this way because we want it to always return `true` until we want it to return `false`. Your solution is brittle - perhaps it works now, but it will break if another `isOpen()` is added later.

artembilan: Ah! That is the point!

Yeah... Would be better do not merge in the evening until this your concern in the morning.

Sorry, will push the revert fix...

artembilan: Addressed via https://github.com/spring-projects/spring-amqp/commit/b238d6a6012d74920d1959fc581effd0d21a2120 and one more especially for `master`: https://github.com/spring-projects/spring-amqp/commit/d7e702a765bbd18b2a64c61297795e1cbe504e69

PR #587, opened 2017-03-14
Remove overridden DMLC methods accepting Queue

panchenko:
As the implementations of those in AbstractMessageListenerContainer
delegate to ones accepting Strings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ sudo: false
 services:
   - rabbitmq
 env:
+  global:
+  - GRADLE_OPTS="-Xmx512m"
   - TERM=dumb
 before_cache:
   - rm -f $HOME/.gradle/caches/modules-2/modules-2.lock
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
@@ -251,7 +251,7 @@ public abstract class AbstractMessageListenerContainer extends RabbitAccessor
 	 * Set the name of the queue(s) to receive messages from.
 	 * @param queues the desired queue(s) (can not be <code>null</code>)
 	 */
-	public void setQueues(Queue... queues) {
+	public final void setQueues(Queue... queues) {
 		setQueueNames(collectQueueNames(queues));
 	}
 
@@ -295,7 +295,7 @@ public abstract class AbstractMessageListenerContainer extends RabbitAccessor
 	 * Add queue(s) to this container's list of queues.
 	 * @param queues The queue(s) to add.
 	 */
-	public void addQueues(Queue... queues) {
+	public final void addQueues(Queue... queues) {
 		this.addQueueNames(collectQueueNames(queues));
 	}
 
@@ -315,7 +315,7 @@ public abstract class AbstractMessageListenerContainer extends RabbitAccessor
 	 * @param queues The queue(s) to remove.
 	 * @return the boolean result of removal on the target {@code queueNames} List.
 	 */
-	public boolean removeQueues(Queue... queues) {
+	public final boolean removeQueues(Queue... queues) {
 		return this.removeQueueNames(collectQueueNames(queues));
 	}
 
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
@@ -296,7 +296,7 @@ public abstract class AbstractMessageListenerContainer extends RabbitAccessor
 	 * @param queues The queue(s) to add.
 	 */
 	public final void addQueues(Queue... queues) {
-		this.addQueueNames(collectQueueNames(queues));
+		addQueueNames(collectQueueNames(queues));
 	}
 
 	/**
@@ -316,7 +316,7 @@ public abstract class AbstractMessageListenerContainer extends RabbitAccessor
 	 * @return the boolean result of removal on the target {@code queueNames} List.
 	 */
 	public final boolean removeQueues(Queue... queues) {
-		return this.removeQueueNames(collectQueueNames(queues));
+		return removeQueueNames(collectQueueNames(queues));
 	}
 
 	/**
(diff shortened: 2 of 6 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -245,21 +244,20 @@ public AcknowledgeMode getAcknowledgeMode() {
...
+		this.queueNames = new CopyOnWriteArrayList<>(queueName);
 	}
 
 	/**
 	 * Set the name of the queue(s) to receive messages from.
 	 * @param queues the desired queue(s) (can not be <code>null</code>)
 	 */
 	public void setQueues(Queue... queues) {
-		List<String> queueNames = new ArrayList<String>(queues.length);
+		final String[] queueNames = new String[queues.length];
```
artembilan: Can't we use `Arrays.stream(queues).map(Queue::getName)` instead ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -196,12 +196,6 @@ public final void setMissingQueuesFatal(boolean missingQueuesFatal) {
 	}
 
 	@Override
-	public void setQueues(Queue... queues) {
-		Assert.state(!isRunning(), "Cannot set queue names while running, use add/remove");
```
artembilan: ??? Why do you remove this method if it has this extra assertion logic ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -212,18 +206,6 @@ public void addQueueNames(String... queueNames) {
 		super.addQueueNames(queueNames);
 	}
 
-	@Override
-	public void addQueues(Queue... queues) {
-		try {
-			addQueues(Arrays.stream(queues)
```
artembilan: ??? This is delegation to the `private void addQueues(Stream<String> queueNameStream)`.
What is the point to get rid of this logic?
Thanks

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -248,19 +230,12 @@ public boolean removeQueueNames(String... queueNames) {
 		return super.removeQueueNames(queueNames);
 	}
 
-	@Override
-	public boolean removeQueues(Queue... queues) {
-		removeQueues(Arrays.stream(queues)
```
artembilan: The same here. To the `private void removeQueues(Stream<String> queueNames) `

What am I missing, please ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
panchenko: Currently both `setQueues(Queue...)` and `setQueueNames(String...)` are overridden with the same checks.

However methods in the base class collect names from `Queue...` and call the corresponding method with `String...` - I tried to mentioned that above, sorry it it was not clear.
So, it's enough to have this extra logic only in the methods with `String...`.

.travis.yml
```diff
@@ -4,6 +4,8 @@ sudo: false
 services:
   - rabbitmq
 env:
+  global:
+  - GRADLE_OPTS="-Xmx512m"
```
artembilan: What is the purpose of this change?

Would be great to have itemization in the commit message with the reason of each modification.
At least those which are not obvious.
Thanks

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
artembilan: You haven't answered to this concerns yet.
Thanks

.travis.yml
panchenko: It failed once in Travis with some "killed" message, my guess is that main gradle process used too much memory.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
panchenko: There are 3 methods - set/add/remove, and those with `Queue...` delegate to ones with `String...`, which actually do all the work.
IMHO methods with `Queue...` can even be declared final.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
panchenko: This one also delegates to `removeQueueNames(String...)` which has this extra logic.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
artembilan: OK! After closer look I really see your logic and what you suggest.
Not enough coffee? Or to many snow outside 😄 

Yes, I think we can go ahead with your suggestion and even really make those `Queue`-based methods as `final`.

Thank you for such a comprehensive analysis and patience to my nit-pickings.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -293,13 +296,7 @@ public void addQueueNames(String... queueNames) {
...
 	 */
 	public void addQueues(Queue... queues) {
-		Assert.notNull(queues, "'queues' cannot be null");
-		Assert.noNullElements(queues, "'queues' cannot contain null elements");
-		String[] queueNames = new String[queues.length];
-		for (int i = 0; i < queues.length; i++) {
-			queueNames[i] = queues[i].getName();
-		}
-		this.addQueueNames(queueNames);
+		this.addQueueNames(collectQueueNames(queues));
```
garyrussell: We don't use `this.` on method calls; previous code was bad; we generally clean these up when modifying code.

artembilan: We can do that on merge.
But since you, @panchenko , are so active contributor, please, consider to follow our code style and overall tracking rules to avoid similar comments again and again.
That should save some time for us, as well as for you 😄

panchenko: Sure I can adjust this :-)

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/eeab98e3657f675126718fdebd896e90b8960872

Issue #589, opened 2017-03-15
Extending Jackson2JsonMessageConverter


calvernaz:
I want to register a new module into the object mapper present in _Jackson2JsonMessageConverter_, but the documentation suggests two ways of doing it: 

- One is to configure the ObjectMapper and inject into _Jackson2JsonMessageConverter_. The documentation in the _setJsonObjectMapper_ suggest an alternative which is to extend the class and override the method _initializeJsonObjectMapper_

- The issue with extending the _Jackson2JsonMessageConverter_ is the lack of access to the ObjectMapper in the parent class, ending up like this:
```
public class JacksonMessageConverter extends Jackson2JsonMessageConverter {
    @Override
    protected void initializeJsonObjectMapper() {
        ObjectMapper objectMapper = new ObjectMapper();
        objectMapper.registerModule(new GuavaModule());
        objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
        this.setJsonObjectMapper(objectMapper);
    }
}
```
So we still have to inject the ObjectMapper, besides overriding the method, it doesn't really look like an alternative.

Changing the ObjectMapper visibility to protected (or a getter) would give us the alternative.

garyrussell:
Yes, it looks like an oversight; there should either be a getter or the mapper should be passed in as an argument to the method.

PR #590, opened 2017-03-15
Expose json object mapper through getter

calvernaz:
This patch address the issue opened in #589

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/support/converter/Jackson2JsonMessageConverter.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/support/converter/Jackson2JsonMessageConverter.java
@@ -65,6 +65,15 @@ public class Jackson2JsonMessageConverter extends AbstractJsonMessageConverter i
 	}
 
 	/**
+	 * Return the default json object mapper.
+	 *
+	 * @return {@link com.fasterxml.jackson.databind.ObjectMapper}
+	 */
+	protected ObjectMapper getJsonObjectMapper() {
+		return jsonObjectMapper;
+	}
+
+	/**
 	 * The {@link com.fasterxml.jackson.databind.ObjectMapper} to use instead of using the default. An
 	 * alternative to injecting a mapper is to extend this class and override
 	 * {@link #initializeJsonObjectMapper()}.
--- a/spring-amqp/src/main/java/org/springframework/amqp/support/converter/Jackson2JsonMessageConverter.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/support/converter/Jackson2JsonMessageConverter.java
@@ -70,7 +70,7 @@ public class Jackson2JsonMessageConverter extends AbstractJsonMessageConverter i
 	 * @return {@link com.fasterxml.jackson.databind.ObjectMapper}
 	 */
 	protected ObjectMapper getJsonObjectMapper() {
-		return jsonObjectMapper;
+		return this.jsonObjectMapper;
 	}
 
 	/**


pivotal-issuemaster:
@calvernaz Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-amqp&pullRequestId=590)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-amqp&pullRequestId=590) to manually synchronize the status of this Pull Request.

artembilan:
@calvernaz,

Please, come back to us with a signed CLA.
Otherwise we can't review and accept your PR.

I understand that it might seem like an "obvious fix", but at a glance it is really with some flaws and would be really better to pass `objectMapper` as an argument to the `initializeJsonObjectMapper()`.
I'd even rename it into `initializeDefaultJsonObjectMapper()`, because that won't be good to let to mutate the provided `ObjectMapper`.
Although the getter must be there anyway.

Thank you for your contribution effort!

calvernaz:
I signed it already, I got the message:

_Thank you for signing the Agreement! The contributor license agreement checks on spring-projects/spring-amqp/pull/590 should now pass._

Yea, I thought about that suggestion, but I had doubts about changing the API and decided to go with the protected getter to leave just enough room for access.

Saying that I can follow up and accommodate all those suggestions.

pivotal-issuemaster:
@calvernaz Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-amqp&pullRequestId=590)!

artembilan:
OK! After manual PR syncing we really see a confirmation.

Right that looks like a breaking change, but my concern to let end user to know that we configure with that method only a default `ObjectMapper` instance. 
If you provide an external one, that is already your responsibility to configure it there.
We can't mutate an external instance unconditionally and changing the API that should be much cleaner now.

I won't mind any other opinion.

artembilan:
@garyrussell ,

Your opinion on the matter?
This is the last community PR (and its issue) we should consider for tomorrow's release.
Thanks.

garyrussell:
@artembilan I am not sure what you mean by "breaking change" - all we're doing is adding a getter to allow a subclass to access the field rather than replacing the default mapper.

The structure of this class is a little odd; normally you should only call final methods from a ctor, explicitly to prevent subclasses from changing expected behavior.

Personally, I would make the mapper final and add another ctor that takes a fully assembled mapper. We could deprecate the setter in 1.7.x and change the javadoc to make it clear the `initialize...` method is only called from the default ctor.

Another alternative would be to deprecate that initialize method and add a new one that takes the mapper as an argument.

artembilan:
Thank you, @garyrussell , for your vision!

Well, at least we are on the same page about the oddity of this class 😄 

How about this: https://github.com/artembilan/spring-amqp/commit/99bc3b95d42b209dcced7c8dda583d5bad2110c7 ?

I didn't make it as a `final` do not break existing extensions. But this might be candidate really for `1.7.x`.
In the `2.0` we can remove those deprecated methods and make the property `final`.

IMO the class isn't designed for extension and it doesn't aim that goal by the purpose.
So, simple injection of externally configured `ObjectMapper` should be as a silver bullet.

garyrussell:
LGTM (with javadocs on the ctors that is 😄 )

artembilan:
> LGTM

Sorry, not enough info. 
Does that mean that we apply it as is for `master`? Or what I did is for `1.7.x`? The `master` version should be provided.
Thanks

garyrussell:
Sorry; yes; I meant apply this with deprecations to 1.7.x and remove them in 2.0.

artembilan:
Good, thank you, Gary, for confirmation!

@calvernaz ,

Sorry, since you see a discussion, this PR becomes unrelated.

Please, close it and let us know if you are good to provide a couple new:
1. For `1.7.x` with my change in that commit
2. For `master` with similar change but with removed `@Deprecated` methods and `final` for the `objectMapper` property.

We need it for tomorrow: the release obligations.

Otherwise let us know and we do that for you.

There are still enough tasks for contribution! 😄

calvernaz:
I've been following along the discussion and the constructor option is my favorite solution for this one.

artembilan:
Superseded with the https://github.com/spring-projects/spring-amqp/commit/710464367e15c9287e4f83de40b833c0e632d66c for `1.7.x` and https://github.com/spring-projects/spring-amqp/commit/13e0b7621a93124938a3e7f8f8ed0ef0b60ca8bb for the `master`.

PR #591, opened 2017-03-15
AMQP-727: RoutingConnectionFactory Fixes

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-727

- Bind the lookup key to any thread in the container that might open a connection
  so that the same connection factory will be used if a connection listener is invoked.
- Add a lookup key qualifier to the container so that the same queue name(s) can be used in
  different virtual hosts.

__cherry-pick to 1.7.x, 1.6.x (minus the DMLC bits)__

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
@@ -197,6 +197,8 @@ public abstract class AbstractMessageListenerContainer extends RabbitAccessor
 
 	private boolean alwaysRequeueWithTxManagerRollback;
 
+	private String lookupKeyQualifier = "";
+
 	/**
 	 * {@inheritDoc}
 	 * @since 1.5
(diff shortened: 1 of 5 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -437,6 +433,24 @@ protected void actualStart() throws Exception {
 		}
 	}
 
+	protected void doRedeclareElementsIfNecessary() {
+		String routingLookupKey = getRoutingLookupKey();
+		if (routingLookupKey != null) {
+			SimpleResourceHolder.bind(getRoutingConnectionFactory(), routingLookupKey);
```
artembilan: Well, actually only `SimpleRoutingConnectionFactory` deals with the `SimpleResourceHolder` maybe that is a bit expensive to store to `ThreadLocal` always unconditionally ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -538,6 +540,43 @@ public ConnectionFactory getConnectionFactory() {
...
+	 * @see #getRoutingLookupKey()
+	 */
+	public void setLookupKeyQualifier(String lookupKeyQualifier) {
+		this.lookupKeyQualifier = lookupKeyQualifier;
+	}
+
+	/**
+	 * Return the lookup key if the connection factory is a
+	 * {@link RoutingConnectionFactory}; null otherwise. The routing key is the
+	 * comma-delimited list of queue names will all spaces removed and bracketed by [...],
```
garyrussell: Spaces is correct here (or perhaps space characters?)

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
garyrussell: True, but I think any meaningful implementation must use a thread local - there is no other way to pass context into the create connection request.

I am not too concerned about the overhead here - it's only done during initialization/restart. The concern about the SI poller last week was because it was applied to every poll.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
artembilan: Yeah. sorry. I meant s/will/with

That why my "with" is bold.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
artembilan: Ah! I see now. Fine with me.
Thank you for explanation!

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
garyrussell: Ah; yes `s/will/with/` 😄

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/0f2c23d0bb0717e65b0366288943bd496fc40523 and cherry-picked to `1.7.x` as https://github.com/spring-projects/spring-amqp/commit/0f2c23d0bb0717e65b0366288943bd496fc40523 and to `1.6.x` as https://github.com/spring-projects/spring-amqp/commit/95c73690fcfb0adfce0a483495f5098c14f2fe08

Issue #592, opened 2017-03-16
Cannot set another host name BrokerRunning 


SirWayne:
It's hard to set another hostname for the BrokerRunning rule.
For the connection factory it's possible to use a static block.

```
    @ClassRule
    public static BrokerRunning brokerRunning = BrokerRunning.isRunning();
    static {
        brokerRunning.setHostName("abc");
    }
```

but the host name for the management client is hardcoded

`
Client client = new Client("http://localhost:15672/api/", "guest", "guest");
`

garyrussell:
Thanks for pointing this out - I raised it as a bug [AMQP-728](https://jira.spring.io/browse/AMQP-728).

PR #593, opened 2017-03-16
AMQP-728: Make BrokerRunning More Configurable

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-728
Resolves: #592

Add more options to configure the `BrokerRunning` JUnit `@Rule`.

Fix missing exclude from build.gradle for the junit subproject.

__cherry-pick to 1.7.x - needs diamond operator fixups__.

### DIFF
--- a/spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/BrokerRunning.java
+++ b/spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/BrokerRunning.java
@@ -79,17 +79,33 @@ import com.rabbitmq.http.client.Client;
  */
 public final class BrokerRunning extends TestWatcher {
 
+	public static final String BROKER_ADMIN_URI = "RABBITMQ_TEST_ADMIN_URI";
+
+	public static final String BROKER_HOSTNAME = "RABBITMQ_TEST_HOSTNAME";
+
+	public static final String BROKER_PORT = "RABBITMQ_TEST_PORT";
+
+	public static final String BROKER_USER = "RABBITMQ_TEST_USER";
+
+	public static final String BROKER_PW = "RABBITMQ_TEST_PASSWORD";
+
+	public static final String BROKER_ADMIN_USER = "RABBITMQ_TEST_ADMIN_USER";
+
+	public static final String BROKER_ADMIN_PW = "RABBITMQ_TEST_ADMIN_PASSWORD";
+
 	public static final String BROKER_REQUIRED = "RABBITMQ_SERVER_REQUIRED";
 
 	private static final String DEFAULT_QUEUE_NAME = BrokerRunning.class.getName();
 
-	private static Log logger = LogFactory.getLog(BrokerRunning.class);
+	private static final Log logger = LogFactory.getLog(BrokerRunning.class);
 
 	// Static so that we only test once on failure: speeds up test suite
-	private static Map<Integer, Boolean> brokerOnline = new HashMap<Integer, Boolean>();
+	private static final Map<Integer, Boolean> brokerOnline = new HashMap<Integer, Boolean>();
 
 	// Static so that we only test once on failure
-	private static Map<Integer, Boolean> brokerOffline = new HashMap<Integer, Boolean>();
+	private static final Map<Integer, Boolean> brokerOffline = new HashMap<Integer, Boolean>();
+
+	private static final Map<String, String> environmentOverrides = new HashMap<>();
 
 	private final boolean assumeOnline;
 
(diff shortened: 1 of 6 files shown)


spring-rabbit-junit/src/test/java/org/springframework/amqp/rabbit/junit/BrokerRunningTests.java
```diff
@@ -0,0 +1,79 @@
...
+	@Test
+	public void testEnvironmentVars() {
+		Map<String, String> vars = new HashMap<>();
+		vars.put("RABBITMQ_TEST_ADMIN_PASSWORD", "FOO");
+		vars.put("RABBITMQ_TEST_ADMIN_URI", "http://foo/bar");
+		vars.put("RABBITMQ_TEST_ADMIN_USER", "BAR");
+		vars.put("RABBITMQ_TEST_HOSTNAME", "BAZ");
+		vars.put("RABBITMQ_TEST_PASSWORD", "QUX");
+		vars.put("RABBITMQ_TEST_PORT", "2345");
+		vars.put("RABBITMQ_TEST_USER", "FIZ");
```
artembilan: I guess we can use constants from the `BrokerRunning` for above definitions, but yeah, the ENV can't use them any way.
So, this is fine for demonstration the target names 😄

spring-rabbit-junit/src/test/java/org/springframework/amqp/rabbit/junit/BrokerRunningTests.java
```diff
@@ -0,0 +1,79 @@
...
+		BrokerRunning.setEnvironmentVariables(vars);
+		BrokerRunning brokerRunning = BrokerRunning.isBrokerAndManagementRunning();
```
artembilan: This doesn't looks like a robust solution from end-user perspective.
There is no guarantee that `setEnvironmentVariables()` will be called before `isRunning()`.
How about an overloaded `isRunning()` variant which accepts that `environmentVariables` Map?
More over I don't see code which clean that `static environmentOverrides` after testing...

spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/BrokerRunning.java
```diff
@@ -192,6 +241,105 @@ public void setHostName(String hostName) {
 		this.hostName = hostName;
 	}
 
+	/**
+	 * Set the user for the amqp connection default "guest".
+	 * @param user the user.
+	 * @since 1.7.2
```
artembilan: Well, OK: this change is for back porting any way.
So, test must be fixed appropriately.

spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/BrokerRunning.java
```diff
@@ -192,6 +241,105 @@ public void setHostName(String hostName) {
...
+	 * Set the user for the management REST API connection default "guest".
+	 * @param user the user.
+	 * @since 1.7.2
+	 */
+	public void setAdminUser(String user) {
+		this.adminUser = user;
+	}
+
+	/**
+	 * Set the paswword for the management REST API connection default "guest".
```
artembilan: Typo: s/paswword/password

spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/BrokerRunning.java
```diff
@@ -389,10 +537,29 @@ public ConnectionFactory getConnectionFactory() {
...
+	 */
+	public String getAdminUri() {
+		if (StringUtils.hasText(this.adminUri)) {
+			return this.adminUri;
+		}
+		else if (!StringUtils.hasText(this.hostName)) {
+			return "http://localhost:15672/api/";
+		}
+		else {
+			return "http://" + this.hostName + ":15672/api/";
```
artembilan: Can't we cache the result of calculations of this method into the `adminUri`?

spring-rabbit-junit/src/test/java/org/springframework/amqp/rabbit/junit/BrokerRunningTests.java
garyrussell: > will be called before isRunning()

It doesn't have to be - just has to be before the first test is run. The actual connection is made in `apply()`.

> More over I don't see code which clean that static environmentOverrides after testing...

That's the whole point - it allows you to configure the environment just once, for a bunch of test classes.

spring-rabbit-junit/src/test/java/org/springframework/amqp/rabbit/junit/BrokerRunningTests.java
garyrussell: I explicitly used literals not constants so I could test the actual values of the constants - in fact I found a bug in one of the constants this way (double underscore).

spring-rabbit-junit/src/test/java/org/springframework/amqp/rabbit/junit/BrokerRunningTests.java
```diff
@@ -0,0 +1,79 @@
...
+
+import org.junit.Test;
+
+import org.springframework.beans.DirectFieldAccessor;
+
+import com.rabbitmq.client.ConnectionFactory;
+
+/**
+ * @author Gary Russell
+ * @since 5.0
```
garyrussell: Oops; yes 1.7.2 (auto comment from IDE).

spring-rabbit-junit/src/test/java/org/springframework/amqp/rabbit/junit/BrokerRunningTests.java
garyrussell: Oh, I see what you mean. I don't want to add a bunch more overloaded statics - we should just document that it must be called first.

artembilan: I can agree about the point to set once, but how about to let to reset any way?

But at the same time I can't agree for your first comment.
Just swap these two lines and see how test works!

Looks what we have there:
```
public static BrokerRunning isRunning() {
	return new BrokerRunning(true);
}
```
So, we create a new instance and initialize its properties from the ENV, not `environmentOverrides`, because we don't have them yet.

artembilan: OK. Let's document for now and come back to the proper fix on demand 😄 

I won't mind if there is better solution then overloaded methods

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/38f38c4d35b27d3f190a848f4e24b6ed0a3c6aa1 and cherry-picked.

PR #595, opened 2017-03-17
Synchronize access to DMLC.consumers in scheduled task

panchenko:
The access to this field is synchronized in other places, only missing here.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
@@ -331,9 +331,13 @@ public class DirectMessageListenerContainer extends AbstractMessageListenerConta
 					this.lastAlertAt = now;
 				}
 			}
-			this.consumers.stream()
-					.filter(c -> !c.getChannel().isOpen())
-					.collect(Collectors.toList()) // needed to avoid ConcurrentModificationException in cancelConsumer()
+			final List<SimpleConsumer> consumersToCancel;
+			synchronized (this.consumersMonitor) {
+				consumersToCancel = this.consumers.stream()
+						.filter(c -> !c.getChannel().isOpen())
+						.collect(Collectors.toList());
+			}
+			consumersToCancel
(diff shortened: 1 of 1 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -331,9 +331,13 @@ protected void actualStart() throws Exception {
 					this.lastAlertAt = now;
 				}
 			}
-			this.consumers.stream()
-					.filter(c -> !c.getChannel().isOpen())
-					.collect(Collectors.toList()) // needed to avoid ConcurrentModificationException in cancelConsumer()
+			final List<SimpleConsumer> consumersToCancel;
+			synchronized (this.consumersMonitor) {
+				consumersToCancel = this.consumers.stream()
```
artembilan: What makes you think that we really should synchronize on this read only stream operation?
Thanks

panchenko: Synchronization is needed to be sure that this operation succeeds even if other threads try to modify it. Please give a look at the class [documentation](https://docs.oracle.com/javase/7/docs/api/java/util/LinkedList.html).

artembilan: OK. Thank you @panchenko !

According to the `Spliterator` JavaDocs (the result of `Collection.stream()`):
```
 * After binding a Spliterator should, on a best-effort basis, throw
 * {@link ConcurrentModificationException} if structural interference is
 * detected.  Spliterators that do this are called <em>fail-fast</em>.
```
When we have similar description on `LinkedList`:
```
The iterators returned by this class's iterator and listIterator methods are fail-fast: if the list is structurally modified at any time after the iterator is created, in any way except through the Iterator's own remove or add methods, the iterator will throw a ConcurrentModificationException. 
```

So, any `iterator` requires strong modification protection during iteration.
Therefore the fix is fine.
Merging...

PR #599, opened 2017-03-28
support custom exchange in @RabbitListener

### LINKED ISSUE #598: Not support binding a queue to a custom Exchange whth Annotation
When binding a queue to the rabbitmq_lvc  exchange,
```
@RabbitListener(
            bindings = @QueueBinding(
                    value = @Queue(),
                    exchange = @Exchange(
                            value = "lvc",
                            type = "'x-lvc'",//ExchangeTypes.TOPIC,
                            durable = "true"
                    ),
                    key="test"
            )
    )
```
An exception will be thrown.
```
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'myBean' defined in file [D:\git\gs-messaging-rabbitmq\complete\target\classes\hello\MyBean.class]: Initialization of bean failed; nested exception is org.springframework.beans.factory.BeanInitializationException: Unexpected exchange type: 'x-lvc'
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:564) ~[spring-beans-4.3.7.RELEASE.jar:4.3.7.RELEASE]
	.................................
Caused by: org.springframework.beans.factory.BeanInitializationException: Unexpected exchange type: 'x-lvc'
	at org.springframework.amqp.rabbit.annotation.RabbitListenerAnnotationBeanPostProcessor.declareExchangeAndBinding(RabbitListenerAnnotationBeanPostProcessor.java:598) ~[spring-rabbit-1.7.1.RELEASE.jar:na]
```

[RabbitListenerAnnotationBeanPostProcessor.java](https://github.com/spring-projects/spring-amqp/blob/master/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java#L595)

panchenko:
Fixes #598

### DIFF
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessorTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessorTests.java
@@ -31,6 +31,7 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.Iterator;
@@ -40,6 +41,7 @@ import java.util.stream.Collectors;
 import org.junit.Test;
 
 import org.springframework.amqp.core.Binding;
+import org.springframework.amqp.core.CustomExchange;
 import org.springframework.amqp.core.Queue;
 import org.springframework.amqp.rabbit.config.MessageListenerTestContainer;
 import org.springframework.amqp.rabbit.config.RabbitListenerContainerTestFactory;
@@ -231,6 +233,20 @@ public class RabbitListenerAnnotationBeanPostProcessorTests {
 		context.close();
 	}
 
+	@Test
+	public void customExhangeTestBean() {
+		ConfigurableApplicationContext context = new AnnotationConfigApplicationContext(Config.class,
+				CustomExchangeTestBean.class);
+
+		final Collection<CustomExchange> exchanges = context.getBeansOfType(CustomExchange.class).values();
+		assertThat(exchanges, hasSize(1));
+		final CustomExchange exchange = exchanges.iterator().next();
+		assertEquals("my_custom_exchange", exchange.getName());
+		assertEquals("custom_type", exchange.getType());
+
+		context.close();
+	}
+
 	@Component
 	static class SimpleMessageListenerTestBean {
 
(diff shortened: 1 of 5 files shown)


panchenko:
If we agree on this, I can file it in JIRA and provide separate branch for 1.7

hyhy01:
Thank you,it works in the spring boot 2.0

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessorTests.java
```diff
@@ -231,6 +233,22 @@ public void multipleRoutingKeysTestBean() {
...
+
+		final Collection<org.springframework.amqp.core.Exchange> exchanges = context
+				.getBeansOfType(org.springframework.amqp.core.Exchange.class).values();
+		assertThat(exchanges, hasSize(1));
+		assertThat(exchanges.iterator().next(), instanceOf(CustomExchange.class));
+		final CustomExchange customExchange = (CustomExchange) exchanges.iterator().next();
+		assertEquals("my_custom_exchange", customExchange.getName());
+		assertEquals("custom_type", customExchange.getType());
+
+		context.close();
```
artembilan: From big height the fix looks reasonable for me.
But does it work really against real Broker?
I'm just curious if that `CustomExchange` is robust.

But, yeah, thank you @panchenko for the effort to support community and improve the Framework!

garyrussell:
Thanks @panchenko - needs a doco update [here](http://docs.spring.io/spring-amqp//reference/html/_reference.html#async-annotation-driven)...

>https://github.com/spring-projects/spring-amqp/blame/master/src/reference/asciidoc/amqp.adoc#L1652

Also javadocs for the `@Exchange.type()`.

garyrussell:
My only concern is this prevents catching typos in exchange types but I suppose that will be obvious when the broker returns an error (and the benefit outweighs that).

artembilan:
Checkstyle failure on the Travis:
```
:spring-rabbit:checkstyleMain[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-amqp/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/Exchange.java:52: Line matches the illegal pattern 'Trailing whitespace'. [Regexp]
```
You can run `./gradlew check` locally instead of `test` before pushing. The `check` task includes `test` in its execution chain: https://docs.gradle.org/3.4.1/userguide/java_plugin.html#sec:java_tasks

Although, frankly, we are also lazy and push to Travis to check the commit :wink 
But, yeah, that isn't good practice, I admit...

panchenko:
I've updated documentation and filed [AMQP-731](https://jira.spring.io/browse/AMQP-731).
Should this also go into 1.7.x?

garyrussell:
Thanks @panchenko; yes, it should go in 1.7.x; but we should be able to cherry-pick this commit, so no need for another PR.

artembilan:
IMO we can't backport it into `1.7.x`. This might be some breaking change for end-users who rely on the `BeanInitializationException("Unexpected exchange type: " + exchangeType)`. 

OTOH the workaround is pretty straightforward - just declare `@Bean` s for `CustomExchange` and `Binding` & `Queue`.

(That's why I was always against such an annotation-based Broker configuration...)

artembilan:
Another "IMHO" since it is still seating on top of my head:
We don't need a JIRA if we have a GH issue, and we don't need a GH issue if we have a PR already.
Unless some special case, when PR might be rejected. 
The single place for discussion is fully enough.
We need JIRA/issue do not forget the feature/bug or when we don't know the solution yet.

garyrussell:
While I would say that's true for small changes like this, for more important changes we should always create a JIRA issue, given that we generally refer to the JIRA release notes (list of closed issues) in our release announcements.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/e020482b59012bcdad3f8e3c25eb679f8f168066

@panchenko , thank you!

PR #601, opened 2017-03-29
AMQP-733: @Exchange.durable() to true by default

artembilan:
JIRA: https://jira.spring.io/browse/AMQP-733

To be consistent with default `AbstractExchange` ctor,
when by default exchanges must be `durable`, make `@Exchange.durable()`
true by default

**Cherry-pick to 1.7.x**

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/Exchange.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/Exchange.java
@@ -58,9 +58,9 @@ public @interface Exchange {
 	String type() default ExchangeTypes.DIRECT;
 
 	/**
-	 * @return true if the exchange is to be declared as durable.
+	 * @return false if the exchange is to be declared as non-durable.
 	 */
-	String durable() default "false";
+	String durable() default "true";
 
 	/**
 	 * @return true if the exchange is to be declared as auto-delete.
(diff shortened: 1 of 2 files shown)


garyrussell:
We can't cherry-pick - recall that when we backported the `ExchangeBuilder` fix, we had to set the default to `false`.

garyrussell:
Needs a What's new; we can add a sentence to the existing

> The ExchangeBuilder now builds durable exchanges by default. See the section called “Builder API for Queues and Exchanges” for more information.

If you agree, I can fix on merge.

artembilan:
Sure thing!

And workaround isn't tough 😄 

Thanks for pointing that out!

artembilan:
>  I can fix on merge.

👍

garyrussell:
Merged as 7cd58daf2a01772c1ba61358b271ea8d71361495

panchenko:
What about `@Queue.durable()` then? 
default is declared as empty string and in code evaluates to false, but `new Queue("foo")` has durable=true.

artembilan:
@panchenko ,

I think that is a good catch!

Would you mind providing the fix?
We are going to release `2.0 M3` today. 

I mean we can wait for a couple hours and then fix it ourselves if no answer.

Thank you!

panchenko:
@artembilan I would like to, but have time only in an hour.

PR #603, opened 2017-04-03
micro-optimization: use parseBoolean()

panchenko:


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -699,7 +699,7 @@ private boolean resolveExpressionAsBoolean(String value, boolean defaultValue) {
 		}
 		else if (resolved instanceof String) {
 			final String s = (String) resolved;
-			return StringUtils.hasText(s) ? Boolean.valueOf(s) : defaultValue;
+			return StringUtils.hasText(s) ? Boolean.parseBoolean(s) : defaultValue;
```
artembilan: This doesn't add much value because `RabbitListenerAnnotationBeanPostProcessor` is performed only once during application start up.
But since you have done this change, I'll just go ahead and merge 😄

PR #604, opened 2017-04-03
AMQP-670: Broker-less @RabbitListener Testing

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-670

Initial PoC.

`TestRabbitTemplate` invokes the listener on the calling test thread.

### DIFF
--- a/spring-rabbit-test/src/main/java/org/springframework/amqp/rabbit/test/TestRabbitTemplate.java
+++ b/spring-rabbit-test/src/main/java/org/springframework/amqp/rabbit/test/TestRabbitTemplate.java
@@ -1,0 +1,130 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.amqp.rabbit.test;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyBoolean;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.BDDMockito.willAnswer;
+import static org.mockito.Mockito.mock;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicReference;
+
+import org.springframework.amqp.core.Message;
+import org.springframework.amqp.core.MessageBuilder;
+import org.springframework.amqp.rabbit.connection.ConnectionFactory;
+import org.springframework.amqp.rabbit.core.RabbitTemplate;
+import org.springframework.amqp.rabbit.listener.AbstractMessageListenerContainer;
+import org.springframework.amqp.rabbit.listener.RabbitListenerEndpointRegistry;
+import org.springframework.amqp.rabbit.listener.adapter.MessagingMessageListenerAdapter;
+import org.springframework.amqp.rabbit.support.CorrelationData;
+import org.springframework.amqp.rabbit.support.RabbitExceptionTranslator;
+import org.springframework.beans.factory.SmartInitializingSingleton;
+import org.springframework.beans.factory.annotation.Autowired;
+
+import com.rabbitmq.client.AMQP.BasicProperties;
+import com.rabbitmq.client.Channel;
+import com.rabbitmq.client.Envelope;
+
+/**
+ * A {@link RabbitTemplate} that invokes {@code @RabbitListener} s directly.
+ * It currently only supports the queue name in the routing key.
+ * It does not currently support publisher confirms/returns.
+ *
+ * @author Gary Russell
+ * @since 2.0
+ *
+ */
+public class TestRabbitTemplate extends RabbitTemplate implements SmartInitializingSingleton {
+
+	private static final String REPLY_QUEUE = "testRabbitTemplateReplyTo";
+
+	private final Map<String, MessagingMessageListenerAdapter> listeners = new HashMap<>();
+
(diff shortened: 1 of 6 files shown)


spring-rabbit-test/src/main/java/org/springframework/amqp/rabbit/test/TestRabbitTemplate.java
```diff
@@ -0,0 +1,130 @@
...
+	@Override
+	public void afterSingletonsInstantiated() {
+		registry.getListenerContainers()
+			.stream()
+			.map(c -> (AbstractMessageListenerContainer) c)
+			.forEach(c -> {
+				for (String queue : c.getQueueNames()) {
+					listeners.put(queue, (MessagingMessageListenerAdapter) c.getMessageListener());
+				}
+				// TODO: support multiple listeners for the same queue(s)? That's a bit odd with @RabbitListener though.
```
artembilan: Well, that's why we all bumped to the fact to support all that Broker's binding tricks 😄 

I think in this simple test way it might be really useful to expect all the listener invoked if they are subscribed to the same queue.

OTOH this might be optional, e.g. `TestRabbitTemplate.setMultiListeners(true/false)`.
But that way `TestRabbitTemplate` should not be claimed as a bean. We should be able just use it from the test code.
If use decides to make it as a bean, that his choice.

Or add `multi` to the `send()` method of this `TestRabbitTemplate`...

Another observation: this `TestRabbitTemplate` should support any `ListenerContainer` in the application context. Build directly or via XML config. As well as Spring Integration's `AmqpInboundChannelAdapter` and `AmqpInboundGateway`

WDYT?

garyrussell:
Pushed support for fanout and other containers.

Be gentle - this is not the final solution, just some building blocks; we can add support for other exchange types (topic etc) later.

artembilan:
There is no issues for me at a glance for the current state.
If you want to have it in the tomorrow's release, I can review thoroughly locally and merge.
Just let me know!

garyrussell:
OK; I will add some docs.

/cc: @mduesterhoeft

mduesterhoeft:
👍 This looks really promising - I really like that we are coming closer to Broker-less testing with spring-amqp.

My thoughts on some details of the PR:

**broadcast flag**
I do not really understand what the `broadcast` flag is good for. I think I would always like my message to go to all the listeners that are subscribed. It would be hard for a user to figure out what is happening if there are multiple listeners subscribed to the same queue but only one receives the message

**binding and routing**
I have a little proposal here on how to enhance the binding logic. https://github.com/mduesterhoeft/spring-amqp/tree/AMQP-670-bindings
I implemented a [MessageListenerAccessor](https://github.com/mduesterhoeft/spring-amqp/blob/AMQP-670-bindings/spring-rabbit-test/src/main/java/org/springframework/amqp/rabbit/test/MessageListenerAccessor.java) that takes the `Bindings`s on the application context into account and decides, based on the exchange a message is sent to, to which listener the message should be delivered to. It is matching the binding queue against the listener queue. I think it would need to be enhanced by taking also the routing key into account - the current state of the branch is just to show the idea of the approach. Do you think this is feasible?

garyrussell:
I don't really like the broadcast flag either (but for different reasons 😄 ) - I added it because @artembilan Artem asked for it and I didn't feel like pushing back - it doesn't really simulate a real rabbit feature - multiple consumers on the same queue never get duplicated messages.

When it's false, we round-robin to the listeners, just like rabbit would.

Yes, I definitely think we can incrementally add features - such as detecting exchanges and bindings in the context and build up routing to simulate the broker. It's just a lot more work and I wanted to get something simple into tomorrow's milestone so people can chew on it. We're definitely interested in your collaboration; we have another milestone due at the end of the month, for boot 2.0.0.M1, with a release candidate towards the end of May and GA a few weeks after that.

artembilan:
That's cool, @mduesterhoeft !

I definitely don't see problems to consume your solution here 😄 

I think we definitely might come up with the solution to consult `routingKey` as well.
Since we have a `Binding` in hands, why don't check its `getRoutingKey()`?

Thank you very much for the effort!

artembilan:
> I don't really like the broadcast flag either

Oh! Let's remove it! I don't mind. I more like `routingKey` selection based on the binding.
I mentioned that in JIRA anyway: 
> If the application doesn't provide Binding, we always can supply additional `@Configuration` (or XML) for test suite.

spring-rabbit-test/src/main/java/org/springframework/amqp/rabbit/test/TestRabbitTemplate.java
```diff
@@ -0,0 +1,200 @@
...
+			.forEach(container -> {
+				for (String queue : container.getQueueNames()) {
+					setupListener(container, queue);
+				}
+			});
+	}
+
+	private void setupListener(AbstractMessageListenerContainer container, String queue) {
+		Listeners listenersForQueue = this.listeners.get(queue);
+		if (listenersForQueue == null) {
```
panchenko: A more functional way: `computeIfAbsent(...).listeners.add(...)`

spring-rabbit-test/src/main/java/org/springframework/amqp/rabbit/test/TestRabbitTemplate.java
```diff
@@ -0,0 +1,200 @@
...
+
+		private final List<Object> listeners = new ArrayList<>();
+
+		private volatile Iterator<Object> iterator;
+
+		Listeners(Object listener) {
+			this.listeners.add(listener);
+		}
+
+		private Object next() {
```
panchenko: I would suggest making this method synchronized

mduesterhoeft:
@artembilan @garyrussell I totally agree going step by step here - I just wanted to put my implementation idea into the discussion. And yes - if the rabbit artifacts like binding, queue, and exchange are declared outside the application we are basically out of luck with my proposal. But then we can still ask a user to provide 'Exchange', 'Binding' and 'Queue' bean definitions in a TestConfiguration.

spring-rabbit-test/src/main/java/org/springframework/amqp/rabbit/test/TestRabbitTemplate.java
artembilan: @garyrussell , are you going to address this concern?
Thanks

@panchenko , thank you very much for review!

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/308c8690442c9db12f58d61b5ae9e57b95ac2251

PR #606, opened 2017-04-05
fix concurrency in RabbitListenerAnnotationBeanPostProcessor

panchenko:
Today I've noticed NPE in our logs in `RabbitListenerAnnotationBeanPostProcessor.postProcessAfterInitialization()`
on the `if (multiMethods.size() > 0) {`line.

Some beans in the application are declared lazy, so they are created on demand, and that happens concurrently.

There are some things to do and discuss:
- [x] https://jira.spring.io/browse/AMQP-735
- [x] how to name inner classes
- [x] if `findMethods` should be renamed
- [x] where the new test should go
- [ ] which branches should be fixed? In our system it was 1.6.x, with the test reproduced on master, logically 1.7.x is also affected.
- [ ] what should be in the commit message, as class name is quite long :-)

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
@@ -20,6 +20,7 @@ import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -267,13 +268,13 @@ public class RabbitListenerAnnotationBeanPostProcessor
 	public Object postProcessAfterInitialization(final Object bean, final String beanName) throws BeansException {
 		Class<?> targetClass = AopUtils.getTargetClass(bean);
 		final TypeMetadata metadata = this.typeCache.computeIfAbsent(targetClass, this::buildMetadata);
-		for (ListenerMethod lm : metadata.methods) {
-			for (RabbitListener rabbitListener : lm.listenerAnnotations) {
+		for (ListenerMethod lm : metadata.listenerMethods) {
+			for (RabbitListener rabbitListener : lm.annotations) {
 				processAmqpListener(rabbitListener, lm.method, bean, beanName);
 			}
 		}
-		if (metadata.multiMethods.length > 0) {
-			processMultiMethodListeners(metadata.classLevelListeners, metadata.multiMethods, bean, beanName);
+		if (metadata.handlerMethods.length > 0) {
+			processMultiMethodListeners(metadata.classAnnotations, metadata.handlerMethods, bean, beanName);
 		}
 		return bean;
 	}
@@ -316,7 +317,7 @@ public class RabbitListenerAnnotationBeanPostProcessor
 		}
 		RabbitListeners anns = AnnotationUtils.findAnnotation(clazz, RabbitListeners.class);
 		if (anns != null) {
-			listeners.addAll(Arrays.asList(anns.value()));
+			Collections.addAll(listeners, anns.value());
 		}
 		return listeners;
 	}
@@ -332,7 +333,7 @@ public class RabbitListenerAnnotationBeanPostProcessor
 		}
 		RabbitListeners anns = AnnotationUtils.findAnnotation(method, RabbitListeners.class);
 		if (anns != null) {
-			listeners.addAll(Arrays.asList(anns.value()));
+			Collections.addAll(listeners, anns.value());
 		}
 		return listeners;
 	}
@@ -762,32 +763,44 @@ public class RabbitListenerAnnotationBeanPostProcessor
 	}
 
 	private static class TypeMetadata {
-		final ListenerMethod[] methods;
-		final Method[] multiMethods;
-		final RabbitListener[] classLevelListeners;
+		/**
+		 * Methods annotated with {@link RabbitListener}.
+		 */
+		final ListenerMethod[] listenerMethods;
+		/**
+		 * Methods annotated with {@link RabbitHandler}.
+		 */
+		final Method[] handlerMethods;
+		/**
+		 * Class level {@link RabbitListener} annotations.
+		 */
+		final RabbitListener[] classAnnotations;
 
 		static final TypeMetadata EMPTY = new TypeMetadata();
 
 		private TypeMetadata() {
-			this.methods = new ListenerMethod[0];
-			this.multiMethods = new Method[0];
-			this.classLevelListeners = new RabbitListener[0];
+			this.listenerMethods = new ListenerMethod[0];
+			this.handlerMethods = new Method[0];
+			this.classAnnotations = new RabbitListener[0];
 		}
 
 		TypeMetadata(ListenerMethod[] methods, Method[] multiMethods, RabbitListener[] classLevelListeners) {
-			this.methods = methods;
-			this.multiMethods = multiMethods;
-			this.classLevelListeners = classLevelListeners;
+			this.listenerMethods = methods;
+			this.handlerMethods = multiMethods;
+			this.classAnnotations = classLevelListeners;
 		}
 	}
 
+	/**
+	 * A method annotated with {@link RabbitListener}, together with the annotations.
+	 */
 	private static class ListenerMethod {
 		final Method method;
-		final RabbitListener[] listenerAnnotations;
+		final RabbitListener[] annotations;
 
 		ListenerMethod(Method method, RabbitListener[] annotations) {
 			this.method = method;
-			this.listenerAnnotations = annotations;
+			this.annotations = annotations;
 		}
 	}
 }
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
@@ -20,6 +20,7 @@ import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


panchenko:
Part of the original exception with version 1.6.8:

```
Caused by: java.lang.NullPointerException: null
at org.springframework.amqp.rabbit.annotation.RabbitListenerAnnotationBeanPostProcessor.postProcessAfterInitialization(RabbitListenerAnnotationBeanPostProcessor.java:296)
at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.applyBeanPostProcessorsAfterInitialization(AbstractAutowireCapableBeanFactory.java:423)
at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.initializeBean(AbstractAutowireCapableBeanFactory.java:1594)
at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:554)
```

garyrussell:
I haven't absorbed the totality of the changes, but I can comment on some of the questions:

JIRA: yes
findMethods name: probably
branches: include 1.6, we should do a 1.6.9 release for boot 1.4.6 which is due next week
commit message: we usually abbreviate to `...BPP`.

I am taking off this afternoon so if @artembilan doesn't review, I'll take a look tomorrow.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -276,37 +266,28 @@ public Object postProcessBeforeInitialization(Object bean, String beanName) thro
...
-		List<RabbitListener> classLevelListeners = this.classAnnotationCache.get(targetClass);
-		List<Method> methods = this.methodCache.get(targetClass);
-		List<Method> multiMethods = this.multiMethodCache.get(targetClass);
-		for (Method method : methods) {
-			List<RabbitListener> listenerAnnotations = this.annotationCache.get(method);
-			if (listenerAnnotations != null) {
-				for (RabbitListener rabbitListener : listenerAnnotations) {
-					processAmqpListener(rabbitListener, method, bean, beanName);
-				}
+		final TypeMetadata metadata = this.typeCache.computeIfAbsent(targetClass, TypeMetadataHolder::new).get();
```
artembilan: We can't use `computeIfAbsent()` if we are going to back port it.
Or... we will need one more backport PR.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -779,4 +761,56 @@ private MessageHandlerMethodFactory createDefaultMessageHandlerMethodFactory() {
 
 	}
 
+	private class TypeMetadataHolder {
```
artembilan: Looks like we don't need this intermediate, singleton-aware holder.
We really just can `compute` `TypeMetadata` if it absent in the `typeCache` map.
I'd even wouldn't worry a lot about concurrent override since all the info from the `Class` is idempotent.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
panchenko: Another PR sounds possible :)

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
panchenko: The idea was to minimize time while the Map segment is locked, however I agree thread contention is not that high here.

artembilan: As I said: the target result is idempotent, so, I won't worry if my `TypeMetadata` is filled concurrently, even if we have some extra computation for nothing. But still it might be better then wait for another thread.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessorTests.java
```diff
@@ -273,6 +278,42 @@ public void queuesToDeclare() {
 		context.close();
 	}
 
+	@Test
+	public void concurrency() throws InterruptedException, ExecutionException {
+		final int concurrencyLevel = 8;
+		final ExecutorService executorService = Executors.newFixedThreadPool(concurrencyLevel);
+		try {
+			for (int i = 0; i < 1000; ++i) {
```
panchenko: @artembilan @garyrussell what do you think would be a better place for such a test?
Other tests here are take ~50ms, but this loop takes much longer. Is there a class with similar tests already, or should it be a new one?

artembilan: Since we test in this method really a concurrency of the `RabbitListenerAnnotationBeanPostProcessor`, there is no better place unless this test class.
I'll review the test locally thoroughly for performance.

garyrussell: I would create a new class `RabbitListenerAnnotationBeanPostProcessor2Tests` and add the `@Rule LongRunningIntegrationTest`.

See `SimpleMessageListenerContainerIntegration2Tests`.

garyrussell: We have a nightly build that runs all tests and usually, we as developers, run `./gradlew testAll` at least once before creating a PR.

artembilan: I have modified the test like this:
```
	@Test
	public void concurrency() throws InterruptedException, ExecutionException {
		final ConfigurableApplicationContext context = new AnnotationConfigApplicationContext(Config.class);
		final int concurrencyLevel = 8;
		final ExecutorService executorService = Executors.newFixedThreadPool(concurrencyLevel);
		for (int i = 0; i < 1000; i++) {
			final Callable<?> task = () -> context.getBeanFactory().createBean(BeanForConcurrencyTesting.class);
			final List<? extends Future<?>> futures = executorService
					.invokeAll(Collections.nCopies(concurrencyLevel, task));
			for (Future<?> future : futures) {
				future.get();
			}
		}
		executorService.shutdown();
		context.close();
	}
```
The main point that `new AnnotationConfigApplicationContext(Config.class);` is out of the loop, therefore most performance impact has gone.

Was that really an intention to create a new `applicationContext` on each iteration?
Since the main premise of the issue that concurrently called `createBean()` causes race conditions in the `BPP`.

I understand an aim to catch the issue with as bigger entropy as possible, but there is really no 100 % guaranties that the problem will appear even with the million iterations. 
Typically we test such a race condition with barriers: `CountDownLatch`, `Semaphore` etc...
But I really don't see reason to chase that goal with this fix.

The simplified version of your tests to demonstrate an issue is enough for me.

Just IMHO from my hurt do not spend more time on the fully qualified issue...

panchenko: @artembilan The race conditions was happening because of the contention of the first two requests. After that type information is cached. So it does not make a lot of sense to make 1000 iterations using cached data.

artembilan: But that is your test with `1000` iterations and, therefore, new `applicationContext` on each one.
So, if you can provide simplified (and fast) version of the test, I won't mind.

garyrussell:
I think we can just `@Ignore` it.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessorTests.java
panchenko: The test was initially created to reproduce the problem I had noticed in logs. That;s the only purpose of having it. With the new `applicationContext` on each iteration there are *1000* times that contention happens, so there are more chances to reproduce. 

With the new implementation that problem can't happen, so theoretically test is not needed and can be ignored, as @garyrussell suggested

garyrussell:
I haven't really been following this (aside from the odd comment) but wouldn't it have been simpler to just synchronize the method?

panchenko:
It terms of doing less code changes - probably, in terms of performance it's better when threads don't block each other and map lookup is done once.
Also I think it's easier to understand when all the type information is organized as an inner class, comparing to be spread in 4 maps.

garyrussell:
Yes; I'm fine with the solution and I agree about the metadata; I just can't get too excited about blocking in this case; blocking during message processing would be a very different story.

artembilan:
> I just can't get too excited about blocking in this case

Gary, where are we blocking, please?

If you decided that `@Ignore` is fine way to go,  I don't have objections and pretty sure in the solution to merge already.
In other words: LGTM

garyrussell:
>Gary, where are we blocking, please?

We are not; I was just saying that a synchronized method would not have given me too much concern because it's a one-off deal for each lazy bean, but since @panchenko has done this fine work already, yes, let's `@Ignore` and merge.

artembilan:
> that a synchronized method would not have given me too much concern 

Yes, that was my original thought as well. But since @panchenko has done all the amazing work, I haven't commented that way. And more over one of my comment was about idempotent outcome even if we are not synchronized.

So, merging...

Thank you for all the team work! 😄

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/37090eefad9178d71c42e29db11df1eda0ed0d84 and cherry-picked as https://github.com/spring-projects/spring-amqp/commit/af2272caf73581513de3a3f97485d8d50e555bda.

@panchenko , thank you again!

PR #607, opened 2017-04-20
AMQP-737: SMLC: Fix Consumer Cancel Stack Trace

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-737

Don't attempt to `basicCancel()` a consumer if the channel is already closed.

Also, catch `IllegalStateException` which can occur if the application context is already closed.

__cherry-pick to 1.6.x, master (change to multi-catch on master)__


artembilan:
... and cherry-picked with mentioned polishing for `master`

PR #613, opened 2017-05-08
AMQP-738: Fix prefetchCount with messagesPerAck

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-738

The `prefetchCount` must be at least `messagesPerAck` otherwise the broker
won't send enough messages.

Adjust up the `prefetchCount` if necessary, as does the `SMLC` for `txSize`.

Also fix the condition for sending acks after the consumer is canceled - the current logic caused an extra ack attempt when stopping the container - the `canceled` flag applied unconditionally, even if no acks were pending.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
@@ -338,6 +338,9 @@ public class DirectMessageListenerContainer extends AbstractMessageListenerConta
 	protected void actualStart() throws Exception {
 		this.aborted = false;
 		this.hasStopped = false;
+		if (getPrefetchCount() < this.messagesPerAck) {
+			setPrefetchCount(this.messagesPerAck);
+		}
 		super.doStart();
 		final String[] queueNames = getQueueNames();
 		checkMissingQueues(queueNames);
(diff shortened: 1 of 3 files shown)


artembilan:
Travis doesn't like test somehow:
```
org.springframework.amqp.rabbit.listener.DirectMessageListenerContainerMockTests > testDeferredAcks FAILED

    java.lang.AssertionError at DirectMessageListenerContainerMockTests.java:166
```

garyrussell:
oops

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -338,6 +338,9 @@ protected void doStop() {
 	protected void actualStart() throws Exception {
 		this.aborted = false;
 		this.hasStopped = false;
+		if (getPrefetchCount() < this.messagesPerAck) {
+			setPrefetchCount(this.messagesPerAck);
```
artembilan: How is it going to work if `prefetchCount > messagesPerAck`?
What will happen to those prefetched messages which are not handled because we have reached `messagesPerAck`?
Thanks

garyrussell: It's fine if it's greater - prefetch simply means how many un-ack'd messages can be outstanding at any time. If prefetch is 15 and mPA is 10, it means we'll have 5 unack'd at the time we send the ack (and then the broker will send 10 more, if available).

If it's smaller, we will be deadlocked with the broker waiting for an ack and us waiting for more messages.

artembilan: Good, thanks.
I fully understand the "smaller". I'm only concerted that with "bigger" we don't let other concurrent consumers to consume those extra messages. But yeah, that is already up to end-user. To have keep unacked messages for the next fetch request, or distribute all the message between concurrent consumers making `prefetch = messagesPerAck`.

Merging...

garyrussell: Yeah - most people will have prefetch "slightly" higher than mPA - so a message is immediately available after the ack is sent.

PR #619, opened 2017-05-19
AMQP-742: Log4j2 Appender Thread-Safety

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-742

Log4j2 does not create thread-safe events by default.

Do the publishing on the calling thread but provide a property to revert to previous behavior.

__I suggest reviewing with `?w=1`__

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
@@ -22,7 +22,6 @@ import java.nio.charset.Charset;
 import java.util.Calendar;
 import java.util.Map;
 import java.util.Map.Entry;
-import java.util.Set;
 import java.util.Timer;
 import java.util.TimerTask;
 import java.util.UUID;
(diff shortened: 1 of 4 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
```diff
@@ -237,96 +342,9 @@ public Message postProcessMessageBeforeSend(Message message, Event event) {
 		@Override
 		public void run() {
 			try {
-				RabbitTemplate rabbitTemplate = AmqpAppender.this.rabbitTemplate;
-				rabbitTemplate.setConnectionFactory(AmqpAppender.this.manager.connectionFactory);
 				while (true) {
```
artembilan: Hm... I wonder how is it going to stop looping?..

I see that we perform `senderPool.shutdownNow()` and don't wait for finishing, but that really looks for me that our `EventSender` will never stop.

garyrussell: ? 

`shutDownNow()` will interrupt the thread so we'll get an `InterruptedException` on the next `take()` from the queue.

artembilan: Ha! Indeed!
Not so obvious, but true.
Will add some comment on merge

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/7bbe11b0b2cb34c412d21d623f1f863008b4bbf4 and cherry-picked to `1.7.x` as https://github.com/spring-projects/spring-amqp/commit/69cd2196e92a98812c6dad0a015d67108077645f and to `1.6.x` as https://github.com/spring-projects/spring-amqp/commit/4a71d4295fc275553772f46c34ad5902188af75f

PR #620, opened 2017-05-20
support URI in configuration

doumdoum:
add support for URI in configuration
use default values from ConnectionFactory

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
@@ -429,24 +429,28 @@ public class RabbitConnectionFactoryBean extends AbstractFactoryBean<ConnectionF
 
 	/**
 	 * @param uri the uri.
-	 * @throws URISyntaxException invalid syntax.
-	 * @throws NoSuchAlgorithmException no such algorithm.
-	 * @throws KeyManagementException key management.
 	 * @see com.rabbitmq.client.ConnectionFactory#setUri(java.net.URI)
 	 */
-	public void setUri(URI uri) throws URISyntaxException, NoSuchAlgorithmException, KeyManagementException {
-		this.connectionFactory.setUri(uri);
+	public void setUri(URI uri) {
+		try{
+			this.connectionFactory.setUri(uri);
+		}
+		catch( URISyntaxException | NoSuchAlgorithmException | KeyManagementException e ){
+			throw new IllegalArgumentException(String.format("Unable to set uri: {}", e.getLocalizedMessage()), e);
+		}
 	}
 
 	/**
 	 * @param uriString the uri.
-	 * @throws URISyntaxException invalid syntax.
-	 * @throws NoSuchAlgorithmException no such algorithm.
-	 * @throws KeyManagementException key management.
 	 * @see com.rabbitmq.client.ConnectionFactory#setUri(java.lang.String)
 	 */
 	public void setUri(String uriString) throws URISyntaxException, NoSuchAlgorithmException, KeyManagementException {
-		this.connectionFactory.setUri(uriString);
+		try{
+			this.connectionFactory.setUri(uriString);
+		}
+		catch( URISyntaxException | NoSuchAlgorithmException | KeyManagementException e ){
+			throw new IllegalArgumentException(String.format("Unable to set uri: {}", e.getLocalizedMessage()), e);
+		}
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


pivotal-issuemaster:
@doumdoum Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-amqp&pullRequestId=620)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-amqp&pullRequestId=620) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@doumdoum Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-amqp&pullRequestId=620)!

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
```diff
@@ -429,24 +429,28 @@ public void setVirtualHost(String virtualHost) {
...
 	 * @see com.rabbitmq.client.ConnectionFactory#setUri(java.net.URI)
 	 */
-	public void setUri(URI uri) throws URISyntaxException, NoSuchAlgorithmException, KeyManagementException {
-		this.connectionFactory.setUri(uri);
+	public void setUri(URI uri) {
+		try {
+			this.connectionFactory.setUri(uri);
+		}
+		catch (URISyntaxException | NoSuchAlgorithmException | KeyManagementException e) {
+			throw new IllegalArgumentException(String.format("Unable to set uri: {}", e.getLocalizedMessage()), e);
```
artembilan: Well, I don't see any gain from Formatter here, so I'd prefer just string concatenation.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
```diff
@@ -169,8 +173,9 @@ public static AmqpAppender createAppender(
 			theLayout = PatternLayout.createDefaultLayout();
 		}
 		AmqpManager manager = new AmqpManager(configuration.getLoggerContext(), name);
+		manager.uri = uri;
 		manager.host = host;
-		manager.port = port;
+		Optional.ofNullable(port).ifPresent(v -> manager.port = Integers.parseInt(v));
```
artembilan: Well, is it really expectable that we are fine to fall back to `0` if our string is empty?
```
public static int parseInt(final String s, final int defaultValue) {
        return Strings.isEmpty(s) ? defaultValue : Integer.parseInt(s);
    }

    public static int parseInt(final String s) {
        return parseInt(s, 0);
    }
```
I mean maybe it would be better to use `if (StringUtils.hasText())` instead of this `Optional.ofNullable()` and don't use that `Integers` from log4j2 already...
Although that might be a fact that log4j2 XML config parser treats empty attributes (e.g. `port=""`) as nulls anyway...

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
```diff
@@ -612,11 +622,15 @@ protected ConnectionFactory createRabbitConnectionFactory() {
...
-			factoryBean.setPassword(this.password);
-			factoryBean.setVirtualHost(this.virtualHost);
+
+			Optional.ofNullable(this.host).ifPresent(factoryBean::setHost);
+			Optional.ofNullable(this.port).ifPresent(factoryBean::setPort);
+			Optional.ofNullable(this.username).ifPresent(factoryBean::setUsername);
+			Optional.ofNullable(this.password).ifPresent(factoryBean::setPassword);
+			Optional.ofNullable(this.virtualHost).ifPresent(factoryBean::setVirtualHost);
+			// overrides all preceding items when set
+			Optional.ofNullable(this.uri).ifPresent(factoryBean::setUri);
```
artembilan: OK. My comment above becomes redundant.
Very smart do not fail 😄

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/logback/AmqpAppender.java
```diff
@@ -178,27 +180,32 @@
...
 	 * RabbitMQ virtual host to connect to.
 	 */
-	private String virtualHost = "/";
+	private String virtualHost;
 
 	/**
 	 * RabbitMQ port to connect to.
 	 */
-	private int port = 5672;
+	private Integer port;
```
artembilan: M-m-m. Can't we use `Integer` for `port` in the log4j2 appender, too?
Thanks

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/4409edc1acd1b15f74a1efbf706842991c8e7a9f.

Thank you very much, @doumdoum , for the contribution.
Looking forward for more!

PR #624, opened 2017-06-05
SCSt GH-913 Support - Raw Message Header

garyrussell:
See https://github.com/spring-projects/spring-integration/pull/2174

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/support/AmqpHeaders.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/support/AmqpHeaders.java
@@ -111,4 +111,10 @@ public abstract class AmqpHeaders {
 	 */
 	public static final String CONSUMER_QUEUE = PREFIX + "consumerQueue";
 
+	/**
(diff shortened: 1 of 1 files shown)


spring-amqp/src/main/java/org/springframework/amqp/support/AmqpHeaders.java
```diff
@@ -111,4 +111,10 @@
 	 */
 	public static final String CONSUMER_QUEUE = PREFIX + "consumerQueue";
 
+	/**
+	 * The received raw message before conversion; this header usually only appears in
+	 * an {@link org.springframework.messaging.support.ErrorMessage}.
+	 */
+	public static final String RAW_MESSAGE = PREFIX + "raw_message";
```
artembilan: `@since 2.0` ? 😄

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/c10092d0c54a4337e7b56378392757b3f3419201

PR #625, opened 2017-06-11
always log QueuesNotAvailableException

panchenko:
It makes sense to have an error logged always if queue is missing.

If that's treated as fatal, then the context is not allowed to start. If that's not fatal, then context starts, subsequent attempts to start the listener are made (infinite by default), and something in the application still does not work - that's definitely an error.

To be discussed what the exact message should be in this case.

It would be nice to have this backported.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
@@ -632,6 +632,9 @@ public class BlockingQueueConsumer {
 				if (failures == null) {
 					failures = new DeclarationException(e);
 				}
+				else {
+					failures.addSuppressed(e);
+				}
 				failures.addFailedQueue(queueName);
 			}
 		}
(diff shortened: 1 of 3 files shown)


panchenko:
Alternatively, an error could be logged in `BlockingQueueConsumer.attemptPassiveDeclarations()`, for example in this line `logger.warn("Failed to declare queue:" + queueName)`.

WDYT?

garyrussell:
Thanks @panchenko ; I am fine with either logging solution (probably what you already have in the PR is better, to avoid log noise if there are a lot of missing queues), and it's ok to backport, but the `addSuppressed()` can't be backported because that's a JDK 7 addition to `Throwable` and we have to support java 6.

That said, I am not sure it adds a whole lot of value anyway.

panchenko:
> That said, I am not sure it adds a whole lot of value anyway.

Sure, just simplifies life a bit, so one can search for "ERROR" in logs in order to check if system works correctly.

panchenko:
And a similar suggestion for DMLC - to be discussed.

garyrussell:
Sorry, my "value" comment was just about capturing the suppressed exceptions, not the improved logging.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/29311f8d3a9707bb0c2c174139daf779cd1ced18 and cherry-picked as https://github.com/spring-projects/spring-amqp/commit/0af0fc708e0c9e366b1021bafee115e91fee6ea8.

@panchenko , 
thank you for amazing (as usual) contribution!

PR #626, opened 2017-06-26
AMQP-749: Force Close Channel With Slow Consumer

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-749

If the consumer thread does not respond to a shutdown (e.g. "stuck" in user code),
the channel remains open and messages remain in an unack'd state.

Add a boolean (default true in 2.0) controlling the behaviour and allowing the container
to force close the channel - thus requeuing any unack'd messages.

Since this code runs on the thread that calls `stop()`, we need to always be sure that
the physical close required `ThreadLocal` is cleared.

There are test cases where a mix of mock and real channels are used and the TL was dirty, causing
side-effects; but this could really be a concern for a real application too.

__I will backport to 1.7.x after review/merge__

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
@@ -1815,7 +1815,7 @@ public class RabbitTemplate extends RabbitAccessor implements BeanFactoryAware, 
 				if (channel == null) {
 					throw new IllegalStateException("Connection returned a null channel");
 				}
-				RabbitUtils.setPhysicalCloseRequired(true);
+				RabbitUtils.setPhysicalCloseRequired(channel, true);
 				this.dedicatedChannels.set(channel);
 			}
 			catch (RuntimeException e) {
@@ -1834,9 +1834,6 @@ public class RabbitTemplate extends RabbitAccessor implements BeanFactoryAware, 
 			}
 			else {
 				RabbitUtils.closeChannel(channel);
-				if (!(channel instanceof ChannelProxy)) { // clear the TL flag if the channel is not a proxy
-					RabbitUtils.isPhysicalCloseRequired();
-				}
 				RabbitUtils.closeConnection(connection);
 			}
 		}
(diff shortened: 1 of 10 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -1834,6 +1834,9 @@ public Boolean isMandatoryFor(final Message message) {
 			}
 			else {
 				RabbitUtils.closeChannel(channel);
+				if (!(channel instanceof ChannelProxy)) { // clear the TL flag if the channel is not a proxy
+					RabbitUtils.isPhysicalCloseRequired();
```
artembilan: Don't we need this in the `finally` of the `doExecute()` as well?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
```diff
@@ -661,6 +661,9 @@ public void stop() {
 		}
 		RabbitUtils.setPhysicalCloseRequired(true);
 		ConnectionFactoryUtils.releaseResources(this.resourceHolder);
+		if (!(this.channel instanceof ChannelProxy)) { // clear the TL flag if the channel is not a proxy
+			RabbitUtils.isPhysicalCloseRequired();
```
artembilan: We may consider just don't set TL flag if `channel` isn't `Proxy`.
See a couple lines above - `RabbitUtils.setPhysicalCloseRequired(true)`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -985,6 +998,9 @@ void cancelConsumer(final String eventMessage) {
 		private void finalizeConsumer() {
 			RabbitUtils.setPhysicalCloseRequired(true);
 			RabbitUtils.closeChannel(getChannel());
+			if (!(getChannel() instanceof ChannelProxy)) { // clear the TL flag if the channel is not a proxy
+				RabbitUtils.isPhysicalCloseRequired();
```
artembilan: The same here: do not call `setPhysicalCloseRequired()` instead if not `ChannelProxy`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
garyrussell: Yes; that's better; thanks!

artembilan:
Yes! That is much better, but need project recompilation, at least 😉

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/82f4775de57eb5a009e32d9c1fff5c91ce9e095a

PR #629, opened 2017-07-14
AMQP-747: Add trustedPackages option Jackson

artembilan:
JIRA: https://jira.spring.io/browse/AMQP-747

To have a fix for the CVE-2017-4995 Jackson vulnerability add
`trustedPackages` option to the `Jackson2JsonMessageConverter`,
`DefaultJackson2JavaTypeMapper` and `DefaultClassMapper` and
consult it for the `Class` of the body during deserialization

**Cherry-pick to 1.7.x and 1.6.x**

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/support/converter/DefaultClassMapper.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/support/converter/DefaultClassMapper.java
@@ -16,10 +16,14 @@
 
 package org.springframework.amqp.support.converter;
 
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
 
 import org.springframework.amqp.core.MessageProperties;
 import org.springframework.beans.factory.InitializingBean;
@@ -35,15 +39,24 @@ import org.springframework.util.ClassUtils;
  * {@value #DEFAULT_CLASSID_FIELD_NAME} header to classes. If this class is not a
  * Spring-managed bean, call {@link #afterPropertiesSet()} to set up the class to id
  * mapping.
+ *
  * @author Mark Pollack
  * @author Gary Russell
- *
+ * @author Artem Bilan
  */
 public class DefaultClassMapper implements ClassMapper, InitializingBean {
 
 	public static final String DEFAULT_CLASSID_FIELD_NAME = "__TypeId__";
 
 	private static final String DEFAULT_HASHTABLE_TYPE_ID = "Hashtable";
+
+	private static final List<String> TRUSTED_PACKAGES =
+			Arrays.asList(
+					"java.util",
+					"java.lang"
+			);
+
+	private final Set<String> trustedPackages = new LinkedHashSet<String>(TRUSTED_PACKAGES);
 
 	private volatile Map<String, Class<?>> idClassMapping = new HashMap<String, Class<?>>();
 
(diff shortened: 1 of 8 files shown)


spring-amqp/src/main/java/org/springframework/amqp/support/converter/DefaultClassMapper.java
```diff
@@ -186,4 +199,48 @@ public void fromClass(Class<?> clazz, MessageProperties properties) {
...
 
+	private Class<?> toClass(String classId) {
+		if (this.idClassMapping.containsKey(classId)) {
+			return this.idClassMapping.get(classId);
+		}
+		if (classId.equals(DEFAULT_HASHTABLE_TYPE_ID)) {
+			return this.defaultMapClass;
+		}
+		try {
+			Class<?> aClass = ClassUtils.forName(classId, getClass().getClassLoader());
```
garyrussell: Do you think it's safe to instantiate the class? In Kafka I compared the packages as String.

artembilan: I think this is good concern and should not be so hard to redo, but at the same time that is what we have in SI:
```
DeserializationConfig config = (DeserializationConfig) context.getConfig();
JavaType result = this.delegate.typeFromId(context, id);

Package aPackage = result.getRawClass().getPackage();
if (aPackage == null || isTrustedPackage(aPackage.getName())) {
		return result;
}
```

At least that is how we copied/pasted from @rwinch 😄 : https://github.com/spring-projects/spring-security/blob/master/core/src/main/java/org/springframework/security/jackson2/SecurityJackson2Modules.java#L180

artembilan: OK. Pushed the fix.

spring-amqp/src/main/java/org/springframework/amqp/support/converter/Jackson2JsonMessageConverter.java
```diff
@@ -53,23 +53,36 @@
 	private boolean typeMapperSet;
 
 	/**
+	 * Construct with an internal {@link ObjectMapper} instance and trusted packed to all ({@code *}).
+	 * @since 1.6.11
+	 */
+	public Jackson2JsonMessageConverter() {
+		this("*");
+	}
```
garyrussell: @artembilan I think we need this on the default class mapper (at least for backports). Agree?

artembilan: Yeah... OK. I just thought that nobody use Jackson 1.x any more.
But right, backward compatibility process doesn't imply that.

Do you need commit on the matter or will do that on merge/cherry-picking?

I think `DefaultJackson2JavaTypeMapper` suffers the same condition.

Thanks

garyrussell: Fixing on backport

garyrussell:
Merged as 9ba9c2a81c4dd54698f848df175ee5357b3fb0a3

I had a few problems with the backport; please review the commit before I merge and cherry-pick to 1.6.x https://github.com/garyrussell/spring-amqp/commit/08c0f74262d3129e923bc5890853731ee816e1f0

In particular, pay attention to restoring the defaults if `trustedPackages` is empty when adding new trusts.

garyrussell:
Cherry picked to 1.7.x as bec645e7c8cf061de6f1d02a600b588f26a35e13 , 1.6.x as 7759b771168b4adbf4d2e9fdb8c55409fe94a525

PR #633, opened 2017-07-19
AMQP-750: Add possibleAuthenticationFailureFatal

artembilan:
JIRA: https://jira.spring.io/browse/AMQP-750

In some cases `PossibleAuthenticationFailureException` is not really
fatal error, so add `possibleAuthenticationFailureFatal` option to the
ListenerContainer to disable a fatal behavior on startup

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
@@ -24,6 +24,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.Executor;
@@ -62,6 +63,7 @@ import org.springframework.amqp.support.converter.MessageConverter;
 import org.springframework.amqp.support.postprocessor.MessagePostProcessorUtils;
 import org.springframework.aop.framework.ProxyFactory;
 import org.springframework.aop.support.DefaultPointcutAdvisor;
+import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.BeanNameAware;
 import org.springframework.beans.factory.DisposableBean;
 import org.springframework.context.ApplicationContext;
@@ -92,7 +94,7 @@ import com.rabbitmq.client.ShutdownSignalException;
  */
 public abstract class AbstractMessageListenerContainer extends RabbitAccessor
 		implements MessageListenerContainer, ApplicationContextAware, BeanNameAware, DisposableBean,
-				ApplicationEventPublisherAware {
+		ApplicationEventPublisherAware {
 
 	static final int DEFAULT_FAILED_DECLARATION_RETRY_INTERVAL = 5000;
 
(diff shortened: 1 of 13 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -1233,7 +1258,7 @@ protected void executeListener(Channel channel, Message messageIn) throws Except
 					if (length < 0 || length > byteBuffer.remaining()) {
 						throw new ListenerExecutionFailedException("Bad batched message received",
 								new MessageConversionException("Insufficient batch data at offset " + byteBuffer.position()),
-										message);
+								message);
 					}
```
garyrussell: Bad indentation

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -1256,7 +1281,7 @@ protected void executeListener(Channel channel, Message messageIn) throws Except
 				else {
 					throw new ListenerExecutionFailedException("Cannot retry message more than once without an ID",
 							new AmqpRejectAndDontRequeueException("Not retryable; rejecting and not requeuing", ex),
-									messageIn);
+							messageIn);
```
garyrussell: ditto

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -1068,7 +1090,10 @@ public void shutdown() {
 	 *
 	 * @throws Exception Any Exception.
 	 */
-	protected abstract void doInitialize() throws Exception;
+	protected void doInitialize() throws Exception {
+		checkMissingQueuesFatalFromProperty();
+		checkPossibleAuthenticationFailureFatalFromProperty();
+	}
```
garyrussell: Why not do this in `initialize` rather than removing the abstract method (which could be dangerous)?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -1602,6 +1627,47 @@ protected void prepareHolderForRollback(RabbitResourceHolder resourceHolder, Run
...
+				if (applicationContext != null) {
+					Properties properties = applicationContext.getBean("spring.amqp.global.properties", Properties.class);
+					String missingQueuesFatal = properties.getProperty("mlc.missing.queues.fatal");
+					if (StringUtils.hasText(missingQueuesFatal)) {
+						setMissingQueuesFatal(Boolean.parseBoolean(missingQueuesFatal));
+					}
+				}
+			}
+			catch (BeansException be) {
+				if (logger.isDebugEnabled()) {
```
garyrussell: `isEnabled()` not needed for simple strings.

src/reference/asciidoc/amqp.adoc
```diff
@@ -4460,7 +4460,36 @@ You can also use a properties bean to set the property globally for all containe
 [source,xml]
 ----
 <util:properties id="spring.amqp.global.properties">
-    <prop key="smlc.missing.queues.fatal">false</prop>
+    <prop key="mlc.missing.queues.fatal">false</prop>
```
garyrussell: Need to document this change in what's new and/or still recognize the old property for backwards compatibility.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
artembilan: Why? That was bad indention before. The `message` arg is fully identical for the `new MessageConversionException` .
Therefore, that is logical that they should be aligned

garyrussell:
Merged as 444ac6a38aa4f64906ac8e112282ac33919969de

PR #636, opened 2017-07-27
AMQP-754 Expose AbstractRabbitListenerContainerFactory's adviceChain

jkuipers:
This allows adding additional advice if another piece of code like
Boot's SimpleRabbitListenerContainerFactoryConfigurer already set the
advice chain.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/AbstractRabbitListenerContainerFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/AbstractRabbitListenerContainerFactory.java
@@ -176,6 +176,14 @@ public abstract class AbstractRabbitListenerContainerFactory<C extends AbstractM
 	}
 
 	/**
+	 * @return the advice chain that was set. Defaults to {@code null}.
+	 * @since 1.7.4
+	 */
+	public Advice[] getAdviceChain() {
+		return this.adviceChain;
+	}
+
+	/**
 	 * @param adviceChain the advice chain to set.
 	 * @see SimpleMessageListenerContainer#setAdviceChain
 	 */
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/config/RabbitListenerContainerFactoryTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/config/RabbitListenerContainerFactoryTests.java
@@ -17,6 +17,7 @@
 package org.springframework.amqp.rabbit.config;
 
 
+import static org.junit.Assert.assertArrayEquals;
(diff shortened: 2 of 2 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/AbstractRabbitListenerContainerFactory.java
```diff
@@ -176,6 +176,14 @@ public void setDefaultRequeueRejected(Boolean requeueRejected) {
 	}
 
 	/**
+	 * @return the advice chain that was set. Defaults to {@code null}.
+	 * @since 1.7.4
+	 */
+	public Advice[] getAdviceChain() {
+		return adviceChain;
```
artembilan: Must be with `this.`

Also how about `addAdvices(Advice... advices)` ?

jkuipers: Adding an adder with an array as the backing field is kinda awkward, since you need to allocate a new array. The whole factory API simply mirrors the methods from the listener container everywhere and doesn't try to add any convenience methods anywhere else. Also, my use case is a rather exotic one; it isn't even obvious from the documentation *what* you're advising with the advice chain, you have to check the source code for that, and even then you only need this when you have multiple methods that want to add advice (like with Boot), so not many people will need this. Therefore I felt that going beyond adding a simple getter might be overkill. 

What exactly are you missing in the commit message?

jkuipers:
Added a {{this}} ref and test assertion with slightly updated commit message: please inform me what, if anything, you're still missing in the message.

The factory API closely mirrors the actual listener container methods and doesn't add convenience methods anywhere else. Also, the use case is a bit esoteric as the current docs don't even inform developers to *what* they're advising when adding an advice chain, you have to read the source code for that, and even then you only need this when you have more than one place where you want to add advice using the factory. Therefore I figured adding an adder method here would probably be overkill and a diversion of the overall API style.

artembilan:
Yes, fully agree with your explanation.
Merging with simple polishing about your name to the `@author` list.

My main concern for commit message was about the header line length - not more than 50 symbols.
"Why?" is explained in the mentioned article.

Not critical, but lets to avoid similar comments and discussions in the future.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/5c1c61f1a37313066fa90dc97a3bb596a58e23ac and cherry-picked to `1.7.x` as https://github.com/spring-projects/spring-amqp/commit/ebb1dfe0a028806f30d7580aad891271d3b5fbd7

Thank you, @jkuipers , for contribution!
Looking forward for more!

PR #638, opened 2017-08-02
AMQP-756: Add support for no-local consumers.

johnou:
https://jira.spring.io/browse/AMQP-756

Currently the no-local boolean is hardcoded when creating consumers and the class cannot be easily extended to add support, this pull request adds support for configuring the no-local flag.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/ListenerContainerFactoryBean.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/ListenerContainerFactoryBean.java
@@ -103,6 +103,8 @@ public class ListenerContainerFactoryBean extends AbstractFactoryBean<AbstractMe
 
 	private Map<String, Object> consumerArgs;
 
+	private Boolean noLocal;
+
 	private Boolean exclusive;
 
 	private Boolean defaultRequeueRejected;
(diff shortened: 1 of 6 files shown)


johnou:
Let me know if you require a second pull request for the 1.7.x branch!

artembilan:
Thank you for contribution!

I can't say why we have there it as a `false` by default, but it looks like won't hurt to allow to configure this option.
Pay attention, though, that there is `DirectMessageListenerContainer` which performs `channel.basicConsume()` from its `doConsumeFromQueue()` as well and with the same `false` now.
So, should be fixed, too.

Also I would like to see at least some simple mock test on the matter. 

The `ListenerContainerFactoryBean` should suffer of this new option, too.

johnou:
@artembilan just noticed it myself when trying to figure out something else which I will create a JIRA ticket for soon. Will ping you when changes are ready.

johnou:
@artembilan PTAL. I am a little unsure about the test, seems brittle.. any pointers?

artembilan:
No sure what you mean, @johnou , but would be better do not squash commits, so we can trace the history of changes.

johnou:
@artembilan oops, I will avoid that in the future for the spring-amqp project.

johnou:
@artembilan thanks!

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/87ead8544910cc35d8cda5d69ff49edf53e483c3

@johnou ,
thank you for contribution.
Looking forward for more!

Pay attention, please, to the commit message though.
It  becomes a PullRequest subject and description and is useful in the future for tracking changes outside of GitHub.

Issue #637, opened 2017-08-01
DirectMessageListenerContainer does not restart consumers after a lost connection has been recovered


Monofraps:
So we've been playing around with version spring-amqp/-rabbit 2.0.0 M5 and have a weird issue with `DirectMessageListenerContainer`s.

We use a `CachingConnectionFactory` with all values at their defaults. The `DirectMessageListenerContainer`s also have everything on default except missingQueuesFatal set to false.

To test connection and consumer recovery, we stop the RabbitMQ server (sends connection close because of shutdown). All `DirectMessageListenerContainer` (DMLC) detect the lost connection properly. However, once the server comes back up, there's always one DMLC which does not restart its consumers. 

I've been debugging a little:
It looks like the first DMLC to run the monitor task calls `DirectMessageListenerContainer#doConsumeFromQueue` from the consumer monitor task and still gets a connection object in [DirectMessageListenerContainer.java#L553](https://github.com/spring-projects/spring-amqp/blob/0d341a1bfa30ba612db6a372f1f37cdb0e1750ff/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java#L553) form the connection factory. This raises an Exception in [DirectMessageListenerContainer#L567](https://github.com/spring-projects/spring-amqp/blob/0d341a1bfa30ba612db6a372f1f37cdb0e1750ff/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java#L567) when creating a channel. This exception is caught in the monitor thread [DirectMessageListenerContainer.java#L416](https://github.com/spring-projects/spring-amqp/blob/0d341a1bfa30ba612db6a372f1f37cdb0e1750ff/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java#L416) which just logs an error. However, the consumer which failed to start is not added back to the list of `consumersToRestart`.

Let me know if you're having trouble reproducing this issue; I'll try to create a minimal example application then.

EDIT: Logs at https://gist.github.com/Monofraps/152a6d95d63252c4eae62f310a84c50a

artembilan:
Thank you for heads up, @Monofraps !

Looks like I see what we are missing there.
So, we have this code so far:
```
try {
	channel = connection.createChannel(isChannelTransacted());
     ...
}
catch (AmqpApplicationContextClosedException e) {
[...]

PR #639, opened 2017-08-02
Properly handle connection errors and `null` consumers in DirectMessageListenerContainer#doConsumeFromQueue 

Monofraps:
DirectMessageListenerContainer#doConsumeFromQueue now properly adds the consumer back to the list of consumers to restart in case of an AmqpConnectException

Fixes spring-projects/spring-amqp#637

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
@@ -595,8 +595,14 @@ public class DirectMessageListenerContainer extends AbstractMessageListenerConta
 			else if (this.logger.isWarnEnabled()) {
 				this.logger.warn("basicConsume failed, scheduling consumer " + consumer + " for restart", e);
 			}
-			this.consumersToRestart.add(consumer);
-			consumer = null;
+
+			if (consumer == null) {
+				this.consumersToRestart.add(new SimpleConsumer(null, null, queue));
+			}
+			else {
+				this.consumersToRestart.add(consumer);
+				consumer = null;
+			}
 		}
 		synchronized (this.consumersMonitor) {
 			if (consumer != null) {


pivotal-issuemaster:
@Monofraps Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-amqp&pullRequestId=639)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-amqp&pullRequestId=639) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@Monofraps Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-amqp&pullRequestId=639)!

Monofraps:
Thinking about it - it would probably be to only add to the consumers to restart list if DirectMessageListenerContainer.this.started is set. Otherwise the list would infinitely grow if now RabbitMQ is available on container startup.

This does not clear `consumersToRestart`, However, `doConsumeFromQueue` will add new entries to the list each time the loop is executed:
```
while (!DirectMessageListenerContainer.this.started && isRunning()) {
	this.cancellationLock.reset();
	try {
		for (String queue : queueNames) {
			consumeFromQueue(queue);
		}
	}
	catch (AmqpConnectException | AmqpIOException e) {
		long nextBackOff = backOffExecution.nextBackOff();
		if (nextBackOff < 0 || e.getCause() instanceof AmqpApplicationContextClosedException) {
			DirectMessageListenerContainer.this.aborted = true;
			shutdown();
			this.logger.error("Failed to start container - fatal error or backOffs exhausted",
											e);
			this.taskScheduler.schedule(this::stop, new Date());
			break;
		}
		this.logger.error("Error creating consumer; retrying in " + nextBackOff, e);
		doShutdown();
		try {
			Thread.sleep(nextBackOff);
		}
		catch (InterruptedException e1) {
		}
		continue; // initialization failed; try again having rested for backOff-interval
	}
	DirectMessageListenerContainer.this.started = true;
	DirectMessageListenerContainer.this.startedLatch.countDown();
}
```

artembilan:
What you tell, @Monofraps , might make sense.
But as I see in the code having unconditional `consumerMonitorTask` brings us this logic:
```
List<SimpleConsumer> restartableConsumers = new ArrayList<>(this.consumersToRestart);
this.consumersToRestart.clear();
if (this.started) {
...
}
```
So, we periodically clear that `consumersToRestart`.
The next task:
```
try {
	for (String queue : queueNames) {
		consumeFromQueue(queue);
	}
}
catch (AmqpConnectException | AmqpIOException e) {
```
May start some consumers and then fail on the next one, so we go to the `doShutdown();` and `continue; // initialization failed; try again having rested for backOff-interval`. And as you see there we don't go to the `DirectMessageListenerContainer.this.started = true;`. 
Therefore if we would have the `started` condition for adding to the `consumersToRestart`, we may end up with the fact that we have dangling consumers or lost during the proper restart.

I won't mind if I miss anything.

I don't have any objection for your current fix though.
When ever you ready, please, let me know that I can go ahead and merge it.

Monofraps:
You can merge the current fix, @artembilan. - As for the other issue I mentioned, it seems you're right that the list is periodically cleared by the monitor thread. I did however see multiple (more than I expected) consumer start messages in my logs if I create the listener container with the RabbitMQ server shut down. (See here: https://gist.github.com/Monofraps/085fde534b118b98758ee19c010e341b this is with 2 listener containers with one registered consumer each) I'll look into this again this weekend.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/545c8c5caf8188586b1428eb38f5c3ffd20e2f05.

The problem with the several consumers on the same queue mostly looks like a side-effect of a desynchronization between `consumerMonitorTask` and the warm up loop with the `while (!DirectMessageListenerContainer.this.started && isRunning()) {`. We iterate and fail, and add to the `consumersToRestart`, but the clean and `doConsumeFromQueue()` may happen a bit later when we have already populated several unconditional instances of consumer for the same queue.

Just my feeling how it looks

PR #640, opened 2017-08-02
AMQP-756: Add support for no-local consumers

johnou:
JIRA: https://jira.spring.io/browse/AMQP-756


johnou:
@artembilan as promised, here it is for the 1.7.x branch.

Issue #641, opened 2017-08-05
Possible memory leak during DirectMessageListenerContainer start phase if no RabbitMQ server is available


Monofraps:
During `DirectMessageListenerContainer` startup, a loop continuously tries to start consumers for each registered queue name.
```
for (String queue : queueNames) {
    consumeFromQueue(queue);
}
```

`consumeFromQueue` will internally invoke `doConsumeFromQueue` if there's no active listener for the given queue. However, `doConsumeFromQueue` will append an entry to `consumersToRestart` every time it fails to start a consumer. During the whole start phase the `this.consumersMonitor` is acquired (Thread.sleep does not unlock monitor objects). This results in the consumer monitor task to block at `synchronized (this.consumersMonitor)` until the container has started (i.e. a consumer for every queue has been created without error). Since the monitor task is blocked, the following code to clean the list of consumers to restart is not cleared resulting in `consumersToRestart` growing until either the container starts up or the JVM crashes due to OOM errors.

```
synchronized (this.consumersMonitor) {
	List<SimpleConsumer> restartableConsumers = new ArrayList<>(this.consumersToRestart);
	this.consumersToRestart.clear();
[...]
}
```

PR #642, opened 2017-08-05
Fix possible memory leak during DirectMessageListenerContainer startup

Monofraps:
DirectMessageListenerContainer now clears the list of consumers to restart if a consumer fails to start during startup. This is necessary to prevent `consumersToRestart` from growing indefinitely while the container is in startup phase.

Fixes spring-projects/spring-amqp#641

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
@@ -554,7 +554,7 @@ public class DirectMessageListenerContainer extends AbstractMessageListenerConta
 			connection = getConnectionFactory().createConnection();
 		}
 		catch (Exception e) {
-			this.consumersToRestart.add(new SimpleConsumer(null, null, queue));
+			addConsumerToRestart(new SimpleConsumer(null, null, queue));
 				throw e instanceof AmqpConnectException
 						? (AmqpConnectException) e
 						: new AmqpConnectException(e);
@@ -598,10 +598,10 @@ public class DirectMessageListenerContainer extends AbstractMessageListenerConta
 			}
 
 			if (consumer == null) {
-				this.consumersToRestart.add(new SimpleConsumer(null, null, queue));
+				addConsumerToRestart(new SimpleConsumer(null, null, queue));
 			}
 			else {
-				this.consumersToRestart.add(consumer);
+				addConsumerToRestart(consumer);
 				consumer = null;
 			}
 		}
@@ -699,6 +699,12 @@ public class DirectMessageListenerContainer extends AbstractMessageListenerConta
 		finally {
 			this.consumers.remove(consumer);
 			consumerRemoved(consumer);
+		}
+	}
+
+	private void addConsumerToRestart(SimpleConsumer consumer) {
+		if (this.started) {
+			this.consumersToRestart.add(consumer);
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -459,6 +459,7 @@ protected void actualStart() throws Exception {
 								}
 								this.logger.error("Error creating consumer; retrying in " + nextBackOff, e);
 								doShutdown();
+								this.consumersToRestart.clear();
```
artembilan: Hm. Looking to what you've done and what we have discussed before, plus the current logic, I'd say that you were right: we shouldn't add to `consumersToRestart` if we are not in the `DirectMessageListenerContainer.this.started`.
Just because this `while (!DirectMessageListenerContainer.this.started && isRunning()) {` are not going to exit until `DirectMessageListenerContainer.this.started = true;`.

WDYT?

Good catch though!
I like your brilliant vision 😄

Monofraps: While I don't see any apparent issues with the solution I've implemented in this PR, it might be safer to use an `if(started)` switch in `doConsumeFromQueue`. 
Though, `org.springframework.amqp.rabbit.listener.DirectMessageListenerContainer.SimpleConsumer#cancelConsumer` does also add the current consumer to the list of consumers to restart. If this methods were invoked while the listener container is looping in its startup phase, there would still be a memory leak. So I suggest that we either wrap all `consumersToRestart.add()` calls in an `if(started)` or keep the `consumersToRestart.clear()` in the start loop.

artembilan: OK. Thank you for your thoughts.
Well, IMO it looks more logical do not add to the `consumersToRestart` because of some condition. 
While  this `consumersToRestart.clear()` works because of the synchronization blocking in the monitor thread, that doesn't mean that it is going to work in the future because we will change the threading model somehow. So, our lucky state here might not be possible and monitor thread may simply go ahead and iterate over `consumersToRestart` in parallel with this start loop.

The `if (started)` is much robust IMO.

Does it make sense to you?

Monofraps: Thanks for your feedback - I see that adding `if(started)` is more robust than the solution above. Though, if someone introduces a new location where consumers are added back to the restart list, he needs to keep in mind that an `if(started)` might be required to prevent memory leaks. 
I suppose a generally cleaner solution would be to have a fixed collection of consumers which are supposed to be running and then have everything else just update states on these instead of creating new `SimpleConsumer`.

I'll update the PR with `if(started)` checks later today .

artembilan: We can't have fixed collection of the consumers because they can be adjusted at runtime: 
```
	private void addQueues(Stream<String> queueNameStream) {
		if (isRunning()) {
			synchronized (this.consumersMonitor) {
				checkStartState();
				Set<String> current = getQueueNamesAsSet();
				queueNameStream.forEach(queue -> {
					if (current.contains(queue)) {
						this.logger.warn("Queue " + queue + " is already configured for this container: "
								+ this + ", ignoring add");
					}
					else {
						consumeFromQueue(queue);
					}
				});
			}
		}
	}
```

We may just introduce a `private void addConsumersToRestart(Consumer)` and encapsulate the proper logic there with an appropriate JavaDocs to explain the purpose and reason.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/21ed8685621daaf36ddbdc559dfaec972785e396.

@Monofraps ,
thank you for such a valuable contribution.
Looking forward for more!

Monofraps:
Thanks for merging. I'll see if I can contribute some more to this and other Spring projects; I kinda like working on open source again.

PR #645, opened 2017-08-16
AmqpAppender: Add thread name to message header

Monofraps:
AmqpAppender for log4j2 and Logback now add the thread name of the log event as a header field as suggested in [AMQP-761](https://jira.spring.io/browse/AMQP-761) and spring-projects/spring-amqp-samples#27

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
@@ -102,6 +102,11 @@ public class AmqpAppender extends AbstractAppender {
 	public static final String CATEGORY_LEVEL = "level";
 
 	/**
+	 * Key name for the thread name in the message properties.
+	 */
+	public static final String THREAD_HEADER_KEY = "thread";
+
+	/**
 	 * Tha manager.
 	 */
 	private final AmqpManager manager;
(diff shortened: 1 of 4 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
```diff
@@ -102,6 +102,11 @@
 	public static final String CATEGORY_LEVEL = "level";
 
 	/**
+	 * Key name for the thread name in the message properties.
+	 */
+	public static final String THREAD_HEADER_KEY = "thread";
```
artembilan: @Monofraps ,

sorry for such a long delay.

I would prefer to have it as simple `THREAD_NAME` name:
1. It is constant, so it is definitely already `KEY`
2. The usage in the code uniquely says that it is used for headers.

That's why I don't see reason to say too much in the constant if it doesn't bring much value or even overcomplicate it.

So, let me to fix it during merge!

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/625063889ffaac37ae6fe600c4219654bee816c9.

@Monofraps ,

thank you for contribution!

In the future, please, consider to follow this instruction: https://github.com/spring-projects/spring-amqp/blob/master/CONTRIBUTING.adoc and don't PR against your `master`. Now you have to `reset hard` your local version to incorporate the `upstream`.

PR #646, opened 2017-08-16
[1.7.x] AmqpAppender for log4j, log4j2 and Logback now add the thread name as a message header field

Monofraps:
AmqpAppender for log4j, log4j2 and Logback now add the thread name of the log event as a header field as suggested in [AMQP-761](https://jira.spring.io/browse/AMQP-761) and spring-projects/spring-amqp-samples#27


artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/51128a27a112d8b26b270ce1772b64626af63581.

The same is applied here as I mentioned in the #645 - use top level branches for the PRs. Much easier to track, merge and then adapt to the `upstream`.

Plus I forgot to mention that I also added a note into the documentation what message properties we have for log events: https://github.com/spring-projects/spring-amqp/commit/51128a27a112d8b26b270ce1772b64626af63581#diff-a9efe5e8a2cf027abf3c5466821127ceR192

Thank you for contribution.
Come back to us, please, for more!

PR #647, opened 2017-08-17
Remove Bogus DEBUG log.

garyrussell:
"Queue with name that starts with 'amq.' cannot be declared." logged unconditionally.

__Cherry-pick to 1.6.x, 1.7.x__


artembilan:
... and cherry-picked.

Issue #651, opened 2017-08-23
Send message on close


Hronom:
I want to send message by using `RabbitTemplate` inside method `close` of my singleton `AutoCloseable` bean, but when I call `rabbitTemplate.convertAndSend` I got an exception:
```
 Exception: java.lang.IllegalStateException: The ApplicationContext is closed and the ConnectionFactory can no longer create connections.
```

Why `CachingConnectionFactory` not stopped in destroy method?

garyrussell:
The connection is closed in `SmartLifecycle.stop()` because of a [deadlock problem where a late consumer reopened the connection](https://github.com/spring-projects/spring-amqp/pull/292).

We could probably move the logic to `destroy()` but it is generally not a good idea to interact with beans in the application context after the context has been closed. However, it is also unlikely that your auto close will occur before the `destroy()`, since that method is also called when the context is closed.

You may need another connection factory that is __not managed by the application context__ to send your message (remember to `destroy()` it though).

Hronom:
Well from logical point of view, when close is called on a bean by Spring, the logic in this method must be executed including communication with beans on that current bean depends. Spring closes beans respectivly to theirs dependencies, so if your application has core service bean this bean will be closed in the end.

Current realisation of `CachingConnectionFactory` not respects closing sequence of Spring beans.

I have created test app https://github.com/Hronom/spring-rabbitmq-issue-test
Here is the closing sequence:
```
2017-08-24 16:12:35.432  INFO 4070 --- [       Thread-3] ationConfigEmbeddedWebApplicationContext : Closing org.springframework.boot.context.embedded.AnnotationConfigEmbeddedWebApplicationContext@247310d0: startup date [Thu Aug 24 16:12:28 EEST 2017]; root of context hierarchy
2017-08-24 16:12:35.433 DEBUG 4070 --- [       Thread-3] o.s.b.f.s.DefaultListableBeanFactory     : Returning cached instance of singleton bean 'connectionFactory'
2017-08-24 16:12:35.434 DEBUG 4070 --- [       Thread-3] o.s.b.f.s.DefaultListableBeanFactory     : Returning cached instance of singleton bean 'org.springframework.amqp.rabbit.config.internalRabbitListenerEndpointRegistry'
2017-08-24 16:12:35.434 DEBUG 4070 --- [       Thread-3] o.s.b.f.s.DefaultListableBeanFactory     : Returning cached instance of singleton bean 'lifecycleProcessor'
[...]

PR #653, opened 2017-08-24
AMQP-764: Move Setting stopped to destroy() [Backport]

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-764

Fixes https://github.com/spring-projects/spring-amqp/issues/651

Previously, `stopped` was set in `stop()` after the context was closed.
This was done to prevent inadvertent re-opening of the connection which
would prevent the application from terminating.

This was too early since other beans destroyed before the factory might
need to send messages.

Remove `Smartlifecycle` and move setting `stopped` to `destroy()`, but
still only after the context is closed; since `destroy()` can be called
by users to reset the connection (although `resetConnection()` is now the
preferred way to do that.

__cherry-pick to 1.6.x__


artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/fed485ae043d74292f9d19bc3f4b88f74ac5d29b and cherry-picked to `1.6.x` as https://github.com/spring-projects/spring-amqp/commit/5623996180df6ff5bec293790a9766de085a055d.

PR #661, opened 2017-09-21
AMQP-770: Re-add constructor with ObjectMapper

ArloL:
Adding the varargs constructor argument breaks reflection instantiation and therefore is a breaking change to the API.


artembilan:
Merged and cherry-picked to `1.7.x` as https://github.com/spring-projects/spring-amqp/commit/a723b58621366abb7b4ef3089f3e86d257f1ec05 . Backported to `1.6.x` as https://github.com/spring-projects/spring-amqp/commit/21a9efc43893656e8f5092f9434b9967c1f317c3.

@ArloL ,

Thank you very for the contribution.
Looking forward for more!

PR #673, opened 2017-10-03
AMQP-757: Add connection BlockedListener support

artembilan:
JIRA: https://jira.spring.io/browse/AMQP-757

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
@@ -35,12 +35,19 @@ import org.apache.commons.logging.LogFactory;
 import org.springframework.amqp.rabbit.support.RabbitExceptionTranslator;
 import org.springframework.beans.factory.BeanNameAware;
 import org.springframework.beans.factory.DisposableBean;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
+import org.springframework.context.ApplicationEventPublisher;
+import org.springframework.context.ApplicationEventPublisherAware;
+import org.springframework.context.ApplicationListener;
+import org.springframework.context.event.ContextClosedEvent;
 import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
 import org.springframework.util.Assert;
 import org.springframework.util.ObjectUtils;
 import org.springframework.util.StringUtils;
 
 import com.rabbitmq.client.Address;
+import com.rabbitmq.client.BlockedListener;
 import com.rabbitmq.client.Recoverable;
 import com.rabbitmq.client.RecoveryListener;
 import com.rabbitmq.client.impl.recovery.AutorecoveringConnection;
@@ -52,7 +59,10 @@ import com.rabbitmq.client.impl.recovery.AutorecoveringConnection;
  * @author Artem Bilan
  *
  */
-public abstract class AbstractConnectionFactory implements ConnectionFactory, DisposableBean, BeanNameAware {
+public abstract class AbstractConnectionFactory implements ConnectionFactory, DisposableBean, BeanNameAware,
+		ApplicationContextAware, ApplicationEventPublisherAware, ApplicationListener<ContextClosedEvent> {
+
+	public static final int DEFAULT_CLOSE_TIMEOUT = 30000;
 
 	private static final String BAD_URI = "setUri() was passed an invalid URI; it is ignored";
 
@@ -84,20 +94,24 @@ public abstract class AbstractConnectionFactory implements ConnectionFactory, Di
 
 	};
 
-	private volatile ExecutorService executorService;
+	private ExecutorService executorService;
 
-	private volatile Address[] addresses;
+	private Address[] addresses;
 
-	public static final int DEFAULT_CLOSE_TIMEOUT = 30000;
-
-	private volatile int closeTimeout = DEFAULT_CLOSE_TIMEOUT;
+	private int closeTimeout = DEFAULT_CLOSE_TIMEOUT;
 
 	private ConnectionNameStrategy connectionNameStrategy =
 			connectionFactory -> (this.beanName != null ? this.beanName : "SpringAMQP") +
 					"#" + ObjectUtils.getIdentityHexString(this) + ":" +
 					this.defaultConnectionNameStrategyCounter.getAndIncrement();
 
-	private volatile String beanName;
+	private String beanName;
+
+	private ApplicationContext applicationContext;
+
+	private ApplicationEventPublisher applicationEventPublisher;
+
+	private volatile boolean contextStopped;
 
 	/**
 	 * Create a new AbstractConnectionFactory for the given target ConnectionFactory.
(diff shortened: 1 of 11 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
```diff
@@ -84,20 +94,24 @@ public void handleRecovery(Recoverable recoverable) {
 
 	};
 
-	private volatile ExecutorService executorService;
+	private ExecutorService executorService;
```
dturanski: Just curious, why remove volatile?

src/reference/asciidoc/amqp.adoc
```diff
@@ -361,6 +361,19 @@ The `ConnectionFactory` argument can be used to distinguish target connection na
 By default a `beanName` of the `AbstractConnectionFactory` and an internal counter are used to generate `connection_name`.
 The `<rabbit:connection-factory>` namespace component is also supplied with the `connection-name-strategy` attribute.
 
+
+The connection might be blocked for interaction from the Broker according to the https://www.rabbitmq.com/memory.html[Memory Alarm].
+Starting with _version 2.0_, the `org.springframework.amqp.rabbit.connection.Connection` can be supplied with the `com.rabbitmq.client.BlockedListener` s to to be notified for connection block and unblock events.
```
dturanski: can be supplied with ~~the~~ `com.rabbitmq...`   
connection block`ed` and unblock`ed`

src/reference/asciidoc/amqp.adoc
```diff
@@ -361,6 +361,19 @@ The `ConnectionFactory` argument can be used to distinguish target connection na
...
+In addition the `AbstractConnectionFactory` emits a `ConnectionBlockedEvent` and `ConnectionUnblockedEvent`, respectively, via its internal `BlockedListener` implementation.
+These can be useful in application logic to react appropriately to the problems on the Broker and perform some maintenance actions for example.
```
dturanski: ~~These can be useful in application logic to react appropriately to the problems on the Broker and perform some maintenance actions for example.~~
These allow you to provide application logic to react appropriately to problems on the broker and take some corrective actions for example.

src/reference/asciidoc/amqp.adoc
```diff
@@ -361,6 +361,19 @@ The `ConnectionFactory` argument can be used to distinguish target connection na
...
+
+IMPORTANT: When single `CachingConnectionFactory` is used in the application, e.g. in case of auto-configuration by Spring Boot, it is important to understand that by default it provides a single shared connection instance.
```
dturanski: ~~IMPORTANT: When single `CachingConnectionFactory` is used in the application, e.g. in case of auto-configuration by Spring Boot, it is important to understand that by default it provides a single shared connection instance.~~
IMPORTANT: When the application is configured with a single `CachingConnectionFactory` ...  ,as is the default with Spring Boot auto-configuration,  the application will stop working when the connection is blocked by the broker.

src/reference/asciidoc/amqp.adoc
```diff
@@ -361,6 +361,19 @@ The `ConnectionFactory` argument can be used to distinguish target connection na
...
+And when it is blocked by the Broker, any its clients stop to work.
+If we deal with producers and consumers in the same application, we may end up with the dead lock situation when producers are blocking connection because no resources on Broker any more and consumers can't free them because connection is blocked.
+To mitigate the problem, there is just enough to have one more separate `CachingConnectionFactory` instance with the same options - one for producers and one for consumers.
```
dturanski: If we have producers and consumers in the same application, we may end up with a ~~dead lock~~deadlock when producers are blocking _the_ connection because there are no resources on _the_ Broker ~~any more~~ anymore and consumers can't free them because _the_ connection is blocked.

src/reference/asciidoc/amqp.adoc
```diff
@@ -361,6 +361,19 @@ The `ConnectionFactory` argument can be used to distinguish target connection na
...
+At the same time this isn't recommended if we use transactional producer and it is important to let that producer to reuse a `Channel` associated with the consumer transaction.
+
```
dturanski: This reads better, but I'm not sure what you mean:  What isn't recommended?
```At the same time this isn't recommended when we use a transactional producer and it is important to let the producer reuse a `Channel` associated with the consumer transaction.```

File: src/reference/asciidoc/amqp.adoc [hunk 7]
```diff
@@ -588,6 +601,8 @@ public interface ConnectionListener {
 }
 ----
 
+Starting with _version 2.0_, the `org.springframework.amqp.rabbit.connection.Connection` object can be supplied with the `com.rabbitmq.client.BlockedListener` s to to be notified for connection block and unblock events.
+
```
dturanski: object can be supplied with ~~the~~ `com.rabbitmq.client.BlockedListener`

dturanski: connection blocked and unblocked events

src/reference/asciidoc/whats-new.adoc
```diff
@@ -158,6 +158,12 @@ To revert to a `Hashtable` use `setDefaultMapType` on the `DefaultClassMapper`.
 When parsing `Queue` and `Exchange` XML components, the parsers no longer register the `name` attribute value as a bean alias if an `id` attribute is present.
 See <<note-id-name>> for more information.
 
+===== Blocked Connection
+The `com.rabbitmq.client.BlockedListener` can now be injected into the `org.springframework.amqp.rabbit.connection.Connection` object.
+Also the `ConnectionBlockedEvent` and `ConnectionUnblockedEvent` events are emitted by the `ConnectionFactory`, when connection is blocked and unblocked by the Broker.
```
dturanski: ... when _the_ connection is  blocked ~~and~~ or unblocked ...

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
artembilan: The properties from setters are considered as configuration options which mostly are specified only during application start up. They are not intended to be changed concurrently at runtime.
And even if they may be changed at runtime concurrently that is not so critical do not get an actual value immediately in other thread.
Without `volatile` we get some performance gain. 

Those properties which are really critical for runtime changes remain with `volatile`. For example `active`, `stopped`, `running` etc.

I do such a refactoring only in the effected code though and only when I remember 😄

src/reference/asciidoc/amqp.adoc
artembilan: Good. How about this?

> The separate `CachingConnectionFactory` isn't recommended for transactional producers, since they should reuse a `Channel` associated with the consumer transaction.

artembilan:
@dturanski ,

thank you for review!

I have just addressed all your comments.
Let me know if anything else or I can go ahead with merging.

artembilan:
Merging since everything has been addressed and need to release today because of Spring Boot 2.0 M5 tomorrow...

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/b3063dc8f1e74fd248f3d4048d939b5ed762ceef

Issue #672, opened 2017-10-02
Queues property is not correctly resolved on the @org.springframework.amqp.rabbit.annotation.RabbitListener class


novtor:
When the "queues" value is configured via application.yml with conditionals inside, it is not correctly resolved
[RabbitListenerIssue.zip](https://github.com/spring-projects/spring-amqp/files/1349143/RabbitListenerIssue.zip)

novtor:
I think, the issue is in this file:
https://github.com/spring-projects/spring-amqp/blob/ffff7d7c12ca1eb12004f3b899f5c810e68999c8/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java#L729

should be replaced by
if (!(resolvedValue.**contains**("#{") && value.**contains**("}"))) {

garyrussell:
I don't think changing it `contains(()` is the right solution.

I think we should simply document that the SpEL evaluation is done after the property placeholder resolution and must match the pattern `^#{...}$`.

My 2c.

PR #675, opened 2017-10-03
GH-672: Fix SpEL parsing for @RabbitListener

artembilan:
Fixes spring-projects/spring-amqp#672

The `StandardBeanExpressionResolver` is based on the
`TemplateAwareExpressionParser` which can parse any string to the
`LiteralExpression` or to the `CompositeStringExpression` if
template tokens (`#{` & `}`) are present in the source expression.
This way we don't need to deal with literal concatenations within
templated expression.
The expression like `#{'foo.' + myBean.bar}` can be replaced
with the `foo.#{myBean.bar}`


artembilan:
Do not merge yet: need team agreement.

PR #682, opened 2017-11-01
AMQP-779: User correlationId with reply container

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-779

Allow the use of a user-supplied correlationId instead of one
generated by the `RabbitTemplate`.

It is the user's responsibility to ensure uniqueness.

__The backport to 1.7.x will need some work because of the `String` Vs. `byte[]`
correlationId changes__

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
@@ -240,6 +240,8 @@ public class RabbitTemplate extends RabbitAccessor implements BeanFactoryAware, 
 
 	private Executor taskExecutor;
 
+	private boolean userCorrelationId;
+
 	/**
 	 * Convenient constructor for use with setter injection. Don't forget to set the connection factory.
 	 */
@@ -650,6 +652,18 @@ public class RabbitTemplate extends RabbitAccessor implements BeanFactoryAware, 
 	 */
 	public void setTaskExecutor(Executor taskExecutor) {
 		this.taskExecutor = taskExecutor;
+	}
+
+	/**
+	 * Set to true to use correlation id provided by the message instead of generating
+	 * the correlation id for request/reply scenarios. The correlation id must be unique
+	 * for all in-process requests to avoid cross talk.
+	 * <p>
+	 * <b>Users must therefore take create care to ensure uniqueness.</b>
+	 * @param userCorrelationId true to use user correlation data.
+	 */
+	public void setUserCorrelationId(boolean userCorrelationId) {
+		this.userCorrelationId = userCorrelationId;
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


artembilan:
I still not sure that we have to do that, but I definitely against backport at all.

And we if already are here, we need to think about similar solution for the `JmsOutboundGateway`

artembilan:
@garyrussell ,

Do you still insist on merging this?
How about some note in the Docs in addition then?
May I don't backport it though? There are plenty of workarounds.

Thanks

garyrussell:
I don't see any hard in merging; let's not backport for now.

kingkongprab:
Hi Gary,

I tried to compile and use 2.0.1 snapshot version to test my use case, But during the start up, I got this below error. any help would be appricaited

org.springframework.context.ApplicationContextException: Failed to start bean 'replyContainer'; nested exception is java.lang.NoSuchMethodError: org.springframework.util.Assert.state(ZLjava/util/function/Supplier;)V

artembilan:
You need to upgrade to Spring Framework `5.0` as well.

kingkongprab:
Could you please provide spring boot bill of material version? I'm currently using this

    <parent>
		<groupId>org.springframework.boot</groupId>
		<artifactId>spring-boot-starter-parent</artifactId>
		<version>1.5.8.RELEASE</version>
	</parent>

artembilan:
`2.0.0.M6`: https://spring.io/blog/2017/11/06/spring-boot-2-0-0-m6-available-now.

However you should take a look into JIRA for this PR for my advice how to overcome your problem without any changes to the framework.

kingkongprab:
if i'm able to correlate the messages using the correlation id, that will solve my problem.

artembilan:
You still do, but you just don’t need to use that AMQP Outbound Gateway.  I merged this, but I still don’t understand your problem and this correlationId became such an obstacle for you. You should just rethink your use-case and don’t try to override built-in logic because you can correlate some other way. 
We spent too much time solving the problem where we don’t see it.
That’s just IMO and I won’t mind to hear some other point of view.

kingkongprab:
The solution that you had suggested may not work for our use case, as the processing of the message happens outside of container and they cannot retain those header values to send it back here,  and the correlation id is residing within the response payload itself. So, considering that, I think correlation id mapping would be best approach for us. At the same time, I haven't tried your approach, feel free to make changes here https://github.com/kingkongprab/spring-amqp-outbound-gateway whatever you think is the best solution for this.

artembilan:
> happens outside of container and they cannot retain those header values

Sorry, not clear how "container" is related to "them".

I may agree that "they" somehow can't transport headers. Although the same "correlationId" is a header as well...

> and the correlation id is residing within the response payload itself.

Not sure what is that since "correlationId" is an AMQP message property and it is mapped to the headers in Spring Integration.

>  I think correlation id mapping would be best approach for us. 

Well, consider that we can't change the Framework code. You, first of all, should design your application to use existing tool and don't push us for the unnecessary change. If your solution is based on the built-in "correlationId" in the Framework, you should consider to change your solution not Framework, because this tool is not only for your application and it is here for a while already. Plus you haven't convinced us how your previous JMS solution worked since the behavior there is exactly the same like in this AMQP component.

Having all those arguments I really doubt in the solution and still feel that your application is wrong not the Framework.

Well, let it be: we have merged this `useUserCorrelationId()` and you have desired solution. So, let us do some other work, but not only support you.

Thank you for understanding

PR #683, opened 2017-11-01
AMQP-763: Allow custom queue implementation for AmqpAppender

Monofraps:
JIRA: https://jira.spring.io/browse/AMQP-763

Some issues I still have with the current solution:
* I want to ask for your opinion on `BlockingQueue#add` vs. `BlockingQueue#offer` in this case. While `offer` is usually the preferred option if potentially bounded queues are expected, I feel like it's not the appenders task to handle out-of-capacity situations so failing with a RuntimeException is fine.
* Sadly, I'm not familiar with Log4j2 but I'd like to provide the same functionality there too. Since Log4j2 does not support subclassing, my current plan is to add a `@PluginAttribute Class ...` to the appender factory and then instantiate the provided class reflectively. Does Spring provide any class instantiation utilities which should be used for this purpose? Is there a more straightforward way to do this?
* With regards to Log4j2 support, would it be more desirable to make both solution behave the same? - i.e. Instead of subclassing for Logback, also add a `queue` configuration option to the appender which accepts a class name and works the same as the Log4j2 version

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
@@ -27,6 +27,7 @@ import java.util.Optional;
 import java.util.Timer;
 import java.util.TimerTask;
 import java.util.UUID;
+import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingQueue;
@@ -40,6 +41,7 @@ import org.apache.logging.log4j.core.LogEvent;
 import org.apache.logging.log4j.core.LoggerContext;
 import org.apache.logging.log4j.core.appender.AbstractAppender;
 import org.apache.logging.log4j.core.appender.AbstractManager;
+import org.apache.logging.log4j.core.async.BlockingQueueFactory;
 import org.apache.logging.log4j.core.config.Configuration;
 import org.apache.logging.log4j.core.config.plugins.Plugin;
 import org.apache.logging.log4j.core.config.plugins.PluginAttribute;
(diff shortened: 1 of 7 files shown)


artembilan:
I think in case of Log4J2 we good to go with its `BlockingQueueFactory` support: https://logging.apache.org/log4j/2.x/manual/appenders.html.

There are some other inconsistency between Logback and Log4J2 appenders, so, I wouldn't consider alignment here as well.

Thank you for contribution!

Will come back for review when you'are done with the Log4J2 solution.

Monofraps:
Thanks for your comment, I'll update the PR within the next couple of hours.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
```diff
@@ -171,7 +174,9 @@ public static AmqpAppender createAppender(
 			@PluginAttribute("contentEncoding") String contentEncoding,
 			@PluginAttribute("clientConnectionProperties") String clientConnectionProperties,
 			@PluginAttribute("async") boolean async,
-			@PluginAttribute("charset") String charset) {
+			@PluginAttribute("charset") String charset,
+			@PluginAttribute(value = "bufferSize", defaultInt = Integer.MAX_VALUE) int bufferSize,
+			@PluginElement(BlockingQueueFactory.ELEMENT_TYPE) BlockingQueueFactory<Event> blockingQueueFactory) {
```
artembilan: Please, consider do not squash commits in the future.
That way it is easier to track the changes.
We squash them on merge any way.

More over I somehow didn't get a notification about this your change.

Reviewing now... Looks promising!

Monofraps: Alright, I'll stop squashing - I'm not getting updates on this PR either, though.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/ad1e2576ba8a6df942cf41d994a7ce118279aa96 after simple code style polishing.

@Monofraps ,
thank you for contribution again.
Looking forward for more!

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
artembilan: I guess something wrong was with GitHub email server yesterday.
This is not only PR without notification though.

PR #685, opened 2017-11-09
AMQP-780: Configurable consumer start timeout

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-780

Log an error if a consumer doesn't start within the timeout; make the timeout configurable.

This can happen if the task executor doesn't have enough threads to support the concurrency.

__cherry-pick to 1.7.x__


artembilan:
... and cherry-picked to `1.7.x` after resolving conflicts as: https://github.com/spring-projects/spring-amqp/commit/c320c2463eece4c9125921bd0dfcc3fed05f4827

PR #687, opened 2017-11-17
AMQP-784: Fix multi method @RabbitListener

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-784

When looking for matching methods, we matched on no annotation or @Payload.

Match should have been no annotation(s) or not @Header.

__cherry-pick to 1.7.x__


spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java
```diff
@@ -110,6 +110,7 @@
 import org.springframework.core.convert.ConversionService;
 import org.springframework.core.convert.converter.Converter;
 import org.springframework.core.convert.support.DefaultConversionService;
+import org.springframework.lang.NonNull;
```
artembilan: This cannot be backported to `1.7.x`  - Spring `5.0` matter.
Will be replaced with the `@Payload`.

artembilan:
... and cherry-picked as https://github.com/spring-projects/spring-amqp/commit/4f4d0c57de4cb9040823c868f0b14f7157ec3370 after replacing `@NonNull` (Spring 5) to `@Validated` (Spring 4).

PR #695, opened 2017-12-14
AMQP-788 Add delegate publisher connection factory

garyrussell:
#JIRA: https://jira.spring.io/browse/AMQP-788

To avoid deadlocks, it is best to use a different connection for producers and
consumers (unless the producer is partiticipating in a consumer transaction).

- Add a delegate `publisherConnectionFactory` to the `CachingConnecetionFactory`.
- Use the same underlying `com.rabbitmq.client.ConnectionFactory` in each.
- Propagate all properties to the delegate.
 - Except, enhance the connection name and bean name with `.publisher`.
- Add a boolean `usePublisherConnection` to the `RabbitTemplate`.
 - If true, use `createPublisherConnection()` when appropriate.

TODO: Docs (after review)

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
@@ -432,6 +432,11 @@ public abstract class AbstractConnectionFactory implements ConnectionFactory, Di
 		return this.publisherConnectionFactory != null;
 	}
 
+	@Override
+	public ConnectionFactory getPublisherConnectionFactory() {
+		return this.publisherConnectionFactory;
+	}
+
 	protected final Connection createBareConnection() {
 		try {
 			String connectionName = this.connectionNameStrategy.obtainNewConnectionName(this);
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
@@ -416,7 +416,7 @@ public abstract class AbstractConnectionFactory implements ConnectionFactory, Di
 		this.connectionNameStrategy = connectionNameStrategy;
 		if (this.publisherConnectionFactory != null) {
 			this.publisherConnectionFactory.setConnectionNameStrategy(
-					cf -> connectionNameStrategy + PUBLISHER_SUFFIX);
+					cf -> connectionNameStrategy.obtainNewConnectionName(cf) + PUBLISHER_SUFFIX);
 		}
 	}
 
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
@@ -63,6 +63,8 @@ import com.rabbitmq.client.impl.recovery.AutorecoveringConnection;
 public abstract class AbstractConnectionFactory implements ConnectionFactory, DisposableBean, BeanNameAware,
 		ApplicationContextAware, ApplicationEventPublisherAware, ApplicationListener<ContextClosedEvent> {
 
+	private static final String PUBLISHER_SUFFIX = ".publisher";
+
 	public static final int DEFAULT_CLOSE_TIMEOUT = 30000;
 
 	private static final String BAD_URI = "setUri() was passed an invalid URI; it is ignored";
@@ -76,6 +78,8 @@ public abstract class AbstractConnectionFactory implements ConnectionFactory, Di
 	private final CompositeChannelListener channelListener = new CompositeChannelListener();
 
 	private final AtomicInteger defaultConnectionNameStrategyCounter = new AtomicInteger();
+
+	private AbstractConnectionFactory publisherConnectionFactory;
 
 	private RecoveryListener recoveryListener = new RecoveryListener() {
 
@@ -115,7 +119,8 @@ public abstract class AbstractConnectionFactory implements ConnectionFactory, Di
 	private volatile boolean contextStopped;
 
 	/**
-	 * Create a new AbstractConnectionFactory for the given target ConnectionFactory.
+	 * Create a new AbstractConnectionFactory for the given target ConnectionFactory,
+	 * with no publisher connection factory.
 	 * @param rabbitConnectionFactory the target ConnectionFactory
 	 */
 	public AbstractConnectionFactory(com.rabbitmq.client.ConnectionFactory rabbitConnectionFactory) {
@@ -123,9 +128,17 @@ public abstract class AbstractConnectionFactory implements ConnectionFactory, Di
 		this.rabbitConnectionFactory = rabbitConnectionFactory;
 	}
 
+	protected final void setPublisherConnectionFactory(
+			AbstractConnectionFactory publisherConnectionFactory) {
+		this.publisherConnectionFactory = publisherConnectionFactory;
+	}
+
 	@Override
 	public void setApplicationContext(ApplicationContext applicationContext) {
 		this.applicationContext = applicationContext;
+		if (this.publisherConnectionFactory != null) {
+			this.publisherConnectionFactory.setApplicationContext(applicationContext);
+		}
 	}
 
 	protected ApplicationContext getApplicationContext() {
@@ -135,6 +148,9 @@ public abstract class AbstractConnectionFactory implements ConnectionFactory, Di
 	@Override
 	public void setApplicationEventPublisher(ApplicationEventPublisher applicationEventPublisher) {
 		this.applicationEventPublisher = applicationEventPublisher;
+		if (this.publisherConnectionFactory != null) {
+			this.publisherConnectionFactory.setApplicationEventPublisher(applicationEventPublisher);
+		}
 	}
 
 	protected ApplicationEventPublisher getApplicationEventPublisher() {
@@ -145,6 +161,9 @@ public abstract class AbstractConnectionFactory implements ConnectionFactory, Di
 	public void onApplicationEvent(ContextClosedEvent event) {
 		if (getApplicationContext() == event.getApplicationContext()) {
 			this.contextStopped = true;
+		}
+		if (this.publisherConnectionFactory != null) {
+			this.publisherConnectionFactory.onApplicationEvent(event);
 		}
 	}
 
@@ -261,6 +280,9 @@ public abstract class AbstractConnectionFactory implements ConnectionFactory, Di
 			Address[] addressArray = Address.parseAddresses(addresses);
 			if (addressArray.length > 0) {
 				this.addresses = addressArray;
+				if (this.publisherConnectionFactory != null) {
+					this.publisherConnectionFactory.setAddresses(addresses);
+				}
 				return;
 			}
 		}
@@ -288,21 +310,34 @@ public abstract class AbstractConnectionFactory implements ConnectionFactory, Di
 
 	public void setConnectionListeners(List<? extends ConnectionListener> listeners) {
 		this.connectionListener.setDelegates(listeners);
+		if (this.publisherConnectionFactory != null) {
+			this.publisherConnectionFactory.setConnectionListeners(listeners);
+		}
 	}
 
 	@Override
 	public void addConnectionListener(ConnectionListener listener) {
 		this.connectionListener.addDelegate(listener);
+		if (this.publisherConnectionFactory != null) {
+			this.publisherConnectionFactory.addConnectionListener(listener);
+		}
 	}
 
 	@Override
 	public boolean removeConnectionListener(ConnectionListener listener) {
-		return this.connectionListener.removeDelegate(listener);
+		boolean result = this.connectionListener.removeDelegate(listener);
+		if (this.publisherConnectionFactory != null) {
+			this.publisherConnectionFactory.removeConnectionListener(listener); // NOSONAR
+		}
+		return result;
 	}
 
 	@Override
 	public void clearConnectionListeners() {
 		this.connectionListener.clearDelegates();
+		if (this.publisherConnectionFactory != null) {
+			this.publisherConnectionFactory.clearConnectionListeners();
+		}
 	}
 
 	public void setChannelListeners(List<? extends ChannelListener> listeners) {
@@ -316,10 +351,16 @@ public abstract class AbstractConnectionFactory implements ConnectionFactory, Di
 	 */
 	public void setRecoveryListener(RecoveryListener recoveryListener) {
 		this.recoveryListener = recoveryListener;
+		if (this.publisherConnectionFactory != null) {
+			this.publisherConnectionFactory.setRecoveryListener(recoveryListener);
+		}
 	}
 
 	public void addChannelListener(ChannelListener listener) {
 		this.channelListener.addDelegate(listener);
+		if (this.publisherConnectionFactory != null) {
+			this.publisherConnectionFactory.addChannelListener(listener);
+		}
 	}
 
 	/**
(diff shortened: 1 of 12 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -1762,15 +1799,21 @@ public Boolean isMandatoryFor(final Message message) {
...
 				channel = resourceHolder.getChannel();
 				if (channel == null) {
 					ConnectionFactoryUtils.releaseResources(resourceHolder);
 					throw new IllegalStateException("Resource holder returned a null channel");
 				}
 			}
 			else {
-				connection = connectionFactory.createConnection(); // NOSONAR - RabbitUtils
+				if (publisherConnectionIfPossible) {
+					connection = connectionFactory.createPublisherConnection(); // NOSONAR - RabbitUtils
```
artembilan: I think we don't need all this "PublisherConnection" kung-fu, when we just can obtain a `publisherConnectionFactory` from the provided CF if `usePublisherConnection == true`.
Or am I missing anything else?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
```diff
@@ -367,11 +414,22 @@ public int getCloseTimeout() {
...
 
 	@Override
 	public void setBeanName(String name) {
 		this.beanName = name;
+		if (this.publisherConnectionFactory != null) {
+			this.publisherConnectionFactory.setBeanName(name + PUBLISHER_SUFFIX);
+		}
+	}
+
+	public boolean hasPublisherConnectionFactory() {
```
artembilan: I think we may have `PublisherConnectionFactoryAware` marker on this factory and use it from the producers to obtain this `publisherConnectionFactory` instance. See my comment in the `RabbitTemplate`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/ConnectionFactory.java
```diff
@@ -46,4 +46,14 @@
 
 	void clearConnectionListeners();
 
+	/**
+	 * Return a separate connection for publishers (if implemented).
+	 * @return the connection.
+	 * @throws AmqpException an exception.
+	 * @since 2.0.2
+	 */
+	default Connection createPublisherConnection() throws AmqpException {
```
artembilan: I don't think we need this since we could just get direct access to the publisherCF delegate in the producers.

garyrussell:
>What am I missing?

It's very simple (in my mind) and quite different to what I was proposing at SpringOne (SimpleConnection having two delegate connections internally). That solution became very complicated, with regard to caching channels.

This simply uses 2 connection factories so the caches are completely independent.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
garyrussell: I don't think we want to expose the sub-factory externally - otherwise they can configure them differently and the order in which they configure them would matter.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/ConnectionFactory.java
garyrussell: See my comment above.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
garyrussell: Again, comment above.

artembilan:
> otherwise they can configure them differently and the order in which they configure them would matter.

Not sure how it is related to us. They always can configure two different CFs and inject them to different components. That's already end-user code will be. What we suggest here is for convenience: the configuration of two CFs in a single place. They always can do that themselves like we recommend in docs. 
More over how they can re-configure a result of the `public ConnectionFactory getPublisherConnectionFactory()`? Unless they are too cool to cast do the harm for themselves 😄 


I absolutely agree about copy CF in the CCF, but I don't get yet one would we need that `createPublisherConnection()`. That's what I don't like because it breaks CF contract per se.

That's my feeling

garyrussell:
Yes; convinced - it also makes the backport easier 😄 

Thanks!

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
```diff
@@ -174,6 +176,23 @@ public CachingConnectionFactory() {
 		this((String) null);
 	}
 
+	/**
+	 * Create a new CachingConnectionFactory given a host name.
+	 * @param hostname the host name to connect to
```
artembilan: `@since 1.7.6` ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/ConnectionFactoryUtils.java
```diff
@@ -72,6 +72,22 @@ public static boolean isChannelTransactional(Channel channel, ConnectionFactory
...
+	 * @param connectionFactory the ConnectionFactory to obtain a Channel for
+	 * @param synchedLocalTransactionAllowed whether to allow for a local RabbitMQ transaction that is synchronized with
+	 * a Spring-managed transaction (where the main transaction might be a JDBC-based one for a specific DataSource, for
+	 * example), with the RabbitMQ transaction committing right after the main transaction. If not allowed, the given
+	 * ConnectionFactory needs to handle transaction enlistment underneath the covers.
+	 * @param publisherConnectionIfPossible obtain a connection from a separate publisher connection
+	 * if possible.
+	 * @return the transactional Channel, or <code>null</code> if none found
+	 */
+	public static RabbitResourceHolder getTransactionalResourceHolder(final ConnectionFactory connectionFactory,
```
artembilan: Why this code is still here?
I thought the `getPublisherConnectionFactory()` would be enough to distinct the CF instance inside the `RabbitTemplate`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitOperations.java
```diff
@@ -42,6 +42,20 @@
...
+	 * Execute the callback with a channel and reliably close the channel afterwards.
+	 * @param action the call back.
+	 * @param publisherConnectionIfPossible true to use a separate publisher connection if possible.
+	 * @param <T> the return type.
+	 * @return the result from the
+	 * {@link ChannelCallback#doInRabbit(com.rabbitmq.client.Channel)}.
+	 * @throws AmqpException if one occurs.
+	 * @since 2.0.2
+	 */
+	default <T> T execute(ChannelCallback<T> action, boolean publisherConnectionIfPossible) throws AmqpException {
```
artembilan: Why it is still so overcomplicated?
Why there is just not enough to extract the publisherCF in the `RabbitTemplate.afterPropertiesSet()` and everything rest will remain the same as before?

No? Any explanations, please?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
garyrussell: This is an existing CTOR; I moved it to comply with our normal ordering. See the deleted code below.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitOperations.java
garyrussell: Good catch. This is an artifact of some earlier refactoring; it can be removed now.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/ConnectionFactoryUtils.java
garyrussell: No; we can't do that - the connection factory might be a `RoutingConnectionFactory` so there is not one publishing CF.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitOperations.java
garyrussell: Good catch. This is an artifact of some earlier refactoring; it can be removed now. I have to step out for a while; I'll push another commit tonight.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
```diff
@@ -367,11 +414,27 @@ public int getCloseTimeout() {
 	 */
 	public void setConnectionNameStrategy(ConnectionNameStrategy connectionNameStrategy) {
 		this.connectionNameStrategy = connectionNameStrategy;
+		if (this.publisherConnectionFactory != null) {
+			this.publisherConnectionFactory.setConnectionNameStrategy(
+					cf -> connectionNameStrategy + PUBLISHER_SUFFIX);
```
artembilan: Must be:
```
cf -> connectionNameStrategy.obtainNewConnectionName(cf) + PUBLISHER_SUFFIX);
```

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/ConnectionFactoryUtils.java
artembilan: Still unclear how this transactional stuff is related to the `RoutingConnectionFactory`.
In the end you perform:
```
if (publisherConnectionIfPossible) {
			ConnectionFactory publisherFactory = connectionFactory.getPublisherConnectionFactory();
```
But if we select the `publsherCF` before calling this `getTransactionalResourceHolder()`, we don't need the logic around `publisherConnectionIfPossible` here at all. Isn't it?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -801,10 +825,10 @@ protected boolean useDirectReplyTo() {
 		}
 		if (this.replyAddress == null || Address.AMQ_RABBITMQ_REPLY_TO.equals(this.replyAddress)) {
 			try {
-				execute(channel -> {
+				doExecute(channel -> {
 					channel.queueDeclarePassive(Address.AMQ_RABBITMQ_REPLY_TO);
 					return null;
-				});
+				}, true);
```
artembilan: Why is `true` explicit?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -983,7 +1008,7 @@ else if (isChannelTransacted()) {
 				return RabbitTemplate.this.buildMessageFromResponse(response);
 			}
 			return null;
-		}, obtainTargetConnectionFactory(this.receiveConnectionFactorySelectorExpression, queueName));
+		}, obtainTargetConnectionFactory(this.receiveConnectionFactorySelectorExpression, queueName), false);
```
artembilan: Why is `false` here explicit?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -1726,15 +1757,22 @@ public Boolean isMandatoryFor(final Message message) {
 
 	@Override
 	public <T> T execute(ChannelCallback<T> action) {
-		return execute(action, getConnectionFactory());
+		return doExecute(action, false);
```
artembilan: Why `false` again?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -1745,12 +1783,13 @@ public Boolean isMandatoryFor(final Message message) {
...
 		else {
-			return doExecute(action, connectionFactory);
+			return doExecute(action, connectionFactory, publisherConnectionIfPossible);
 		}
 	}
 
 
-	private <T> T doExecute(ChannelCallback<T> action, ConnectionFactory connectionFactory) {
+	private <T> T doExecute(ChannelCallback<T> action, ConnectionFactory connectionFactory,
+			boolean publisherConnectionIfPossible) {
```
artembilan: Why can't we just rely on the `usePublisherConnection` property? Why do we need the method param context?

garyrussell: The rationale I used was to use the publisher channel only for `send()` operations, not `receive()` operations.

Now I remember why I added an overloaded `execute()` operation with the boolean, so the user can choose.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
garyrussell: Because it's a `receive()` operation, not a publish.

garyrussell:
I don't know why GitHub won't allow me to reply to this comment.

>But if we select the publsherCF before calling this getTransactionalResourceHolder(), we don't need the logic around publisherConnectionIfPossible here at all. Isn't it?

No, because when a transaction is in process, the bound channel to use is from the outer connection factory. The `RabbitTemplate` has to lookup the transactional channel based on the container's connection factory, not the publisher CF.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -666,6 +668,28 @@ public void setUserCorrelationId(boolean userCorrelationId) {
 		this.userCorrelationId = userCorrelationId;
 	}
 
+	/**
+	 * True if separate publisher connection(s) are used.
+	 * @return true or false.
+	 * @since 1.7.5
```
artembilan: `@since 1.7.6`, although looking at this complex solution I would stay away from back porting since we always have a workaround as really separately configured `CF` .

I understand that you still think about Rabbit Binder, but I have answer there for you how may we proceed further for the proper fix: https://github.com/spring-cloud/spring-cloud-stream-binder-rabbit/pull/115#issuecomment-351727547

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
artembilan: Well, why?

The current solution and recommendation is to use a separate CCF for the `RabbitTemplate` to avoid dead-lock on the consumer side. This way any template operations are performed on the same injected CF. So, I would expect here consistency as well. Right, I understand that these `receive()` operations are not publishing per se, but does it matter in the end? We have a dead lock in the consumer when producer is blocked on the same connection. But in these `receive()` operations we don't have any producing functionality and in most cases they are in the beginning of the logic. Therefore, IMO, we are good to perform them in any connection and from here our API would be still simple.

WDYT?

garyrussell: OK; you've worn me down.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
garyrussell: >at this complex solution

I don't know why you keep saying it's complex; it's rather simple, I think.

But OK; no backport; go ahead with the binder fix.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/4b78c200012bbe7ac8112bcd4753d0f111de3762

PR #697, opened 2017-12-18
AMQP-794: CRErrorHandler: Traverse cause tree

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-794

In the `ConditionalRejectingErrorHandleri.DefaultExceptionStrategy`,
if the `cause` of the `ListenerExecutionFailedException` is a `MessagingException`,
traverse the `cause` tree to find the root cause of all such exceptions unless
it's a fatal cause itself (`MethodArgumentResolutionException` or `MessageConversionException`).
The final fatal check is performed on the root cause.


garyrussell:
I think we might need to cherry-pick this (for the SCSt binder).

artembilan:
... and cherry-picked to `1.7.x`

PR #698, opened 2017-12-19
AMQP-793: Exception on ack for closed Channel

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-793

Throw an exception to the caller when attempting to ack/nack a
message on a closed channel.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
@@ -96,8 +96,11 @@ public class CachingConnectionFactory extends AbstractConnectionFactory
 
 	private static final int DEFAULT_CHANNEL_CACHE_SIZE = 25;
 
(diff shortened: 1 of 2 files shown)


dyhack:
Is it compatible with spring boot 1.5.6 release ？

garyrussell:
No; it works with boot 2.0 and higher.

Follow the links to [here](https://stackoverflow.com/questions/47831554/is-there-a-reliable-way-to-know-when-a-message-is-successfully-acknowledged/47835319#47835319) for a work around for earlier versions.

By the way, the current boot 1.5.x is 1.5.18 - it is recommended that you keep more up to date with releases.

dyhack:
but ,when I use  spring-rabbit 2.0.3.It ok,Should I do not update the dependies to 2.0.3.RELEASE

garyrussell:
Boot 1.5.x uses spring-rabbit 1.7.x; we don't test 2.0.x with boot 1.5.x; if it works, that's fine.

dyhack:
At present, there is no problem in using it.Thanks

dyhack:
Why don't you fix spring-rabbit 1.7.X?

garyrussell:
We typically only back-port fixes for critical problems that don't have a work-around.

PR #705, opened 2018-01-23
AMQP-796: Fix Admin Transaction

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-796

If an admin uses a transactional `RabbitTemplate` it will start a transaction.
If the connection was opened due to a `RabbitTemplate` operation it should participate
in the same transaction.
Previously, the template used a second channel and treated it as a local transaction.

Also fix the `RabbitAdmin` so it does no work if there is nothing to declare.

__cherry-pick to 1.7.x__

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/ConnectionFactoryUtils.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/ConnectionFactoryUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2017 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


artembilan:
... and cherry-picked as https://github.com/spring-projects/spring-amqp/commit/f786c5b59dfa02e4abda544dad73f98ac26558fb.

I have accepted `main` changes, but rejected some `test` changes since it looks like in `1.7.x` we still don't have `RabbitTemplate`-based ctor in the `RabbitAdmin`, so the reported problem might not be related since we don't start TX from `RabbitAdmin`. At least that is how I feel it. Anyway the changes to the `ConnectionFactoryUtils` are good and won't hurt to prevent any possible issues in the future.

garyrussell:
Right. It is possible a user has a custom listener and does some otger tx template operations. So definitely needs to be in 1.7.

Thanks.

PR #707, opened 2018-02-02
Improve rabbit template

marknorkin:
Hello, Gary and Artem

Please take a look at several small improvements to rabbit template.
You can find explanation for each change in commit messages.

Happy to resolve any comments if you find this PR useful.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
@@ -342,11 +342,9 @@ public class RabbitTemplate extends RabbitAccessor implements BeanFactoryAware, 
 	 * exchange and routing key.
 	 * @param replyAddress the replyAddress to set
 	 */
-	public void setReplyAddress(String replyAddress) {
-		synchronized (this) {
-			this.replyAddress = replyAddress;
-			this.evaluatedFastReplyTo = false;
-		}
+	public synchronized void setReplyAddress(String replyAddress) {
+		this.replyAddress = replyAddress;
+		this.evaluatedFastReplyTo = false;
 	}
 
 	/**
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
@@ -1206,15 +1206,8 @@ public class RabbitTemplate extends RabbitAccessor implements BeanFactoryAware, 
 		catch (TimeoutException e) {
 			// no result in time
 		}
-		try {
-			if (exception == null || !(exception instanceof ConsumerCancelledException)) {
-				channel.basicCancel(consumer.getConsumerTag());
-			}
-		}
-		catch (Exception e) {
-			if (this.logger.isDebugEnabled()) {
-				this.logger.debug("Failed to cancel consumer: " + consumer, e);
-			}
+		if (exception == null || !(exception instanceof ConsumerCancelledException)) {
+			cancelConsumerQuietly(channel, consumer);
 		}
 		if (exception != null) {
 			if (exception instanceof RuntimeException) {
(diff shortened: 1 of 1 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -1627,25 +1627,23 @@ private Message doSendAndReceiveWithDirect(String exchange, String routingKey, M
...
-						container.setTaskExecutor(this.taskExecutor);
-					}
-					if (this.afterReceivePostProcessors != null) {
-						container.setAfterReceivePostProcessors(this.afterReceivePostProcessors
-								.toArray(new MessagePostProcessor[this.afterReceivePostProcessors.size()]));
-					}
-					container.start();
-					this.directReplyToContainers.put(connectionFactory, container);
-					this.replyAddress = Address.AMQ_RABBITMQ_REPLY_TO;
+		DirectReplyToMessageListenerContainer container;
```
artembilan: Would you mind explaining why this change, please?

The idea is like try to get `container` from the map concurrently without any locks.
And only if it is null, go to the `synchronized (this.directReplyToContainers)` and try to do that again.

But now, with your, change we always block all other threads because the first one has decided to go to the `synchronized (this.directReplyToContainers)` block. The problem is going to be exponential for all the waiting threads and actually for nothing, if we have already populated `container` before.

We borrowed this style from the Singleton Pattern implementation: https://www.journaldev.com/1377/java-singleton-design-pattern-best-practices-examples#thread-safe-singleton. This is called "double checked locking principle is used".

What is idea do not do that?

Thanks

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -342,8 +343,10 @@ public String getEncoding() {
 	 * @param replyAddress the replyAddress to set
 	 */
 	public void setReplyAddress(String replyAddress) {
-		this.replyAddress = replyAddress;
-		this.evaluatedFastReplyTo = false;
+		synchronized (this) {
```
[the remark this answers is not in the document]
garyrussell: Yes, won't hurt, but why not synchronize the method instead of `synchronized (this)`.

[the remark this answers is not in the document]
marknorkin: @artembilan yeah, that was the thought process behind the change. just to ensure consistency, even though it most likely be thread confined

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
marknorkin: I understand that this is an attempt to go with fastpath first, but the reasoning behind removing this is that underlying map implementation is not thread safe and so every access (read and write) should be done behind the lock, like in other places, otherwise we could see stale values. Basically it is the same as if we would synchronize setter but not a getter.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -1202,34 +1198,20 @@ private Delivery consumeDelivery(Channel channel, String queueName, long timeout
...
+			Throwable cause = e.getCause();
+			this.logger.error("Consumer failed to receive message: " + consumer, cause);
+			exception = RabbitExceptionTranslator.convertRabbitAccessException(cause);
+			throw exception;
 		}
 		catch (InterruptedException e) {
 			Thread.currentThread().interrupt();
 		}
 		catch (TimeoutException e) {
 			// no result in time
```
marknorkin: @artembilan @garyrussell also I'm not sure why the TimeoutException gets swallowed here and not for example rethrown as in `waitForConfirms`, same applies to InterruptedException, so I left it as is though there will be failing tests if this change will be performed. Just again the thought for consistency.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
artembilan: Yeah... That's good point. 
But again: for consistency with the approach.

However in our case we never see stale value, because we never override it.
Only what we do is cleaning the map during `stop()`.

Does it make sense to keep "unsafe" `get()` ?

Thanks

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
artembilan: I think this was decision of say first and made like this to treat these type of exceptions as "no result".

We are waiting for the `Delivery` object here and if we don't have it in time or our thread is interrupted, let's just return with `null`!

That's only an explanation and this comment confirm that.

We may reconsider this logic, but it is definitely a breaking for the current point version...

marknorkin: @artembilan  thank you for elaborating on this one, more clear to me now.

marknorkin:
Hi, @garyrussell, just would like to kindly ask if there are any updates regarding this changes ?

garyrussell:
@marknorkin I prefer to keep the optimization to avoid grabbing the `directReplyToContainers` monitor unnecessarily. You did not address that in your polishing commit.

garyrussell:
Sorry - didn't see the 3rd commit.

garyrussell:
@artembilan I am ok with this now.

PR #711, opened 2018-02-22
AMQP-800: LocalizedQueueConnectionFactory Fix

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-800

The `LocalizedQueueConnectionFactory` incorrectly destroyed the injected default
connection factory. It should only destroy factories that it creates.

**cherry-pick to 1.7.x**


artembilan:
There is nothing to cherry-pick. That code has been introduced via: https://github.com/spring-projects/spring-amqp/commit/b5f530d80cec0b9851a2a82a9e96d8e17c594e20

So, now I think we need to reconsider some tests to destroy an external `CF` manually.

garyrussell:
Yes; will push. Thanks.

PR #716, opened 2018-03-05
Upgrade to Gradle 4.6

garyrussell:
- add support for JUnit5 tests
- resume gradle testing of JUnit5 tests - `RabbitTemplateMPPIntegrationTests.java`, `SimpleMessageListenerContainerLongTests.java`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -173,7 +173,6 @@ subprojects { subproject ->
 			// restrict memory usage, so tests don't fail with exit code 137 on travis
 			maxHeapSize = '256m'
 		}
-		useJUnitPlatform()
 	}
 
 	task testAll(type: Test, dependsOn: check)
@@ -184,10 +183,6 @@ subprojects { subproject ->
 		}
 	}
 
-	testAll {
-		useJUnitPlatform()
-	}
-
 	tasks.withType(Test).all {
 		// suppress all console output during testing unless running `gradle -i`
 		logging.captureStandardOutput(LogLevel.INFO)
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -183,6 +184,10 @@ subprojects { subproject ->
 		}
 	}
 
+	testAll {
+		useJUnitPlatform()
```
artembilan: Looks like this can go to the next `tasks.withType(Test).all` block.

build.gradle
```diff
@@ -89,9 +89,9 @@ subprojects { subproject ->
 		hamcrestVersion = '1.3'
 		jackson2Version = '2.9.1'
 		junit4Version = '4.12'
-		junitJupiterVersion = '5.0.2'
-		junitPlatformVersion = '1.0.2'
-		junitVintageVersion = '4.12.2'
+		junitJupiterVersion = '5.1.0'
+		junitPlatformVersion = '1.1.0'
```
artembilan: Is it OK to upgrade to the `minor` version in our point release? 
Junit is not only a test dependency. The Cairo platform must provide these versions as well.

What am I missing in our backward compatibility policy?

Thanks

garyrussell: Cairo gets its versions from [boot](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-dependencies/pom.xml#L114) so it seems we were wrong anyway.

garyrussell: Yes, it's a compile dependency for spring-rabbit-junit, but `optional` too; so I think we're ok.

artembilan:
Good, but how about those two test classes. 
I still don't understand how your fix is related to them.

Thanks

garyrussell:
Sorry, I missed your earlier comment.

Since they are JUnit5 tests, they have not been running in the gradle build. I was just commenting that they are now being tested again after this change is applied.

PR #718, opened 2018-03-08
AMQP-801-2: Introduce ConsumerDecorator

artembilan:
JIRA: https://jira.spring.io/browse/AMQP-801

To properly assign the queue to the `ConsumeOkEvent`, we need perform
such a logic in the `Consumer.handleConsumeOk()`.

* Introduce `BlockingQueueConsumer.ConsumerDecorator` to be created on
each `channel.basicConsume()` for wrapping the target `InternalConsumer`
per queue
* Add getters to the `ConsumeOkEvent` for better interoperability
* Assert assigned queue names for the `ConsumeOkEvent`s in the
`SimpleMessageListenerContainerIntegration2Tests`

**Cherry-pick to 1.7.x**

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
@@ -63,6 +63,7 @@ import org.springframework.util.backoff.BackOffExecution;
 import com.rabbitmq.client.AMQP;
 import com.rabbitmq.client.AlreadyClosedException;
 import com.rabbitmq.client.Channel;
+import com.rabbitmq.client.Consumer;
 import com.rabbitmq.client.DefaultConsumer;
 import com.rabbitmq.client.Envelope;
 import com.rabbitmq.client.Recoverable;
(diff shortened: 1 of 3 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
```diff
@@ -952,6 +950,52 @@ public void handleDelivery(String consumerTag, Envelope envelope, AMQP.BasicProp
...
+			this.delegate.handleShutdownSignal(consumerTag, sig);
+		}
+
+		public void handleCancel(String consumerTag) throws IOException {
+			this.delegate.handleCancel(consumerTag);
+		}
+
+		public void handleCancelOk(String consumerTag) {
+			this.delegate.handleCancelOk(consumerTag);
+		}
```
garyrussell: Perhaps we should also store the tag in a field; if only so that it's visible in a debugger (save having to look it up in the map). Given that the tag in the delegate gets overwritten when there are multiple queues.

Also implement `toString()` with the queue and tag (again, just for a debugger).

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/ConsumeOkEvent.java
```diff
@@ -20,6 +20,8 @@
 
 /**
  * @author Gary Russell
```
garyrussell: Hmmm - shouldn't checkstyle bark at no description?

artembilan: I think we don't have it enabled too many classes are without JavaDocs.

```
<module name="JavadocStyle">
	<property name="checkEmptyJavadoc" value="false"/>
</module>
```
That's what we have in the `checkstyle.xml`

garyrussell: ok; yeah; I guess we have it in the newer projects.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/5ce8529738daf5a6d0eeb1c1cb1f4275c5af02f8 and cherry-picked

Issue #719, opened 2018-03-09
MessagePostprocessor seems not to get called on reply


konraifen88:
I am trying to invoke an `MessagePostProcessor` to apply/verify a header before sending/after receiving an message.

I have managed to add the PostProcessor functionality for following situations:
* Before send (in publisher)
* After receive (in publisher)
* After receive (in consumer with an AdviceChain, but I do not think that this is the meant sollution)

But I am not able to set it up for sending reply. When looking into the code I can see that the reply is handled inside the `AbstractAdaptableMessageListener` where an `replyPostProcessor` is available.

But this `replyPostProcessor` never is set, so it is always null and because of that never executed.

Little bit more detail:
The `AbstractAdaptableMessageListener` is created in the `MethodRabbitListenerEndpoint.createMessageListener` method. After creating a new `MessagingMessageListenerAdapter` `replyToAddress`, `messageConverter` and `beanResolver` are set and the listener is returned. I think that here the `replyPostProcessor` also should be set, too.

But I do not have this great understanding of the code to say this for sure.

The version I am using is 1.7.6.RELEASE

artembilan:
Yeah... Looks like this is an omission on the Framework side.

Has to be fixed.

As a workaround I suggest to return the whole `org.springframework.amqp.core.Message` from your `@RabbitListener` method.

Or extend `SimpleRabbitListenerContainerFactory` and in the end of overloaded `initializeContainer` call the `SimpleMessageListenerContainer.getListener()`, cast it to the `AbstractAdaptableMessageListener` and then inject `MessagePostProcessor`  via the mentioned `setReplyPostProcessor()`.

garyrussell:
>After receive (in consumer with an AdviceChain, but I do not think that this is the meant sollution)

In 2.0 we added `setAfterReceivePostProcessors` to the container factory.

If we fix this (replyMPP) in 1.7.x, we should probably back port that too.

konraifen88:
Thanks for the quick reply. For the moment I've managed to do my post processing during, or better before/after the message conversion. For that I've created my own `Jackson2JsonMessageConverter` and add/process my header in this step.

garyrussell:
@konraifen88 - please open a [JIRA "Improvement" issue](https://jira.spring.io/browse/AMQP) against Spring AMQP; when that is implemented, we'd then need a change to Spring Boot, to auto configure them. They can probably be simple "marker" interfaces that extend `MessagePostProcessor`.

[...]

PR #720, opened 2018-03-12
GH-719: Container Factory: Add replyPostProcessor

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/719

Add `replyPostProcessor` to the container factory.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2017 the original author or authors.
+ * Copyright 2014-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
```diff
@@ -173,12 +175,21 @@ public void setMessageConverter(MessageConverter messageConverter) {
...
+	 * Set a post processor to process the reply immediately before
+	 * {@code Channel#basicPublish()}. Often used to compress the data.
 	 * @param replyPostProcessor the reply post processor.
+	 * @deprecated in favor of
+	 * {@link #setBeforeSendReplyPostProcessors(MessagePostProcessor...)}.
 	 */
+	@Deprecated
 	public void setReplyPostProcessor(MessagePostProcessor replyPostProcessor) {
-		this.replyPostProcessor = replyPostProcessor;
+		this.beforeSendReplyPostProcessors = new MessagePostProcessor[] { replyPostProcessor };
```
artembilan: I think we can just call `setBeforeSendReplyPostProcessors(replyPostProcessor)` from here for consistency.

And that new `setBeforeSendReplyPostProcessors()` deserves some JavaDocs.

artembilan:
Do we back-port this fix?

garyrussell:
I don't think so; he has a workaround via the `MessageConverter`.

PR #721, opened 2018-03-13
AMQP-798: Master to 2.1.x; Fix Tangles

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-798

Also update amqp-client to 5.2.0

Don't use `MessageConverter` in `Message.toString()` (cycle)

Fix cycle between template and admin (via containers)

Fix outbound reference from support to connection

Fix cycle between connection and listener via RabbitUtils

Fix cycle between listener and core via ChannelAwareMessageListener and admin

Fix cycle between listener.adapter and listener via RabbitListenerErrorHandler
Fix cycle between amqp.support and amqp.core via Correlation

Fix cycle between listener and transaction via ListenerFailedRuleBasedTransactionAttribute

Fix Tests


artembilan:
There are a couple failed tests:
```
org.springframework.amqp.rabbit.listener.ContainerShutDownTests > testUninterruptibleListenerDMLC FAILED
    java.lang.AssertionError at ContainerShutDownTests.java:96
org.springframework.amqp.rabbit.listener.ContainerShutDownTests > testUninterruptibleListenerSMLC FAILED
    java.lang.AssertionError at ContainerShutDownTests.java:93
```
See Travis report

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitUtils.java
```diff
@@ -324,21 +324,11 @@ public static boolean isExchangeDeclarationFailure(Exception e) {
...
-			if (t instanceof AmqpRejectAndDontRequeueException) {
-				shouldRequeue = false;
-			}
-			t = t.getCause();
-		}
-		if (logger.isDebugEnabled()) {
-			logger.debug("Rejecting messages (requeue=" + shouldRequeue + ")");
-		}
-		return shouldRequeue;
+		throw new UnsupportedOperationException("Use ContainerUtils");
```
artembilan: Why can't we delegate here to the mentioned `ontainerUtils.shouldRequeue()` then?
This seems for me as breaking change now...

I think even if we need a reflection here, it still would be better to keep functionality unbroken.

Although we are moving stuff around anyway to fix tangles... :smile:

src/reference/asciidoc/whats-new.adoc
```diff
@@ -1,180 +1,20 @@
...
 === What's New
 
-==== Changes in 2.0 Since 1.7
-
-===== CachingConnectionFactory
-
-Starting with _version 2.0.2_, the `RabbitTemplate` can be configured to use a different connection to that used by listener containers.
-This is to avoid deadlocked consumers when producers are blocked for any reason.
-See <<separate-connection>> for more information.
+==== Changes in 2.1 Since 2.1
```
artembilan: Typo: Since `2.0`, I guess :smile: 

I think we need to do similar refactoring for SI as well

artembilan:
BTW we also need to rename XSD to version `2.1`.

garyrussell:
screen shot 2018-03-14 at 11 11 15 am

garyrussell:
Polishing looks good; thanks (one comment made).

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/4464d4d8a5397e90807294f45b33c3063adae6a0 with addressing mentioned comment about deprecating `setRabbitAdmin()` in favor of new `setAmqpAdmin()`.

PR #726, opened 2018-03-19
Update RabbitAdmin.java

ooooow:
java.lang.NullPointerException

```kotlin
demo : Map<String,Collection> map = new HashMap<>();
        map.put("AAA",new ArrayList<>());
        map.put("BBB",null);
        Collection<Collection> collections = map.values();
        System.out.println(collections.size());
        for (Collection<?> collection : collections) {
            if(collection.size()>0){
                System.out.println(collection.size());
            }
        }
```

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitAdmin.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitAdmin.java
@@ -435,7 +435,8 @@ public class RabbitAdmin implements AmqpAdmin, ApplicationContextAware, Applicat
 		Collection<Collection> collections = this.applicationContext.getBeansOfType(Collection.class, false, false)
 				.values();
 		for (Collection<?> collection : collections) {
-			if (collection.size() > 0 && collection.iterator().next() instanceof Declarable) {
+			//NullPointerException
+			if (collection != null && collection.size() > 0 && collection.iterator().next() instanceof Declarable) {
 				for (Object declarable : collection) {
 					if (declarable instanceof Exchange) {
 						contextExchanges.add((Exchange) declarable);


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitAdmin.java
```diff
@@ -435,7 +435,8 @@ public void initialize() {
 		Collection<Collection> collections = this.applicationContext.getBeansOfType(Collection.class, false, false)
 				.values();
 		for (Collection<?> collection : collections) {
-			if (collection.size() > 0 && collection.iterator().next() instanceof Declarable) {
+			//NullPointerException
+			if (collection != null && collection.size() > 0 && collection.iterator().next() instanceof Declarable) {
```
panchenko: @ooooow Could you please elaborate update what circumstances `getBeansOfType()` can return `null` bean instances?

ooooow: Suppose there is a bean below:
@Bean
    public List<String> initList(){
        List<String> stringList = null;
        if(1==2){
            stringList.add("1111");
        }
        return stringList;
    }

ooooow: if(1==2)  is a  example

garyrussell: It is invalid to define a `null` bean

artembilan:
@ooooow ,

Looks like your PR is going to be closed as *Invalid* without an appropriate feedback from you.
We are going to release `2.0.3` in a couple days and that would be great to incorporate a fix for some issue if that has a place. Otherwise - sorry.

Anyway there are many things to contribute to.
So, you are always welcome to help us this or other way!

artembilan:
Closed as `Invalid` with unreasonable changes and with the lack of response.

Thanks for understanding!

Issue #722, opened 2018-03-15
Make AmqpAppender.Event public


Romster:
In the official documentation  it is recommended to override [AmqpAppender#postProcessMessageBeforeSend](https://docs.spring.io/spring-amqp/api/org/springframework/amqp/rabbit/logback/AmqpAppender.html#postProcessMessageBeforeSend-org.springframework.amqp.core.Message-org.springframework.amqp.rabbit.logback.AmqpAppender.Event-) to customise the message.
But the second parameter of this method is an [AmqpAppender.Event](https://docs.spring.io/spring-amqp/api/org/springframework/amqp/rabbit/logback/AmqpAppender.Event.html), which is declared as _protected static_.
It seems strange to use a class with that declaration in a public method, which is designed to be overridden in client code, don't you think so?

Also I faced [the problem ](https://stackoverflow.com/questions/49284094/inheritance-from-java-class-with-a-public-method-accepting-a-protected-class-in) when I tried to implement my AmqpAppender in Kotlin.

garyrussell:
>It seems strange to use a class with that declaration in a public method, which is designed to be overridden in client code, don't you think so?

Why would it seem "strange" when it works fine in java?

```java
@SpringBootApplication
public class Rgh722Application {

	public static void main(String[] args) {
		SpringApplication.run(Rgh722Application.class, args);
	}

	public static class MyAppender extends AmqpAppender {

		@Override
		public Message postProcessMessageBeforeSend(Message message, Event event) {
			message.getMessageProperties().setHeader("foo", "foobar");
			return message;
		}

	}

}
```

The class wasn't designed for use in Kotlin; we can certainly open it up to make it usable in Kotlin, given that it has different semantics to Java. [Contributions are welcome](https://github.com/spring-projects/spring-amqp/blob/master/CONTRIBUTING.adoc).

Romster:
@garyrussell Well.
Maybe it seems strange to me because I haven't written in Java for some time.
But it's great if you are ready to open for Kotlin interoperability. And I would be glad to contribute this change, even if it looks insignificant.

artembilan:
> But it's great if you are ready to open for Kotlin interoperability

Well, @Romster , I would suggest you to ask Kotlin guys on the matter to determine why it isn't interoperable with Java. This is JVM language, so what is the point to fix foundation to tolerate a flaw in the target implementation? 

IMO, the problem is in Kotlin if it isn't compatible with Java.

[...]

PR #728, opened 2018-03-22
Made AmqpAppender.postProcessMessageBeforeSend  protected

Romster:
[Jira](https://jira.spring.io/browse/AMQP-805)
[Issue](https://github.com/spring-projects/spring-amqp/issues/722)

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
@@ -231,7 +231,7 @@ public class AmqpAppender extends AbstractAppender {
 	 * @param event The event.
 	 * @return The modified message.
 	 */
-	public Message postProcessMessageBeforeSend(Message message, Event event) {
+	protected Message postProcessMessageBeforeSend(Message message, Event event) {
 		return message;
 	}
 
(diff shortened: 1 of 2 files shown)


artembilan:
@Romster ,

Thank you for contribution!

Why did you do the Pull Request against `1.7.x` branch? Wouldn't you like to have it in the `2.1` or at least in the `2.0.x`? I mean why PR is not against `master`? If we need to back port the fix, there is just need to ask to do that. Otherwise it sounds like only the fix for `1.7.x`. Am I missing anything? Thanks

artembilan:
I think the problem was really in a mix of `public` and `protected`.

Yes, it compiles in Java, indeed, but you can't use that `public` method outside of the class because it requires a `protected` type. That's really wrong from usage perspective, but we all live with that in Java for a while. :smile: 

Although I still think that Kotlin has to follow Postel's Law and be tolerant to Java classes it uses.

artembilan:
Thanks, @Romster , for a quick turnaround and explanation!

Well, so far so good: your change not so big to be cherry-picked to `master` and `2.0.x`.

I just wanted to know if we should do that or you had something specific in the mind for them.
Nothing more: just wanted do not confuse if I would going to cherry-pick without confirmation.

So, LGMT, merging...
(With commit message polishing :smile: )

Romster:
@artembilan I'm sorry - what does LGMT mean?:) Never saw this abbreviation.

artembilan:
Oh! My bad. Typo. I wanted to say: Looks Good To Me. 
Gotcha! :smile:

Romster:
@artembilan XD
Great! Thank you! Should I close my issue or do anything else?

artembilan:
No. You are good to go.
Although, wait!

There are many other things requiring a contribution.
So, please, stay and PR them! :wink: 

Thank you so far!

artembilan:
... and cherry-picked to `master` and `2.0.x`

PR #731, opened 2018-03-28
AMQP-806: Simple Publisher Confirms

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-806

Avoid complex infrastructure for simple use cases.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
@@ -96,6 +96,7 @@ import com.rabbitmq.client.AMQP;
 import com.rabbitmq.client.AMQP.BasicProperties;
 import com.rabbitmq.client.AMQP.Queue.DeclareOk;
 import com.rabbitmq.client.Channel;
+import com.rabbitmq.client.ConfirmListener;
 import com.rabbitmq.client.DefaultConsumer;
 import com.rabbitmq.client.Envelope;
 import com.rabbitmq.client.GetResponse;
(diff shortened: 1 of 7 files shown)


src/reference/asciidoc/amqp.adoc
```diff
@@ -996,6 +998,37 @@ For example, when running on a transacted listener container thread and performi
...
+Furthermore, the `RabbitTemplate` sets the property `publisherSequenceNumber` in the sent message `MessageProperties`.
+If you wish to check (or log, etc) specific confirmations, you can do so with an overloaded `invoke` method:
+
+[source, java]
+----
+public <T> T invoke(OperationsCallback<T> action, com.rabbitmq.client.ConfirmCallback acks,
+        com.rabbitmq.client.ConfirmCallback nacks);
+----
+
+NOTE: These `ConfirmCallback` objects (for acks and nacks) are the Rabbit client callbacks, not the template callback.
```
artembilan: Why we really can't use there our own `org.springframework.amqp.rabbit.core.RabbitTemplate.ConfirmCallback` ?
It doesn't look too hard to delegate calls from the internal Rabbit client `ConfirmCallback` instances.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -1848,6 +1849,21 @@ public Boolean isMandatoryFor(final Message message) {
...
+	}
+
+	/**
+	 * Invoke operations on the same channel.
+	 * If callbacks are needed, both callbacks must be supplied.
+	 * @param action the callback.
+	 * @param acks a confirm callback for acks.
+	 * @param nacks a confirm callback for nacks.
+	 * @param <T> the return type.
+	 * @return the result of the action method.
```
artembilan: `@since 2.1` ?

src/reference/asciidoc/amqp.adoc
garyrussell: It doesn't really fit; we don't have correlation data or a cause or a way to convey `multiple`.

I suppose I could subclass `CorrelationData` to provide the tag and `multiple`. WDYT?

artembilan: Ah! I see what you mean.
No, I think it's really not necessary to overhead here: since we don't have anything to wait for, therefore we don't need correlation.

So, the current solution is fine.

Will pull locally for deeper review and possible merging...

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/3da5931567b69af5e793ba41f3bf1cd41207457e

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
garyrussell: We need to add this method to `RabbitOperations` - can you do that if you find no more problems in the review?

garyrussell: Oops - too late.

artembilan: I think we can just add subsequent commit.
Not a big deal though.

Thanks for the pointer! My bad :smile:

artembilan: Addressed via https://github.com/spring-projects/spring-amqp/commit/bf5498db8025be7c604b6a55804145de7e5a47d2

PR #733, opened 2018-03-30
AMQP-807: Use method genericReturnType() for conversion

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-807

When converting `@RabbitListener` results, use the generic return type
of the method, instead of `object.getClass()` to properly convey
type information in message headers.

Otherwise, `List<Foo>` becomes `List<Map>` on the receiving side, since
the list content type is erased.

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/support/converter/AbstractMessageConverter.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/support/converter/AbstractMessageConverter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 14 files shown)


artembilan:
Looks like too much changes for such a simple problem.
Would you mind to elaborate what else you did here?
Meanwhile I pull it locally: really some points are unclear...
Thanks

spring-amqp/src/main/java/org/springframework/amqp/support/converter/AbstractMessageConverter.java
```diff
@@ -50,26 +52,46 @@ protected boolean isCreateMessageIds() {
...
 	 * @return a message
+	 * @since 2.1
 	 */
-	protected abstract Message createMessage(Object object, MessageProperties messageProperties);
+	protected Message createMessage(Object object, MessageProperties messageProperties, @Nullable Type genericType) {
+		return createMessage(object, messageProperties);
+	}
 
 	@Override
 	public abstract Object fromMessage(Message message) throws MessageConversionException;
```
artembilan: Sounds like we don't this method declaration in this class.
The same meaning we get from the an interface contract.

garyrussell:
Yes; you are correct; it can be removed (legacy code).

The change looks more than it really is because I had to refactor the invocations to always return an `InvocationResult` (which used to be `ResultHolder` that was only returned from a multi-method listener).

I then had to make it a top level class, to avoid a class tangle.

I should have added this information to the commit comment; my apologies.

artembilan:
OK. I understand now. No problem.
Merging with mentioned polishing and more info (your comment) into the commit message...

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/38f59a2537e835e1b9e444ed089a8bfae9fa9588

Issue #734, opened 2018-03-30
IllegalStateException on Transaction Scoped Beans


seandcrouse:
https://github.com/spring-projects/spring-amqp/blob/bf5498db8025be7c604b6a55804145de7e5a47d2/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitAdmin.java#L438

Beans using org.springframework.transaction.support.SimpleTransactionScope fail to get instantiated because RabbitAdmin is not operating inside a transaction.  I have a message producer that filters out duplicate messages per transaction.  This is done with a transaction scoped Set bean.  I know that my producer will only use this bean within that transaction.  However, RabbitAdmin is getting every Collection bean, calling the size() method, which causes the class scoped proxy to try to instantiate the bean.  This fails because it's not happening inside a transaction.

Sample code:

```java
    public static String TRANSACTION_SCOPE = "transaction";
    
    @Bean(name = "sentBundles")
    @Scope(value = TRANSACTION_SCOPE, proxyMode = ScopedProxyMode.TARGET_CLASS)
    public Set<MessageBundle> sentBundles() {
        return new HashSet<>();
    }

    @Bean
    public static CustomScopeConfigurer customScope() {
        CustomScopeConfigurer configurer = new CustomScopeConfigurer();
        configurer.addScope(TRANSACTION_SCOPE, new SimpleTransactionScope());
        return configurer;
    }
```

garyrussell:
I am not sure there's much we can do about this, except, perhaps, add a boolean property to the admin to disable provisioning of `Collection<Declarable>`.

In the meantime, if you can live without the Admin declaring the exchanges/queues/bindings automatically, set its `autoStartup` property to `false`; you would then have to declare your elements yourself using the Admin's API.

artembilan:
I think this is exactly the way we should go in the Framework. 
Since we can't predict what end-user can do with his/her beans, we have only the way as simple `Declarables` object with the some fluent builder API to add instances to declare in the end.
And, therefore, deprecate `Collection` approach to avoid similar problems in the future.
Plus the custom type will make the solution robust.

WDYT?

garyrussell:
Issued a PR with the boolean; we can consider a strong type in future.

That said, @seandcrouse - what version are you using? We already suppressed fetching non-singletons here: https://github.com/spring-projects/spring-amqp/pull/406

seandcrouse:
We're using:

[...]

PR #735, opened 2018-03-30
GH-734: Option to suppress declaring Collections

garyrussell:
Fixes https://github.com/spring-projects/spring-amqp/issues/734

Add `declareCollections` flag to admin (default false).

__cherry-pick to 2.0.x, 1.7.x__


artembilan:
... and cherry-picked to `2.0.x` and `1.7.x`.

Let's reconsider for strong type in the next version!

PR #742, opened 2018-04-17
[For Team Consideration/Discussion] Add Kotlin test case

garyrussell:
Given the perceived increase in Kotlin users, perhaps we should consider adding some test cases.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 buildscript {
-	ext.kotlin_version = '1.2.31'
+	ext.kotlinVersion = '1.2.31'
 	repositories {
 		maven { url 'https://repo.spring.io/plugins-release' }
 	}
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -1,11 +1,13 @@
 buildscript {
+	ext.kotlin_version = '1.2.31'
```
artembilan: I would prefer `kotlinVersion` for consistency with all other similar variables.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitKotlinTests.kt
```diff
@@ -0,0 +1,74 @@
...
+import org.springframework.amqp.rabbit.junit.RabbitAvailableCondition
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import org.springframework.test.annotation.DirtiesContext
+import org.springframework.test.context.junit.jupiter.SpringJUnitConfig
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+
+@SpringJUnitConfig
```
artembilan: JavaDocs?
At least with your name and `@since`

Also: shound't this class be placed into the `kotlin` faceset instead of `java`?

build.gradle
garyrussell: Good catch - this was generated by IDEA 😄

artembilan: Ha! Don't tell me that STS doesn't support Kotlin yet! :wink:

garyrussell: It's buggy - I figured the support in IDEA would be first class.

PR #754, opened 2018-05-23
Changes to build with JDK10

garyrussell:
- update jaCoCo version
- new `Integer(n)` is deprecated
- @PreDestroy no longer present
- SSL change

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -95,7 +95,7 @@ subprojects { subproject ->
 	}
 
 	jacoco {
-		toolVersion = '0.7.9'
+		toolVersion = '0.8.1'
 	}
 
 	// dependencies that are common across all java projects
--- a/spring-amqp/src/test/java/org/springframework/amqp/support/SimpleAmqpHeaderMapperTests.java
+++ b/spring-amqp/src/test/java/org/springframework/amqp/support/SimpleAmqpHeaderMapperTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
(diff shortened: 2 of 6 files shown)


artembilan:
> `@PreDestroy` no longer present

That's funny! That's not what Java was all these years... :smile: 

Should I download and install JDK 10 locally as well?

`javax.annotation-api` must be there is CP, though: https://stackoverflow.com/questions/46502001/how-to-get-access-to-javax-annotation-resource-at-run-time-in-java-9.

artembilan:
I wonder if we have to do anything with the Travis config to rely on Java 10 already, but still check compatibility with Java 8.
Although that should be guaranteed by the `sourceCompatibility` and `targetCompatibility` Gradle props.

garyrussell:
>Should I download and install JDK 10 locally as well?

Up to you; I wasn't planning to change the build to use JDK10; I just wanted to make sure our tests pass with it.

OTOH, I suppose changing the build will ensure we don't introduce any new problems.

artembilan:
>  I suppose changing the build will ensure we don't introduce any new problems.

That's correct. So, or Travis or CI plan must be changed to JDK 10 therefore.

I guess we have more similar issues with SI :smile: . However that is already a different story...

garyrussell:
Only had to change JaCoCo for kafka; checking SI now...

artembilan:
One more time:

Are we going to do anything with CI plan or Travis config?

Let me know and this will be ready for merge!

garyrussell:
Looks like it's not supported on travis yet: https://github.com/travis-ci/travis-ci/issues/9368

Let's go ahead and merge; we can defer the build change.

PR #755, opened 2018-05-25
AMQP-816; Support consume from broker-named queues

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-816

- Admin now populates the declared name on the `Queue`
- Containers now maintain a `Queue` list instead of a list of names
- Containers use the declared name if the defined name is empty
- Don't allow adding broker-named queues when the container is running
- Add code to the DMLC to check the name change; not needed for SMLC since
   it refetches the queue names during restart attempts

**cherry-pick to 2.0.x**


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -318,16 +353,31 @@ public final void addQueues(Queue... queues) {
 	public boolean removeQueueNames(String... queueNames) {
 		Assert.notNull(queueNames, "'queueNames' cannot be null");
 		Assert.noNullElements(queueNames, "'queueNames' cannot contain null elements");
-		return this.queueNames.removeAll(Arrays.asList(queueNames));
+		AtomicBoolean result = new AtomicBoolean();
+		if (this.queues.size() > 0) {
+			Set<String> toRemove = new HashSet<>(Arrays.asList(queueNames));
+			this.queues.stream().forEach(q -> {
+				if (toRemove.contains(q.getName()) || toRemove.contains(q.getDeclaredName())) {
+					this.queues.remove(q);
```
artembilan: Is it really safe to remove from the collection during its iteration?
Seems for me in other places we use `Iterator` on the matter.

On the other hand there is a `removeIf()` on the `Collection`:
```
 /**
     * Removes all of the elements of this collection that satisfy the given
     * predicate.  Errors or runtime exceptions thrown during iteration or by
     * the predicate are relayed to the caller.
```

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -318,16 +353,31 @@ public final void addQueues(Queue... queues) {
...
 	 */
-	public final boolean removeQueues(Queue... queues) {
-		return removeQueueNames(collectQueueNames(queues));
+	public boolean removeQueues(Queue... queues) {
+		Assert.notNull(queues, "'queues' cannot be null");
+		Assert.noNullElements(queues, "'queues' cannot contain null elements");
+		return removeQueueNames(Arrays.stream(queues)
+				.map(q -> q.getName())
+				.collect(Collectors.toList())
+				.toArray(new String[0]));
```
artembilan: `toArray(String[]::new)` instead of these two operators ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -587,7 +637,11 @@ public void setForceCloseChannel(boolean forceCloseChannel) {
 	 */
 	protected String getRoutingLookupKey() {
 		return super.getConnectionFactory() instanceof RoutingConnectionFactory
-				? (this.lookupKeyQualifier + this.queueNames.toString().replaceAll(" ", ""))
+				? (this.lookupKeyQualifier + this.queues.stream()
+								.map(q -> q.getName())
+								.collect(Collectors.toList())
+							.toString()
+							.replaceAll(" ", ""))
```
artembilan: Maybe `Collectors.joining(",")` instead?

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/BrokerDeclaredQueueNameTests.java
```diff
@@ -0,0 +1,200 @@
...
+		testBrokerNamedQueue(this.dmlc, this.latch3, this.latch4, this.queue2);
+	}
+
+	private void testBrokerNamedQueue(AbstractMessageListenerContainer container,
+			CountDownLatch latch1, CountDownLatch latch2, Queue queue) throws Exception {
+		container.start();
+		String firstDeclaredName = queue.getDeclaredName();
+		this.template.convertAndSend(firstDeclaredName, "foo");
+		assertThat(latch1.await(10, TimeUnit.SECONDS)).isTrue();
+		assertThat(this.message.get().getBody()).isEqualTo("foo".getBytes());
```
artembilan: You don't reset this `AtomicReference` in between tests: some stale data may remain.

spring-amqp/src/main/java/org/springframework/amqp/core/Queue.java
```diff
@@ -39,6 +39,9 @@
 
 	private final java.util.Map<java.lang.String, java.lang.Object> arguments;
 
+	private volatile String declaredName = AnonymousQueue.Base64UrlNamingStrategy.DEFAULT.generateName()
```
artembilan: Why can't we re-use an existing `name`?
If it is volatile `Queue` object, it doesn't matter what we have set to the property, but when it is an active, broker-connected entity, we will get a declared name there.

Anything specific in a favor of the separate property?
Thanks

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
garyrussell: Good catch.

spring-amqp/src/main/java/org/springframework/amqp/core/Queue.java
garyrussell: When the connection is lost and re-established, we can't declare the queue with a broker-declared name (illegal to name a queue starting with `amq.gen`. So we need to keep name as `""` so the broker will give us another new name.

Also, with the DMLC, I needed a unique value here in case the container is started before the queues could be declared - perhaps because the broker is down.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitAdmin.java
```diff
@@ -239,7 +240,11 @@ public String declareQueue(final Queue queue) {
 		try {
 			return this.rabbitTemplate.execute(channel -> {
 				DeclareOk[] declared = declareQueues(channel, queue);
-				return declared.length > 0 ? declared[0].getQueue() : null;
+				String declaredName = declared.length > 0 ? declared[0].getQueue() : null;
+				if (StringUtils.hasText(declaredName)) {
+					queue.setDeclaredName(declaredName);
```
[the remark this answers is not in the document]
garyrussell: Doh

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -318,16 +353,31 @@ public final void addQueues(Queue... queues) {
...
 	 * @param queues The queue(s) to remove.
 	 * @return the boolean result of removal on the target {@code queueNames} List.
 	 */
-	public final boolean removeQueues(Queue... queues) {
-		return removeQueueNames(collectQueueNames(queues));
+	public boolean removeQueues(Queue... queues) {
+		Assert.notNull(queues, "'queues' cannot be null");
+		Assert.noNullElements(queues, "'queues' cannot contain null elements");
+		return removeQueueNames(Arrays.stream(queues)
+				.map(q -> q.getName())
```
artembilan: Well, that's confusing. In other places you also consult a `delcaredName` property. Why not here?
More over we have just realized that for broker-based names we always are going to have the `name` as an empty string.

Would you mind to elaborate more what's going on here?

Something similar I see in the `DMLC` - we remove using the `name` and don't check a `declaredName`

spring-amqp/src/main/java/org/springframework/amqp/core/Queue.java
```diff
@@ -145,6 +151,16 @@ public String getDeclaredName() {
...
 	}
 
+	/**
+	 * Return the name provided to the constructor or the broker-generated name
+	 * if that name is an empty String.
+	 * @return the name.
+	 * @since 2.0.4
+	 */
+	public String getActualName() {
+		return StringUtils.hasText(this.name) ? this.name : this.declaredName;
```
artembilan: Can't we have a property on the matter and update it when `setDeclaredName()` is called?
That's might be over-optimization, but still - curious what is your opinion :smile:

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -353,17 +345,12 @@ public void addQueues(Queue... queues) {
 	public boolean removeQueueNames(String... queueNames) {
 		Assert.notNull(queueNames, "'queueNames' cannot be null");
 		Assert.noNullElements(queueNames, "'queueNames' cannot contain null elements");
-		AtomicBoolean result = new AtomicBoolean();
 		if (this.queues.size() > 0) {
```
artembilan: Why do we still need this `if`? Won't `removeIf()` returns `false` if there is nothing removed

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -637,11 +623,10 @@ public void setForceCloseChannel(boolean forceCloseChannel) {
...
 	protected String getRoutingLookupKey() {
 		return super.getConnectionFactory() instanceof RoutingConnectionFactory
-				? (this.lookupKeyQualifier + this.queues.stream()
+				? (this.lookupKeyQualifier + "[" + this.queues.stream()
 								.map(q -> q.getName())
-								.collect(Collectors.toList())
-							.toString()
-							.replaceAll(" ", ""))
+								.collect(Collectors.joining(",")) + "]")
+						.replaceAll(" ", "")
```
artembilan: Why do we need to replace whitespaces if there is no one anymore? You join over just `,`. No `toString()` leftovers also.

garyrussell: Because that's how we tell them to configure the keys on the factory...

>Also starting with version 1.4, you can configure a routing connection factory in a listener container. In that case, the list of queue names is used as the lookup key. For example, if you configure the container with setQueueNames("foo", "bar"), the lookup key will be "[foo,bar]" (no spaces).

artembilan: Right, but I still don't understand from where those spaces come since we have changed algorithm from `List.toString()` to the `Collectors.joining(",")`.

:man_shrugging:

garyrussell: Hehe - there was a bug in the test

    container.setQueueNames("foo, bar");

instead of

    container.setQueueNames("foo", "bar");

So that's where the space came from 😄

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
garyrussell: If there are no queues, it avoids creating the set.

garyrussell: But it's an unlikely  condition - feel free to change on merge.

garyrussell:
>Do we really need to back-port it?
>Doesn't it look like a behavior breaking change?

It's not breaking; it is a new feature though.

My concern is his comment on the SO answer

>Thanks, I agree with you, but here I have to comply with the specs which mandate the empty name. It's been adopted by a wider community and we can't change it. ... Looks like will have to use a Java Client unfortunately.

But I guess we can release an M1 soon, if you think it's too much code to backport.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -260,36 +261,53 @@ public AcknowledgeMode getAcknowledgeMode() {
...
-	}
-
-	private static String[] collectQueueNames(Queue... queues) {
 		Assert.notNull(queues, "'queues' cannot be null");
 		Assert.noNullElements(queues, "'queues' cannot contain null elements");
-		String[] queueNames = new String[queues.length];
-		for (int i = 0; i < queues.length; i++) {
-			queueNames[i] = queues[i].getName();
+		if (isRunning()) {
+			for (Queue queue : queues) {
```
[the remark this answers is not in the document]
garyrussell: ? queues is `Queue[]`. It would have to be `Arrays.stream(queues).forEach()`; for loop is simpler.

https://stackoverflow.com/questions/35518471/in-java-8-why-were-arrays-not-given-the-foreach-method-of-iterable

spring-amqp/src/main/java/org/springframework/amqp/core/AnonymousQueue.java
```diff
@@ -64,17 +67,42 @@ public AnonymousQueue(NamingStrategy namingStrategy) {
...
+
 	/**
 	 * A strategy to name anonymous queues.
+	 * @deprecated - use the {@link org.springframework.amqp.core.NamingStrategy}.
 	 * @since 1.5.3
 	 *
 	 */
+	@Deprecated
 	@FunctionalInterface
 	public interface NamingStrategy {
```
artembilan: Can't we make this as `extends org.springframework.amqp.core.NamingStrategy` for more smooth migration in the future?

PR #757, opened 2018-05-29
AMQP-797: Defer caching publisher callback channel

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-797

Defer caching publisher callback channels until acks received.

- another user might perform an erroneous call that forces the channel closed and the confirmations
  will be lost.
- if the factory is destroyed with channels in that state, force them closed to generate nacks

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
@@ -1098,14 +1098,14 @@ public class CachingConnectionFactory extends AbstractConnectionFactory
 			}
 		}
 
-		private void returnToCache(ChannelProxy proxy) {
+		private void returnToCache(Channel proxy) {
 			synchronized (this.channelList) {
 				// Allow for multiple close calls...
 				if (CachingConnectionFactory.this.active && !this.channelList.contains(proxy)) {
 					if (logger.isTraceEnabled()) {
 						logger.trace("Returning cached Channel: " + this.target);
 					}
-					this.channelList.addLast(proxy);
+					this.channelList.addLast((ChannelProxy) proxy);
 					setHighWaterMark();
 					this.theConnection.channelsAwaitingAcks.remove(proxy);
 				}
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
@@ -34,6 +34,7 @@ import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.BlockingDeque;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingDeque;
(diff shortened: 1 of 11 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
```diff
@@ -1088,13 +1088,27 @@ private void logicalClose(ChannelProxy proxy) throws Exception {
...
 				}
 			}
-			// Allow for multiple close calls...
-			if (!this.channelList.contains(proxy)) {
-				if (logger.isTraceEnabled()) {
-					logger.trace("Returning cached Channel: " + this.target);
+			if (CachingConnectionFactory.this.active && this.publisherConfirms
+					&& proxy instanceof PublisherCallbackChannel) {
+				this.theConnection.channelsAwaitingAcks.add(proxy);
+				((PublisherCallbackChannel) proxy).setAfterAckCallback(this::returnToCache, proxy);
```
artembilan: This a bit awkward, but I guess we don't have choice since `PublisherCallbackChannelImpl` is not `ChannelProxy`  instance.
I mean we we might not be able just to use `this` there in the `processAck()`. Correct ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/PublisherCallbackChannel.java
```diff
@@ -74,6 +83,14 @@
 	 */
 	Channel getDelegate();
 
+	/**
+	 * Set a callback to be invoked after the ack/nack has been handled.
+	 * @param callback the callback.
+	 * @param proxy the proxy.
+	 * @since 2.1
+	 */
+	void setAfterAckCallback(Consumer<ChannelProxy> callback, ChannelProxy proxy);
```
artembilan: Let's hope we don't have any package tangles with this !

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
garyrussell: Yeah - `this` doesn't work (that's what I had initially) - we need a reference to the proxy. We can look at reworking in 2.1.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/PublisherCallbackChannel.java
garyrussell: Good point - I'll check 😄

garyrussell:
Yes; docs needed after review 😄 and before merge.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
artembilan: ... or we need to keep a dictionary for `PublisherCallbackChannel <-> ChannelProxy`  cross-referencing.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/PublisherCallbackChannel.java
```diff
@@ -86,10 +84,10 @@
 	/**
 	 * Set a callback to be invoked after the ack/nack has been handled.
 	 * @param callback the callback.
-	 * @param proxy the proxy.
+	 * @param proxyForThis the proxy for this channel.
 	 * @since 2.1
 	 */
-	void setAfterAckCallback(Consumer<ChannelProxy> callback, ChannelProxy proxy);
+	void setAfterAckCallback(Consumer<Channel> callback, Channel proxyForThis);
```
artembilan: Yeah... Still confusing: the `PublisherCallbackChannel` is a `Channel` per se. 

Doesn't my idea about relationship `Map` work?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/PublisherCallbackChannelImpl.java
```diff
@@ -90,11 +90,24 @@
...
+	private volatile java.util.function.Consumer<Channel> afterAckCallback;
+
 	public PublisherCallbackChannelImpl(Channel delegate) {
 		delegate.addShutdownListener(this);
 		this.delegate = delegate;
 	}
 
+	@Override
+	public synchronized void setAfterAckCallback(java.util.function.Consumer<Channel> callback) {
+		if (getPendingConfirmsCount() == 0) {
```
artembilan: `Assert.notNull()` for the `callback` ?

src/reference/asciidoc/amqp.adoc
```diff
@@ -839,8 +839,7 @@ public AmqpTemplate rabbitTemplate() {
 ----
 
 Starting with _version 1.4_, in addition to the `retryTemplate` property, the `recoveryCallback` option is supported on the `RabbitTemplate`.
-It is used as a second argument for the `RetryTemplate.execute(RetryCallback<T, E> retryCallback,
-			RecoveryCallback<T>recoveryCallback)`.
+It is used as a second argument for the `RetryTemplate.execute(RetryCallback<T, E> retryCallback, RecoveryCallback<T>recoveryCallback)`.
```
artembilan: Whitespace before `recoveryCallback` word.

Will fix on merge

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/2e99acd3335d1c0fb794eaf0340723f6944caabf.

Looks like I combined the previous commit as well, but I don't think that there are too much destruction.

PR #758, opened 2018-05-30
AMQP-817: Synthetic NACKs on separate thread

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-817

Previously, synthetic nacks were processed on the client thread, which caused
a deadlock if the user attempted some operation such as creating a new channel
on that thread.

Also, when generated from a application-level `close()` a similar deadlock between
the connection factory and `PublisherCallbackChannelImpl` could occur.

Synthetic nacks are now always processed on a separate thread.


artembilan:
Gary, there is some conflict with already merged `RabbitTemplatePublisherCallbacksIntegrationTests3`.

Thanks

artembilan:
Forgot to ask though: shouldn't we consider this for back-port?
Looks like a serious hard issue...

artembilan:
OK. Fine with me. Please, fix conflicts and we are good to go

garyrussell:
Done

PR #765, opened 2018-06-21
AMQP-821: Repub Recoverer limit stack trace header

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-821

Since headers are not fragmented, limit the stack trace header length to
prevent failures.

**cherry-pick to 2.0.x**


artembilan:
Back-ported to `2.0.x` as https://github.com/spring-projects/spring-amqp/commit/6b2f48c33d91a88f350fe8432dfdbc0d8f641ac6 after resolving AssertJ conflicts, which has been introduced exactly since `2.1`.

PR #766, opened 2018-06-22
AMQP-811: Listener container factory improvements

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-811

- `ChannelAwareMessageListener` now inherits from `MessageListener`
  - enables `SimpleRabbitListenerEndpoint` to be used with either
  - deprecate setters for CAML
- Support creating a container with no listener

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/ListenerContainerFactoryBean.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/ListenerContainerFactoryBean.java
@@ -22,6 +22,7 @@ import java.util.concurrent.Executor;
 import org.aopalliance.aop.Advice;
 
 import org.springframework.amqp.core.AcknowledgeMode;
+import org.springframework.amqp.core.MessageListener;
 import org.springframework.amqp.core.MessagePostProcessor;
 import org.springframework.amqp.core.Queue;
 import org.springframework.amqp.rabbit.connection.ConnectionFactory;
@@ -82,7 +83,7 @@ public class ListenerContainerFactoryBean extends AbstractFactoryBean<AbstractMe
 
 	private Boolean exposeListenerChannel;
 
-	private Object messageListener;
+	private MessageListener messageListener;
 
 	private ErrorHandler errorHandler;
 
@@ -211,7 +212,7 @@ public class ListenerContainerFactoryBean extends AbstractFactoryBean<AbstractMe
 		this.exposeListenerChannel = exposeListenerChannel;
 	}
 
-	public void setMessageListener(Object messageListener) {
+	public void setMessageListener(MessageListener messageListener) {
 		this.messageListener = messageListener;
 	}
 
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
@@ -394,15 +394,16 @@ public abstract class AbstractMessageListenerContainer extends RabbitAccessor
 	/**
 	 * Set the message listener implementation to register. This can be either a Spring
 	 * {@link MessageListener} object or a Spring {@link ChannelAwareMessageListener}
-	 * object. Using the strongly typed
-	 * {@link #setChannelAwareMessageListener(ChannelAwareMessageListener)} is preferred.
+	 * object.
 	 *
 	 * @param messageListener The listener.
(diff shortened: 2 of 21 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/ListenerContainerFactoryBean.java
```diff
@@ -211,7 +212,7 @@ public void setExposeListenerChannel(boolean exposeListenerChannel) {
 		this.exposeListenerChannel = exposeListenerChannel;
 	}
 
-	public void setMessageListener(Object messageListener) {
+	public void setMessageListener(MessageListener messageListener) {
```
artembilan: This is going to be a bit of breaking change, but I'm fine with the fix as far as the XML parser is not confused with two different setters.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -394,23 +394,24 @@ public void setExposeListenerChannel(boolean exposeListenerChannel) {
...
 	 * @param messageListener The listener.
 	 * @throws IllegalArgumentException if the supplied listener is not a
 	 * {@link MessageListener} or a {@link ChannelAwareMessageListener}
+	 * @deprecated use {@link #setMessageListener(MessageListener)}.
 	 * @see MessageListener
 	 * @see ChannelAwareMessageListener
 	 */
+	@Deprecated
 	public void setMessageListener(Object messageListener) {
 		checkMessageListener(messageListener);
```
artembilan: This `checkMessageListener()` method has to be modified as well: there is no reason to check for the `ChannelAwareMessageListener`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectReplyToMessageListenerContainer.java
```diff
@@ -82,33 +82,40 @@ public final boolean removeQueueNames(String... queueNames) {
 	}
 
 	@Override
+	@SuppressWarnings("deprecation")
```
artembilan: I somehow think that this and the next `setChannelAwareMessageListener` have to be marked as `@Deprecated` as well...

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -394,23 +394,24 @@ public void setExposeListenerChannel(boolean exposeListenerChannel) {
...
 	 *
 	 * @param messageListener The listener.
 	 * @throws IllegalArgumentException if the supplied listener is not a
 	 * {@link MessageListener} or a {@link ChannelAwareMessageListener}
+	 * @deprecated use {@link #setMessageListener(MessageListener)}.
 	 * @see MessageListener
 	 * @see ChannelAwareMessageListener
 	 */
+	@Deprecated
 	public void setMessageListener(Object messageListener) {
```
artembilan: There are a couple JavaDocs which refer to this method. I think they have to be fixed appropriately as well

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/api/ChannelAwareMessageListener.java
```diff
@@ -27,7 +28,7 @@
  * @author Gary Russell
  */
 @FunctionalInterface
-public interface ChannelAwareMessageListener {
+public interface ChannelAwareMessageListener extends MessageListener {
```
artembilan: Having this change, I guess we should reconsider logic in the `AbstractAdaptableMessageListener`.

The `MessageListenerAdapter` has also redundant check now as well:
```
else if (!(delegate instanceof MessageListener)) {
					throw new AmqpIllegalStateException("MessageListenerAdapter cannot handle a "
							+ "ChannelAwareMessageListener delegate if it hasn't been invoked with a Channel itself");
}
```
which is not true after `if (delegate instanceof ChannelAwareMessageListener)` any more.

garyrussell:
Looks like it was a travis glitch; it's impossible for latch1 to countdown and not latch2 - rebuild was successful.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/68d74bdac3f52e2e4463bb79c1edfbbcae1d056b

PR #767, opened 2018-06-26
AMQP-820: SMLC: Remove InternalConsumer decorator

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-820

Use a discrete `InternalConsumer` for each queue.


artembilan:
Not sure if that is related to your work, but this is what we have in Travis:
```
org.springframework.amqp.rabbit.listener.BrokerDeclaredQueueNameTests > testBrokerNamedQueueDMLC() FAILED
    org.junit.ComparisonFailure at BrokerDeclaredQueueNameTests.java:110
```

garyrussell:
That's the same failure as [this PR](https://github.com/spring-projects/spring-amqp/pull/766#issuecomment-400325656) - see my comment there. But I see now that the second latch has a count of 2. I will take a look.

PR #771, opened 2018-07-10
AMQP-808: Add Declarables for declaring collection

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-808

Deprecate the use of `Collection<Declarable>` in favor of a dedicated
`Declarables` container object to avoid unexpected side-effects when
getting all `Collection` beans from the application context.

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/core/Declarables.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/core/Declarables.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


spring-amqp/src/main/java/org/springframework/amqp/core/Declarables.java
```diff
@@ -0,0 +1,46 @@
...
+ * broker using a single bean declaration for the collection.
+ *
+ * @author Gary Russell
+ * @since 2.1
+ */
+public class Declarables {
+
+	private final Collection<Declarable> declarables = new ArrayList<>();
+
+	public Declarables(Collection<Declarable> declarables) {
```
artembilan: Why not vararg? At least as a variant...

spring-amqp/src/main/java/org/springframework/amqp/core/Declarables.java
```diff
@@ -30,7 +34,14 @@
 
 	private final Collection<Declarable> declarables = new ArrayList<>();
 
+	public Declarables(Declarable... declarables) {
```
artembilan: I think we need to add here this:
```
@SafeVarargs
@SuppressWarnings("varargs")
```
See the same `Arrays.asList()`

garyrussell: I don't see any warnings - I believe the Array case is because `T` is generic; here `Declarable` is a raw type.

PR #772, opened 2018-07-10
AMQP-804: Add BrokerEventListener

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-804

Bind an auto-delete queue to the `amq.rabbitmq.events` exchange
with a configurable list of event keys.

Initial commit.

Docs/javadocs to follow after initial review.


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BrokerEventListener.java
```diff
@@ -0,0 +1,174 @@
...
+		TopicExchange exchange = new TopicExchange("amq.rabbitmq.event");
+		try {
+			this.admin.declareQueue(this.eventQueue);
+			Arrays.stream(this.eventKeys).forEach(k -> {
+				Binding binding = BindingBuilder.bind(this.eventQueue).to(exchange).with(k);
+				this.admin.declareBinding(binding);
+			});
+		}
+		catch (Exception e) {
+			logger.error("failed to declare event queue/bindings - is the plugin enabled?");
```
artembilan: Missed `e` in the `error()`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BrokerEvent.java
```diff
@@ -0,0 +1,58 @@
...
+import org.springframework.amqp.event.AmqpEvent;
+import org.springframework.util.Assert;
+
+/**
+ * @author Gary Russell
+ * @since 2.1
+ *
+ */
+@SuppressWarnings("serial")
+public class BrokerEvent extends AmqpEvent {
```
artembilan: Are you going to fine-grain to the possible event types?

I mean that from the target application perspective it really not so useful just to catch all the possible `BrokerEvent` in one place. Even if we could have separate listeners for different routing keys, we still will emit only `BrokerEvent`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BrokerEventListener.java
```diff
@@ -0,0 +1,174 @@
...
+	@Override
+	public void stop(Runnable callback) {
+		stop();
+		callback.run();
+	}
+
+	@Override
+	public void onMessage(Message message) {
+		if (this.applicationEventPublisher != null) {
+			this.applicationEventPublisher.publishEvent(new BrokerEvent(this, message.getMessageProperties()));
```
artembilan: Isn't there some useful information in the message `body` to transfer to the event as well?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BrokerEventListener.java
```diff
@@ -0,0 +1,174 @@
...
+		try {
+			this.admin.declareQueue(this.eventQueue);
+			Arrays.stream(this.eventKeys).forEach(k -> {
+				Binding binding = BindingBuilder.bind(this.eventQueue).to(exchange).with(k);
+				this.admin.declareBinding(binding);
+			});
+		}
+		catch (Exception e) {
+			logger.error("failed to declare event queue/bindings - is the plugin enabled?");
+			this.bindingsFailed = true;
```
artembilan: Shouldn't we store a failure message as well for case `bindingsFailed` is really `true`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BrokerEvent.java
garyrussell: Well, the event selection in the configured event keys can be very narrow `user.password.changed` or wildcarded `user.#`.

I was thinking that further selection can be done in the listener...

```java
@EventListener(condition = "event.eventType == 'queue.created'")
```

I really don't think we want to create a large number of java events to map the broker events - we'd have to keep them in sync each time they add a new one.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BrokerEventListener.java
garyrussell: No; see https://www.rabbitmq.com/event-exchange.html

>Each event has various properties associated with it. These are translated into AMQP 0-9-1 data encoding and inserted in the message headers. **The message body is always blank.** (empty `byte[]`).

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BrokerEventListener.java
```diff
@@ -0,0 +1,218 @@
...
+
+	private ApplicationEventPublisher applicationEventPublisher;
+
+	/**
+	 * Construct an instance using the supplied connection factory and event keys. Event
+	 * keys are patterns to match routing keys for events published to the
+	 * {@code amq.rabbitmq.event} topic exchange. They can therefore match wildcards;
+	 * examples are {@code user.#, queue.created}. Refer to the plugin documentation for
+	 * information about available events. A single-threaded
+	 * {@link DirectMessageListenerContainer} will be created; its lifecyle will be
```
artembilan: Typo: lifecycle 

Will fix on merge if only concern.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BrokerEventListener.java
```diff
@@ -0,0 +1,218 @@
...
+
+	@Override
+	public void stop(Runnable callback) {
+		stop();
+		callback.run();
+	}
+
+	@Override
+	public void onMessage(Message message) {
+		if (this.applicationEventPublisher != null) {
```
artembilan: I think we need to add a WARN log in the `else`.
show the event we have received and mention the it would be better to add an `ApplicationEventPublisher`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BrokerEventListener.java
```diff
@@ -0,0 +1,218 @@
...
+	public BrokerEventListener(ConnectionFactory connectionFactory, String... eventKeys) {
+		this(new DirectMessageListenerContainer(connectionFactory), true, eventKeys);
+	}
+
+	/**
+	 * Construct an instance using the supplied listener container factory and event keys.
+	 * Event keys are patterns to match routing keys for events published to the
+	 * {@code amq.rabbitmq.event} topic exchange. They can therefore match wildcards;
+	 * examples are {@code user.#, queue.created}. Refer to the plugin documentation for
+	 * information about available events. The container's lifecyle will be not be
```
artembilan: DITTO

Issue #712, opened 2018-02-28
Send & Receive with Temporary queues is create a noLocal consumer


sarah-micallef:
When connecting to Qpid, the reply message is not being received when using the send and receive methods of the `RabbitTemplate`. I suspect that this is because the consumer created to receive the replies is configured as `noLocal` (see https://github.com/spring-projects/spring-amqp/blob/1.7.x/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java#L1297).

This article (https://www.rabbitmq.com/interoperability.html#qpid-python-tests) points out that no-local consumers are not supported by RabbitMQ which explains why replies are received irrespective of the no-local property when using RabbitMQ.

Should the consumer created to receive the reply be configured with the `noLocal` property set to `false`?

artembilan:
Related SO question: https://stackoverflow.com/q/49028292/2756547

artembilan:
Confirmed via JavaDocs from the RabbitMQ Java Client on the matter:
https://github.com/rabbitmq/rabbitmq-java-client/blob/master/src/main/java/com/rabbitmq/client/Channel.java#L1129-L1130

```
* @param noLocal True if the server should not deliver to this consumer
* messages published on this channel's connection. Note that the RabbitMQ server does not support this flag.
```

I think we should make it `false` by default and expose something like `noLocalReplyConsumer` on the `RabbitTemplate` to let to retain the previous behavior.

Contributions are welcome!

garyrussell:
Copying the SO dialog here...

Gary: 
>t's a bit unusual to have the server side running in the same application as the client side. That's the only way no-local should affect this scenario since, normally, the client-side doesn't publish the reply.

>We don't claim compatibility with QPID (certainly not AMQP 1.0) but I see QPID claims support for AMQP 0.9.x so feel free to open a JIRA Issue and we'll take a look as time permits.

Sarah:
>Thanks for your reply. This actually happened during a spring boot test which makes use of qpid as an embedded amqp broker. So in my case I have both the server and the client running within the same app. 

BTW, I was not aware that using an embedded QPID broker would work with Spring AMQP (never tried it) - we should explore that as a first class feature.

sarah-micallef:
[...]

PR #773, opened 2018-07-10
GH-712: Add RabbitTemplate.noLocalReplyConsumer

artembilan:
Fixes https://github.com/spring-projects/spring-amqp/issues/712

For controlling `noLocal` flag for reply consumers in the `sendAndReceive()`
operations introduce a `noLocalReplyConsumer` option for the `RabbitTemplate`.
Make it `false` by default since it does not have any value in RabbitMQ,
but matters in QPid for temporary queues


garyrussell:
Merged as 5989d03f03c7b024c47276bec47d29abecd3c964

PR #775, opened 2018-07-12
AMQP-824 -Name the thread pool and avoid unnecessary creation

wdroste:


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
```diff
@@ -170,8 +171,28 @@
...
+	 * @return specified executor service otherwise the default one is created and returned.
+	 */
+	protected ExecutorService getDeferredCloseExecutor() {
+		if (null != getExecutorService()) {
+			return getExecutorService();
+		}
+        synchronized (this.connectionMonitor) {
+            if (null == this.deferredCloseExecutor) {
+                this.deferredCloseExecutor =
+                        Executors.newCachedThreadPool(new AmqpNamedThreadPoolFactory());
```
garyrussell: Inconsistent tab/space indentation - we exclusively use tabs.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
```diff
@@ -1295,4 +1316,23 @@ else if (!RabbitUtils.isNormalChannelClose(cause)) {
 
 	}
 
+
+    /**
+     * Name each create thread by this class.
+     */
+    private static class AmqpNamedThreadPoolFactory implements ThreadFactory {
+        private AtomicInteger threadId = new AtomicInteger();
+
```
garyrussell: Need 
```java
AmqpNamedThreadPoolFactory() {
    super();
}
```

(package protected - to avoid the compiler creating a synthetic constructor).

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
```diff
@@ -1295,4 +1316,23 @@ else if (!RabbitUtils.isNormalChannelClose(cause)) {
...
+        @Override
+        public Thread newThread(Runnable r) {
+            Thread t = new Thread(r);
+            t.setDaemon(true);
+            t.setName("spring-rabbit-deferred-pool-"
+                    + threadPoolId.incrementAndGet()
+                    + "-thread-"
```
garyrussell: We should include the connection factory beanName (if present) in the thread names. Will need a protected `getBeanName()` on the abstract factory and pass it into this class.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
```diff
@@ -1295,4 +1316,23 @@ else if (!RabbitUtils.isNormalChannelClose(cause)) {
...
+    /**
+     * Name each create thread by this class.
+     */
+    private static class AmqpNamedThreadPoolFactory implements ThreadFactory {
+        private AtomicInteger threadId = new AtomicInteger();
+
+        @Override
+        public Thread newThread(Runnable r) {
+            Thread t = new Thread(r);
+            t.setDaemon(true);
```
artembilan: Why does it become `daemon`?
Does looks like a `DefaultThreadFactory` does the same:
```
    if (t.isDaemon())
                t.setDaemon(false);
```
Well, it does exactly opposite.

On the other hand I wonder why `CustomizableThreadFactory` doesn't work here for us?..

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
artembilan: I think we don't need `stsatic threadPoolId` counter if we are going to use a `beanName` as base.

garyrussell: (Except when it's not a bean 😄 )

artembilan: Then call `setBeanName()` manually! 😃

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/7bf27de1528ab7520f7303099aae078f4dbe194b.

Cherry-picked to `master` and `2.0.x` as: https://github.com/spring-projects/spring-amqp/commit/b896c89f18d2536231f7068c712af424029346d9

@wdroste ,

Thank you for the contribution, but consider the next time to PR against `master`. There is a pretty big delta between branches and it is hard to cherry-pick from the bottom.
Also, please, pay attention how I polished your `null` conditions: we prefer `null` on the right hand.

Looking forward for more!

PR #783, opened 2018-07-30
Add Future<?> to CorrelationData

garyrussell:
As an alternative to using a template callback, add a `ListenableFuture<?>` to the
`CorrelationData` base class.


artembilan:
Some tests failures on Travis.

garyrussell:
I have another idea about returned messages, but that should be a different PR.

PR #784, opened 2018-07-31
Add returned message to correlation data

garyrussell:
If confirms and returns are enabled, a returned message is sent before the
ack; add the returned message to the correlation data, if provided.


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/CorrelationData.java
```diff
@@ -36,6 +37,10 @@
 
 	private volatile String id;
 
+	private volatile String returnCorrelation;
+
+	private volatile Message returnedmessage;
```
artembilan: Should camelCase - `returnedMessage`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/CorrelationData.java
```diff
@@ -70,10 +75,44 @@ public void setId(String id) {
...
+	 * returnCorrelation}. Guaranteed to be populated before the future is set.
+	 * @return the message or null.
+	 * @since 2.1
+	 */
+	@Nullable
+	public Message getReturnedmessage() {
+		return this.returnedmessage;
+	}
+
+	public void setReturnedmessage(Message returnedmessage) {
```
artembilan: Shouldn't this be package protected do not let end-user to override?
Or is it harmless from the Framework perspective and that's already end-user responsibility for the damage (s)he does for their application?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/PublisherCallbackChannel.java
```diff
@@ -37,6 +37,8 @@
 
 	String RETURN_CORRELATION_KEY = "spring_listener_return_correlation";
 
+	String RETURNED_MESSAGE_CORRELATION_KEY = "spring_returned_message_correlation";
```
artembilan: Why do we need this extra property?
Maybe the purpose of the existing `RETURN_CORRELATION_KEY` should be extended and its real name change to reflect reality?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/PublisherCallbackChannelImpl.java
```diff
@@ -872,6 +882,10 @@ public void addListener(Listener listener) {
 				if (pendingConfirm.getTimestamp() < cutoffTime) {
 					expired.add(pendingConfirm);
 					iterator.remove();
+					CorrelationData correlationData = pendingConfirm.getCorrelationData();
+					if (correlationData != null && StringUtils.hasText(correlationData.getReturnCorrelation())) {
+						this.pendingReturns.remove(correlationData.getReturnCorrelation());
```
artembilan: Why an existing `CorrelationData.id` can't be used for this return purpose as well?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/PublisherCallbackChannel.java
garyrussell: No; they are different; the first one tells us which template (listener) to call for returned messages; the new one enables us to get a reference to the `PendingConfirm` so we can add the returned message to its correlation data (if present). See the new `pendingReturns` map.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/PublisherCallbackChannelImpl.java
garyrussell: Good idea!

PR #785, opened 2018-07-31
@SendTo: Add property-placeholders support

artembilan:
Add property-placeholders resolution for the `@SendTo` processors for
consistency with all other Spring AMQP annotations support

External request on the matter: https://stackoverflow.com/questions/51620793/use-property-placeholder-with-spring-amqp-sendto-annotation

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/DelegatingInvocableHandler.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/DelegatingInvocableHandler.java
@@ -180,11 +180,12 @@ public class DelegatingInvocableHandler {
 			}
 			replyTo = destinations.length == 1 ? resolve(destinations[0]) : null;
 		}
-		return this.beanExpressionContext.getBeanFactory().resolveEmbeddedValue(replyTo);
+		return replyTo;
 	}
 
 	private String resolve(String value) {
 		if (this.resolver != null) {
+			value = this.beanExpressionContext.getBeanFactory().resolveEmbeddedValue(value);
 			Object newValue = this.resolver.evaluate(value, this.beanExpressionContext);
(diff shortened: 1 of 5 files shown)


garyrussell:
We need a similar fix in `DelegatingInvocableHandler` for multi-method listeners.

garyrussell:
Perhaps move this to a utility so we don't have the same code in 2 places?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/DelegatingInvocableHandler.java
```diff
@@ -184,6 +185,7 @@ private String extractSendTo(String element, SendTo ann) {
 
 	private String resolve(String value) {
 		if (this.resolver != null) {
+			value = this.beanExpressionContext.getBeanFactory().resolveEmbeddedValue(value);
```
artembilan: Does this work for you, Gary?

Also pay attention to the test changes.

garyrussell: Sorry - I missed that; yes; it's ok with me - not worth pulling out to a utility (maybe next time if we need a change).

garyrussell: Waiting for green 😄

artembilan: 😃 . Right, just was going to same the same.

Not sure though that it is going to happen... I mean more changes in that part 😉

garyrussell:
Merged as 0a943f64f94406fe0c64c0551d528460f8ca4526 after removing cherry-pick comment.

PR #792, opened 2018-08-08
AMQP-825: Add Reply RetryTemplate

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-825

__cherry-pick to 2.0.x__

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/support/SendRetryContextAccessor.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/support/SendRetryContextAccessor.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.amqp.support;
+
+import org.springframework.amqp.core.Address;
+import org.springframework.amqp.core.Message;
+import org.springframework.retry.RetryContext;
+
+/**
+ * Type safe accessor for retried message sending.
+ *
+ * @author Gary Russell
+ * @since 2.0.6
+ *
+ */
+public class SendRetryContextAccessor {
+
+	/**
+	 * Key for the message we tried to send.
+	 */
+	public static final String MESSAGE = "message";
+
+	/**
+	 * Key for the Address we tried to send to.
+	 */
+	public static final String ADDRESS = "address";
+
+	private final RetryContext context;
+
+	public SendRetryContextAccessor(RetryContext context) {
+		this.context = context;
+	}
+
+	public void setMessage(Message message) {
+		this.context.setAttribute(MESSAGE, message);
+	}
+
+	public Message getMessage() {
+		return (Message) this.context.getAttribute(MESSAGE);
+	}
+
+	public void setAddress(Address address) {
+		this.context.setAttribute(ADDRESS, address);
+	}
+
+	public Address getAddress() {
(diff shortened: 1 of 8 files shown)


spring-amqp/src/main/java/org/springframework/amqp/support/SendRetryContextAccessor.java
```diff
@@ -0,0 +1,64 @@
...
+	 */
+	public static final String ADDRESS = "address";
+
+	private final RetryContext context;
+
+	public SendRetryContextAccessor(RetryContext context) {
+		this.context = context;
+	}
+
+	public void setMessage(Message message) {
```
artembilan: Are we really going to use a setter?

spring-amqp/src/main/java/org/springframework/amqp/support/SendRetryContextAccessor.java
```diff
@@ -0,0 +1,64 @@
...
+		this.context.setAttribute(MESSAGE, message);
+	}
+
+	public Message getMessage() {
+		return (Message) this.context.getAttribute(MESSAGE);
```
artembilan: Sounds like this could be just simple static type-safe utility:
```
public static Message getMessage(AttributeAccessor attributes) {
		return (Message) attributes.getAttribute(MESSAGE);
}
```

I mean there is no need to to require from end-user to create an new instance of some extra class for the help to access to attributes.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/AbstractRabbitListenerContainerFactory.java
```diff
@@ -295,6 +301,26 @@ public void setBeforeSendReplyPostProcessors(MessagePostProcessor... beforeSendR
 		this.beforeSendReplyPostProcessors = beforeSendReplyPostProcessors;
 	}
 
+	/**
+	 * Set a {@link RetryTemplate} to use when sending replies.
+	 * @param retryTemplate the template.
+	 * @since 2.0.6
+	 * @see #setReplyRecoveryCallback(RecoveryCallback)
```
artembilan: How about to also add a `@see AbstractAdaptableMessageListener#setReplyRecoveryCallback(RecoveryCallback)` for better navigation?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/AbstractRabbitListenerContainerFactory.java
```diff
@@ -370,10 +396,18 @@ public C createListenerContainer(RabbitListenerEndpoint endpoint) {
...
+			AbstractAdaptableMessageListener messageListener = (AbstractAdaptableMessageListener) instance
+					.getMessageListener();
+			if (this.beforeSendReplyPostProcessors != null) {
+				messageListener.setBeforeSendReplyPostProcessors(this.beforeSendReplyPostProcessors);
+			}
+			if (this.retryTemplate != null) {
+				messageListener.setRetryTemplate(this.retryTemplate);
+			}
+			if (this.recoveryCallback != null) {
+				messageListener.setRecoveryCallback(this.recoveryCallback);
```
artembilan: How about to propagate it only `if (this.retryTemplate != null)`?
I believe a `recoveryCallback` is not going to be used when otherwise...

And that might be reflected in the Javadocs as well.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
```diff
@@ -414,15 +444,33 @@ protected void sendResponse(Channel channel, Address replyTo, Message messageIn)
...
+				final Message messageToSend = message;
+				final AtomicBoolean first = new AtomicBoolean(true);
+				this.retryTemplate.execute(ctx -> {
+					if (this.recoveryCallback != null && first.getAndSet(false)) {
+						ctx.setAttribute(SendRetryContextAccessor.MESSAGE, messageToSend);
+						ctx.setAttribute(SendRetryContextAccessor.ADDRESS, replyTo);
+					}
+					doPublish(channel, replyTo, messageToSend);
+					return null;
+				}, this.recoveryCallback);
```
artembilan: Can we do here a wrapping to populate those attributes instead of `AtomicBoolean` logic in the `RetryCallback` impl above?
However no: we are going to have extra object for each message...

So, WDYT?

PR #793, opened 2018-08-10
AMQP-828: AutoRecovery with template.receive()

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-828

Close auto-recoving channels during recovery since the consumer is
no longer there.

This was previously fixed for the `BlockingQueueConsumer`, but not for
`template.receive()` operations.


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
```diff
@@ -86,7 +84,7 @@
  * @author Alex Panchenko
  * @author Johno Crawford
  */
-public class BlockingQueueConsumer implements RecoveryListener {
+public class BlockingQueueConsumer {
```
artembilan: This is some kind of breaking change for back-porting...

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/ClosingRecoveryListener.java
```diff
@@ -0,0 +1,97 @@
...
+	@Override
+	public void handleRecoveryStarted(Recoverable recoverable) {
+		if (logger.isDebugEnabled()) {
+			logger.debug("Closing an autorecovered channel: " + recoverable);
+		}
+		try {
+			((Channel) recoverable).close();
+		}
+		catch (IOException | TimeoutException e) {
+			logger.debug("Error closing an autorecovered channel");
```
artembilan: "Error closing" with `debug`? 🤷‍♂️

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/ClosingRecoveryListener.java
```diff
@@ -0,0 +1,97 @@
...
+ * @since 1.7.10
+ *
+ */
+public final class ClosingRecoveryListener implements RecoveryListener {
+
+	private static final Log logger = LogFactory.getLog(ClosingRecoveryListener.class);
+
+	private static final RecoveryListener INSTANCE = new ClosingRecoveryListener();
+
+	private static final ConcurrentMap<AutorecoveringChannel, Boolean> hasListener = new ConcurrentHashMap<>();
```
artembilan: Looks like you want to have a `Collections.newSetFromMap()`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/ClosingRecoveryListener.java
```diff
@@ -0,0 +1,97 @@
...
+		}
+	}
+
+	/**
+	 * Add a listener if necessary so we can immediately close an autorecovered
+	 * channel if necessary since the actual consumer will no longer exist.
+	 * @param channel the channel.
+	 */
+	public static void addRecoveryListenerIfNecessary(Channel channel) {
+		if (channel instanceof ChannelProxy) {
```
artembilan: Don't we need to deal with an `else` variant as well?

On the other hand why `ChannelProxy` can't implement `AutorecoveringChannel` and delegate respectively?
Although I believe there is going to be some overhead since we may disable auto-recovery, but we still fill up this cache almost for nothing...

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/ClosingRecoveryListener.java
```diff
@@ -0,0 +1,97 @@
...
+				AutorecoveringChannel autorecoveringChannel = (AutorecoveringChannel) ((ChannelProxy) channel)
+						.getTargetChannel();
+				if (hasListener.putIfAbsent(autorecoveringChannel, Boolean.TRUE) == null) {
+					autorecoveringChannel.addRecoveryListener(INSTANCE);
+				}
+			}
+		}
+	}
+
+	public static void removeChannel(AutorecoveringChannel channel) {
```
artembilan: This needs Javadocs.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
garyrussell: Why do you consider this a breaking change? The listener implementation was just moved to another class.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/ClosingRecoveryListener.java
garyrussell: This was the original code moved from the BQC.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/ClosingRecoveryListener.java
garyrussell: I don't understand, sorry; we always have a `ChannelProxy` (except in some mock test cases).

Right, we only want to track actual auto recovery channels.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
artembilan: Well, I mean an interaction exactly with this class, if someone somehow extract it from the container and cast to the `RecoveryListener` for his/her own peculiar logic.

But yeah... Should not be so critical do not proceed with the fix you have done so far...

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/ClosingRecoveryListener.java
garyrussell: No; I originally had `ConcurrentHashMap.newKeySet()`, but I need `putIfAbsent()`.

artembilan: Well, but our `boolean` value is idempotent there. Only what bother us is exactly if there was value before or not.
See `HashSet` impl:
```
  * @return <tt>true</tt> if this set did not already contain the specified
  * element
  */
  public boolean add(E e) {
        return map.put(e, PRESENT)==null;
  }
```

For me this is exactly what you do with:
```
if (hasListener.putIfAbsent(autorecoveringChannel, Boolean.TRUE) == null) {
```

So, essentially you implement here some small `Set` logic, but let it be if you insist. At least we get some performance gain skipping `Set` operation and moving to `Map` one directly. 😉

garyrussell: Yes; that was my thought, and `newKeySet()` does exactly that anyway (back the set with a map with boolean values, always true).

artembilan:
Cherry-picked to 2.0.x as https://github.com/spring-projects/spring-amqp/commit/e51bf2f57d1623346c098672711ad198e8ec2a39

This can't be cherry-picked clearly to `1.7.x`. Need separate PR.

garyrussell:
Cherry-picked to 1.7.x as 159d2b43158834f86c018635c12ea2d99930dd27 after resolving conflicts.

PR #795, opened 2018-08-22
AMQP-829: RabbitTemplate Req/Resp Fail Fast

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-829

If the channel closes while we are waiting for a reply, throw an
exception instead of waiting to time out.

### DIFF
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitTemplateTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitTemplateTests.java
@@ -395,7 +395,8 @@ public class RabbitTemplateTests {
 		RabbitTemplate template = new RabbitTemplate(connectionFactory);
 		template.setReplyTimeout(60_000);
 		Message input = new Message("Hello, world!".getBytes(), new MessageProperties());
-		Executors.newSingleThreadExecutor().execute(() -> {
+		ExecutorService exec = Executors.newSingleThreadExecutor();
+		exec.execute(() -> {
 			try {
 				shutdownLatch.await(10, TimeUnit.SECONDS);
 			}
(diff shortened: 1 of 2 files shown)


spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitTemplateTests.java
```diff
@@ -365,6 +373,46 @@ public void testNestedTxBinding() throws Exception {
...
+			listener.set(invocation.getArgument(0));
+			shutdownLatch.countDown();
+			return null;
+		}).given(mockChannel).addShutdownListener(any());
+		SingleConnectionFactory connectionFactory = new SingleConnectionFactory(mockConnectionFactory);
+		connectionFactory.setExecutor(mock(ExecutorService.class));
+		RabbitTemplate template = new RabbitTemplate(connectionFactory);
+		template.setReplyTimeout(60_000);
+		Message input = new Message("Hello, world!".getBytes(), new MessageProperties());
+		Executors.newSingleThreadExecutor().execute(() -> {
```
artembilan: Don't we need to kill the thread pool in the end of test?
The point is because of this:

```
public static ExecutorService newSingleThreadExecutor() {
        return new FinalizableDelegatedExecutorService
            (new ThreadPoolExecutor(1, 1,
                                    0L, TimeUnit.MILLISECONDS,
                                    new LinkedBlockingQueue<Runnable>()));
    }
```
I mean it dies, of course, in the end of the JVM, but it still going to hang on during test suit execution.

artembilan:
Gary, how do you think: is this fix going to cover the recent failure on CI: https://build.spring.io/browse/AMQP-MEIGHT-1531 ?

garyrussell:
I don't think so, but I have added diagnostic logs to try to understand why we have started getting those failures; I'll take a look.

Issue #804, opened 2018-09-03
Wrong consumer canceled using direct-reply-to


gasserus:
For short period of time two parallel requests are sent, which lead to two parallel consumers.
Afterwards I send requests which need only one consumer (the other consumer switches to idle) for at least `>idle time`.
After this time one consumer is canceled. Instead of canceling the idle consumer the running consumer is canceled. The request never gets an answer and after the reply timeout the consumer returns null.

In the PR #703 you mentioned `In the end it doesn't matter which consumers remain in the container`.
Looking at the adjustConsumers method in DirectMessageListenerContainer there is written that always the first consumer is canceled instead of the idled.
**Shouldn't only the idled consumers be removed or do I miss something?**

The remote procedure call (RPC) szenario is used with direct-reply-to (https://docs.spring.io/spring-amqp/reference/htmlsingle/#direct-reply-to).
No dedicated reply queue is used.

Library:
* spring-boot-starter-amqp 2.0.4.RELEASE
* spring-amqp 2.0.5.RELEASE

artembilan:
How about to use a `DirectReplyToMessageListenerContainer` instead?

I see @garyrussell agreed to consider this as a bug, but maybe meanwhile there is some workaround with another component...

Thanks for the report!

garyrussell:
@artembilan the bug *is* in the DRTMLC.

PR #805, opened 2018-09-04
GH-804: DRTMLC - don't cancel in-use consumer

garyrussell:
Fixes https://github.com/spring-projects/spring-amqp/issues/804

The `DirectReplyToMessageListenerContainer` shuts down consumers that haven't
been used for the `idleEventInterval`. However, the superclass simply cancels
the first consumer in the list, which may actually being used in the DRTMLC.

Change the logic to find a consumer that is not actually being used.

**cherry-pick to 2.0.x**


spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainerIntegrationTests.java
```diff
@@ -551,12 +551,16 @@ public void testReplyToConsumersReduced() throws Exception {
...
 		container.afterPropertiesSet();
 		container.start();
-		ChannelHolder channelHolder = container.getChannelHolder();
-		assertTrue(activeConsumerCount(container, 1));
-		container.releaseConsumerFor(channelHolder, false, null);
+		ChannelHolder channelHolder1 = container.getChannelHolder();
+		ChannelHolder channelHolder2 = container.getChannelHolder();
+		assertTrue(activeConsumerCount(container, 2));
+		container.releaseConsumerFor(channelHolder2, false, null);
+		Thread.sleep(500);
```
artembilan: I understand that this is quick and simple fix, but as experience shows we can't rely on the system clock: no guarantee that CPU will get access to the target process exactly after this timeout. 
In the end we are going to have sporadic test failures.

The best way to sleep with short period and check the state of the component we would like to assert afterward.

If you have some argument to proceed with this simple sleep, let me know here.

Thanks

artembilan:
... and cherry-picked to `2.0.x`.

PR #809, opened 2018-09-07
AMQP-833: DLQ and Expiry - Discard

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-833

A common pattern is to add time-to-live to a DLQ and route
expired messages back to the main queue.

If a message fails with a fatal exception, such messages will
cycle around for ever; the application has no mechanism to
look at the `x-death` header.

Add logic to the `ConditionalRejectingErrorHandler` to look for
an `x-death` header and discard the message completely by throwing
an `ImmediateAcknowledgeAmqpException`.

Add a property to the error handler to revert to the previous
behavior.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/ConditionalRejectingErrorHandler.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/ConditionalRejectingErrorHandler.java
@@ -16,10 +16,15 @@
 
 package org.springframework.amqp.rabbit.listener;
 
+import java.util.List;
+import java.util.Map;
+
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
 import org.springframework.amqp.AmqpRejectAndDontRequeueException;
+import org.springframework.amqp.ImmediateAcknowledgeAmqpException;
+import org.springframework.amqp.core.Message;
 import org.springframework.amqp.rabbit.listener.exception.ListenerExecutionFailedException;
 import org.springframework.amqp.support.converter.MessageConversionException;
 import org.springframework.messaging.MessagingException;
@@ -54,6 +59,8 @@ public class ConditionalRejectingErrorHandler implements ErrorHandler {
 
 	private final FatalExceptionStrategy exceptionStrategy;
 
+	private boolean discardFatalsWithXDeath = true;
+
 	/**
 	 * Create a handler with the {@link ConditionalRejectingErrorHandler.DefaultExceptionStrategy}.
 	 */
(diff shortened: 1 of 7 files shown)


spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DlqExpiryTests.java
```diff
@@ -0,0 +1,120 @@
...
+public class DlqExpiryTests {
+
+	@Autowired
+	private Config config;
+
+	@Test
+	public void testExpiredDies() throws Exception {
+		this.config.template().convertAndSend("test.expiry.main", "foo");
+		assertThat(this.config.latch.await(10, TimeUnit.SECONDS)).isTrue();
+		Thread.sleep(300);
```
artembilan: This looks suspicious.
As experience shows we are going to have a race condition sporadically.

Why a single `latch` not enough to verify this test-case?

Shouldn't we check instead that we get an `ImmediateAcknowledgeAmqpException` and there is no more messages in the `test.expiry.main` and stop container, of course before verification. 

Well, if you can convince me that this is OK what you have done so far, we can proceed with merging.

Thanks

src/reference/asciidoc/amqp.adoc
```diff
@@ -4280,6 +4280,10 @@ The default `FatalExceptionStrategy` logs a warning message when an exception is
 
 Since _version 1.6.3_ a convenient way to add user exceptions to the fatal list is to subclass `ConditionalRejectingErrorHandler.DefaultExceptionStrategy` and override the method `isUserCauseFatal(Throwable cause)` to return true for fatal exceptions.
 
+A common pattern for handling DLQ messages is to set a time-to-live on those messages as well as additional DLQ configuration such that these messages expire and are routed back to the main queue for retry.
+The problem with this technique is that messages that cause fatal exceptions will loop forever.
+Starting with version 2.1, the the `ConditionalRejectingErrorHandler` detects an `x-death` header on a message that causes a fatal exception to be thrown, the message will be logged and discarded.
```
artembilan: Double `the`

src/reference/asciidoc/whats-new.adoc
```diff
@@ -74,3 +74,8 @@ See <<async-return>> for more information.
 ===== Connection Factory Bean Changes
 
 The `RabbitConnectionFactoryBean` now calls `enableHostnameVerification()` by default; to revert to the previous behavior, set the `enabaleHostnameVerification` property to `false`.
+
+===== Listener Container Changes
+
+The default `ConditionalRejectingErrorHandler` will now completely discard messages that cause fatal errors if an `x-death` header is present.
+See <<>> for more information.
```
artembilan: Missed link to the target chapter.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/ConditionalRejectingErrorHandler.java
```diff
@@ -69,10 +76,33 @@ public ConditionalRejectingErrorHandler(FatalExceptionStrategy exceptionStrategy
 		this.exceptionStrategy = exceptionStrategy;
 	}
 
+	/**
+	 * Set to false to disable the, now, default behavior of logging and discarding
```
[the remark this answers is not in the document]
garyrussell: It reads OK to me; what is the objection?

The `, now,` is intended to emphasize that this default behavior is new `@since 2.1`.

cc/ @buzzardo - Jay, can you help us out? Thanks.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DlqExpiryTests.java
garyrussell: Because it's specifically testing that the count doesn't increase after the latch has counted down.

Running the test with the old code fails.

While it is true that we could get a "false" successful test because of a busy CI server, it would be a false positive, not a false negative (which is what we need to avoid).

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/ConditionalRejectingErrorHandler.java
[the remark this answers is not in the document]
artembilan: Ah! I see what you mean.

In our language I would place that `now` into the brackets - `to disable (now) the default behavior`.

What confuses me here is a comma after `the` 😄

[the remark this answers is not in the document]
garyrussell: I am ok with  "the (now) default" if you want to change it on merge and fix the others. Or I will fix tomorrow when I am home.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/90f8aa03da03759f96247561a69d17fa2dfea6e7

PR #819, opened 2018-09-20
Close auto-recovering connection

garyrussell:
When a connection is auto-recovered, the `RabbitAdmin` is not
invoked to re-declare auto-delete queues because the connection
listeners are not invoked.

Close an auto-recovered connection before it is recovered.

Tested with a stand-alone spring-cloud-bus application.

**Cherry-pick to 2.0.x, 1.7.x**


artembilan:
... and cherry-picked to `2.0.x`.
Back-ported to `1.7.x` as  https://github.com/spring-projects/spring-amqp/commit/1283edfb35f8466e0b9e6c89442d425046e821ff after resolving conflicts.

PR #821, opened 2018-09-21
AMQP-813: Requeue after retry exhausted option

artembilan:
JIRA: https://jira.spring.io/browse/AMQP-813

* Add `ImmediateRequeueAmqpException` to let `ContainerUtils.shouldRequeue()`
to return `true` immediately and requeue the message in the container
* Add `ImmediateRequeueMessageRecoverer` to throw a mentioned above
`ImmediateRequeueAmqpException`
* Refactor `RetryInterceptorBuilder` to avoid duplicated code
* Mention changes in the Docs

### DIFF
--- a/src/reference/asciidoc/amqp.adoc
+++ b/src/reference/asciidoc/amqp.adoc
@@ -5231,6 +5231,8 @@ RepublishMessageRecoverer recoverer = new RepublishMessageRecoverer(amqpTemplate
 };
 ----
 
+Starting with version 2.1, an `ImmediateRequeueMessageRecoverer` is  added to throw an `ImmediateRequeueAmqpException` which notifies a listener container to requeue the current failed message.
+
 ===== Exception Classification for Retry
 
 Spring Retry has a great deal of flexibility for determining which exceptions can invoke retry.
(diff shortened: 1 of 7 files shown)


src/reference/asciidoc/amqp.adoc
```diff
@@ -5231,6 +5231,8 @@ RepublishMessageRecoverer recoverer = new RepublishMessageRecoverer(amqpTemplate
 };
 ----
 
+Starting with version 2.1, an `ImmediateRequeueMessageRecoverer` is  added to throw an `ImmediateRequeueAmqpException` which notifies a listener container to requeue the current failed message.
+
 ===== Exception Classification for Retry
 
 Spring Retry has a great deal of flexibility for determining which exceptions can invoke retry.
```
garyrussell: I think we should mention elsewhere in the docs (somewhere where we talk about `defaultRequeueRejected` ?) that this new exception can be thrown by a listener to override the behavior if the `defaultRequeueRejected` has been changed from its default (`true`) to `false`.

artembilan: Doh. Has meant that when developed this exception, but been side-tracked with other stuff...
Will add soon.

Thanks

src/reference/asciidoc/amqp.adoc
```diff
@@ -5188,6 +5180,8 @@ When using RabbitMQ 2.8.x or higher, this also facilitates delivering the messag
 
 Or, you can throw a `AmqpRejectAndDontRequeueException`; this prevents message requeuing, regardless of the setting of the `defaultRequeueRejected` property.
 
+Starting with version 2.1, an `ImmediateRequeueAmqpException` is introduced to perform exactly opposite logic: the message will be required regardless of the setting of the `defaultRequeueRejected` property.
```
garyrussell: s/required/requeued/ - will fix during merge.

garyrussell:
Merged as d75777de389ae07d2bfd5f9f28db0de2702bfeaf

garyrussell:
Doh - should have waited for Travis.

PR #831, opened 2018-10-19
AMQP-838: Deprecate container.setMessageConverter

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-838

The `setMessageConverter()` on the listener container was only used
to convey the converter from the container factory to the listener
adapter. This is confusing to users hand-wiring a container since
there is an implication it is used by the container itself.

Deprecate the method, and use the endpoint (by default) to convey
the converter to the endpoint's adapter.

**cherry-pick to 2.0.x**


artembilan:
... and cherry-picked to `2.0.x`

Issue #841, opened 2018-11-01
Unable to successfully stop tomcat server


Kmegumi:
if i had execute the method of amqpTemplate.convertAndSend, when i stop the tomcat server,I will find the exception a thread named [pool-3-thread-1] has failed to stop it, but i cannot find which thread pool own it，this is my amqp configurations and used demo

![c436a89b-2b6c-4cc5-ae3f-6345d08345ff](https://user-images.githubusercontent.com/26270054/47833813-d1101780-ddd7-11e8-9b90-36de64c06455.png)
![9361ed8d-5f9a-489c-934c-b271eb376750](https://user-images.githubusercontent.com/26270054/47833824-db321600-ddd7-11e8-88f5-625a6f9f3664.png)

![b7b57c63-80de-44fd-a738-c59d32b29b22](https://user-images.githubusercontent.com/26270054/47833830-e08f6080-ddd7-11e8-9c1b-9eb4286f923e.png)

![716f5ce8-1d06-4e3e-b4c3-23d01a4c4d47](https://user-images.githubusercontent.com/26270054/47833939-55629a80-ddd8-11e8-9f12-56941ba49491.png)

artembilan:
You need to consider to add a `ThreadPoolTaskExecutor` `@Bean` and inject it into the `ConnectionFactory` bean.

The Tomcat is very sensitive to the threads which are not closed properly during application shutdown.

garyrussell:
If the `ApplicationContext` is `close()` d properly, these threads will terminate, This should happen automatically if the context is managed by the web container and properly configured.

Kmegumi:
I have found, if I had set the property publisherConfirms be true, when execute the method doCreateBareChannel will encapsulation the channel to be PublisherCallbackChannelImpl,if we not add ThreadPoolTaskExecutor into CachingConnectionFactory the SingleThreadExecutor in PublisherCallbackChannelImpl will not shutdown properly, Can we shutdown it properly in the destroy phase?

garyrussell:
Yes; we should shut that one down (or use a daemon thread).

artembilan:
I'll take it.

PR #843, opened 2018-11-05
GH-841: Close executor in the PublisherConChannel

artembilan:
Fixes https://github.com/spring-projects/spring-amqp/issues/841

**cherry-pick to 2.0.x**

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/PublisherCallbackChannelImpl.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/PublisherCallbackChannelImpl.java
@@ -862,8 +862,7 @@ public class PublisherCallbackChannelImpl
 	@Override
 	public synchronized int getPendingConfirmsCount() {
 		return this.pendingConfirms.values().stream()
-				.map(Map::size)
-				.mapToInt(Integer::valueOf)
+				.mapToInt(Map::size)
 				.sum();
 	}
 
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/PublisherCallbackChannelImpl.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/PublisherCallbackChannelImpl.java
@@ -86,13 +86,13 @@ import com.rabbitmq.client.impl.recovery.AutorecoveringChannel;
  *
  * @author Gary Russell
  * @author Arnaud Cogoluègnes
+ * @author Artem Bilan
+ *
  * @since 1.0.1
  *
  */
 public class PublisherCallbackChannelImpl
 		implements PublisherCallbackChannel, ConfirmListener, ReturnListener, ShutdownListener {
-
-	private static final ExecutorService DEFAULT_EXECUTOR = Executors.newSingleThreadExecutor();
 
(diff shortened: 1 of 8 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/PublisherCallbackChannelImpl.java
```diff
@@ -854,7 +862,7 @@ public synchronized int getPendingConfirmsCount(Listener listener) {
 	@Override
 	public synchronized int getPendingConfirmsCount() {
 		return this.pendingConfirms.values().stream()
-				.map(m -> m.size())
+				.map(Map::size)
```
panchenko: mapToInt() directly here?

artembilan:
OK. No cherry-pikcing to `1.7.x`. Looks like there is no that internal executor yet...

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/PublisherCallbackChannelImpl.java
```diff
@@ -817,7 +820,12 @@ public void close() throws IOException, TimeoutException {
 	}
 
 	private void shutdownCompleted(String cause) {
-		this.executor.execute(() -> generateNacksForPendingAcks(cause));
+		if (!this.executor.isShutdown()) {
+			this.executor.execute(() -> generateNacksForPendingAcks(cause));
+			if (!this.executorExplicitlySet) {
+					this.executor.shutdown();
```
panchenko: style: indentation looks wrong here.

artembilan:
@garyrussell ,

Any thoughts how the test may fail:
```
org.springframework.amqp.rabbit.listener.DirectMessageListenerContainerMockTests > testRemoveQueuesWhileNotConnected FAILED
    org.mockito.exceptions.verification.TooLittleActualInvocations at DirectMessageListenerContainerMockTests.java:255
```
?

It doesn't for me locally 😢 

Thanks

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/PublisherCallbackChannelImpl.java
```diff
@@ -119,7 +121,8 @@ public PublisherCallbackChannelImpl(Channel delegate) {
 	public PublisherCallbackChannelImpl(Channel delegate, @Nullable ExecutorService executor) {
 		delegate.addShutdownListener(this);
 		this.delegate = delegate;
-		this.executor = executor != null ? executor : DEFAULT_EXECUTOR;
+		this.executor = executor != null ? executor : Executors.newSingleThreadExecutor();
```
garyrussell: No; we don't want a new executor for every channel; that's why it's static.

I suggest we move the static executor to the CCF (non-static there though) and always pass it into the ctor; then shut down the executor in the `CCF.destroy()`.

garyrussell:
>Any thoughts how the test may fail:

Yes; it's a race in the tests; I am counting down latch3 in the mock for `basicQos` instead of `basicConsume`.

garyrussell:
Needs a backport - many conflicts.

artembilan:
Well, the workaround is very easy, as we all know, so let’s don’t back-port!

garyrussell:
👍

PR #844, opened 2018-11-08
AMQP-840: RabbitTemplate add hook for timeouts

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-840

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
@@ -1810,7 +1810,19 @@ public class RabbitTemplate extends RabbitAccessor implements BeanFactoryAware, 
 		if (this.logger.isDebugEnabled()) {
 			this.logger.debug("Reply: " + reply);
 		}
+		if (reply == null) {
+			replyTimedOut(message.getMessageProperties().getCorrelationId());
+		}
 		return reply;
+	}
+
+	/**
+	 * Subclasses can implement this to be notified that a reply has timed out.
+	 * @param correlationId the correlationId
+	 * @since 2.1.2
+	 */
+	protected void replyTimedOut(String correlationId) {
+		// NOSONAR
 	}
 
 	/**
--- a/src/reference/asciidoc/amqp.adoc
+++ b/src/reference/asciidoc/amqp.adoc
@@ -3204,7 +3204,7 @@ This is `false` now by default.
 
 By default, the send and receive methods will timeout after 5 seconds and return null.
 This can be modified by setting the `replyTimeout` property.
-Starting with _version 1.5_, if you set the `mandatory` property to true (or the `mandatory-expression` evaluates to
+Starting with version 1.5, if you set the `mandatory` property to true (or the `mandatory-expression` evaluates to
 `true` for a particular message), if the message cannot be delivered to a queue an `AmqpMessageReturnedException` will
 be thrown.
 This exception has `returnedMessage`, `replyCode`, `replyText` properties, as well as the `exchange` and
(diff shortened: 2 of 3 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -1810,9 +1810,21 @@ private Message exchangeMessages(final String exchange, final String routingKey,
...
+		}
 		return reply;
 	}
 
+	/**
+	 * Subclasses can implement this to be notified that a reply has timed out.
+	 * @param correlationId the correlationId
+	 * @since 2.1.2
+	 */
+	protected void replyTimedOut(String correlationId) {
```
artembilan: Can't we emit something like `ReplyTimedOutEvent` instead and don't require to subsclass a `RabbitTemplate`?

garyrussell: See the referenced SO answer; I changed my sample to this...

```java
class MultiReplyTemplate extends RabbitTemplate {

	private static final byte[] NOBODY = new byte[0];

	private final ConcurrentMap<String, Message> replies = new ConcurrentHashMap<>();

	@Override
	public void onMessage(Message message) {
		// Not thread-safe but that's ok since the DRTMLC is single-threadded.
		String corr = message.getMessageProperties().getCorrelationId();
		Message combined = this.replies.get(corr);
		if (combined == null) {
			combined = new Message(NOBODY, new MessageProperties());
			combined.getMessageProperties().getHeaders().put("replies", new ArrayList<Message>());
			this.replies.put(corr, combined);
		}
		List<Message> list = (List<Message>) combined.getMessageProperties().getHeaders().get("replies");
		list.add(message);
		if (list.size() == 3) {
			this.replies.remove(corr);
			combined.getMessageProperties().setCorrelationId(corr);
			super.onMessage(combined);
		}
	}

	@Override
	protected void replyTimedOut(String correlationId) {
		this.replies.remove(correlationId);
		System.out.println("timed out");
	}

}
```

But I suppose we could emit an event as well. WDYT?

artembilan: Well, such an extension also can rely on the event emitting trick.

Or we can just provide an `protected emitReplyTimedOutEvent(request, correlationId)`  to allow to override it for such a exotic logic...

garyrussell: However, the `correlationId` is meaningless outside of the template since it is an internal value. So I think this is the right solution because it only makes sense to a subclass.

There is another related (but different) stack overflow question [here](https://stackoverflow.com/questions/53043123/spring-amqp-request-reply-how-to-manage-the-timeout#comment93250648_53043123), where he wants a notification of late replies so he can send them someplace. Emitting those as an event makes sense. But he never opened a JIRA issue.

panchenko:
To support that use case directly - there can be a variant of convertSendAndReceive() which passes messages to a Predicate until `false` is returned (true = continue, false = stop).

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
artembilan: Gary, I think I am convinced, unless you see something from @panchenko 's comment.

The `LateReplyEvent` we can address in the separate issue...

garyrussell:
@panchenko True, but yikes!!, we already have 46 overloaded `convertSendAndReceive()` methods; we would have to double that.

Plus, this is a bit of a corner case - first time it's come up in 7+ years.

Issue #846, opened 2018-11-14
 Listener not registered exception from RabbitTemplate.sendAndReceive


ssinger:
I have code that is as follows

~~~      
            Message reply = template.sendAndReceive(exchange, routingKey,
                message);
~~~

where template is a RabbitTemplate instance.

This sometimes throws the following exception.

~~~
Caused by: java.lang.IllegalArgumentException: Listener not registered: org.springframework.amqp.rabbit.core.RabbitTemplate@4525951e []
        at org.springframework.util.Assert.notNull(Assert.java:193) ~[spring-core-5.0.9.RELEASE.jar!/:5.0.9.RELEASE]
        at org.springframework.amqp.rabbit.support.PublisherCallbackChannelImpl.addPendingConfirm(PublisherCallbackChannelImpl.java:944) ~[spring-rabbit-2.0.6.RELEASE.jar!/:2.0.6.RELEASE]
        at sun.reflect.GeneratedMethodAccessor257.invoke(Unknown Source) ~[?:?]
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:1.8.0_171]
        at java.lang.reflect.Method.invoke(Method.java:498) ~[?:1.8.0_171]
   at com.sun.proxy.$Proxy116.addPendingConfirm(Unknown Source) ~[?:?]
        at org.springframework.amqp.rabbit.core.RabbitTemplate.setupConfirm(RabbitTemplate.java:2026) ~[spring-rabbit-2.0.6.RELEASE.jar!/:2.0.6.RELEASE]
        at org.springframework.amqp.rabbit.core.RabbitTemplate.doSend(RabbitTemplate.java:1998) ~[spring-rabbit-2.0.6.RELEASE.jar!/:2.0.6.RELEASE]
        at org.springframework.amqp.rabbit.core.RabbitTemplate.exchangeMessages(RabbitTemplate.java:1751) ~[spring-rabbit-2.0.6.RELEASE.jar!/:2.0.6.RELEASE]
        at org.springframework.amqp.rabbit.core.RabbitTemplate.doSendAndReceiveAsListener(RabbitTemplate.java:1702) ~[spring-rabbit-2.0.6.RELEASE.jar!/:2.0.6.RELEASE]
[...]

PR #847, opened 2018-11-15
GH-846: Fix send and receive with confirms

garyrussell:
Fixes https://github.com/spring-projects/spring-amqp/issues/846

Send and receive with a direct container (the default) fails on the second
operation because the remplate is not registered as a listener with the
callback channel.

`doSendAndReceiveWithDirect` does not invoke `doSendAndReceiveAsListener`
within `execute` because the channel is obtained from the container.

The first send and receive succeeds because `execute` is invoked once in the
`useDirectReplyTo` which determines whether the broker supports direct
reply-to.

`addListener()` is called from `execute()` after a channel has been received
from the cache.

Fix is to call `addListener()` from `doSendAndReceiveWithDirect`.

**cherry-pick to 2.0.x**

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
@@ -1769,8 +1769,11 @@ public class RabbitTemplate extends RabbitAccessor implements BeanFactoryAware, 
 		}
 		ChannelHolder channelHolder = container.getChannelHolder();
 		try {
-			return doSendAndReceiveAsListener(exchange, routingKey, message, correlationData,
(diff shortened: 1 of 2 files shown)


artembilan:
Not sure if related, but Travis has a failure:
```
org.springframework.amqp.rabbit.listener.BrokerDeclaredQueueNameTests > testBrokerNamedQueueDMLC() FAILED
    org.junit.ComparisonFailure at BrokerDeclaredQueueNameTests.java:132
```

Thanks

artembilan:
... and back-ported as https://github.com/spring-projects/spring-amqp/commit/1317215572afbf2dec91393858b65523c1a56cd6 after removing `RabbitTemplatePublisherCallbacksIntegrationTests3`, which doesn't exist in `2.0.x` yet.

Issue #849, opened 2018-11-16
Producer appeared Channel shutdown: clean channel shutdown; protocol method: #method<channel.close>(reply-code=200, reply-text=Closed due to exception from ConfirmListener.handle{N,A}ck, class-id=0, method-id=0)


johnbukaixin:
The environment is spring-amqp:2.0.8，amqp-client:5.4.3. So,does anyone know the possible reason?

garyrussell:
It means we had some problem processing the acknowledgment; you should see the root cause elsewhere in the log; I just ran a test...

```
2018-11-16 09:54:44.863 ERROR 58565 --- [ 127.0.0.1:5672] c.r.c.impl.ForgivingExceptionHandler     : ConfirmListener.handle{N,A}ck threw an exception for channel AMQChannel(amqp://guest@127.0.0.1:5672/,1)

java.lang.RuntimeException: test fail ack
	at org.springframework.amqp.rabbit.connection.PublisherCallbackChannelImpl.handleAck(PublisherCallbackChannelImpl.java:923) ~[spring-rabbit-2.1.2.BUILD-SNAPSHOT.jar:2.1.2.BUILD-SNAPSHOT]
	at com.rabbitmq.client.impl.ChannelN.callConfirmListeners(ChannelN.java:492) [amqp-client-5.4.1.jar:5.4.1]
	at com.rabbitmq.client.impl.ChannelN.processAsync(ChannelN.java:369) [amqp-client-5.4.1.jar:5.4.1]
	at com.rabbitmq.client.impl.AMQChannel.handleCompleteInboundCommand(AMQChannel.java:178) [amqp-client-5.4.1.jar:5.4.1]
	at com.rabbitmq.client.impl.AMQChannel.handleFrame(AMQChannel.java:111) [amqp-client-5.4.1.jar:5.4.1]
	at com.rabbitmq.client.impl.AMQConnection.readFrame(AMQConnection.java:670) [amqp-client-5.4.1.jar:5.4.1]
	at com.rabbitmq.client.impl.AMQConnection.access$300(AMQConnection.java:48) [amqp-client-5.4.1.jar:5.4.1]
	at com.rabbitmq.client.impl.AMQConnection$MainLoop.run(AMQConnection.java:597) [amqp-client-5.4.1.jar:5.4.1]
	at java.lang.Thread.run(Thread.java:748) [na:1.8.0_131]

[...]

PR #851, opened 2018-11-16
GH-849: Pub. Confirm/Return defensive code

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/849

Add defensive catch blocks; we currently catch exceptions in user `handleConfirm`
calls, but not for returns, or for the code that determines which listener to
call.

Throwing exceptions to the client causes the channel to be killed.

**cherry-pick to 2.0.x**


artembilan:
... and cherry-picked to `2.0.x` after resolving conflicts.

PR #858, opened 2018-11-29
Sonar Fixes

garyrussell:


spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/BrokerRunning.java
```diff
@@ -614,4 +611,14 @@ private void closeResources(Connection connection, Channel channel) {
 		}
 	}
 
+	private static class BrokerNotAliveException extends RuntimeException {
```
artembilan: You throw this one from `public isUp()` method, so, there is not going to a simple way to analyze such a `private` exception. What does hold you back to have it as `public`?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/DeclarationExceptionEvent.java
```diff
@@ -30,7 +30,7 @@
 
 	private static final long serialVersionUID = -8367796410619780665L;
 
-	private final Declarable declarable;
+	private transient Declarable declarable;
```
artembilan: Can't it be `final transient` ?

PR #860, opened 2018-11-30
AMQP-846: Fix NPE (adviceChain)

garyrussell:
JIRA: https://jira.spring.io/browse/AMQP-846

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/AbstractRabbitListenerContainerFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/AbstractRabbitListenerContainerFactory.java
@@ -41,6 +41,7 @@ import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.context.ApplicationEventPublisher;
 import org.springframework.context.ApplicationEventPublisherAware;
+import org.springframework.lang.Nullable;
 import org.springframework.retry.RecoveryCallback;
 import org.springframework.retry.support.RetryTemplate;
 import org.springframework.transaction.PlatformTransactionManager;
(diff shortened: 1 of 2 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/AbstractRabbitListenerContainerFactory.java
```diff
@@ -194,8 +195,9 @@ public void setDefaultRequeueRejected(Boolean requeueRejected) {
 	 * @return the advice chain that was set. Defaults to {@code null}.
 	 * @since 1.7.4
 	 */
+	@Nullable
 	public Advice[] getAdviceChain() {
-		return Arrays.copyOf(this.adviceChain, this.adviceChain.length);
+		return this.adviceChain == null ? null : Arrays.copyOf(this.adviceChain, this.adviceChain.length);
```
artembilan: Why do we need copy here at all?
Since I see we do it anyway in the `setAdviceChain()`.

Looks like this is an issue only for the current `2.1.x` version

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -512,8 +512,9 @@ public void setAdviceChain(Advice... adviceChain) {
 		this.adviceChain = Arrays.copyOf(adviceChain, adviceChain.length);
 	}
 
+	@Nullable
 	protected Advice[] getAdviceChain() {
-		return this.adviceChain;
+		return this.adviceChain == null ? null : Arrays.copyOf(this.adviceChain, this.adviceChain.length);
```
artembilan: Right. And this is exactly what I commented above. 
Why do we need to do this?
Is this something SonarQube complains, too?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/AbstractRabbitListenerContainerFactory.java
garyrussell: Because the user could mutate the array but it won't take any effect because we built the proxy only during initialization.

PR #863, opened 2018-12-01
Upgrade to Gradle 5

garyrussell:
- annotation processors are now on their own path instead of the compile path
- Log4J's annotation processor detects the appender


artembilan:
How can we be sure that all plugins we use in the build are really compatible with Gradle 5 and works well?

Thanks

garyrussell:
Good point; I am just tired of seeing the deprecation warning; perhaps we should just add the `annotationProcessor` path for now.

artembilan:
I guess we can move to Gradle 5 when Spring Framework is ready on the matter.
Then we can borrow their experience to upgrade/replace appropriate plugins.

garyrussell:
Adding `annotationProcessor` does not stop the deprecation warning; I guess I'll live with it for now.

artembilan:
On the other hand I see that Spring security has done that already: https://github.com/spring-projects/spring-security/issues/6148

So, maybe not so big pain doing that on our side.

I worry more about docs plugins though...

PR #867, opened 2018-12-07
Use Context CL whenever it is applicable

artembilan:
* Some code style refactoring: mostly lambdas to method references

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
@@ -57,6 +57,7 @@ import org.springframework.jmx.export.annotation.ManagedResource;
 import org.springframework.lang.Nullable;
 import org.springframework.scheduling.concurrent.CustomizableThreadFactory;
 import org.springframework.util.Assert;
+import org.springframework.util.ClassUtils;
 import org.springframework.util.ObjectUtils;
 import org.springframework.util.StringUtils;
 
@@ -617,7 +618,7 @@ public class CachingConnectionFactory extends AbstractConnectionFactory
 		else {
 			interfaces = new Class<?>[] { ChannelProxy.class };
 		}
-		return (ChannelProxy) Proxy.newProxyInstance(ChannelProxy.class.getClassLoader(),
+		return (ChannelProxy) Proxy.newProxyInstance(ClassUtils.getDefaultClassLoader(),
 				interfaces, new CachedChannelInvocationHandler(connection, targetChannel, channelList,
 						transactional));
 	}
@@ -1317,7 +1318,7 @@ public class CachingConnectionFactory extends AbstractConnectionFactory
 
 		private volatile Connection target;
 
-		ChannelCachingConnectionProxy(Connection target) {
+		ChannelCachingConnectionProxy(@Nullable Connection target) {
 			this.target = target;
 		}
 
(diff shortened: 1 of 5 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
```diff
@@ -617,7 +618,7 @@ private ChannelProxy getCachedChannelProxy(ChannelCachingConnectionProxy connect
 		else {
 			interfaces = new Class<?>[] { ChannelProxy.class };
 		}
-		return (ChannelProxy) Proxy.newProxyInstance(ChannelProxy.class.getClassLoader(),
+		return (ChannelProxy) Proxy.newProxyInstance(ClassUtils.getDefaultClassLoader(),
 				interfaces, new CachedChannelInvocationHandler(connection, targetChannel, channelList,
```
garyrussell: According to Andy's comments, we don't really need it here, right?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -1131,7 +1132,7 @@ protected void initializeProxy(Object delegate) {
 		}
 		factory.addInterface(ContainerDelegate.class);
 		factory.setTarget(delegate);
-		this.proxy = (ContainerDelegate) factory.getProxy(ContainerDelegate.class.getClassLoader());
+		this.proxy = (ContainerDelegate) factory.getProxy(ClassUtils.getDefaultClassLoader());
```
garyrussell: ditto

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -583,8 +584,9 @@ private void checkMissingQueues(String[] queueNames) {
 				 * Use reflection to avoid class tangles.
 				 */
 				try {
-					Class<?> clazz = ClassUtils.forName("org.springframework.amqp.rabbit.core.RabbitAdmin",
-							getClass().getClassLoader());
+					Class<?> clazz =
+							ClassUtils.forName("org.springframework.amqp.rabbit.core.RabbitAdmin",
+									ClassUtils.getDefaultClassLoader());
```
garyrussell: and here

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
artembilan: I'd rather have these classes loaded in the `RestartClassLoader` to be able to pick the changes up, than once sticked with the `Launcher$AppClassLoader` instead. However I might be wrong and will change this back, if you insist and have some simple explanation why we don't need this change here and there.

My understanding it is OK for `static` blocks, but if we do something at runtime would be better to rely on the context classloader. 

We can use `getApplicationContext().getClassLoader()` instead though...

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
artembilan: The same here about CL from AC.

PR #889, opened 2019-01-24
Modernizing the look and feel and removing Docbook

Buzzardo:
I set things up to match the new standard we have for producing reference guides. Specifically, it now fetches the necessary resources from spring-doc-resourcesand uses them to put a collapsing ToC on the left and use the modern look and feel. I also drove out Docbook and an old stylesheet.


build.gradle
```diff
@@ -4,8 +4,8 @@ buildscript {
 		maven { url 'https://repo.spring.io/plugins-release' }
 	}
 	dependencies {
-		classpath 'io.spring.gradle:docbook-reference-plugin:0.3.1'
 		classpath 'org.asciidoctor:asciidoctor-gradle-plugin:1.5.8'
+		classpath 'org.asciidoctor:asciidoctorj-pdf:1.5.0-alpha.15'
```
artembilan: It seems to me I've asked this question in other place, but let me to repeat it here anyway for better coverage and context.

Do we really still support PDF view form our Reference Manuals?
Is there a reason to keep this `alpha` dependency?

build.gradle
```diff
@@ -329,40 +327,70 @@ project('spring-rabbit-test') {
 
 }
 
+configurations {
+	docs
+}
+
+def docResourcesVersion = '0.1.0.RELEASE'
```
artembilan: This should go to `ext {` block instead.

build.gradle
```diff
@@ -329,40 +327,70 @@ project('spring-rabbit-test') {
...
+task prepareAsciidocBuild(type: Sync) {
+	dependsOn configurations.docs
+	// copy doc resources
+	from {
+		configurations.docs.collect { zipTree(it) }
+	}
+	// and doc sources
+	from "src/reference/asciidoc/"
+	// to a build directory of your choice
+	into "$buildDir/asciidoc/build"
```
artembilan: Isn't it going to sound awkward: `/build/asciidoc/build`?
Can we just live with the `$buildDir/asciidoc` ?

build.gradle
Buzzardo: We produce PDF for people who want to read it as a book and for people who want to read offline (on airplanes or whatever). I don't know why that's still alpha. It's been around for years now, so it seems like there should be a release version of it.

build.gradle
Buzzardo: That second `build` is to have a working directory for the Asciidoc conversion. I would have called it `working`, but Brian Clozel did that bit. We can change it, of course.

build.gradle
Buzzardo: Fixed. And now I know what the ext block is for. Thanks. :)

build.gradle
Buzzardo: I removed the trailing `build`. It works.

artembilan:
Good. Let me pull it locally, review one more time and build to see a result.
Then we will know how to proceed next.

Thanks so far!

artembilan:
Looks nice, but I see missed Copyright, authors list and version:
image


Is it expected or are we missing anything?

The PDF is in similar problem.
For the note the current view is like this: 

image

Any comments, @Buzzardo, please ?

Buzzardo:
Those were in the XML file for Docbook, I think. I'll add them to the Asciidoc.

src/reference/asciidoc/index.adoc
```diff
@@ -5,6 +5,14 @@
 :numbered:
 :icons: font
 :hide-uri-scheme:
+:spring-amqp-version: 2.1.3
```
artembilan: That's not good. No one is going to support this variable here.
We have a Gradle on top of project build, so would be great to propagate its `project-version` over here into the docs as well.

In Spring Integration we did it like this:
```
[source, groovy, subs="normal"]
----
compile "org.springframework.integration:spring-integration-amqp:{project-version}"
----
```

Hope it can help somehow...
`subs="normal"` is a trick over here.

artembilan: You don't need to sent anything from the Gradle file.
According their docs the `project-version` attribute matches a `$project.version` variable: https://asciidoctor.org/docs/asciidoctor-gradle-plugin/

Buzzardo: That took a little doing, but I got it. It turns out that I had to pass the build.gradle file itself to the Asciidoctor attributes list. That seems odd to me, but it works (and is what the Asciidoctor folks say to do).

build.gradle
```diff
@@ -358,7 +358,9 @@ task reference(type: org.asciidoctor.gradle.AsciidoctorTask) {
 		}
 	options doctype: 'book', eruby: 'erubis'
 	attributes 'source-highlighter': 'coderay',
-		'icons': 'font'
+		'icons': 'font',
+		'build-gradle': file('build.gradle'),
+		'spring-amqp-version': "${project.version}"
```
artembilan: Why can't we just live with an existing `project-version` property?
Why do we need to introduce a new one for the same meaning?

We are here really in the scope of the current project and don't build any composite docs with any other projects...

build.gradle
```diff
@@ -329,40 +328,72 @@ project('spring-rabbit-test') {
...
+task reference(type: org.asciidoctor.gradle.AsciidoctorTask) {
+	backends = ["pdf"]
+	sourceDir "$buildDir/asciidoc"
+	sources {
+		include "index.adoc"
+		}
+	options doctype: 'book', eruby: 'erubis'
+	attributes 'source-highlighter': 'coderay',
+		'icons': 'font',
+		'build-gradle': file('build.gradle'),
```
artembilan: Wait. Why do we need this variable?

src/reference/asciidoc/index.adoc
```diff
@@ -1,8 +1,17 @@
...
+:icons: font
+:hide-uri-scheme:
+Mark Pollack; Mark Fisher; Oleg Zhurakousky; Dave Syer; Gary Russell; Gunnar Hillert; Artem Bilan; Stéphane Nicoll; Arnaud Cogoluègnes; Jay Bryant
+
+(C) 2010 - 2019 by Pivotal Software, Inc.
+
+Copies of this document may be made for your own use and for distribution to others, provided that you do not charge any fee for such copies and further provided that each copy contains this Copyright Notice, whether distributed in print or electronically.
 
-:toc:
+Version {project-version}
```
artembilan: I think we don't need to say `version` here. Plus it would be great to have it in **bold** style.
Also it looks like currently we have it before the Copyright.

Also a new PDF view doesn't show the version, neither Copyright.

I might agree though do not show `Authors` word any more:

image


@garyrussell , WDYT?

garyrussell: Looks nice - thanks Jay!

> think we don't need to say version here

I presume you mean just the word "version" (like with the current docs).

I honestly don't mind either way, but I think having the word is probably better than not.

artembilan: OK! 

Then it still has to be moved before Copyright and be bold anyway! 😄 

Although the problem with PDF remains...

src/reference/asciidoc/index.adoc
```diff
@@ -5,13 +5,13 @@
 :numbered:
 :icons: font
 :hide-uri-scheme:
-Mark Pollack; Mark Fisher; Oleg Zhurakousky; Dave Syer; Gary Russell; Gunnar Hillert; Artem Bilan; Stéphane Nicoll; Arnaud Cogoluègnes; Jay Bryant
+Authors: Mark Pollack; Mark Fisher; Oleg Zhurakousky; Dave Syer; Gary Russell; Gunnar Hillert; Artem Bilan; Stéphane Nicoll; Arnaud Cogoluègnes; Jay Bryant
```
artembilan: I would prefer without `Authors:` since it is pretty nice on the final page after the tittle.

Hence my point do not have `Version:` label as well. 

It is fully clear what is that in bold on its own line 😄 

But I still would like to see version before Copyright:

image


If you all are OK, I can merge with that my polishing

artembilan:
I think we can live with that bug in PDF view...

src/reference/asciidoc/index.adoc
Buzzardo: If you're comfortable making those changes, that's OK by me. Thanks, Artem.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/454e30e1b871d16963488b8142428b6f2abe807c

Note: I've upgrade to the latest Asciidoc plugins and had to add `"project-version=$project.version"`. Looks like it is not propagated any more or it is done some other way. Their documentation doesn't reflect that 🤷‍♂️ 

@Buzzardo ,

Thank you very much for hard work!
Now, I guess, we are good to proceed with the same task for Spring Integration 😄

garyrussell:
Nice - and we're live: https://docs.spring.io/spring-amqp/docs/2.1.4.BUILD-SNAPSHOT/reference/

Great work @Buzzardo ; thanks!

garyrussell:
(and @artembilan for all the review work 😄 )

Buzzardo:
Huge thanks to you fellows. It's nice to get another one modernized. :)

PR #895, opened 2019-01-31
Introducing addBeforePublishPostProcessors method to the RabbitTemplate

mhewedy:
Sometimes we need a way to append our `BeforePublishPostProcessors` to the list ([that might be manipulated by libraries](https://github.com/openzipkin/brave/blob/master/instrumentation/spring-rabbit/src/main/java/brave/spring/rabbit/SpringRabbitTracing.java#L106)).
By calling `setBeforePublishPostProcessors`, it overrides other  `BeforePublishPostProcessors` .

So, this handy method might come to the rescue.


mhewedy:
>Also: why only BeforePublishPostProcessors? How about afterReceivePostProcessors as well?
Shouldn't that Brave library take care after receiving, too, on the other hand?

For the listening part, Brave [advising](https://github.com/openzipkin/brave/blob/master/instrumentation/spring-rabbit/src/main/java/brave/spring/rabbit/SpringRabbitTracing.java#L167) the SimpleRabbitListenerContainerFactory

The point is, my code will run after thier Spring BeanPostProcessors (by definition), unless I put my code in some Spring BeanPostProcessors (with higher precedence) (which I think is a workaround).

If it is fine to proceed, I can fix the PR and add the remove method and the add/remoev for afterReceivePostProcessors.

What do you think?

mhewedy:
I think we might not need to add remove, we are not doing CRUD here.

artembilan:
> might not need to add remove method, we are not doing CRUD here.

Well, if you add your own processor at runtime, you really might need to remove it afterwards in some scenarios.

We have similar consistency in many other places in other projects, and no one called it CRUD yet, but add/remove hooks really do their trick in some cases.

I think remove must be object based and index based.

See `org.springframework.integration.channel.ChannelInterceptorAware` for example.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -584,13 +586,35 @@ public void setBeanFactory(BeanFactory beanFactory) throws BeansException {
...
+	 * {@link BasicProperties} from {@link MessageProperties}. May be used for operations
+	 * such as compression. Processors are invoked in order, depending on {@code PriorityOrder},
+	 * {@code Order} and finally unordered.
+	 * <p>
+	 * In contrast to {@link #setBeforePublishPostProcessors(MessagePostProcessor...)}, this
+	 * method does not override the previously added beforePublishPostProcessors.
+	 * @param beforePublishPostProcessor the post processor.
+	 * @since 2.1.4
+	 */
+	public void addBeforePublishPostProcessors(MessagePostProcessor beforePublishPostProcessor) {
```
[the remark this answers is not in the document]
mhewedy: fixed

artembilan:
The Checkstyle violation:
```
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-amqp/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java:653: At-clauses have to appear in the order '[@param, @return, @throws, @since, @deprecated, @see]'. [AtclauseOrder]
```

You can run `./gradlew check` instead of `test` locally before pushing commits to the PR.

artembilan:
Please, don't squash your commits before pushing to the PR.
There is just enough to add them as is.
This way we keep history for your commits and also a review properly.

We are going to squash them in the end before merging to the master.

mhewedy:
Aha ... Ok Will do.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -584,27 +586,126 @@ public void setBeanFactory(BeanFactory beanFactory) throws BeansException {
...
+
+	/**
+	 * Remove the provided {@link MessagePostProcessor} from the {@link #beforePublishPostProcessors} list.
+	 * @param beforePublishPostProcessor the MessagePostProcessor to remove.
+	 * @since 2.1.4
+	 * @see #addBeforePublishPostProcessor(MessagePostProcessor)
+	 */
+	public void removeBeforePublishPostProcessor(MessagePostProcessor beforePublishPostProcessor) {
+		Assert.notNull(beforePublishPostProcessor, "'beforePublishPostProcessor' cannot be null");
+		if (this.beforePublishPostProcessors != null) {
```
artembilan: There is no reason to have this `if`, since we are covered with the previous `Assert.notNull()`.
You can return `boolean` removal result though.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -584,27 +586,126 @@ public void setBeanFactory(BeanFactory beanFactory) throws BeansException {
...
+	 * Remove a {@link MessagePostProcessor} from the {@link #beforePublishPostProcessors} list for specific index.
+	 * @param index the index for the MessagePostProcessor to remove.
+	 * @return the removed MessagePostProcessor
+	 * @since 2.1.4
+	 * @see #addBeforePublishPostProcessor(MessagePostProcessor)
+	 */
+	@Nullable
+	public MessagePostProcessor removeBeforePublishPostProcessor(int index) {
+		if (this.beforePublishPostProcessors != null) {
+			return new ArrayList<>(this.beforePublishPostProcessors).remove(index);
```
artembilan: Doesn't look good. you remove from a newly created `ArrayList` meanwhile the original one is still going to an item on that position.

src/reference/asciidoc/amqp.adoc
```diff
@@ -3168,6 +3168,9 @@ For example, after your POJO has been converted, the `MessagePostProcessor` enab
 Starting with _version 1.4.2_, additional extension points have been added to the `RabbitTemplate` - `setBeforePublishPostProcessors()` and `setAfterReceivePostProcessors()`.
 The first enables a post processor to run immediately before sending to RabbitMQ. When using batching (see <<template-batching>>), this is invoked after the batch is assembled and before the batch is sent. The second is invoked immediately after a message is received.
 
+_version 2.1.4_ introduced `addBeforePublishPostProcessor()` and `addAfterReceivePostProcessor()` to the `RabbitTemplate` to allow appending new post processor to the list of before publish and after receive post processors respectively. Also there are methods provided to remove the post processors.
```
artembilan: I don't think that sentence should start like this.
Also don't use that `_` symbol to wrap version anymore.

Plus, please, follow "One sentence per line" recommendation: https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
mhewedy: the `notNull` checks the input beforePublishPostProcessor, while the `if` checks the internal list.

artembilan: Oh, right! Pure vision. So, then just `Boolean` return concern

spring-amqp/src/main/java/org/springframework/amqp/support/postprocessor/MessagePostProcessorUtils.java
```diff
@@ -58,6 +61,25 @@ else if (processor instanceof Ordered) {
 		return sorted;
 	}
 
+	/**
+	 * @param collection non nullable collection.
+	 * @param index      the index of the element to be removed.
+	 * @return element that has been removed.
+	 */
+	@Nullable
+	public static MessagePostProcessor remove(Collection<MessagePostProcessor> collection, int index) {
```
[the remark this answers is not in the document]
mhewedy: I am thinking of the remove with an index with the existence of sorting... 

Because also of the following issue:

Say I add 3 Post Processors, and depending on the order, they being shovelled when saved into the instance var `beforePublishPostProcessors`. 

Then I wanted to delete the one at index 2, but because of the sorting shovel, the element at the index 2 is not in place. so it ends up removing another element than what I indented to delete.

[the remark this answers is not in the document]
artembilan: Wow! Exactly! 

Since we are sorting we definitely can't remove by index afterwards.
Please, drop it off altogether.

Good catch!

Thanks for your sagacity 😄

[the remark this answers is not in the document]
mhewedy: methods removed.

garyrussell:
LGTM; except I wonder if the add... methods should have varargs like the set... methods (not critical, just pointing it out). Apologies if it's already been discussed/discarded - I didn't read all the previous comments.

Also, the `AbstractMessageListenerContainer` also has `afterReceivePostProcessors` so we should probably apply the same treatment there; not necessarily part of this PR, however.

artembilan:
Good catch, Gary!

@mhewedy ,

any thoughts if you can apply the same change to the `AbstractMessageListenerContainer` and also reconsider an implementation in favor of vararg?

Thanks

mhewedy:
Yes I can do the suggested updates by Gary. Will work on it and update the PR.

mhewedy:
Thanks for giving me the chance. 😊

Issue #897, opened 2019-02-01
Listener container cannot be created with advice when TCCL is a child of the container's class loader


wilkinsona:
There's a regression in Spring AMQP 2.1 that means that a listener container cannot be created with advice when the TCCL is a child of the container's class loader. This class loader arrangement occurs with Spring Boot's DevTools but will also occur in other environments such as an app server or servlet container where Spring AMQP is installed as a shared library. Please see https://github.com/spring-projects/spring-boot/issues/15697 for more details.

PR #898, opened 2019-02-01
GH-897: Fix CL issue for AMLC$ContainerDelegate

artembilan:
Fixes https://github.com/spring-projects/spring-amqp/issues/897

The `AbstractMessageListenerContainer.ContainerDelegate` is not end-user
class, so proxy for it has to be created by its normal `ClassLoader`


artembilan:
I hope I understood the issue properly...

PR #901, opened 2019-02-09
Reduce factory method complexity

garyrussell:
- proposed new model for non-null property propagation

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/DirectRabbitListenerContainerFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/DirectRabbitListenerContainerFactory.java
@@ -18,6 +18,7 @@ package org.springframework.amqp.rabbit.config;
 
 import org.springframework.amqp.rabbit.listener.DirectMessageListenerContainer;
 import org.springframework.amqp.rabbit.listener.RabbitListenerEndpoint;
+import org.springframework.amqp.utils.JavaUtils;
 import org.springframework.scheduling.TaskScheduler;
 
 /**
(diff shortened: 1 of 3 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/SimpleRabbitListenerContainerFactory.java
```diff
@@ -134,43 +136,37 @@ protected SimpleMessageListenerContainer createContainerInstance() {
...
+
+	private <T> void setIf(boolean predicate, T value, Consumer<T> instance) {
+		if (predicate) {
+			instance.accept(value);
 		}
-		if (this.deBatchingEnabled != null) {
-			instance.setDeBatchingEnabled(this.deBatchingEnabled);
+	}
+
+	private <T> void setIfNotNull(T value, Consumer<T> instance) {
```
artembilan: Sounds like a public utility which can be applied in many other places. However its name sounds better like `acceptIfNotNull()` 😄

garyrussell: Agreed - maybe `JavaUtils`, `CodeUtils` ?

`configureIfNotNull` ?

artembilan: In Spring Boot they have something similar in the `PropertyMapper`.
We may think about something similar and evolve eventually.

From the current perspective an implementation is not really about *configure*, but just a conditional call for the provided `Consumer`, but I can live with whatever you think is better for us

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/DirectRabbitListenerContainerFactory.java
```diff
@@ -104,12 +105,10 @@ protected DirectMessageListenerContainer createContainerInstance() {
 	protected void initializeContainer(DirectMessageListenerContainer instance, RabbitListenerEndpoint endpoint) {
 		super.initializeContainer(instance, endpoint);
 
-		if (this.taskScheduler != null) {
-			instance.setTaskScheduler(this.taskScheduler);
-		}
-		if (this.monitorInterval != null) {
-			instance.setMonitorInterval(this.monitorInterval);
-		}
+		JavaUtils javaUtils = JavaUtils.INSTANCE.acceptIfNotNull(this.taskScheduler, instance::setTaskScheduler)
```
artembilan: I wonder if we need to promote such an approach to Spring Framework directly.
Or if they have already something on the matter...

I have similar method complexity problem in Spring Integration and some common utility would save me.

PR #903, opened 2019-02-11
More JavaUtils - hasText and BiConsumer Flavors

garyrussell:
- also fix race in `RabbitTemplateTests`

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/utils/JavaUtils.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/utils/JavaUtils.java
@@ -86,49 +86,49 @@ public final class JavaUtils {
 	 * Invoke {@link BiConsumer#accept(Object, Object)} with the arguments if the
 	 * condition is true.
 	 * @param condition the condition.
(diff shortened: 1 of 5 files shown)


spring-amqp/src/main/java/org/springframework/amqp/utils/JavaUtils.java
```diff
@@ -66,4 +69,68 @@ private JavaUtils() {
...
+		}
+		return this;
+	}
+
+	/**
+	 * Invoke {@link BiConsumer#accept(Object, Object)} with the arguments if the
+	 * condition is true.
+	 * @param condition the condition.
+	 * @param value the second consumer argument
+	 * @param t the first consumer argument
```
artembilan: The new Spring Check Style doesn't like single-letter variables.
Maybe it would be better to work out a good manner for us in our projects as well.
Plus it isn't clear why you chave this order for arguments while it is much natural to have the "first" as first and "second" as second. I mean how about to swap `value` and `t` in the method contract?

Issue #905, opened 2019-02-15
@RabbitListener use SimpleAsyncTaskExecutor crete the same name thread，when has many SimpleMessageListenerContainer


xxw1754352621:
**Affects Version(s):** 2.1.3.RELEASE

---

## Question
when we use the @RabbitListener ，for create the container，there is a question 。The All Thread name  would be same。Because they use  the below code to create。

RabbitListenerContainerFactory#createListenerContainer(org.springframework.amqp.rabbit.listener.RabbitListenerEndpoint)
![image](https://user-images.githubusercontent.com/8134326/52839627-c3399480-3131-11e9-907a-9363f6efb2cf.png)

But actually，we want the different SimpleAsyncTaskExecutor's thread has its own name，like the below picture. They always name like 'SimpleAsyncTaskExecutor-1',but we  want 'specialSimpleAsyncTaskExecutor-1''

![image](https://user-images.githubusercontent.com/8134326/52839730-0d227a80-3132-11e9-912b-496f33d3b2a8.png)

Just you use the @Bean to create the SimpleMessageListenerContainer  ，the SimpleAsyncTaskExecutor's thread can be name the thread with different name。
we want to name is different ，because we want to analyze the dump file。

garyrussell:
You currently need to create a different container factory for each listener and inject a different `TaskExecutor` into each one. Spring names the threads using the executor's bean name by default.

I will treat this issue as a new feature request to add an `executor` property to `@RabbitListener` to enable different `TaskExecutor` beans to be injected into containers so they can share the same factory.

garyrussell:
I have found a much simpler way to fix this.

garyrussell:
In the mean time; here is a simple work-around...

```properties
spring.rabbitmq.listener.simple.auto-startup=false
```

```java
@SpringBootApplication
public class Rgh905Application {

	public static void main(String[] args) {
		SpringApplication.run(Rgh905Application.class, args);
	}

	@RabbitListener(id = "listen1", queues = "foo")
	public void listen(String in) {
		System.out.println(in + ":" + Thread.currentThread().getName());
	}

	@Bean
	public ApplicationRunner runner1(RabbitTemplate template) {
		return args -> template.convertAndSend("foo", "bar");
	}

	@Bean
	public ApplicationRunner runner2(RabbitListenerEndpointRegistry registry) {
		return args -> {
			registry.getListenerContainers().stream()
				.map(c -> (AbstractMessageListenerContainer) c)
[...]

PR #908, opened 2019-02-15
GH-905: Fix @RabbitListener Thread Names

garyrussell:
Fixes https://github.com/spring-projects/spring-amqp/issues/905

Default container thread names are based on the bean name. `@RabbitListener`
containers are not beans; use `getListenerId()` instead (which falls back
to bean name for other containers).

**cherry-pick to all supported branches**


artembilan:
... and cherry-picked to `2.1.x` and `2.0.x`.
Back ported pt `1.7.x` as https://github.com/spring-projects/spring-amqp/commit/9f7b65ba273caf4df5311479f1163fd34ffebf1c after fixing conflicts and move a fix to the proper class.

Issue #910, opened 2019-02-19
Fix __TypeId__ header for async @RabbitListener return types


garyrussell:
The return value type was incorrectly set to `Mono<?>` or `Future<?>` rather than the generic type.

Using `((ParameterizedType) resultArg.getReturnType()).getActualTypeArguments()[0])` works for concrete types but not wildcards - so use the class of the actual completed `Future/Mono`.

PR #911, opened 2019-02-19
GH-910: Fix return type headers for async returns

garyrussell:
Fixes https://github.com/spring-projects/spring-amqp/issues/910

__cherry-pick to 2.1__


artembilan:

PR #912, opened 2019-02-20
GH-910: Fix async return type erasure

garyrussell:
Fixes https://github.com/spring-projects/spring-amqp/issues/910

Further to the previous patch, avoid async return type erasure by using
the actual generic type of the async result, unless it is a wildcard type.
If it's a wildcard, fall back to the actual return object's type.


artembilan:
Do we back-port this to the previous version?

garyrussell:
Oh, yes please; forgot to add that.

artembilan:

Issue #914, opened 2019-02-22
spring-rabbit-test: rabbitListenerAnnotationProcessor has already been defined


liias:
I am using:
spring boot 2.1.2.RELEASE
spring-rabbit-2.1.3.RELEASE
spring-rabbit-test-2.1.3.RELEASE

When running tests I get an error:

`The bean 'org.springframework.amqp.rabbit.config.internalRabbitListenerAnnotationProcessor', defined in class path resource [org/springframework/amqp/rabbit/test/RabbitListenerTestBootstrap.class], could not be registered. A bean with that name has already been defined in class path resource [org/springframework/amqp/rabbit/annotation/RabbitBootstrapConfiguration.class] and overriding is disabled.`

Spring suggests to use "spring.main.allow-bean-definition-overriding=true" to override the default behaviour and it does work. However, I find it a bit bad solution to change global spring configuration.

There is also possibility that I am doing something wrong.

liias:
Just looked into how **RabbitBootstrapConfiguration** gets loaded in my tests.

Using @RabbitListenerTest does two things:
Enables @EnableRabbit (which does @Import(**RabbitBootstrapConfiguration**.class))
Adds @Import(RabbitListenerTestBootstrap.class)

Now, I can't do @Import(RabbitListenerTestBootstrap.class) instead of @RabbitListenerTest, because 
rabbit-test checks the annotation specifically in org.springframework.amqp.rabbit.test.RabbitListenerTestHarness#RabbitListenerTestHarness.

And even if I could, there are probably some parts of RabbitBootstrapConfiguration I do want?

garyrussell:
I am not sure why you opened an issue here; this has nothing to do with this project.

Can you explain what you think we can do about it?

liias:
Hey.

So, I don't know the project well to state facts, so here are my guesses:
You could take away `@EnableRabbit` from `@RabbitListenerTest`, so RabbitBootstrapConfiguration is not loaded in. In order to get other beans from RabbitBootstrapConfiguration, the RabbitBootstrapConfiguration could be split up or you could redefine all beans in RabbitListenerTestBootstrap, so RabbitListenerTestBootstrap would be a full replacement for RabbitBootstrapConfiguration.

Does that make any sense?

garyrussell:
It makes sense for Boot users but it will break every application that uses `@RabbitListenerTest` but not Spring Boot.

[...]

PR #915, opened 2019-02-22
GH-914: Honor AllowBeanDefOverriding = false

artembilan:
Fixes https://github.com/spring-projects/spring-amqp/issues/914

* Rework `RabbitBootstrapConfiguration` into the `ImportBeanDefinitionRegistrar`
and check for bean definitions presence it is going to register.
This way an override from the `RabbitListenerTestBootstrap` is going to
have a precedence and its `RabbitListenerTestHarness` won't allow a
regular `RabbitListenerAnnotationBeanPostProcessor` to be registered
* Rework `TestRabbitTemplate` to gather all the listener container
from the `ContextRefreshedEvent` instead of the `SmartInitializingSingleton`.
Looks like Spring doesn't care about their order, therefore we need to
be sure that `RabbitListenerAnnotationBeanPostProcessor` has populated
its `registry` with containers before.
* Demonstrate that `allowBeanDefinitionOverriding = false` works well
now in the `ExampleRabbitListenerCaptureTest`
* Optimize `checkTestConfigs` and `updateCopyrights` Gradle tasks for
their `inputs` and `outputs` for better `UP-TO-DATE` handling

**Cherry-pick to 2.1.x without `build.gradle` changes**

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -140,21 +140,25 @@ subprojects { subproject ->
 	[compileJava, compileTestJava]*.options*.compilerArgs = [xLintArg]
 
 	task checkTestConfigs {
+		inputs.files(
+				sourceSets.test.java.srcDirs.collect {
+					fileTree(it)
+							.include('**/*.xml')
+							.exclude('**/log4j2-test.xml')
+				})
+		outputs.dir('build/resources')
+		
 		doLast {
-			def configFiles = []
-			sourceSets.test.allSource.srcDirs.each {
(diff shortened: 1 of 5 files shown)


garyrussell:
See travis errors.

artembilan:
Looks like now its' green.

garyrussell:
And cherry-picked as 5fccf26ee03798171798682d6b45d79be8be562a after removing gradle changes.

Issue #923, opened 2019-03-12
@Lazy @RabbitListener deadlocks during start()


garyrussell:
```java
class LazyListenerStarter implements SmartInitializingSingleton, ApplicationContextAware {

	private final ConnectionFactory cf;

	private final String queue;

	private final String beanName;

	private ApplicationContext applicationContext;

	LazyListenerStarter(ConnectionFactory cf, String queue, String beanName) {
		this.cf = cf;
		this.queue = queue;
		this.beanName = beanName;
	}

	@Override
	public void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
		this.applicationContext = applicationContext;
	}

	@Override
	public void afterSingletonsInstantiated() {
		Executors.newSingleThreadExecutor().execute(() -> {
			try {
				Connection conn = this.cf.createConnection();
				Channel channel = conn.createChannel(false);
				GetResponse got = channel.basicGet(this.queue, false);
				if (got != null) {
					startIt(conn, channel, got.getEnvelope().getDeliveryTag());
				}
				else {
					channel.basicConsume(this.queue, new DefaultConsumer(channel) {

						@Override
						public void handleDelivery(String consumerTag, Envelope envelope, BasicProperties properties,
								byte[] body) throws IOException {

							try {
								getChannel().basicCancel(getConsumerTag());
								startIt(conn, getChannel(), envelope.getDeliveryTag());
							}
							catch (TimeoutException e) {
								e.printStackTrace();
							}
						}
					});
				}
			}
			catch (Exception e) {
				e.printStackTrace();
			}
		});
	}

	private void startIt(Connection conn, Channel channel, long deliveryTag) throws IOException, TimeoutException {
		channel.basicReject(deliveryTag, true);
		channel.close();
		conn.close();
		this.applicationContext.getBean(this.beanName); // DEADLOCK here
		System.out.println("Started");
	}

}
```

The deadlock is because `getBean()` locks `this.singletonObjects()` and the container waits 60 seconds for the container to start; the consumers call `getBean()` so are blocked until that 60 seconds expires.

dsyer:
I ran this code and it printed "Started" when I sent it a message. That means it didn't deadlock right? Here's my main:

```java
@SpringBootApplication
public class LazyApplication {

	public static void main(String[] args) throws Exception {
		new SpringApplicationBuilder(LazyApplication.class).run(args);
	}

}

@Configuration
class RabbitConfiguration {

	@Bean
[...]

PR #925, opened 2019-03-13
GH-923: Fix start delay for @Lazy @RabbitListener

garyrussell:
Fixes https://github.com/spring-projects/spring-amqp/issues/923

While checking for missing or mis-matched queues, a lazily-loaded
listener container can deadlock for 60 seconds.

This occurs if the `allBeanNamesByType` cache does not currently have
an entry for `Queue` (e.g. cleared by registering a singleton).
When lazy beans are referenced, the `RabbitListenerEndpointRegistry`
starts the container and `start()` waits for the consumers to start.
Getting a reference to the lazy bean holds the `singletonObjects`
lock, which is required by the consumer(s) to get the `Queue` beans
to check.

Add a test case to demonstrate the issue.

Disable the redeclaration logic during the initial start of such a
container.

**cherry-pick to 2.1.x**

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
@@ -222,6 +222,8 @@ public abstract class AbstractMessageListenerContainer extends RabbitAccessor
 
 	private String errorHandlerLoggerName = getClass().getName();
 
+	private volatile boolean lazyLoad;
+
 	@Override
 	public void setApplicationEventPublisher(ApplicationEventPublisher applicationEventPublisher) {
 		this.applicationEventPublisher = applicationEventPublisher;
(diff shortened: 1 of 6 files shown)


artembilan:

kushal154:
Can this also be pushed to 2.0.x?

garyrussell:
@kushal154 2.0.x is no longer supported but it was supported when this fix was applied, so it should have been back-ported. I'll see what I can do. But, if you can upgrade to a supported version, that would be better.

garyrussell:
@kushal154 I have back ported the fix to the 2.0.x branch, if you want to try 2.0.14.BUILD-SNAPSHOT from `https://repo.spring.io/snapshot`.

I will build 2.0.14.RELEASE tomorrow (US EST).

kushal154:
Thanks @garyrussell . Will try with the RELEASE tomorrow. Unfortunately our project is stuck with 2.0.x release for now because of some other dependencies.

kushal154:
It's working fine with 2.0.14.BUILD-SNAPSHOT

garyrussell:
2.0.14 is in maven central and the spring release repo.

Issue #935, opened 2018-12-10
Problems after upgrading to spring integration 5.0


bojanv55:
We are using "<spring.integration.version>5.0.8.RELEASE</spring.integration.version>" in combination with "<spring.version>5.0.9.RELEASE</spring.version>".

The problem we are facing is that there are a lot of messages in Unacked state on channels that are without consumers. Java shows that this socket connection is still open, but I guess somehow ListenerContainer is not running.

When we restart the component, all messages are sent to dead-letter queue. 

I know that this is not that much info, but wanted to ask if you already experienced this issues before?

![image](https://user-images.githubusercontent.com/1422698/49744438-09d7c200-fc9d-11e8-9da0-178187df0211.png)

garyrussell:
Moved from SI https://github.com/spring-projects/spring-integration/issues/2658

PR #936, opened 2019-03-19
GH-935: Handle all exceptions in handleDelivery

garyrussell:
Fixes https://github.com/spring-projects/spring-amqp/issues/935

- Don't call basicCancel if already canceled
- Catch all `Exception`s

**cherry-pick to all supported** (may need a separate PR, especially for 1.7.x)


artembilan:

Looks like `2.0.x` version doesn't know about `queueName` at all, so no way to back-port clearly.

Please, proceed with a separate PR.
That one might be "cherry-pickable" already 😄

garyrussell:
OK; will do - will also submit one for 1.7.x (remove lambdas).

PR #964, opened 2019-03-28
Add type-safe queue properties to AmqpAdmin

garyrussell:
Retain (don't deprecate) the current method for JMX users.


spring-amqp/src/main/java/org/springframework/amqp/core/QueueInformation.java
```diff
@@ -0,0 +1,89 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.amqp.core;
+
+/**
+ * Information about a queue, resulting from a passive declaration.
+ *
+ * @author Gary Russell
+ * @since 2.1
```
artembilan: I guess `2.2` already.

Issue #920, opened 2019-03-07
Reuse afterReceivePostProcessors AsyncRabbitTemplate from the provided RabbitTemplate


artembilan:
https://stackoverflow.com/questions/55030829/reply-message-post-processing-in-asyncrabbittemplate

PR #965, opened 2019-03-28
GH-920: Propagate afterReceive to AsyncRTemplate

artembilan:
Fixes https://github.com/spring-projects/spring-amqp/issues/920

The `AsyncRabbitTemplate` is missing an after receive post-processing
in its internal containers for replies

* Reuse `afterPostProcessors` from the provided `RabbitTemplate` and
propagate them into the internal listener containers for replies
* Fix `AbstractCompressingPostProcessor` do not mutate provided
`MessageProperties` and build a fresh instance.
* Demonstrate post-processors propagation by the `GZipPostProcessor`
& `GUnzipPostProcessor` configuration in the `AsyncRabbitTemplateTests`

**Cherry-pick to 2.1.x**

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/support/postprocessor/AbstractCompressingPostProcessor.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/support/postprocessor/AbstractCompressingPostProcessor.java
@@ -27,8 +27,10 @@ import org.apache.commons.logging.LogFactory;
 import org.springframework.amqp.AmqpException;
 import org.springframework.amqp.AmqpIOException;
 import org.springframework.amqp.core.Message;
+import org.springframework.amqp.core.MessageBuilderSupport;
 import org.springframework.amqp.core.MessagePostProcessor;
 import org.springframework.amqp.core.MessageProperties;
+import org.springframework.amqp.core.MessagePropertiesBuilder;
 import org.springframework.core.Ordered;
 import org.springframework.util.FileCopyUtils;
 
@@ -38,11 +40,13 @@ import org.springframework.util.FileCopyUtils;
  * present.
  *
  * @author Gary Russell
+ * @author Artem Bilan
+ *
  * @since 1.4.2
  */
 public abstract class AbstractCompressingPostProcessor implements MessagePostProcessor, Ordered {
 
-	private final Log logger = LogFactory.getLog(this.getClass());
+	protected final Log logger = LogFactory.getLog(getClass()); // NOSONAR final
 
 	private final boolean autoDecompress;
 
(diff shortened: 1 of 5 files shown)


artembilan:
This is not going to pass on Travis though, because of https://github.com/spring-projects/spring-framework/issues/22702

philwebb:
@artembilan I just pushed a fix for https://github.com/spring-projects/spring-framework/issues/22702

artembilan:
Fully not sure why it still fails on Travis.
Everything is OK locally. 
Please, check on merge.
Thanks

spring-amqp/src/main/java/org/springframework/amqp/support/postprocessor/AbstractCompressingPostProcessor.java
```diff
@@ -74,13 +78,19 @@ public Message postProcessMessage(Message message) throws AmqpException {
 		try {
 			OutputStream zipper = getCompressorStream(zipped);
 			FileCopyUtils.copy(new ByteArrayInputStream(message.getBody()), zipper);
-			MessageProperties messageProperties = message.getMessageProperties();
-			String currentEncoding = messageProperties.getContentEncoding();
-			messageProperties
-					.setContentEncoding(getEncoding() + (currentEncoding == null ? "" : ":" + currentEncoding));
+			MessageProperties originalProperties = message.getMessageProperties();
+			MessageBuilderSupport<MessageProperties> messagePropertiesMessageBuilder =
```
garyrussell: Why the abstract type here instead of `MessagePropertiesBuilder` ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/AsyncRabbitTemplate.java
```diff
@@ -157,6 +158,11 @@ public AsyncRabbitTemplate(ConnectionFactory connectionFactory, String exchange,
 		this.template.setExchange(exchange == null ? "" : exchange);
 		this.template.setRoutingKey(routingKey);
 		this.container = new SimpleMessageListenerContainer(connectionFactory);
+		Collection<MessagePostProcessor> afterReceivePostProcessors = this.template.getAfterReceivePostProcessors();
+		if (afterReceivePostProcessors != null) {
+			this.container.setAfterReceivePostProcessors(
```
garyrussell: If not null? `JavaUtils` ?

spring-amqp/src/main/java/org/springframework/amqp/support/postprocessor/AbstractCompressingPostProcessor.java
```diff
@@ -74,13 +78,19 @@ public Message postProcessMessage(Message message) throws AmqpException {
...
+					MessagePropertiesBuilder.fromClonedProperties(originalProperties)
```
garyrussell: It's not at all clear why this change is needed. Seems like unnecessary overhead.

spring-amqp/src/main/java/org/springframework/amqp/support/postprocessor/AbstractCompressingPostProcessor.java
artembilan: That is what returned by the `setContentEncoding()`.

I guess I can rework the logic here into the `MessageBuilder` instead...

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/AsyncRabbitTemplate.java
artembilan: Good point!

spring-amqp/src/main/java/org/springframework/amqp/support/postprocessor/AbstractCompressingPostProcessor.java
artembilan: We build here a new message and modify existing properties new new content encoding like `gzip:`.
So, the original message gets wrong properties. 
We don't here on the producer side a guarantee that the provided message is not built by end-user.
In fact the `AsyncRabbitTemplateTests` uses the same `fooMessage` for different sends and I only figured out in debug that it fails because of `gzip:gzip:` prefix in the content encoding for the second send.
It is OK on the consumer side though, since message is created by the Framework and we really don't propagate it to the target end-user code.

Does it make sense?

garyrussell: Got it; thanks, but perhaps it should be an option (default dont clone,)? In most cases, users won't keep the original message around - especially if `convertAndSend()` is used.

garyrussell:
Merged as 38358d306b04c8df47da44970669933e473b5856 with minor doc polishing, and cherry-picked to 2.1.x as 142fd7f44322703b9a932d2fe285413717c974a9

Issue #973, opened 2019-04-10
RabbitListenerTestHarness not autowiring


matthewh86:
**Affects Version(s):** 2.1.5

---
I've seen issue #969 which has the same surface issue that I'm encountering, but the fix advised is not relevant as I have not defined `@EnableRabbit` in the TestConfiguration. It is however defined in the actual app configuration.

**stacktrace**
```
org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'com.example.EventListenerTest': Unsatisfied dependency expressed through field 'harness'; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'org.springframework.amqp.rabbit.test.RabbitListenerTestHarness' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {@org.springframework.beans.factory.annotation.Autowired(required=true)}

	at org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor$AutowiredFieldElement.inject(AutowiredAnnotationBeanPostProcessor.java:596)
	at org.springframework.beans.factory.annotation.InjectionMetadata.inject(InjectionMetadata.java:90)
	at org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor.postProcessProperties(AutowiredAnnotationBeanPostProcessor.java:374)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.populateBean(AbstractAutowireCapableBeanFactory.java:1411)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.autowireBeanProperties(AbstractAutowireCapableBeanFactory.java:391)
	at org.springframework.test.context.support.DependencyInjectionTestExecutionListener.injectDependencies(DependencyInjectionTestExecutionListener.java:119)
	at org.springframework.test.context.support.DependencyInjectionTestExecutionListener.prepareTestInstance(DependencyInjectionTestExecutionListener.java:83)
	at org.springframework.boot.test.autoconfigure.SpringBootDependencyInjectionTestExecutionListener.prepareTestInstance(SpringBootDependencyInjectionTestExecutionListener.java:44)
	at org.springframework.test.context.TestContextManager.prepareTestInstance(TestContextManager.java:246)
	at org.springframework.test.context.junit.jupiter.SpringExtension.postProcessTestInstance(SpringExtension.java:97)
[...]

PR #975, opened 2019-04-10
GH-973: Higher order for RabbitListenerTestHarness

artembilan:
Fixes https://github.com/spring-projects/spring-amqp/issues/973

When we use `@EnableRabbit` and `@RabbitListenerTest` in the same
configuration set, e.g. mixing real `@Configuration` and test one for
`@RabbitListenerTest`, we may end up with the case when `@EnableRabbit`
is processed before `@RabbitListenerTest`, so, `RabbitListenerTestHarness`
bean is not going to appear in the application context.

* Override `getOrder()` for the `RabbitListenerTestHarness` to give it
higher priority than regular `RabbitListenerAnnotationBeanPostProcessor`,
so it is registered first and then the last one won't override existing
bean

**Cherry-pick to 2.1.x**


garyrussell:
Cherry-picked as 1333fdabd2790c1e1b14afb4a38ed65168bf97d4

Issue #922, opened 2019-03-12
Support Batch @RabbitListener


garyrussell:
Add an option to move debatching logic from the container to the listener adapter so that we can support `List<Foo>` as the listener method payload argument.

PR #980, opened 2019-04-11
GH-922: Add Batch-mode @RabbitListener

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/922

- remove unnecessary back-ticks from literal table columns


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/AbstractRabbitListenerContainerFactory.java
```diff
@@ -356,30 +383,33 @@ public C createListenerContainer(RabbitListenerEndpoint endpoint) {
...
+			.acceptIfNotNull(this.phase, instance::setPhase)
+			.acceptIfNotNull(this.afterReceivePostProcessors, instance::setAfterReceivePostProcessors);
+		instance.setDeBatchingEnabled(!this.batchListener);
 		if (endpoint != null) { // endpoint settings overriding default factory settings
 			javaUtils
-					.acceptIfNotNull(endpoint.getAutoStartup(), instance::setAutoStartup)
-					.acceptIfNotNull(endpoint.getTaskExecutor(), instance::setTaskExecutor);
+				.acceptIfNotNull(endpoint.getAutoStartup(), instance::setAutoStartup)
+				.acceptIfNotNull(endpoint.getTaskExecutor(), instance::setTaskExecutor);
+			javaUtils
```
artembilan: Looks like this can be combined with the previous chain.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/AbstractRabbitListenerContainerFactory.java
```diff
@@ -344,6 +349,28 @@ public void setContainerConfigurer(Consumer<C> configurer) {
...
+	 * Set a {@link BatchingStrategy} to use when debatching messages.
+	 * @param batchingStrategy the batching strategy.
+	 * @since 2.2
+	 * @see #setBatchListener(boolean)
+	 */
+	public void setBatchingStrategy(BatchingStrategy batchingStrategy) {
+		this.batchingStrategy = batchingStrategy;
+	}
+
+	@SuppressWarnings("deprecation")
```
artembilan: Doesn't look like there is some in the method below to suppress.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/BatchMessagingMessageListenerAdapter.java
```diff
@@ -0,0 +1,78 @@
...
+
+			if (this.converterAdapter.isMessageList()) {
+				List<Message<?>> messages = new ArrayList<>();
+				this.batchingStrategy.deBatch(amqpMessage, fragment -> {
+					messages.add(super.toMessagingMessage(fragment));
+				});
+				return new GenericMessage<>(messages);
+			}
+			else {
+				List<Object> list = new ArrayList<Object>();
```
artembilan: We can use diamond operator.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitBatchIntegrationTests.java
```diff
@@ -0,0 +1,156 @@
...
+		@Bean
+		public Listener listener() {
+			return new Listener();
+		}
+
+	}
+
+	public static class Listener {
+
+		List<Foo> foos;
```
artembilan: Time to ditch `foos` and `bars`? 😉

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/AbstractRabbitListenerContainerFactory.java
garyrussell: I thought it was a bit dirty to do that since the property receiver is a different object; I kept changing my mind but ended up with this.

Feel free to change if you feel strongly.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitBatchIntegrationTests.java
garyrussell: While I understand Jay's point about docs, I don't think we have to change all of our tests.

PR #982, opened 2019-04-16
Add typed header getter method

garyrussell:


spring-amqp/src/main/java/org/springframework/amqp/core/MessageProperties.java
```diff
@@ -136,6 +136,19 @@ public void setHeader(String key, Object value) {
...
+	/**
+	 * Typed getter for a header.
+	 * @param headerName the header name.
+	 * @param clazz the type.
+	 * @param <T> the type.
+	 * @return the header value
+	 * @since 2.2
+	 */
+	@SuppressWarnings("unchecked")
+	public <T> T getHeader(String headerName, @SuppressWarnings("unused") Class<T> clazz) {
```
artembilan: If you don't use the `Class<T>`, it doesn't looks like you need it at all.
See Mockito's `InvocationOnMock.getArgument()` and their JavaDocs.

PR #990, opened 2019-04-19
MessageListenerAdapter improve

### LINKED ISSUE #972: Receiver fail to get Channel
**Affects Version(s):*1.7.6* \<>

---
my receiver like:
```
public void processCDN(Object obj1, Object message, Object channel) throws IOException {
        logger.info("consumerContent:" + message);
//        this.msgNullRefusal(content, channel, message);
//        channel.basicAck(message.getMessageProperties().getDeliveryTag(), false);
    }
```
my config like:

```
    @Bean("simpleMessageListenerContainer")
    public SimpleMessageListenerContainer simpleMessageListenerContainer(ConnectionFactory cachingConnectionFactory, DirectConsumerListener directConsumerListener) {
        SimpleMessageListenerContainer simpleMessageListenerContainer = new SimpleMessageListenerContainer();
        simpleMessageListenerContainer.setQueueNames(MQInfoEnum.PACK_CDN.getQueueName() + (MQInfoEnum.PACK_CDN.isSuffixIpHostOnStart() ? NetUtil.getHostIp() : ""));
        simpleMessageListenerContainer.setDefaultRequeueRejected(false);
        simpleMessageListenerContainer.setAcknowledgeMode(AcknowledgeMode.MANUAL);
        simpleMessageListenerContainer.setExposeListenerChannel(true);
        simpleMessageListenerContainer.setChannelTransacted(true);
        simpleMessageListenerContainer.setConnectionFactory(cachingConnectionFactory);
        MessageListenerAdapter messageListenerAdapter = new MessageListenerAdapter(directConsumerListener);
        HashMap<String, String> queueOrTagToMethodName = Maps.newHashMap();
        queueOrTagToMethodName.put(MQInfoEnum.PACK_CDN.getQueueName() + (MQInfoEnum.PACK_CDN.isSuffixIpHostOnStart() ? NetUtil.getHostIp() : ""), "processCDN");
        messageListenerAdapter.setQueueOrTagToMethodName(queueOrTagToMethodName);
        simpleMessageListenerContainer.setMessageListener(messageListenerAdapter);
        return simpleMessageListenerContainer;
    }
```

There are the only message could be received because of MessageListenerAdapter,
so I cannot have channel.basicAck

kc910521:
about closed #972
I must use ->

+ the old style listener adapter i.e MessageListenerAdapter (because i need queueName to be a string that concat queue name with the current server name )
+ i need manual ACK

so I try to extend MessageListenerAdapter, but that is not so convenient, i must copy all code from method onMessage, if this pull request pass, i could finish former biz by override repackArgument with code ->

> if (listenerArguments != null) {
List resList = new LinkedList(Arrays.asList(listenerArguments));
resList.add(channel);
resList.add(message);
return resList.toArray();
}
return listenerArguments;


garyrussell:
This change is reasonable; I am not keen on the method name "repack...", though. How about "enhanceListenerArguments"?

Also, a test case would be good, together with some documentation on how a user might subclass the adapter to use this.

Finally, I am curious as to why you can't use the new-style adapter.

>because i need queueName to be a string that concat queue name with the current server name

doesn't make sense to me.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/MessageListenerAdapter.java
```diff
@@ -344,6 +344,24 @@ protected String getListenerMethodName(Message originalMessage, Object extracted
...
+	 * <p>
+	 * This can be overridden to add argument for receiving such as Channel or Message.
+	 *
+	 * @param listenerArguments the content of the message
+	 * @param channel           the Rabbit channel to operate on
+	 * @param message           the incoming Rabbit message
+	 * @return the array of arguments to be passed into the listener method (each element of the array corresponding to
+	 * a distinct method argument)
+	 */
+	protected Object[] repackArgument(Object[] listenerArguments, Channel channel, Message message) {
```
artembilan: It looks like would be better to make an existing `buildListenerArguments` to accept these extra arguments for the `Channel` and `Message`.

Although it looks like `MessagingMessageListenerAdapter` should be fully enough for your use-case...

kc910521:
@garyrussell @artembilan 

thx for your reply and method name suggests, but I don't think MessagingMessageListenerAdapter could meet my needs.
my project **has many msg listeners with likely naming rules in a class**, but I think there are only carry one method in a MessagingMessageListenerAdapter you declare,(not confirmed)
so i suppose that i must declare 10 MessagingMessageListenerAdapter s for 10 consumers of the queue,  that cause 10 SimpleMessageListenerContainer to receive it.

about use case ... 


 
``public class YourMessageListenerAdapter extends MessageListenerAdapter {

    @Override
    protected Object[] repackArgument(Object[] listenerArguments, Channel channel, Message message) {
        if (listenerArguments != null) {
            List<Object> resList = new LinkedList<Object>(Arrays.asList(listenerArguments));
            // put anything u want
            resList.add(channel);
            resList.add(message);
            return resList.toArray();
        }
        return listenerArguments;
    }

I think change parameters of buildListenerArguments is not good for the older version, because the method is protected, may in use by extends

artembilan:
> I think there are only carry one method in a MessagingMessageListenerAdapter 

That's not correct. See `@RabbitHandler` approach in the Reference Manual: https://docs.spring.io/spring-amqp/docs/2.1.5.RELEASE/reference/#annotation-method-selection

Re. `may in use by extends`, I would say that you try to fix a bug for missed arguments to use for building.

This way I would suggest then to deprecate a `buildListenerArguments()` method and make end-user to use a new one like just `listenerArguments()` with an appropriate set of parameters to consider.

With two methods from API you are going to confuse end-users what and how to implement/override.

However we still can just break that API in the current version and change a `buildListenerArguments()` signature. in the end we will just mention it in the *Migration Guide*.

Let's hear what @garyrussell  thinks on the matter and then ,please, also share some test-case and some docs are needed for this any way.

Thanks

garyrussell:
>However we still can just break that API in the current version and change a buildListenerArguments() signature. in the end we will just mention it in the Migration Guide.

I have no real problems with a breaking change to that method in 2.2.

We don't hear much about this old adapter these days anyway.

However, why not simply add 

```java
protected Object[] buildListenerArguments(Object extractedMessage, Channel channel, Message message) {
    return buildListenerArguments(extractedMessage);
}
```

??

artembilan:
Doh! Of course! We can rely on the polymorphism!
And still deprecate an old method 😄 

I just don't see reason in confusing API...

kc910521:
> also share some test-case and some docs are needed for this any way.

OK, I have pushed test case, but I do not know where should i submit documents, or u mean comment?

kc910521:
> > However we still can just break that API in the current version and change a buildListenerArguments() signature. in the end we will just mention it in the Migration Guide.
> 
> I have no real problems with a breaking change to that method in 2.2.
> 
> We don't hear much about this old adapter these days anyway.
> 
> However, why not simply add
> 
> ```java
> protected Object[] buildListenerArguments(Object extractedMessage, Channel channel, Message message) {
>     return buildListenerArguments(extractedMessage);
> }
> ```
> 
> ??

thx, good suggestion

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/MessageListenerAdapter.java
```diff
@@ -344,6 +344,25 @@ protected String getListenerMethodName(Message originalMessage, Object extracted
...
+	 * mode
+	 *
+	 * @param extractedMessage the content of the message
+	 * @param channel the Rabbit channel to operate on
+	 * @param message the incoming Rabbit message
+	 * @return the array of arguments to be passed into the listener method (each element of the array corresponding to
+	 * a distinct method argument)
+	 */
+	protected Object[] buildListenerArguments(Object extractedMessage, Channel channel, Message message) {
+		return buildListenerArguments(extractedMessage);
```
artembilan: Do we deprecate this old method or not?

For end-users it is still going to be a confusing which one to implement.

garyrussell: Yes, we should deprecate it.

kc910521:
doc uploaded

src/reference/asciidoc/amqp.adoc
```diff
@@ -1705,6 +1705,8 @@ new MessageListenerAdapter((ReplyingMessageListener<String, String>) data -> {
 ----
 ====
 
+Starting with version 2.2, we deprecate method `buildListenerArguments` and provide new one with complete arguments including `Channel` and `Message`, which helping listener to receive customized arguments, such as getting `Channel` to basicReject in manual acknowledge mode by means of doing extends `MessageListenerAdapter` and overriding `buildListenerArguments(Object, Channel, Message)` then return array carried `Channel` and `Message`.
```
artembilan: Please, be more specific which method we deprecate: `buildListenerArguments(Object)`.
"and provide a new one" ?

The `basicReject` must be wrapped into the code, too.

src/reference/asciidoc/whats-new.adoc
```diff
@@ -16,3 +16,8 @@ When using <<receiving-batch,batching>>, `@RabbitListener` methods can now recei
 
 The Log4J and Logback `AmqpAppender` s now support a `verifyHostname` SSL option.
 See <<logging>> for more information.
+
+===== MessageListenerAdapter Changes
+
+You can now receive Channel and Message objects in listener by means of doing extends MessageListenerAdapter.
```
artembilan: I would just say that `MessageListenerAdapter` provides now a new method and an old one is deprecated.

What you say here is just true even now.

The `MessageListenerAdapter` must be wrapped to the code, too.

src/reference/asciidoc/amqp.adoc
kc910521: changed

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/MessageListenerAdapter.java
```diff
@@ -339,11 +339,33 @@ protected String getListenerMethodName(Message originalMessage, Object extracted
 	 * @param extractedMessage the content of the message
 	 * @return the array of arguments to be passed into the listener method (each element of the array corresponding to
 	 * a distinct method argument)
+	 * @deprecated use @{org.springframework.amqp.rabbit.listener.adapter.MessageListenerAdapter#buildListenerArguments
+	 * (java.lang.Object, com.rabbitmq.client.Channel, org.springframework.amqp.core.Message)} to get complete arguments
 	 */
+	@SuppressWarnings("deprecation")
```
panchenko: should be on the new method

kc910521: does old method need any @Deprecated or Supperwarning?

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/adapter/MessageListenerAdapterTests.java
```diff
@@ -205,4 +231,12 @@ public String notDefinedOnInterface(String input) {
 		}
 
 	}
+
+	private class ExtendedListenerAdapter extends MessageListenerAdapter {
```
panchenko: can be an anonymous inner class

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/MessageListenerAdapter.java
panchenko: and here `@Deprecated` as an annotation

kc910521: ok

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/MessageListenerAdapter.java
```diff
@@ -339,11 +339,34 @@ protected String getListenerMethodName(Message originalMessage, Object extracted
 	 * @param extractedMessage the content of the message
 	 * @return the array of arguments to be passed into the listener method (each element of the array corresponding to
 	 * a distinct method argument)
+	 * @deprecated use @{org.springframework.amqp.rabbit.listener.adapter.MessageListenerAdapter#buildListenerArguments
```
panchenko: `{@link`, can be without classname, only a method reference starting with `#`

kc910521: thx

src/reference/asciidoc/amqp.adoc
```diff
@@ -1705,6 +1705,12 @@ new MessageListenerAdapter((ReplyingMessageListener<String, String>) data -> {
 ----
 ====
 
+Starting with version 2.2, we deprecate method `buildListenerArguments(Object)` and provide
```
artembilan: I prefer the formal language in the docs without referring to any person.
So, not `we`, but "the `MessageListenerAdapter.buildListenerArguments(Object)`"  has been deprecated.

Also , please, follow the rule *One sentence per like*: https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

Your next sentence after period starts with lower case, for example.

src/reference/asciidoc/amqp.adoc
```diff
@@ -1705,6 +1705,12 @@ new MessageListenerAdapter((ReplyingMessageListener<String, String>) data -> {
...
+`buildListenerArguments(Object, Channel, Message)`. the new method not only treat `Object` like `buildListenerArguments(Object)` but new arguments
```
artembilan: Typo: `treats`

src/reference/asciidoc/amqp.adoc
```diff
@@ -1705,6 +1705,12 @@ new MessageListenerAdapter((ReplyingMessageListener<String, String>) data -> {
...
+`Channel` and `Message`, which helping listener to custom operations, such as getting `Channel` object to `basicReject(long, boolean)` in manual acknowledge mode
```
artembilan: `to customize` ?

src/reference/asciidoc/whats-new.adoc
```diff
@@ -16,3 +16,9 @@ When using <<receiving-batch,batching>>, `@RabbitListener` methods can now recei
 
 The Log4J and Logback `AmqpAppender` s now support a `verifyHostname` SSL option.
 See <<logging>> for more information.
+
+===== MessageListenerAdapter Changes
+
+`MessageListenerAdapter` provides now a new method to build an array of arguments to be passed into target listener
+and an old one is deprecated.
```
artembilan: Would be better to point at least method name.

Imaging the situation when I already know the Framework and their is just enough for me to read `What's New` to start using new features. I don't see reason to send me to the big chapter if by the method name I would know what to look for in the source code.

kc910521:
@artembilan OK, I'm trying to make the doc simple and supply sample code

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/8ac8ea1f64b17a77afade9f0bcb6b00fda55d33a.

@kc910521 ,

kc910521:
> Merged as [8ac8ea1](https://github.com/spring-projects/spring-amqp/commit/8ac8ea1f64b17a77afade9f0bcb6b00fda55d33a).
> 
> @kc910521 ,
> 
> thank you for contribution; looking forward for more!

My pleasure.

PR #991, opened 2019-04-22
Add Spring Data projection support

garyrussell:
Add support for converting JSON message bodies to Spring Data Projection interfaces.
This allows very selective, and low-coupled bindings to data, including
the lookup of values from multiple places inside the JSON document.
For example the following interface can be defined as a message payload type:

```
interface SomeSample {

  @JsonPath({ "$.username", "$.user.name" })
  String getUsername();

}
```

Accessor methods will be used to lookup the property name as field in the
received JSON document by default.
The @JsonPath expression allows customization of the value lookup, and even
to define multiple JSONPath expressions, to lookup values from multiple
places until an expression returns an actual value.

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/support/converter/AbstractJackson2MessageConverter.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/support/converter/AbstractJackson2MessageConverter.java
@@ -18,6 +18,8 @@ package org.springframework.amqp.support.converter;
 
 import java.io.IOException;
 import java.lang.reflect.Type;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -53,9 +55,7 @@ public abstract class AbstractJackson2MessageConverter extends AbstractMessageCo
 
 	protected final Log log = LogFactory.getLog(getClass()); // NOSONAR protected
 
-	public static final String DEFAULT_CHARSET = "UTF-8";
-
-	private volatile String defaultCharset = DEFAULT_CHARSET;
+	public static final Charset DEFAULT_CHARSET = StandardCharsets.UTF_8;
 
 	/**
 	 * The supported content type; only the subtype is checked, e.g. *&#47;json,
@@ -68,6 +68,8 @@ public abstract class AbstractJackson2MessageConverter extends AbstractMessageCo
 	@Nullable
 	private ClassMapper classMapper = null;
 
+	private Charset defaultCharset = DEFAULT_CHARSET;
+
 	private boolean typeMapperSet;
 
 	private ClassLoader classLoader = ClassUtils.getDefaultClassLoader();
@@ -77,6 +79,8 @@ public abstract class AbstractJackson2MessageConverter extends AbstractMessageCo
 	private boolean useProjectionForInterfaces;
 
 	private ProjectingMessageConverter projectingConverter;
+
+	private boolean standardCharset;
 
 	/**
 	 * Construct with the provided {@link ObjectMapper} instance.
@@ -111,12 +115,15 @@ public abstract class AbstractJackson2MessageConverter extends AbstractMessageCo
 	 * @param defaultCharset The default charset.
 	 */
 	public void setDefaultCharset(@Nullable String defaultCharset) {
-		this.defaultCharset = (defaultCharset != null) ? defaultCharset
+		this.defaultCharset = (defaultCharset != null) ? Charset.forName(defaultCharset)
 				: DEFAULT_CHARSET;
+		if (this.defaultCharset.equals(StandardCharsets.UTF_8)) {
+			this.standardCharset = true;
+		}
 	}
 
 	public String getDefaultCharset() {
-		return this.defaultCharset;
+		return this.defaultCharset.name();
 	}
 
 	@Override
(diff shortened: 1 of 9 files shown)


spring-amqp/src/main/java/org/springframework/amqp/support/converter/AbstractJackson2MessageConverter.java
```diff
@@ -205,7 +223,15 @@ public Object fromMessage(Message message, @Nullable Object conversionHint) thro
 					encoding = getDefaultCharset();
 				}
 				try {
-					if (conversionHint instanceof ParameterizedTypeReference) {
+					JavaType inferredType = this.javaTypeMapper.getInferredType(properties);
+					if (inferredType != null && this.useProjectionForInterfaces && inferredType.isInterface()
+							&& !inferredType.getRawClass().getPackage().getName().startsWith("java.util")) { // List etc
+						if (this.projectingConverter == null) {
+							this.projectingConverter = new ProjectingMessageConverter(this.objectMapper);
```
artembilan: Don't we need to reject such a configuration if there is no `spring-data-common` in the classpath?

Well, I believe we can do that check in the `setUseProjectionForInterfaces()` proactively.

spring-amqp/src/main/java/org/springframework/amqp/support/converter/AbstractJackson2MessageConverter.java
```diff
@@ -265,9 +291,14 @@ protected Message createMessage(Object objectToConvert, MessageProperties messag
 
 		byte[] bytes;
 		try {
-			String jsonString = this.objectMapper
-					.writeValueAsString(objectToConvert);
-			bytes = jsonString.getBytes(getDefaultCharset());
+			if (getDefaultCharset().equals("UTF-8")) {
```
artembilan: There is `StandardCharsets.UTF_8.UTF_8` to consider instead.

Although it even might be better to have that `defaultCharset` property as a `Charset` instance instead to fail early if provided string representation is not supported.

spring-amqp/src/main/java/org/springframework/amqp/support/converter/DefaultJackson2JavaTypeMapper.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2017 the original author or authors.
```
artembilan: Why?

spring-amqp/src/main/java/org/springframework/amqp/support/converter/ProjectingMessageConverter.java
```diff
@@ -0,0 +1,59 @@
...
+ * @author Gary Russell
+ * @since 2.2
+ *
+ */
+public class ProjectingMessageConverter {
+
+	private final ProjectionFactory projectionFactory;
+
+	public ProjectingMessageConverter(ObjectMapper mapper) {
+		JacksonMappingProvider provider = new JacksonMappingProvider(mapper);
```
artembilan: `Assert.notNull(mapper)` ?

src/reference/asciidoc/amqp.adoc
```diff
@@ -3277,6 +3277,44 @@ converter to determine the type.
...
+@RabbitListener(queues = "projection")
+public void projection(SomeSample in) {
+    String username = in.getUsername();
+    ...
+}
+----
+====
+
+Accessor methods will be used to lookup the property name as field in the received JSON document by default.
+The @JsonPath expression allows customization of the value lookup, and even to define multiple JSONPath expressions, to lookup values from multiple places until an expression returns an actual value.
```
artembilan: Wrap `@JsonPath` to the code for consistency.

spring-amqp/src/main/java/org/springframework/amqp/support/converter/AbstractJackson2MessageConverter.java
garyrussell: Yeah - I just copied this from Ollie's code in Kafka,

spring-amqp/src/main/java/org/springframework/amqp/support/converter/DefaultJackson2JavaTypeMapper.java
garyrussell: Wow - I have no idea how that happened.

spring-amqp/src/main/java/org/springframework/amqp/support/converter/AbstractJackson2MessageConverter.java
```diff
@@ -53,9 +55,7 @@
 
 	protected final Log log = LogFactory.getLog(getClass()); // NOSONAR protected
 
-	public static final String DEFAULT_CHARSET = "UTF-8";
-
-	private volatile String defaultCharset = DEFAULT_CHARSET;
+	public static final Charset DEFAULT_CHARSET = StandardCharsets.UTF_8;
```
artembilan: Need JavaDoc for `public` constant.

spring-amqp/src/main/java/org/springframework/amqp/support/converter/AbstractJackson2MessageConverter.java
```diff
@@ -184,6 +195,26 @@ public void setTypePrecedence(Jackson2JavaTypeMapper.TypePrecedence typePreceden
...
+	/**
+	 * Set to true to use Spring Data projection to create the object if the inferred
+	 * parameter type is an interface.
+	 * @param useProjectionForInterfaces true to use projection.
+	 * @since 2.2
+	 */
+	public void setUseProjectionForInterfaces(boolean useProjectionForInterfaces) {
+		this.useProjectionForInterfaces = useProjectionForInterfaces;
+		try {
+			ClassUtils.forName("org.springframework.data.projection.ProjectionFactory", this.classLoader);
```
artembilan: There is `isPresent()` instead.

spring-amqp/src/main/java/org/springframework/amqp/support/converter/AbstractJackson2MessageConverter.java
artembilan: Can't we create this directly in the `setUseProjectionForInterfaces()` ?

Issue #994, opened 2019-05-10
Shutdown channelExecutor problem in CachingConnectionFactory.destroy() with RejectedExecutionException


mehdichitforoosh:
Hi.

**Affects Version(s):** 2.1.4
**Spring Boot**: 2.1.3
## Bug report
When  **publish-returns** or **publish-confirms** properties are set in rabbit properties ,
`CachingConnectionFactory` creates a `PublisherCallbackChannelImpl` with a `channelExecutor` injected in it.
When the app is stopping, `PublisherCallbackChannelImpl.shutdownCompleted() `method runs a task in executor , this causes `RejectedExecutionException` in multichannel connection.
I think we should call `ExecutorService.awaitTermination()` after `ExecutorService.shutdownNow()` in `CachingConnectionFactory.destroy()` .

[How to solve RejectedExecutionException](https://examples.javacodegeeks.com/core-java/util/concurrent/rejectedexecutionexception/java-util-concurrent-rejectedexecutionexception-how-to-solve-rejectedexecutionexception/)

garyrussell:
@mehdichitforoosh Do you have a stack trace? As far as I can tell, all the tasks should have been enqueued before the executor is shut down.

I agree we should use `shutDown()` instead as well as `awaitTermination()` because it's possible some queued tasks will be skipped, but I don't see where a `RejectectExcecutionException` will be thrown; hence I'd like to see a stack trace.

mehdichitforoosh:
```
java.util.concurrent.RejectedExecutionException: Task org.springframework.amqp.rabbit.connection.PublisherCallbackChannelImpl$$Lambda$1233/161497017@4a27c318 rejected from java.util.concurrent.ThreadPoolExecutor@44e0342c[Shutting down, pool size = 2, active threads = 2, queued tasks = 0, completed tasks = 4]
2019-05-10 17:54:04.747 DEBUG 23 --- [      Thread-13] o.s.i.monitor.IntegrationMBeanExporter   : Unregistering JMX-exposed beans on shutdown
    at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution (ThreadPoolExecutor.java:2063)
    at java.util.concurrent.T[WARNING] 
[...]

PR #995, opened 2019-05-10
Fix RejectedExecutionException during close

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/994

`CCF.destroy()` abruptly shuts down the `channelsExecutor` (if present).
This can cause `RejectedExecutionException` because the publisher
callback channel uses the same executor to schedule nacks if necessary.

- Use an `ActiveObjectCounter` to keep track of in-flight async closes
- Wait until there are not active objects before calling `shutDown()` (instead of `shutDownNow()`)
- Wait until all queued tasks are complete
- Add `PublisherCallbackChannelFactory` to make testing possible

__cherry-pick to 2.1.x__

**Suggest reviewing with `?w=1`**


artembilan:

PR #996, opened 2019-05-10
GH-994: Improve test

garyrussell:
See https://github.com/spring-projects/spring-amqp/issues/994

Move the `closingLatch.countDown()` to `waitForConfirmsOrDie()` so we
are sure the in-flight async close is running.
Previously, there was a race because the latch was counted down just
before the close was issued on the async thread.

**cherry-pick to 2.1.x**


artembilan:

Issue #999, opened 2019-05-15
The channelMax limit is reached with rabbit having higher channel max than client


ldai100:
Hi, I've been getting this error for a while and I don't see anything wrong in my configuration. Here's my stack trace:
```
org.springframework.amqp.AmqpResourceNotAvailableException: The channelMax limit is reached. Try later.
	at org.springframework.amqp.rabbit.connection.SimpleConnection.createChannel(SimpleConnection.java:59)
	at org.springframework.amqp.rabbit.connection.CachingConnectionFactory$ChannelCachingConnectionProxy.createBareChannel(CachingConnectionFactory.java:1326)
	at org.springframework.amqp.rabbit.connection.CachingConnectionFactory$ChannelCachingConnectionProxy.access$200(CachingConnectionFactory.java:1312)
	at org.springframework.amqp.rabbit.connection.CachingConnectionFactory.doCreateBareChannel(CachingConnectionFactory.java:654)
	at org.springframework.amqp.rabbit.connection.CachingConnectionFactory.createBareChannel(CachingConnectionFactory.java:637)
	at org.springframework.amqp.rabbit.connection.CachingConnectionFactory.getCachedChannelProxy(CachingConnectionFactory.java:607)
	at org.springframework.amqp.rabbit.connection.CachingConnectionFactory.getChannel(CachingConnectionFactory.java:498)
	at org.springframework.amqp.rabbit.connection.CachingConnectionFactory.access$1600(CachingConnectionFactory.java:101)
	at org.springframework.amqp.rabbit.connection.CachingConnectionFactory$ChannelCachingConnectionProxy.createChannel(CachingConnectionFactory.java:1331)
	at org.springframework.amqp.rabbit.core.RabbitTemplate.doExecute(RabbitTemplate.java:2067)
	at org.springframework.amqp.rabbit.core.RabbitTemplate.lambda$execute$13(RabbitTemplate.java:2024)
	at org.springframework.retry.support.RetryTemplate.doExecute(RetryTemplate.java:287)
	at org.springframework.retry.support.RetryTemplate.execute(RetryTemplate.java:180)
	at org.springframework.amqp.rabbit.core.RabbitTemplate.execute(RabbitTemplate.java:2023)
	at org.springframework.amqp.rabbit.core.RabbitTemplate.send(RabbitTemplate.java:982)
	at org.springframework.amqp.rabbit.core.BatchingRabbitTemplate.send(BatchingRabbitTemplate.java:67)
[...]

PR #1004, opened 2019-05-17
GH-999: Release permit when channel is closed

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/999

The previous commit fixes permit releases when the close is deferred
waiting for confirms.

For logical closes, the `releasePermitIfNecessary()` was moved to
`returnToCache`. However, if `logicalClose()` detects the underlying
channel is already closed, it is not returned to the cache, it is
discarded.

In this case, `logicalClose()` must release the permit.

Add a test case to expose the issue and verify it's corrected.

**cherry-pick to 2.1.x**


artembilan:

PR #1005, opened 2019-05-21
Handle async result with null completion

garyrussell:
If a listener method completes its async result with a null
value, do not attempt to send the result.

**cherry-pick to 2.1.x**


artembilan:
... and cherry-picked to `2.1.x` as https://github.com/spring-projects/spring-amqp/commit/6a872ad447cad602f59cffaa36594b3814615f4b after fixing conflicts.

Issue #1006, opened 2019-05-22
RabbitListener's returning Mono<Void> doesn't ack the message


johanhaleby:
As discussed on [stackoverflow](https://stackoverflow.com/questions/56231398/how-to-ack-nack-with-reactive-rabbitlistener-in-spring-amqp) a `RabbitListener` returning `Mono<Void>` will not ack the message. For example:

```java
@RabbitListener
public Mono<Void> myListener(MyMessage myMessage) {
    Mono<Void> mono = myService.doSomething(myMessage);
    return mono;
}
```

The reason is that `spring-rabbit` doesn't seem to handle empty streams.

jndietz:
This still appears to be a problem.

```groovy
    @RabbitListener(queues = "audit")
    Mono<Void> listen(CustomerActivityEventWrapper wrapper) {
        log.debug("Received message {}", wrapper.toString())
        strategies
                .find { it.handles(wrapper.payload) }
                .handleEvent(wrapper.payload)
                      // 👆 returns a Mono<Void>
    }
```

The API call happening within my strategy occurs, but I still see this in the console:

```
2022-08-04 12:15:20.056 DEBUG 21028 --- [ntContainer#0-2] com.somecompany.listener.EventListener          : Received message com.somecompany.model.rabbit.event.CustomerActivityEventWrapper@150f7163
2022-08-04 12:15:20.184  WARN 21028 --- [ntContainer#0-2] .a.r.l.a.MessagingMessageListenerAdapter : Container AcknowledgeMode must be MANUAL for a Mono<?> return type; otherwise the container will ack the message immediately
2022-08-04 12:15:20.669 DEBUG 21028 --- [ntContainer#0-2] o.s.w.r.f.client.ExchangeFunctions       : [386278] HTTP POST https:/some-api.somecompany.com/api/v1/customer-events
2022-08-04 12:15:22.303 DEBUG 21028 --- [ctor-http-nio-2] o.s.http.codec.json.Jackson2JsonEncoder  : [386278] Encoding [com.somecompany.model.jsite.event.JsiteAuditEvent@4aac9d27]
2022-08-04 12:15:22.408 DEBUG 21028 --- [ctor-http-nio-2] o.s.w.r.f.client.ExchangeFunctions       : [386278] [1d990f1a-1, L:/10.5.26.18:53692 - R:somecompany.somecompany.com/10.72.93.31:443] Response 201 CREATED
[...]

PR #1011, opened 2019-05-30
GH-1006: Fix listener ack for `Mono<Void>`

artembilan:
Fixes https://github.com/spring-projects/spring-amqp/issues/1006

When listener method returns `Mono<Void>`, the `success` callback is
not called from the Reactor because there is no value to handle.

* Move `basicAck()` to the `completeConsumer` which is called when `Mono`
is completed successfully with value or without

**Cherry-pick to 2.1.x**


garyrussell:
Chery-picked as 3bbd1aa08e83f474d4e76607c2a373088defe936

Issue #1014, opened 2019-06-06
Update MDC handling in the AmqpAppender


egusev:
For proper logging additional information of calling paramenters/context is required. The best option to do that is MDC. 
The current implementation copies MDC into the message's headers.
The main problem with it that Rabbit has limited size for header and default size is pretty small - 131072 bytes (128 kb). unlike of message body it can't be fragmented https://github.com/spring-cloud/spring-cloud-stream-binder-rabbit/issues/159

so the proposals are:
1. remove (or make it optional) serializating MDC into headers
2. add possibility to add MDC into the message body
3. make AmqpAppender.sendEvent protected (instead of private) to allow override it to do custom stuff

egusev:
I mean this class org.springframework.amqp.rabbit.log4j2.AmqpAppender

egusev:
update: JsonLayout already serializes MDC into the body. so only points 1 and 3 are actual

garyrussell:
Sounds reasonable; contributions are welcome: https://github.com/spring-projects/spring-amqp/blob/master/CONTRIBUTING.adoc

PR #1015, opened 2019-06-07
GH-1014 introduced new parameter `serializeMdc` for log4j2 appender

egusev:
GH-1014 introduced new parameter `serializeMdc` for log4j2 appender to turn on the serialization of MDC into header. by default it turned off

https://github.com/spring-projects/spring-amqp/issues/1014

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
@@ -183,7 +183,8 @@ public class AmqpAppender extends AbstractAppender {
 			@PluginAttribute("async") boolean async,
 			@PluginAttribute("charset") String charset,
 			@PluginAttribute(value = "bufferSize", defaultInt = Integer.MAX_VALUE) int bufferSize,
-			@PluginElement(BlockingQueueFactory.ELEMENT_TYPE) BlockingQueueFactory<Event> blockingQueueFactory) {
+			@PluginElement(BlockingQueueFactory.ELEMENT_TYPE) BlockingQueueFactory<Event> blockingQueueFactory,
+  		    @PluginAttribute(value = "serializeMdc") boolean serializeMdc) {
 		if (name == null) {
 			LOGGER.error("No name for AmqpAppender");
 		}
@@ -226,6 +227,7 @@ public class AmqpAppender extends AbstractAppender {
 		manager.clientConnectionProperties = clientConnectionProperties;
 		manager.charset = charset;
 		manager.async = async;
+		manager.serializeMdc = serializeMdc;
 
 		BlockingQueue<Event> eventQueue;
 		if (blockingQueueFactory == null) {
@@ -283,7 +285,7 @@ public class AmqpAppender extends AbstractAppender {
 		return message;
 	}
 
-	private void sendEvent(Event event, Map<?, ?> properties) {
+	protected void sendEvent(Event event, Map<?, ?> properties) {
 		LogEvent logEvent = event.getEvent();
 		String name = logEvent.getLoggerName();
 		Level level = logEvent.getLevel();
@@ -312,8 +314,10 @@ public class AmqpAppender extends AbstractAppender {
 		amqpProps.setTimestamp(tstamp.getTime());
 
 		// Copy properties in from MDC
-		for (Entry<?, ?> entry : properties.entrySet()) {
-			amqpProps.setHeader(entry.getKey().toString(), entry.getValue());
+		if (this.manager.serializeMdc) {
+			for (Entry<?, ?> entry : properties.entrySet()) {
+				amqpProps.setHeader(entry.getKey().toString(), entry.getValue());
+			}
 		}
 		if (logEvent.getSource() != null) {
 			amqpProps.setHeader(
@@ -326,7 +330,7 @@ public class AmqpAppender extends AbstractAppender {
 		doSend(event, logEvent, amqpProps);
 	}
 
-	private void doSend(Event event, LogEvent logEvent, MessageProperties amqpProps) {
+	protected void doSend(Event event, LogEvent logEvent, MessageProperties amqpProps) {
 		StringBuilder msgBody;
 		String routingKey;
 		try {
@@ -604,6 +608,11 @@ public class AmqpAppender extends AbstractAppender {
 		 * the system charset.
 		 */
 		private String charset = Charset.defaultCharset().name();
+
+		/**
+		 * Whether or not add MDC properties into message headers. Default is true to keep backward compatibility.
+		 */
+		private boolean serializeMdc = true;
 
 		private boolean durable = true;
 
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
@@ -184,7 +184,7 @@ public class AmqpAppender extends AbstractAppender {
 			@PluginAttribute("charset") String charset,
 			@PluginAttribute(value = "bufferSize", defaultInt = Integer.MAX_VALUE) int bufferSize,
 			@PluginElement(BlockingQueueFactory.ELEMENT_TYPE) BlockingQueueFactory<Event> blockingQueueFactory,
-  		    @PluginAttribute("serializeMdc") boolean serializeMdc) {
+  		    @PluginAttribute("addMdcAsHeaders") boolean addMdcAsHeaders) {
 		if (name == null) {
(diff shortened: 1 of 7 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
```diff
@@ -603,6 +607,11 @@ public int incrementRetries() {
 		 */
 		private String charset = Charset.defaultCharset().name();
 
+		/**
+		 * Whether or not add MDC properties into message headers. Default is true to keep backward compatibility.
+		 */
+		private boolean serializeMdc = true;
```
garyrussell: I don' think this property name properly reflects what it does.

Maybe `addMdcHeaders` ?

egusev: well, it is not very accurate as while mdc has no headers. maybe addMdcToHeaders?

garyrussell: `addMdcAsHeaders` ?

egusev:
> Also, we need a [reference manual](https://docs.spring.io/spring-amqp/reference/html/#_log4j_2_appender) update that describes this feature and, perhaps, a note about how to use the `JsonLayout` to add MDC to the message body.

How can I do that? is it in this git repo? could you share the link?
Actually the same improvement is required for logback too and it will make this modification general instead of log4j2 specific. I started updating logback appender but have not finished it yet. Thought share first changes to get a feedback

As I remember no specific modification is required to serialize MDC into the message

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
egusev: looks fine

garyrussell:
>As I remember no specific modification is required to serialize MDC into the message

Right, but we need to explain why you would set this property to `false`.

Docs are here: https://github.com/spring-projects/spring-amqp/tree/master/src/reference/asciidoc

Specifically, logging.adoc.

egusev:
> > As I remember no specific modification is required to serialize MDC into the message
> 
> Right, but we need to explain why you would set this property to `false`.
> 
> Docs are here: https://github.com/spring-projects/spring-amqp/tree/master/src/reference/asciidoc
> 
> Specifically, logging.adoc.

Thanks. Updated. Check it please. Feel free to update documentation as far as English is not my native  language

src/reference/asciidoc/logging.adoc
```diff
@@ -168,12 +168,29 @@ The following example shows how to configure a Log4j 2 appender:
...
 
+.Log4j 2 Appender Properties
+[cols="2l,2l,4", options="header"]
+|===
+| Property
+| Default
+| Description
+
+| `addMdcAsHeaders`
+| `false`
```
garyrussell: We don't need these backticks since the columns are marked as `literal` - I pushed a commit this morning to fix this; so you might need to rebase.

You mentioned you are also working on the logback appender. Should we wait for that before merging?

Thanks for your work!

egusev: Hey, I think yes - I've already updated it and implementing and troubleshooting tests

egusev: now it's done, I think

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/logback/AmqpAppender.java
```diff
@@ -810,50 +871,14 @@ public void run() {
...
-								"location",
-								String.format("%s.%s()[%s]", location[0], location[1], location[2]));
-					}
-					String routingKey = AmqpAppender.this.routingKeyLayout.doLayout(logEvent);
-					// Set applicationId, if we're using one
-					if (AmqpAppender.this.applicationId != null) {
-						amqpProps.setAppId(AmqpAppender.this.applicationId);
-					}
+
+					MessageProperties amqpProps = AmqpAppender.this.prepareMessageProperties(event);
```
artembilan: We don't need `this.` prefix for methods.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
```diff
@@ -605,6 +609,11 @@ public int incrementRetries() {
 		 */
 		private String charset = Charset.defaultCharset().name();
 
+		/**
+		 * Whether or not add MDC properties into message headers.
+		 */
+		private boolean addMdcAsHeaders = false;
```
garyrussell: Why false? Originally it was `true` for backwards compatibility.

egusev: Hey, yes, right, it was. I changed my mind by some reasons:
this modification will be included into one of next versions. if someone will update packages and will find missing data (MDC) in logs it is obviosly will be found very soon and it will be clear that it is related to the packages updating. it is not very good but I think it is ok - that person can will check documentation and will find a solution.
And an opposite, if we will keep it as before then people will miss some messages in logs completely while most of messages will be present. It will less observable, can be annoying and they will need much more time to find a solution.

artembilan: I'm fine having it as a `false` by default in the current version. 
Doesn't look like we are going to back-port it...

src/reference/asciidoc/logging.adoc
```diff
@@ -235,7 +244,7 @@ In addition they populate headers with the following values:
 * The level of the log event
 * `thread`: the name of the thread where log event happened
 * The location of the stack trace of the log event call
-* A copy of all the MDC properties
+* A copy of all the MDC properties (in the case if `addMdcAsHeaders` is set to `true`)
```
artembilan: We need to state in the `whats-new.adoc` that now `MDC` is not copied into headers by default.
If we agree that it has to be `false` though...

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
garyrussell: I think it should be `true`; I don't think most people store large data in MDC.

artembilan: Oh! I didn't look at that like this.
So, fine with me either way.

@egusev, your word is last 😄

egusev: sorry guys, my word could be only a first one :)
the main point from my point of view is that logs miss happens unexpectedly and requires time to investigate. having false by default we  eliminate the problem for futher usages. but `true` is also fine for me

src/reference/asciidoc/logging.adoc
egusev: ok, will add it once we will make the final decision

artembilan: OK. So, final decision is to have it as `true` by default. That's it.
Thanks

garyrussell:
I suppose ee could make it false by default, if the layout is the JsonLayout.

egusev:
it seems something wrong happened here https://travis-ci.org/spring-projects/spring-amqp/builds/545176010?utm_source=github_status&utm_medium=notification
no idea how to fix it or restart this check

garyrussell:
I restarted it.

artembilan:
The Travis is broken somehow.

Since we all are agree with what we have in this PR i pull it locally for final review and some polishing and will merge it after testing.

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/925b8c601c2f9c9e6c4d35c6c3b0bb386b9bb11e after some polishing.

Cherry-picking to all other supported branches...
Will let here to know when ready.

artembilan:
Cherry-picked into `2.1.x` as https://github.com/spring-projects/spring-amqp/commit/c983c0469b24e50895a5a52f1b057a1572960f96 after fixing conflicts.

artembilan:
Cherry-picked to `2.0.x` as https://github.com/spring-projects/spring-amqp/commit/af76b709daee8b8fcc76c2e0e2fd50e22bb7c9a5 after fixing more conflicts.

artembilan:
And finally cherry-picked to `1.7.x` after even more fixing conflicts: https://github.com/spring-projects/spring-amqp/tree/1.7.x

Let's consider next time do not back-port some new feature or do that really simple way.

@egusev ,

thank you for the contribution anyway; looking forward for more!

Issue #1021, opened 2019-06-12
@sendTo does not work with errorHandler


priyanshus1:
**Affects Version(s):** 2.1.5

## Question
I am using @sendTo annotation with RabbitListenerErrorHandler errorHandler().
But I am getting following error:

```
Caused by: org.springframework.amqp.rabbit.listener.adapter.ReplyFailureException: Failed to send reply with payload 'InvocationResult [returnValue=org.springframework.remoting.support.RemoteInvocationResult@2c663a5c, returnType=null]'
	at org.springframework.amqp.rabbit.listener.adapter.AbstractAdaptableMessageListener.doHandleResult(AbstractAdaptableMessageListener.java:385) ~[spring-rabbit-2.1.5.RELEASE.jar:2.1.5.RELEASE]
	at org.springframework.amqp.rabbit.listener.adapter.AbstractAdaptableMessageListener.handleResult(AbstractAdaptableMessageListener.java:339) ~[spring-rabbit-2.1.5.RELEASE.jar:2.1.5.RELEASE]
	at org.springframework.amqp.rabbit.listener.adapter.MessagingMessageListenerAdapter.returnOrThrow(MessagingMessageListenerAdapter.java:162) ~[spring-rabbit-2.1.5.RELEASE.jar:2.1.5.RELEASE]
	at org.springframework.amqp.rabbit.listener.adapter.MessagingMessageListenerAdapter.onMessage(MessagingMessageListenerAdapter.java:147) ~[spring-rabbit-2.1.5.RELEASE.jar:2.1.5.RELEASE]
	at org.springframework.amqp.rabbit.listener.AbstractMessageListenerContainer.doInvokeListener(AbstractMessageListenerContainer.java:1552) ~[spring-rabbit-2.1.5.RELEASE.jar:2.1.5.RELEASE]
	... 10 common frames omitted
Caused by: org.springframework.amqp.AmqpException: Cannot determine ReplyTo message property value: Request message does not contain reply-to property, and no default response Exchange was set.
	at org.springframework.amqp.rabbit.listener.adapter.AbstractAdaptableMessageListener.getReplyToAddress(AbstractAdaptableMessageListener.java:465) ~[spring-rabbit-2.1.5.RELEASE.jar:2.1.5.RELEASE]
[...]

PR #1023, opened 2019-06-13
GH-1021: Fix `@SendTo` after error is handled

garyrussell:
https://github.com/spring-projects/spring-amqp/issues/1021

Sending the result from a `RabbitListenerErrorHandler` was broken
for class-level `@RabbitListener` because the send to expression
was lost.

**cherry-pick to 2.1.x**

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapter.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapter.java
@@ -149,7 +149,9 @@ public class MessagingMessageListenerAdapter extends AbstractAdaptableMessageLis
 							.build();
 					Object errorResult = this.errorHandler.handleError(amqpMessage, message, e);
 					if (errorResult != null) {
-						handleResult(new InvocationResult(errorResult, null, null), amqpMessage, channel, message);
+						handleResult(new InvocationResult(errorResult,
+									this.handlerAdapter.getSendToForPayload(message.getPayload()), null),
+								amqpMessage, channel, message);
 					}
 					else {
 						logger.trace("Error handler returned no result");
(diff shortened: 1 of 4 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -149,7 +149,9 @@ public void onMessage(org.springframework.amqp.core.Message amqpMessage, Channel
 							.build();
 					Object errorResult = this.errorHandler.handleError(amqpMessage, message, e);
 					if (errorResult != null) {
-						handleResult(new InvocationResult(errorResult, null, null), amqpMessage, channel, message);
+						handleResult(new InvocationResult(errorResult,
+									this.handlerAdapter.getSendToForPayload(message.getPayload()), null),
```
priyanshus1: The Type send here is still null. 
Do you think it would work better with POJOs, If type is send here ?

garyrussell: Good point, although it's only important for complex return types with lots of generics; for simple POJOs the actual return value is used.

But I will change it.

priyanshus1: Wonderful but I have a question here:
For POJOs where jackson mapper is being used, a Type_Id is attached to header which is usually the POJO name with its namespace, Does that relate with type here ?

garyrussell: For simple classes the `result.getClass()` is enough for the type ID.

For more complex types, eg. `Map<Foo, Bar>` we need the generic return type so that information is conveyed in other headers.

Of course, the error handler must return the same type as the listener, otherwise the headers won't match (the generic return type of the error handler is `Object`, so we can't use that.

artembilan:

PR #1025, opened 2019-06-14
Use openjdk8 instead of oraclejdk8 for Travis

artembilan:


artembilan:
Looks like still doesn't work with Open JDK anyway.

Closing as `Invalid`.

PR #1031, opened 2019-06-21
Fix `SmartLifecycle.stop(Runnable)` usage

artembilan:
* Also remove redundant `stop(Runnable)` implementations which repeat
 a `default` one in the `SmartLifecycle`

**Cherry-pick to 2.1.x & 2.0.x**


garyrussell:
Can't cherry-pick to 2.0.x; SL does not have a default stop(Runnable) in SF 5.0

artembilan:
OK. Let me take care of that then!
Should be as simple as a two classes with `callback` issue.

artembilan:
Back-ported into `2.0.x` as https://github.com/spring-projects/spring-amqp/commit/ff8a5da5fa89aff91eb99bb4159553dcdbe52337

Issue #1032, opened 2019-06-24
Add support for consumer batching


garyrussell:
Currently, consumer batching (providing the listener with the complete batch in one call) only applies when the producer creates the batch.

With the `SimpleMessageListenerContainer`, it should be relatively easy with its existing infrastructure to support presenting the listener with a batch of discrete messages.

- deprecate `txSize` in favor of `batchSize`
- add `enableConsumerBatching`
- create a "short" batch if no messages received in `receiveTimeout`.
- add a new default methods to `MessageListener` `isBatchConsumer()` and `onMessageBatch()`

garyrussell:
Still needs work for `@RabbitListener`.

PR #1033, opened 2019-06-24
GH-1032: Add consumer-side batching support

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1032

- Needs (a lot) more tests
- Needs support in `@RabbirtListener`

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/core/MessageListener.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/core/MessageListener.java
@@ -49,9 +49,7 @@ public interface MessageListener {
 	 * @since 2.2
 	 */
 	default void onMessageBatch(List<Message> messages) {
-		throw new UnsupportedOperationException(
-				"The container has been configured with 'consumerBatchEnabled' but the listener is not "
-				+ "a 'BatchMessageListener");
+		throw new UnsupportedOperationException("This listener does not support message batches");
 	}
 
 }
--- a/spring-amqp/src/main/java/org/springframework/amqp/core/MessageListener.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/core/MessageListener.java
@@ -49,7 +49,9 @@ public interface MessageListener {
 	 * @since 2.2
 	 */
 	default void onMessageBatch(List<Message> messages) {
-		throw new UnsupportedOperationException("Should never be called by the container");
+		throw new UnsupportedOperationException(
+				"The container has been configured with 'consumerBatchEnabled' but the listener is not "
+				+ "a 'BatchMessageListener");
 	}
 
 }
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/ListenerContainerFactoryBean.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/ListenerContainerFactoryBean.java
@@ -368,6 +368,30 @@ public class ListenerContainerFactoryBean extends AbstractFactoryBean<AbstractMe
 		this.receiveTimeout = receiveTimeout;
 	}
 
+	/**
+	 * This property has several functions.
+	 * <p>
+	 * When the channel is transacted, it determines how many messages to process in a
+	 * single transaction. It should be less than or equal to
+	 * {@link #setPrefetchCount(int) the prefetch count}.
+	 * <p>
+	 * It also affects how often acks are sent when using
+	 * {@link org.springframework.amqp.core.AcknowledgeMode#AUTO} - one ack per BatchSize.
+	 * <p>
+	 * Finally, when {@link #setConsumerBatchEnabled(boolean)} is true, it determines how
+	 * many records to include in the batch as long as sufficient messages arrive within
+	 * {@link #setReceiveTimeout(long)}.
+	 * <p>
+	 * <b>IMPORTANT</b> The batch size represents the number of physical messages
+	 * received. If {@link #setDeBatchingEnabled(boolean)} is true and a message is a
+	 * batch created by a producer, the actual number of messages received by the listener
+	 * will be larger than this batch size.
+	 * <p>
+	 *
+	 * Default is 1.
+	 * @param batchSize the batch size
+	 * @since 2.2
+	 */
 	public void setBatchSize(int batchSize) {
 		this.batchSize = batchSize;
 	}
@@ -377,7 +401,14 @@ public class ListenerContainerFactoryBean extends AbstractFactoryBean<AbstractMe
 		setBatchSize(txSize);
 	}
 
-	protected void setConsumerBatchEnabled(boolean consumerBatchEnabled) {
+	/**
+	 * Set to true to present a list of messages based on the {@link #setBatchSize(int)},
+	 * if the container and listener support it.
+	 * @param consumerBatchEnabled true to create message batches in the container.
+	 * @since 2.2
+	 * @see #setBatchSize(int)
+	 */
+	public void setConsumerBatchEnabled(boolean consumerBatchEnabled) {
 		this.consumerBatchEnabled = consumerBatchEnabled;
 	}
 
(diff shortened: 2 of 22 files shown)


artembilan:
I think that is type: "Needs support in `@RabbitListener`"

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -312,17 +317,45 @@ public void setReceiveTimeout(long receiveTimeout) {
...
 	 */
+	@Deprecated
 	public void setTxSize(int txSize) {
-		Assert.isTrue(txSize > 0, "'txSize' must be > 0");
-		this.txSize = txSize;
+		setBatchSize(txSize);
+	}
+
+
+	public void setConsumerBatchEnabled(boolean consumerBatchEnabled) {
```
artembilan: Need JavaDocs for all new API.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/ListenerContainerFactoryBean.java
```diff
@@ -366,8 +368,17 @@ public void setReceiveTimeout(long receiveTimeout) {
 		this.receiveTimeout = receiveTimeout;
 	}
 
+	public void setBatchSize(int batchSize) {
```
artembilan: Javadoc

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/ListenerContainerFactoryBean.java
```diff
@@ -366,8 +368,17 @@ public void setReceiveTimeout(long receiveTimeout) {
...
+		this.batchSize = batchSize;
+	}
+
+	@Deprecated
 	public void setTxSize(int txSize) {
-		this.txSize = txSize;
+		setBatchSize(txSize);
+	}
+
+	protected void setConsumerBatchEnabled(boolean consumerBatchEnabled) {
```
artembilan: Why is this `protected` ? Are you going to detect this option internally by the `@RabbitListener(Handler)` method signature?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -312,17 +317,45 @@ public void setReceiveTimeout(long receiveTimeout) {
...
+	 * <p>
+	 * It also affects how often acks are sent when using
+	 * {@link org.springframework.amqp.core.AcknowledgeMode#AUTO} - one ack per BatchSize.
+	 * <p>
+	 * Finally, when {@link #setConsumerBatchEnabled(boolean)} is true, it determines how
+	 * many records to include in the batch as long as sufficient messages arrive within
+	 * {@link #setReceiveTimeout(long)}.
+	 * <p>
+	 * Default is 1.
+	 * @param batchSize the batch size
```
artembilan: `@since 2.2` ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/api/ChannelAwareBatchMessagelistener.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.springframework.amqp.core.Message;
+
+import com.rabbitmq.client.Channel;
+
+/**
+ * Used to receive a batch of messages if the container supports it.
+ *
+ * @author Gary Russell
+ * @since 5.2
```
artembilan: `2.2` - this is Spring AMQP

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/ListenerContainerFactoryBean.java
garyrussell: stupid IDE (or IDE driver)

garyrussell:
OK; will issue a new PR for additional tests and `@RabbitListener` work.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -1458,7 +1488,13 @@ protected void doInvokeListener(ChannelAwareMessageListener listener, Channel ch
 			}
 			// Actually invoke the message listener...
 			try {
-				listener.onMessage(message, channelToUse);
+				if (data instanceof List) {
+					((ChannelAwareBatchMessagelistener) listener).onMessageBatch((List<Message>) data, channelToUse);
```
artembilan: We don't need to cast here: The `ChannelAwareMessageListener` has that `onMessageBatch()` method

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -1500,13 +1536,21 @@ private void cleanUpAfterInvoke(@Nullable RabbitResourceHolder resourceHolder, C
...
 	 *
 	 * @see org.springframework.amqp.core.MessageListener#onMessage
 	 */
-	protected void doInvokeListener(MessageListener listener, Message message) {
+	@SuppressWarnings("unchecked")
+	protected void doInvokeListener(MessageListener listener, Object data) {
+		Message message = null;
 		try {
-			listener.onMessage(message);
+			if (listener instanceof BatchMessageListener) {
```
artembilan: Don't we need to do something slightly opposite?
Check for `data instance of `List` and always call `onMessageBatch()` since it is there on the `MessageListener` anyway.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/ListenerContainerFactoryBean.java
```diff
@@ -366,8 +368,48 @@ public void setReceiveTimeout(long receiveTimeout) {
...
+	 *
+	 * Default is 1.
+	 * @param batchSize the batch size
+	 * @since 2.2
+	 */
+	public void setBatchSize(int batchSize) {
+		this.batchSize = batchSize;
+	}
+
+	@Deprecated
```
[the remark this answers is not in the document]
garyrussell: I didn't think it was necessary; none of these factory bean methods have javadocs and we only use it within the framework, but I can add if you insist.

spring-amqp/src/main/java/org/springframework/amqp/core/MessageListener.java
```diff
@@ -49,7 +49,9 @@ default void containerAckMode(AcknowledgeMode mode) {
 	 * @since 2.2
 	 */
 	default void onMessageBatch(List<Message> messages) {
-		throw new UnsupportedOperationException("Should never be called by the container");
+		throw new UnsupportedOperationException(
+				"The container has been configured with 'consumerBatchEnabled' but the listener is not "
```
artembilan: I don't think the `consumerBatchEnabled` context makes sense here.
It is really not a listener contract resposibility to worry about a container state...

garyrussell: Well, "should never be called by the container" doesn't really help a user fix the problem.

What do you suggest?

artembilan: Have an assertion in the container. And here in the `UnsupportedOperationException` just mentioned what to use instead.

garyrussell: I don't want to add an assertion in the happy path. Reverted to a generic message.

garyrussell: Not needed - assertion already done in `doInitialize()`.

PR #1037, opened 2019-06-27
GH-1032: Consumer Batching Phase 2 @RabbitListener

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1032

Add basic consumer batching support to the `@RabbitListener` infrastructure.

**Still needs more tests** But I think this can be merged after review.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/SimpleRabbitListenerContainerFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/SimpleRabbitListenerContainerFactory.java
@@ -38,7 +38,7 @@ import org.springframework.amqp.utils.JavaUtils;
 public class SimpleRabbitListenerContainerFactory
 		extends AbstractRabbitListenerContainerFactory<SimpleMessageListenerContainer> {
 
-	private Integer txSize;
+	private Integer batchSize;
 
 	private Integer concurrentConsumers;
 
@@ -54,14 +54,24 @@ public class SimpleRabbitListenerContainerFactory
 
 	private Long receiveTimeout;
 
-	private Boolean deBatchingEnabled;
+	private Boolean consumerBatchEnabled;
 
 	/**
 	 * @param txSize the transaction size.
-	 * @see SimpleMessageListenerContainer#setTxSize
+	 * @see SimpleMessageListenerContainer#setBatchSize
+	 * @deprecated in favor of {@link #setBatchSize(Integer)}
 	 */
+	@Deprecated
 	public void setTxSize(Integer txSize) {
-		this.txSize = txSize;
+		setBatchSize(txSize);
+	}
+
+	/**
+	 * @param batchSize the transaction size.
+	 * @see SimpleMessageListenerContainer#setBatchSize
+	 */
+	public void setBatchSize(Integer batchSize) {
+		this.batchSize = batchSize;
 	}
 
 	/**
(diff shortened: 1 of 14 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/BatchMessagingMessageListenerAdapter.java
```diff
@@ -49,6 +55,34 @@ public BatchMessagingMessageListenerAdapter(Object bean, Method method, boolean
...
 	}
 
+	@Override
+	public void onMessageBatch(List<org.springframework.amqp.core.Message> messages, Channel channel) {
+		Message<?> converted;
+		if (this.converterAdapter.isAmqpMessageList()) {
+			converted = new GenericMessage<>(messages);
+		}
+		else {
+			List<Message<?>> messagingMessages = messages.stream()
```
artembilan: May we avoid Java Streams on critical Framework paths?

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java
```diff
@@ -1495,6 +1545,20 @@ public DirectRabbitListenerContainerFactory directListenerContainerFactory() {
...
 
+		@Bean
+		public SimpleRabbitListenerContainerFactory consumerBatchContainerFactory() {
+			SimpleRabbitListenerContainerFactory factory = new SimpleRabbitListenerContainerFactory();
+			factory.setConnectionFactory(rabbitConnectionFactory());
+			factory.setConsumerTagStrategy(consumerTagStrategy());
+			factory.setBatchListener(true);
+			factory.setContainerConfigurer(container -> {
+				container.setConsumerBatchEnabled(true);
+				container.setDeBatchingEnabled(true);
```
artembilan: Why these two options can't be populated directly by the `SimpleRabbitListenerContainerFactory`  when that `etBatchListener(true)` is used?

src/reference/asciidoc/amqp.adoc
```diff
@@ -2863,6 +2870,61 @@ This header can be obtained by adding the `@Header(AmqpHeaders.LAST_IN_BATCH)` b
 The header is mapped from `MessageProperties.isLastInBatch()`.
 In addition, `AmqpHeaders.BATCH_SIZE` is populated with the size of the batch in every message fragment.
 
+In addition, a new property `consumerBatchEnabled` has been added to the `SimpleMessageListenerContainer`.
+When this is true, the container will create a batch of messages, up to `batchSize`; a partial batch is delivered if `receiveTimeout` elapses with no new messages arriving.
+If a producer-created batch is received, it is debatched and added to the consumer-side batch; therefore the actual number of messages delivered may exceed `batchSize`, which represents the number of messages received from the broker.
+`deBatchingEnabled` must be true when `consumerBatchEnabled` is true.
```
artembilan: Why can't this happen automaticaly?

garyrussell: Because `batchListener` on the factory serves 2 purposes.

1. configures the `BatchMessagingMessageListenerAdapter`
2. sets deBatchingEnabled to false (for producer batches).

I was not comfortable with this either; I think I need to add `consumerBatchEnabled` as a first class property on the factory.

Will push another commit.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java
garyrussell: Because the original purpose of `batchListener` was for producer batches.

artembilan: M-m-m. Can't we cover those "producer batches" with a condition based on the `BatchingStrategy` provided?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/SimpleRabbitListenerContainerFactory.java
```diff
@@ -54,14 +54,24 @@
...
 	 */
+	@Deprecated
 	public void setTxSize(Integer txSize) {
-		this.txSize = txSize;
+		setBatchSize(txSize);
+	}
+
+	/**
+	 * @param batchSize the transaction size.
+	 * @see SimpleMessageListenerContainer#setBatchSize
```
artembilan: Should we say `@since 2.2` ?

Issue #1038, opened 2019-06-28
Automatic reconnection when RMQ is not available when sending the first message.


dmkost:
Hello. Please help me with the problem.

The problem is related to automatic reconnection.
I use Spring Boot 2.1.3, but it happens in version 2.1.6 and RabbitMQ with spring-boot-starter-amqp

Bin RabbitTemplate looks like this:
```java
public final class ExtendedRabbitTemplate extends RabbitTemplate {

    public ExtendedRabbitTemplate(final ConnectionFactory connectionFactory) {
        super(connectionFactory);
    }

    @Override
    protected final void replyTimedOut(final String correlationId) {
        throw new RpcTimeoutException(String.format("Timeout on receiving reply to message with correlation id = %s", correlationId));
    }

}
```
```java
    @Bean
    public RabbitTemplate rabbitTemplate(ConnectionFactory connectionFactory) {
        RabbitTemplate template = new ExtendedRabbitTemplate(connectionFactory);
        template.setMessageConverter(producerJackson2MessageConverter());
        template.setReplyTimeout(properties.getRpcMessagesReplyTimeout());
        return template;
    }
```

We use the following method to support rpc:
![Снимок экрана 2019-06-28 в 15 32 09](https://user-images.githubusercontent.com/25726407/60342382-fe8a8080-99b9-11e9-90e2-a604f8100d8a.png)
We want to use Direct ReplyTo with pseudo queues.

The scenario in which the normal behavior RabbitTemplate
1. Initial state: Internet is available (access to the RMQ), the application is not running.
2. Run the application.
3. We send several messages and get answers. All is well.
4. Turning off the Internet (access to the RMQ).
5. We send several messages and get an exception after a timeout. Everything is also OK.

Scenario in which abnormal behavior is detected:
1. Initial state: the Internet is not available(access to the RMQ), the application is not running
2. Run the application.
3. We send several messages and get exceptions org.springframework.amqp.AmqpConnectException: java.net.ConnectException: Connection refused.
4. Turn on the Internet (access to the RMQ)
5. We send several messages and receive exceptions after a timeout. Therefore, no reconnection occurs.

The error is as follows:
![Снимок экрана 2019-06-28 в 15 15 38](https://user-images.githubusercontent.com/25726407/60343845-ddc42a00-99bd-11e9-9e19-248246d361a2.png)

[...]

PR #1043, opened 2019-07-01
GH-1038: RT: More evaluateFastReplyTo Fixes

garyrussell:
Fixes https://github.com/spring-projects/spring-amqp/issues/1038

The previous fix to catch `AmqpConnectException` was incorrect - the
`ShutdownSignalException` for the passive queue declaration failure
is wrapped in an `AmqpConnectException` so we would have failed to
detect the failure.

Also, it has been reported that sometimes an `AmqpIOException` is thrown.

- Add `AmqpIOException` to the catch block
- Search for, and explitly check for the queue declaration failed exception
- For all other cases rethrow so we can test again

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
@@ -960,16 +960,16 @@ public class RabbitTemplate extends RabbitAccessor // NOSONAR type line count
 				}
 				if (cause instanceof ShutdownSignalException) {
 					if (RabbitUtils.isPassiveDeclarationChannelClose((ShutdownSignalException) cause)) {
-						if (logger.isDebugEnabled()) {
+						if (logger.isWarnEnabled()) {
 							logger.warn("Broker does not support fast replies via 'amq.rabbitmq.reply-to', temporary "
-									+ "queues will be used:" + ex.getMessage() + ".");
+									+ "queues will be used: " + cause.getMessage() + ".");
 						}
 						this.replyAddress = null;
 						return false;
 					}
 				}
 				if (logger.isDebugEnabled()) {
-					logger.debug("IO error, deferring directReplyTo detection");
+					logger.debug("IO error, deferring directReplyTo detection: " + ex.toString());
 				}
 				throw ex;
 			}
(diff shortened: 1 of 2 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -951,18 +953,25 @@ protected boolean useDirectReplyTo() {
...
-					logger.debug("Connection error, deferring directReplyTo detection");
+			catch (AmqpConnectException | AmqpIOException ex) {
+				Throwable cause = ex;
+				while (cause != null && !(cause instanceof ShutdownSignalException)) {
+					cause = cause.getCause();
+				}
+				if (cause instanceof ShutdownSignalException) {
+					if (RabbitUtils.isPassiveDeclarationChannelClose((ShutdownSignalException) cause)) {
+						if (logger.isDebugEnabled()) {
+							logger.warn("Broker does not support fast replies via 'amq.rabbitmq.reply-to', temporary "
```
artembilan: You say `isDebugEnabled()`, but use a `warn()`. Is it expectable?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -951,18 +953,25 @@ protected boolean useDirectReplyTo() {
...
+									+ "queues will be used:" + ex.getMessage() + ".");
```
artembilan: Need space after column

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -951,18 +953,25 @@ protected boolean useDirectReplyTo() {
...
+						return false;
+					}
 				}
-				throw ex;
-			}
-			catch (Exception e) {
 				if (logger.isDebugEnabled()) {
-					logger.warn("Broker does not support fast replies via 'amq.rabbitmq.reply-to', temporary "
-							+ "queues will be used:" + e.getMessage() + ".");
+					logger.debug("IO error, deferring directReplyTo detection");
```
artembilan: No need to wrap to `isDebugEnabled()` any more

PR #1045, opened 2019-07-09
Add filtered getter to Declarables

garyrussell:


spring-amqp/src/main/java/org/springframework/amqp/core/Declarables.java
```diff
@@ -49,6 +51,21 @@ public Declarables(Collection<Declarable> declarables) {
...
 
+	/**
+	 * Return the elements that are instances of the provided class.
+	 * @param <T> The type.
+	 * @param clazz the class.
+	 * @return the filtered list.
+	 * @since 2.2
+	 */
+	@SuppressWarnings("unchecked")
+	public <T> List<T> getFilteredDeclarables(Class<T> clazz) {
```
artembilan: `getDeclarablesByType()` ? Similar to what we have in the `BeanFactory`

spring-amqp/src/main/java/org/springframework/amqp/core/Declarables.java
```diff
@@ -49,6 +51,21 @@ public Declarables(Collection<Declarable> declarables) {
...
+		return this.declarables.stream()
+				.filter(dec -> clazz.isInstance(dec))
```
artembilan: `clazz::isInstance` ?
Not sure what is going on with STS and why it still doesn't support method reference optimization suggestion...

PR #1047, opened 2019-07-17
Setting for BlockingQueueConsumer.failedDeclarationRetryInterval ignored if logging not enabled

ianroberts:
When a BlockingQueueConsumer fails to passively declare a queue, it is supposed to sleep for `failedDeclarationRetryInterval` milliseconds before retrying.  However the `Thread.sleep` is wrapped inside an `if(logger.isWarnEnabled())` with the effect that, when WARN-level logging is _not_ enabled for this class then it ignores the retry interval and tries again immediately - potentially hundreds of times per second.

This is not caught by the tests, because when the tests are running the root logger is set to INFO.

Fix is to move the sleep outside the logger `if` block.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
@@ -637,15 +637,15 @@ public class BlockingQueueConsumer {
 		if (passiveDeclareRetries > 0 && this.channel.isOpen()) {
 			if (logger.isWarnEnabled()) {
 				logger.warn("Queue declaration failed; retries left=" + (passiveDeclareRetries), e);
-				try {
-					Thread.sleep(this.failedDeclarationRetryInterval);
-				}
-				catch (InterruptedException e1) {
-					this.declaring = false;
-					Thread.currentThread().interrupt();
-					this.activeObjectCounter.release(this);
-					throw RabbitExceptionTranslator.convertRabbitAccessException(e1); // NOSONAR stack trace loss
-				}
+			}
+			try {
+				Thread.sleep(this.failedDeclarationRetryInterval);
+			}
+			catch (InterruptedException e1) {
+				this.declaring = false;
+				Thread.currentThread().interrupt();
+				this.activeObjectCounter.release(this);
+				throw RabbitExceptionTranslator.convertRabbitAccessException(e1); // NOSONAR stack trace loss
 			}
 		}
 		else if (e.getFailedQueues().size() < this.queues.length) {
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@ianroberts Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-amqp&pullRequestId=1047)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-amqp&pullRequestId=1047) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@ianroberts Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-amqp&pullRequestId=1047)!

ianroberts:
Travis failure appears to be unrelated to my changes - seems to have been failing the same way for all PRs since https://travis-ci.org/spring-projects/spring-amqp/builds/546782197

artembilan:
Right. Never mind that Travis failure: we also don't understand why it doesn't allow us to use Java 8 and only in this project.
All others our good.

I think to avoid confusion, I'm going to go ahead and disable Travis integration for this project.
Although it looks like I don't have admin permissions. @garyrussell ? 😄 

Anyway I'll review your PR meanwhile.

ianroberts:
https://travis-ci.community/t/oracle-jdk-11-and-10-are-pre-installed-not-the-openjdk-builds/785/16 suggests Oracle JDK 8 not available on Xenial (the travis default) but you can make it work by forcing trusty instead.  Or switch the build to openjdk but that's a whole other can of worms...

garyrussell:
Yeah - we tried to switch to open jdk but had other issues.

Disabled for now.

artembilan:
Yeah... Looks like it was like that from day first: https://github.com/spring-projects/spring-amqp/commit/fd696bfaf97bfe28fc4fa5f7181169ccef5d05e1#diff-936c7a1b05a983a11d2eb7c57ad0c82cR221.

@garyrussell , any thoughts what could be a reason to sleep only in case of warning logging level?
Isn't that just an omission in the logic and this PR really corrects a code flow?

Thanks

ianroberts:
To give some background - I have a slightly unusual use case in that I control the message _producer_ code entirely but the _consumers_ are running potentially untrusted code from third parties. Therefore I want the producer to handle the creation of all the queues (so we can have the option of providing locked down credentials to each consumer restricting them to their own queue), however this is all running in Kubernetes and I can’t necessarily control the order in which the components start up.

So I thought I could make the consumers allow infinite retries at 30 seconds interval so they will wait until the producer is up and has created the queues. When I tried this in testing I got an avalanche of error messages on my rabbit console, so tried cranking up the spring amqp log level to investigate... and it stopped happening.

garyrussell:
>Isn't that just an omission in the logic and this PR really corrects a code flow?

Yes, clearly a bug.

Thanks for the fix @ianroberts

garyrussell:
Should be cherry-picked to all supported branches.

ianroberts:
@artembilan is that sufficient?

artembilan:

Back-ported to `2.0.x` and `1.7.x` as: https://github.com/spring-projects/spring-amqp/commit/89dac79d76639bcc7ddbc8b33d7598d101e6fb34 after fixing conflicts.

artembilan:
@ianroberts ,

thank you very much for contribution; looking forward for more!

Issue #1049, opened 2019-07-18
expose SASL property to the logging subsystems


bcollard:
**Affects Version(s):** 2.1.6.RELEASE

---

## Enhancement

It would be interesting to enforce the "EXTERNAL" SASL mechanism when a client is using a keystore with a client cert to authenticate against a RabbitMQ server.

See this S.O. discussion for more details: https://stackoverflow.com/questions/42851250/saslconfig-using-spring-rabbitmq-not-working

This behavior should exist by default on the logging subsystems (log4j & logback) or be customizable over a new attribute (~"SaslConfig"). See this class for log4j2: https://github.com/spring-projects/spring-amqp/blob/master/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java

bcollard:
see also https://github.com/spring-projects/spring-boot/issues/6719

garyrussell:
There are many properties on the underlying RabbitMQ connection factory and the factory there is not created by the Spring factory bean (it's created by the appender itself, which is outside of the Spring context).

Since there are so many properties, I propose adding

```java
protected void configureRabbitConnectionFactory(ConnectionFactory cf) {
}
```

so users can subclass the Appenders and perform additional configuration as necessary.

I believe that would suit your needs?

artembilan:
I see there is on the `RabbitConnectionFactoryBean`:
```
/**
	 * @param saslConfig the sasl config.
	 * @see com.rabbitmq.client.ConnectionFactory#setSaslConfig(com.rabbitmq.client.SaslConfig)
	 */
	public void setSaslConfig(SaslConfig saslConfig) {
		this.connectionFactory.setSaslConfig(saslConfig);
	}
```

and this one available for access from the overridden `configureRabbitConnectionFactory(RabbitConnectionFactoryBean)`.

So, I think we are covered already and no need in a new method. Only the problem that Log4J appenders are based on `@PluginFactory` static method to be created. So, provide some customization into an `AmqpAppender` we need to pay some effort into an extension. essentially fully custom `AmqpAppender` .

Any other thoughts how to overcome `static` restrictions?

garyrussell:
I didn't notice that we already have that method.

Looks like having protected methods there is useless though due to the static creation.

Given that we already support other SSL props, we should just go ahead and add this one.

Given the static class, and it already has SSL support, we should just go ahead.

bcollard:
[...]

PR #1051, opened 2019-07-18
GH-1049: Support SaslConfig in logging appenders

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1049

**cherry-pick to 2.1.x**

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitUtils.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitUtils.java
@@ -30,8 +30,12 @@ import org.springframework.util.Assert;
 import com.rabbitmq.client.AMQP;
 import com.rabbitmq.client.AlreadyClosedException;
 import com.rabbitmq.client.Channel;
+import com.rabbitmq.client.DefaultSaslConfig;
+import com.rabbitmq.client.JDKSaslConfig;
 import com.rabbitmq.client.Method;
+import com.rabbitmq.client.SaslConfig;
 import com.rabbitmq.client.ShutdownSignalException;
+import com.rabbitmq.client.impl.CRDemoMechanism;
 import com.rabbitmq.client.impl.recovery.AutorecoveringChannel;
 
 /**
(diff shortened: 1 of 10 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitUtils.java
```diff
@@ -369,4 +373,34 @@ public static int getMaxFrame(ConnectionFactory connectionFactory) {
...
+	 * </ul>
+	 * @param saslConfig the string value.
+	 * @param connectionFactory the connection factory to get the name, pw, host.
+	 * @return the saslConfig.
+	 */
+	public static SaslConfig stringToSaslConfig(String saslConfig,
+			com.rabbitmq.client.ConnectionFactory connectionFactory) {
+
+		switch (saslConfig) {
+		case "DefaultSaslConfig.PLAIN":
```
artembilan: String switch i not going to work with Java 6.
On the other hand  I don't understand why do you require a class name here?
Why just plain `plain` is not enough to request from end-user?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
```diff
@@ -195,41 +198,43 @@ public static AmqpAppender createAppender(// NOSONAR NCSS line count
...
-		manager.durable = durable;
-		manager.autoDelete = autoDelete;
-		manager.contentType = contentType;
-		manager.contentEncoding = contentEncoding;
-		manager.connectionName = connectionName;
-		manager.clientConnectionProperties = clientConnectionProperties;
-		manager.charset = charset;
-		manager.async = async;
-		manager.addMdcAsHeaders = addMdcAsHeaders;
+		JavaUtils.INSTANCE
```
artembilan: `JavaUtils` is there since `2.1.x`. So, this is not going to be clearly back-ported into `2.0.x` and `1.7.x`.
However I can play with `1.7.x` if you confirm that we don't back-port to `2.0.x` because of its EOL 😄

garyrussell: Right; just 2.1.x - it turns out SSL support wasn't added until 2.0 anyway https://jira.spring.io/browse/AMQP-624

I've repushed with a modified commit comment.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitUtils.java
garyrussell: Because `JDKSaslSupport` also supports `PLAIN`.

artembilan:
... and cherry-picked into `2.1.x` after fixing some conflicts in test class.

artembilan:
Added more polishing into `2.1.x` since there is no AssertJ dependency yet: https://github.com/spring-projects/spring-amqp/commit/586fa836982bf44f2da8f2aa130ddf064fb2a008

Issue #1053, opened 2019-07-23
ImmediateRequeueAmqpException doesn't work


dqueffeulouatw:
**Affects Version(s):** 2.1.7

---

Hi, 

I think there is a bug in the org.springframework.amqp.rabbit.listener.ContainerUtils method.
I'm trying to throw an ImmediateRequeueAmqpException from a RabbitListener to requeue a message but it doesn't work.
While debugging I found out that the throwable is not an ImmediateRequeueAmqpException but a ListenerExecutionFailedException with a cause set to ImmediateRequeueAmqpException .
So the test on the throwable doesn't work.

Here is the code :
```
public static boolean shouldRequeue(boolean defaultRequeueRejected, Throwable throwable, Log logger) {
        boolean shouldRequeue = defaultRequeueRejected ||
                throwable instanceof MessageRejectedWhileStoppingException ||
                throwable instanceof ImmediateRequeueAmqpException;
        Throwable t = throwable;
        while (shouldRequeue && t != null) {
            if (t instanceof AmqpRejectAndDontRequeueException) {
                shouldRequeue = false;
            }
            t = t.getCause();
        }
        if (logger.isDebugEnabled()) {
            logger.debug("Rejecting messages (requeue=" + shouldRequeue + ")");
        }
        return shouldRequeue;
    }
```

PR #1054, opened 2019-07-23
Fix ImmediateRequeueAmqpException

garyrussell:
Fixes https://github.com/spring-projects/spring-amqp/issues/1053

`ContainerUtils.shouldRequeue` did not traverse the cause tree so we never
matched on this exception.

**cherry-pick to 2.1.x**


artembilan:
... and cherry-picked into 2.1.x

PR #1061, opened 2019-08-08
Convert more tests to JUnit5

garyrussell:
- some low-hanging fruit
- speed up admin tests with declaration failures with a `NoBackoffPolicy` or remove retry altogether
- speed up some tests that were waiting 2 seconds for no message expected


build.gradle
```diff
@@ -368,6 +368,7 @@ project('spring-rabbit-test') {
 
 		compile project(":spring-rabbit")
 		compile ("junit:junit:$junit4Version") {
+			optional(it)
 			exclude group: 'org.hamcrest', module: 'hamcrest-core'
 		}
 		compile "org.hamcrest:hamcrest-all:$hamcrestVersion"
```
artembilan: May we need to make this dependency optional as well?
The same might be applied for the `mockito-core` one, too...

build.gradle
```diff
@@ -345,18 +345,18 @@ project('spring-rabbit-junit') {
 	dependencies { // no spring-amqp dependencies allowed
 
 		compile "org.springframework:spring-core:$springVersion"
-		compile "junit:junit:$junit4Version"
+		compile ("junit:junit:$junit4Version", optional)
```
artembilan: I'm not sure if you have fixed all the tests into JUnit 5 style, but I guess we need to think eventually about removing `testCompile ("junit:junit:$junit4Version") {` to disallow to use it accidentally in tests

build.gradle
garyrussell: Well, perhaps - but the reason for making JUnit4 optional here is so they can use 5 without excluding 4. Let's think about it later.

build.gradle
garyrussell: Right - but we still have a long way to go before we can do that 😄

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/2c5d0374f81dec95e77b12df5c612d98acb88b9b

Issue #1062, opened 2019-08-09
Potential typo in SimpleMessageListenerContainer when logging exceptions?


Asapin:
Hello.
We recently experienced RabbitMQ outage and got the following logs from our RabbitMQ listener:

> ERROR 6 --- [ntContainer#3-5] o.s.a.r.l.SimpleMessageListenerContainer : Consumer received fatal=false exception on startup
>
> org.springframework.amqp.rabbit.listener.QueuesNotAvailableException: Cannot prepare queue for listener. Either the queue doesn't exist or the broker will not allow us to use it.
>
> ERROR 6 --- [ntContainer#3-5] o.s.a.r.l.SimpleMessageListenerContainer : Stopping container from aborted consumer

And while investigating issues with our RabbitMQ configuration, we got really confused why container was stopped if it says `fatal=false`. Further investigation lead us to this code in `org.springframework.amqp.rabbit.listener.SimpleMessageListenerContainer`:

```
catch (QueuesNotAvailableException ex) {
    logger.error("Consumer received fatal=" + isMismatchedQueuesFatal() + " exception on startup", ex);
    if (isMissingQueuesFatal()) {
        this.startupException = ex;
        // Fatal, but no point re-throwing, so just abort.
        aborted = true;
    }
    publishConsumerFailedEvent("Consumer queue(s) not available", aborted, ex);
}
```

For logging this code checks `isMismatchedQueuesFatal()` method, but for actual logic it uses `isMissingQueuesFatal()`.

Searching for other usages of `isMismatchedQueuesFatal()` shows that it's mainly used when configuring listener and declaring queues. The only other place where its usage is not configuration-related is the snippet above.

Is this intentional or just a typo?

garyrussell:
Yes; it's a typo; thanks for reporting - but it's also wrong to just log the value; the error should explicitly refer to the `missingQueuesFatal` property.

PR #1063, opened 2019-08-09
GH-1062: Fix missing queues log message

garyrussell:
Fixes https://github.com/spring-projects/spring-amqp/issues/1062

**cherry-pick to 2.1.x, 2.0.x**


artembilan:
... and cherry-picked to `2.1.x` and `2.0.x`. Although I think we need to start mentioning that `2.0.x` generation is already EOL...

garyrussell:
No; that's true for Kafka 2.0

https://github.com/spring-projects/spring-boot/blob/2.0.x/spring-boot-project/spring-boot-dependencies/pom.xml#L163

but 2.0 here has to live as long as boot 2.0

https://github.com/spring-projects/spring-boot/blob/2.0.x/spring-boot-project/spring-boot-dependencies/pom.xml#L157

Issue #1065, opened 2019-08-12
Async @RabbitListener does not handle MessageConverter exceptions


peterlgh7:
## Bug report

spring-amqp version: 2.1.8-RELEASE

I'm using `spring.rabbitmq.listener.simple.acknowledge-mode: MANUAL` as per the documentation:

> The listener container factory must be configured with AcknowledgeMode.MANUAL so that the consumer thread will not ack the message; instead, the asynchronous completion will ack or nack the message when the async operation completes.

Using the following simple example. If I publish an invalid JSON string, the message is never acked/nacked.

```java
import org.springframework.amqp.rabbit.annotation.RabbitListener;
import org.springframework.stereotype.Component;
import org.springframework.util.concurrent.ListenableFuture;
import org.springframework.util.concurrent.SettableListenableFuture;

@Component
public class Listener {

  @RabbitListener(queues = "example")
  public ListenableFuture<Void> receive(User user) {
    System.out.println(user);
    SettableListenableFuture<Void> future = new SettableListenableFuture<>();
    future.set(null);
    return future;
  }
}
```

After a bit of debugging I get to:
https://github.com/spring-projects/spring-amqp/blob/2bf1117a7b305d7a0dc9e18c8118cad9ae6ae43d/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapter.java#L130-L133

Where, in the case of an exception being thrown by the `toMessagingMessage`, the `invokeHandlerAndProcessResult`(which is the one to correctly handle acks/nacks for async listeners) never gets called. Afterwards, the message is simply not acked/nacked because the container's acknowledge mode is MANUAL.

garyrussell:
Hmmm - this is not just limited to async reply types - any manual-ack listener behaves this way.

We have an error handler (default) that treats message conversion (and some others) as fatal, but we only NACK if the ackmode is AUTO.

These exceptions must always be considered fatal because they occur before the listener ever gets them.

peterlgh7:
You are right that it happens also without for non-async listeners. My project is currently using manual acks and suffering from the same problem. I passes to async in part because I thought the problem may not exist(since async listeners actually handle acks for me).

garyrussell:
I find it rather intriguing that nobody else has reported this over the years; there are a lot of people using MANUAL acks, but I guess conversion exceptions must be rare.

[...]

PR #1066, opened 2019-08-12
GH-1065: Reject fatal with MANUAL ack mode

garyrussell:
Fixes https://github.com/spring-projects/spring-amqp/issues/1065

Previously, messages with fatal errors (e.g. conversion) would be
left in an un-acked state.

They are now rejected and NOT requeued by default.

**cherry-pick to 2.1.x**


artembilan:
... and cherry-picked as https://github.com/spring-projects/spring-amqp/commit/2d4c3c0373e9f94a9668cd8092da5e85753a29f5 after fixing conflicts.

PR #1069, opened 2019-08-14
Upgrade Jackson version ; use BOM for dependencies

garyrussell:


publish-maven.gradle
```diff
@@ -14,6 +14,11 @@ install {
 
 def customizePom(pom, gradleProject) {
 	pom.whenConfigured { generatedPom ->
+
+		generatedPom.dependencies = generatedPom.dependencies.findAll {
+			dep -> dep.artifactId != "jackson-bom"
```
artembilan: Do we still have a proper BOM import in the generated pom?

artembilan: So, is it expected to have it like this in the final pom:
```
  <dependency>
      <groupId>com.fasterxml.jackson.core</groupId>
      <artifactId>jackson-core</artifactId>
      <scope>compile</scope>
      <optional>true</optional>
    </dependency>
    <dependency>
      <groupId>com.fasterxml.jackson.core</groupId>
      <artifactId>jackson-databind</artifactId>
      <scope>compile</scope>
      <optional>true</optional>
    </dependency>
```
and don't have any `dependencyManagement` block in the pom?

I see SF uses `id "io.spring.dependency-management" version "1.0.7.RELEASE" apply false` instead of `platform` dependency.
Although I believe that we can live with built-in Gradle solution, only the problem is to fix a `publish-maven.gradle` script properly to resolve actual versions for those dependencies.
Unfortunately we just can't have a  `dependencyManagement` in the library's pom, I guess...

garyrussell: Hmmm - good catch - I assumed that would just happen automatically; we could make jackson the parent like I did i the SI samples that use jackson.

If you don't like that solution (I am not keen on it), then, yes, more work is needed. If you have any ideas how to do it, please take a look when you have time, my gradle/groovy skills are lacking.

artembilan: As a quick solution, I suggest to forget a standard Gradle approach and do whatever it is in SF with that `io.spring.dependency-management` plugin. Let's hope it still works well with the latest Gradle version!

build.gradle
```diff
@@ -15,6 +15,7 @@ plugins {
 	id 'org.asciidoctor.convert' version '1.6.1'
 	id 'org.ajoberstar.grgit' version '3.1.1'
 	id "io.spring.nohttp" version "0.0.3.RELEASE"
+	id "io.spring.dependency-management" version "1.0.7.RELEASE" apply false
```
artembilan: `1.0.8.RELEASE` is there already: https://plugins.gradle.org/plugin/io.spring.dependency-management

build.gradle
```diff
@@ -45,6 +66,21 @@ nohttp {
...
+	dependencyManagement {
+		resolutionStrategy {
+			cacheChangingModulesFor 0, 'seconds'
+		}
+		applyMavenExclusions = false
+		generatedPomCustomization {
+			enabled = false
+		}
+		imports {
+			mavenBom "com.fasterxml.jackson:jackson-bom:$jackson2Version"
```
artembilan: I wouldn't do this in all the modules of the project: only in those where we need such a dependency.
What makes you to move all the versions stuff to the top level?

garyrussell: 1. ok
2. because the plugin can't resolve the bom version - the Spring PR did the same.

build.gradle
artembilan: Gary, how about this one to address as well?

Thanks

Issue #1067, opened 2019-08-13
Add an enum to select the type of publisher confirms


garyrussell:
Simplify Boot auto-configuration.

`SIMPLE` and `CORRELATED`.

See https://github.com/spring-projects/spring-boot/pull/17848#discussion_r313381228

snicoll:
Thanks Gary. If none is a valid value, we probably need an option in the enum for that as well.

garyrussell:
Good point!

snicoll:
Can we please reopen this issue? I think deprecated methods that take a boolean are ignoring the boolean value and set the enum to enable the feature. If you call those methods with `false` they'll actually enable the behaviour.

garyrussell:
Doh

PR #1070, opened 2019-08-20
GH-1067: Fix deprecated setters

garyrussell:
Fixes https://github.com/spring-projects/spring-amqp/issues/1067

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
@@ -412,7 +412,7 @@ public class CachingConnectionFactory extends AbstractConnectionFactory
 			setPublisherConfirmType(ConfirmType.CORRELATED);
 		}
 		else if (this.confirmType.equals(ConfirmType.CORRELATED)) {
-			this.confirmType = ConfirmType.NONE;
+			setPublisherConfirmType(ConfirmType.NONE);
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
```diff
@@ -406,9 +406,11 @@ public void setPublisherReturns(boolean publisherReturns) {
 	 */
 	@Deprecated
 	public void setPublisherConfirms(boolean publisherConfirms) {
-		Assert.isTrue(!ConfirmType.SIMPLE.equals(this.confirmType),
+		Assert.isTrue(!publisherConfirms || !ConfirmType.SIMPLE.equals(this.confirmType),
 				"Cannot set both publisherConfirms and simplePublisherConfirms");
-		setPublisherConfirmType(ConfirmType.CORRELATED);
+		if (publisherConfirms) {
```
artembilan: Now let's imaging the situation that I call `setPublisherConfirms(true)` and have a `setPublisherConfirmType(ConfirmType.CORRELATED)` and then I call `setPublisherConfirms(false)`, but I don't have `setPublisherConfirmType(ConfirmType.NONE)`.

Is it possible?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
```diff
@@ -411,6 +411,9 @@ public void setPublisherConfirms(boolean publisherConfirms) {
 		if (publisherConfirms) {
 			setPublisherConfirmType(ConfirmType.CORRELATED);
 		}
+		else if (this.confirmType.equals(ConfirmType.CORRELATED)) {
+			this.confirmType = ConfirmType.NONE;
```
artembilan: I think we still need to call `setPublisherConfirmType()`, since we need to propagate a state into a `this.publisherConnectionFactory` as well...

garyrussell: Doh - good catch - not enough coffee 😦

Issue #1071, opened 2019-08-20
Decouple BrokerRunning from JUnit4


garyrussell:
`BrokerRunning` is used by the JUnit5 `BrokerRunningCondition`; decouple the code so that no JUnit 4 classes are required.

PR #1072, opened 2019-08-20
GH-1071: JUnit 5 Support Improvements

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1071

- Remove JUnit4 dependency from `RabbitAvailableCondition` (minor breaking API change)
- Add `purgeAfterEach` to `@RabbitAvailable`
- tabs not spaces in `RabbitAvailableCondition` (review with `?w=1`)
- `@LogLevels` now requires `level`

### DIFF
--- a/spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/BrokerRunningSupport.java
+++ b/spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/BrokerRunningSupport.java
@@ -96,9 +96,6 @@ public final class BrokerRunningSupport {
 
 	private final String[] queues;
 
-	private final int defaultPort = fromEnvironment(BROKER_PORT, null) == null ? BrokerTestUtils.getPort()
-			: Integer.valueOf(fromEnvironment(BROKER_PORT, null));
-
 	private int port;
 
 	private String hostName = fromEnvironment(BROKER_HOSTNAME, "localhost");
@@ -204,7 +201,9 @@ public final class BrokerRunningSupport {
 		}
 		this.purge = purge;
 		this.management = management;
-		setPort(this.defaultPort);
+		setPort(fromEnvironment(BROKER_PORT, null) == null
+				? BrokerTestUtils.getPort()
+				: Integer.valueOf(fromEnvironment(BROKER_PORT, null)));
 	}
 
 	private BrokerRunningSupport(boolean assumeOnline, String... queues) {
--- a/spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/BrokerRunningSupport.java
+++ b/spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/BrokerRunningSupport.java
@@ -17,7 +17,6 @@
 package org.springframework.amqp.rabbit.junit;
 
 import java.io.IOException;
-import java.net.MalformedURLException;
 import java.net.URISyntaxException;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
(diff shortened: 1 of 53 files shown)


spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/BrokerRunningSupport.java
```diff
@@ -0,0 +1,655 @@
...
+ * <p>If you wish to enforce the broker being available, for example, on a CI server,
+ * set the environment variable {@value #BROKER_REQUIRED} to {@code true} and the
+ * tests will fail fast.
+ *
+ * @author Dave Syer
+ * @author Gary Russell
+ *
+ * @since 1.7
+ * @see Assume
+ * @see org.junit.internal.AssumptionViolatedException
```
artembilan: Please, fix JavaDocs for this new class. It is really not Rule related any more. Plus these JUnit4 dependencies must go from this class.

spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/BrokerRunningSupport.java
```diff
@@ -0,0 +1,655 @@
...
+	public static final String BROKER_REQUIRED = "RABBITMQ_SERVER_REQUIRED";
+
+	public static final String DEFAULT_QUEUE_NAME = BrokerRunningSupport.class.getName();
+
+	private static final String GUEST = "guest";
+
+	private static final Log logger = LogFactory.getLog(BrokerRunningSupport.class); // NOSONAR - lower case
+
+	// Static so that we only test once on failure: speeds up test suite
+	private static final Map<Integer, Boolean> brokerOnline = new HashMap<Integer, Boolean>(); // NOSONAR - lower case
```
artembilan: I guess we can fix all these Sonar smells using upper case already. Plus we can use now diamonds.

spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/BrokerRunningSupport.java
```diff
@@ -0,0 +1,655 @@
...
+
+	private final boolean assumeOnline;
+
+	private final boolean purge;
+
+	private final boolean management;
+
+	private final String[] queues;
+
+	private final int defaultPort = fromEnvironment(BROKER_PORT, null) == null ? BrokerTestUtils.getPort()
```
artembilan: Should be `static`, as well as `fromEnvironment()` method.
Also my IDEA says that `parseInt()` must be used instead of `valueOf()`

spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/BrokerRunningSupport.java
```diff
@@ -0,0 +1,655 @@
...
+	 * @param hostName the hostName to set
+	 */
+	public void setHostName(String hostName) {
+		this.hostName = hostName;
+	}
+
+	/**
+	 * Set the user for the amqp connection default "guest".
+	 * @param user the user.
+	 * @since 1.7.2
```
artembilan: I would remove all these `@since`: the whole class is since `2.2`

spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/BrokerRunningSupport.java
```diff
@@ -0,0 +1,655 @@
...
+		}
+	}
+
+	private Connection getConnection(ConnectionFactory cf) throws IOException, TimeoutException {
+		Connection connection = cf.newConnection();
+		connection.setId(generateId());
+		return connection;
+	}
+
+	private Channel createQueues(Connection connection) throws IOException, MalformedURLException, URISyntaxException {
```
artembilan: A `MalformedURLException` is redundant.

spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/LogLevels.java
```diff
@@ -55,6 +55,6 @@
 	 * The Log4j level name to switch the categories to during the test.
 	 * @return the level.
 	 */
-	String level() default "";
+	String level();
```
artembilan: How about to have it as `debug` by default?

spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/RabbitAvailableCondition.java
```diff
@@ -35,100 +36,110 @@
...
+					throw new IllegalStateException("Required RabbitMQ is not available", e);
+				}
+				return ConditionEvaluationResult.disabled("RabbitMQ is not available");
+			}
+		}
+		return ENABLED;
+	}
+
+	@Override
+	public void afterEach(ExtensionContext context) throws Exception {
```
artembilan: `Exception` is redundant

spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/BrokerRunningSupport.java
garyrussell: We can't make this static because it will be initialized before `ENVIRONMENT_OVERRIDES` is set (but I agree about the method).

artembilan: OK. Then let's make it *local* as it us suggested by the IDE.

artembilan:
> convert more tests.

Does that mean that we are very close to remove hard dependency on JUnit 4 and only use it for rules in the `test` module?

garyrussell:
I wouldn't say "very close", but certainly "closer".

We have one nut to crack - `RepatableProcessor` `@Rule`.

In most cases, it should be easy to replace with `@RepeatableTest` but there is one test case where the `RepeatableProcessor` calls the test method on multiple threads.

spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/BrokerRunningSupport.java
artembilan: OK. I can merge it now if you are not going to do anything with this.

Thanks

garyrussell: I am not sure what you mean by "local".

artembilan: image

garyrussell: Yeah - I just fired up IDEA - fixing...

garyrussell:
I added a couple more conversions with the push; I don't plan on doing any more today, so this can be merged.

PR #1073, opened 2019-08-21
Convert repeated tests to JUnit 5 @RepeatableTest

garyrussell:
- add lifecycle to `@LogLevels` to avoid adjusting the log on each iteration
- with `@RepeatableTest` there is a template context between the class and method contexts

### DIFF
--- a/spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/LogLevels.java
+++ b/spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/LogLevels.java
@@ -22,6 +22,7 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
+import org.junit.jupiter.api.TestInstance;
 import org.junit.jupiter.api.extension.ExtendWith;
 
 /**
@@ -56,5 +57,12 @@ public @interface LogLevels {
 	 * @return the level (default DEBUG).
 	 */
 	String level() default "DEBUG";
+
+	/**
+	 * When used at the class level, set this to PER_CLASS to set/restore
+	 * the levels once only. Useful with {@code @RepeatedTest}.
+	 * @return the Lifecycle.
+	 */
+	TestInstance.Lifecycle lifecycle() default TestInstance.Lifecycle.PER_METHOD;
 
 }
--- a/spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/LogLevels.java
+++ b/spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/LogLevels.java
@@ -22,7 +22,6 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
-import org.junit.jupiter.api.TestInstance;
 import org.junit.jupiter.api.extension.ExtendWith;
 
 /**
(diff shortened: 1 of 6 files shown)


spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/LogLevels.java
```diff
@@ -57,4 +58,11 @@
 	 */
 	String level() default "DEBUG";
```
artembilan: You know, I've checked a `Level.toLevel()` and it defaults to `DEBUG`. For now... It might be changed eventually.
So, let's revert this change to the empty string and really rely on the default in the target Log4J impl!

spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/LogLevels.java
```diff
@@ -57,4 +58,11 @@
...
 
+	/**
+	 * When used at the class level, set this to PER_CLASS to set/restore
+	 * the levels once only. Useful with {@code @RepeatedTest}.
+	 * @return the Lifecycle.
+	 */
+	TestInstance.Lifecycle lifecycle() default TestInstance.Lifecycle.PER_METHOD;
```
artembilan: Can't this be per class by default? In most cases we really use it on the class level without considering it to be changed in between. Well, essentially it does just an idempotent work between methods anyway, so, I even would consider do not have this option at all. On class level it is for the whole class; on method level - just for method and has a precedence over the class one.

What am I missing with this option, please?

Thanks

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitTemplatePerformanceIntegrationTests.java
```diff
@@ -44,87 +49,78 @@
  * @since 1.0
  *
  */
+@RabbitAvailable(queues = RabbitTemplatePerformanceIntegrationTests.ROUTE)
+@LongRunning
```
artembilan: Is it still really *long-runing* ?

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/MessageListenerRecoveryRepeatIntegrationTests.java
```diff
@@ -57,13 +56,19 @@
  * @author Gary Russell
  *
  */
+@RabbitAvailable(queues = MessageListenerRecoveryRepeatIntegrationTests.TEST_QUEUE, purgeAfterEach = false)
+@LogLevels(level = "ERROR", lifecycle = Lifecycle.PER_CLASS, classes = { RabbitTemplate.class,
+		ConditionalRejectingErrorHandler.class,
+		SimpleMessageListenerContainer.class, BlockingQueueConsumer.class,
+		MessageListenerRecoveryRepeatIntegrationTests.class })
+@TestInstance(Lifecycle.PER_CLASS)
```
artembilan: Shouldn't we consider to have this option enabled globally for all the test classes?
I somehow believe that Spring Test Framework makes test instances really per class anyway...

spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/LogLevels.java
garyrussell: I kept it this way so we get the nice delimiter in the log between tests; but I have another idea. Stay tuned...

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitTemplatePerformanceIntegrationTests.java
garyrussell: 7 seconds total.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/MessageListenerRecoveryRepeatIntegrationTests.java
garyrussell: >I somehow believe

I didn't believe so (and I just tested it) - the context is cached but you get a new instance per test method.

We can try setting it globally, but I'd rather wait until all conversion is done before trying it.

artembilan: Yeah... Got it. Thank you!
Yep! Let's deffer it to the later phase!

Looking forward for your final solution for that `lifecycle` concern!

PR #1078, opened 2019-08-30
Fix tangle between listener and listener.adapter

garyrussell:
- Move `ContainerUtils` to `listener.support` (new package)
- Move `MessageRejectedWhileStoppingException` to `listener.exception`


garyrussell:
Before

Before

After

After

So now `listener` and `listener.adapter` depend on `listener.support`.

Issue #1080, opened 2019-09-04
'concurrentConsumers' cannot be more than 'maxConcurrentConsumers'


gaecfov:
the system throw IllegalArgumentException "'concurrentConsumers' cannot be more than 'maxConcurrentConsumers'" when i call the "setConcurrency" in "SimpleMessageListenerContainer".

bacause call the method "setConcurrentConsumers" before the method "setMaxConcurrentConsumers" in "setConcurrency"

for example:

setp 1:
set concurrentConsumer:1
set maxConcurrentConsumer:1

setp 2:
call setConcurrency("2-5")

artembilan:
OK. I think we need to change order of method:
```
if (separatorIndex != -1) {
	setConcurrentConsumers(Integer.parseInt(concurrency.substring(0, separatorIndex)));
	setMaxConcurrentConsumers(
			Integer.parseInt(concurrency.substring(separatorIndex + 1, concurrency.length())));
}
```

To stabilize the API.

gaecfov:
thanks for your response.

PR #1081, opened 2019-09-04
GH-1080: SMLC: Fix concurrency configuration order

artembilan:
Fixes https://github.com/spring-projects/spring-amqp/issues/1080

When we have a configuration like this:
```
container.setConcurrentConsumers(1);
container.setMaxConcurrentConsumers(1);
container.setConcurrency("2-5");
```
we fail with an assertion like
`'concurrentConsumers' cannot be more than 'maxConcurrentConsumers'`

* Change the order in the `SimpleMessageListenerContainer.setConcurrency()`
how we populate `maxConcurrentConsumers` and `concurrentConsumers`

**Cherry-pick to 2.1.x and 1.7.x**

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
@@ -214,11 +214,13 @@ public class SimpleMessageListenerContainer extends AbstractMessageListenerConta
 	 * @since 2.0
 	 */
 	public void setConcurrency(String concurrency) {
+		this.maxConcurrentConsumers = null;
+		this.concurrentConsumers = 1;
 		try {
 			int separatorIndex = concurrency.indexOf('-');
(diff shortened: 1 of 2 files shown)


artembilan:
Thanks, Gary, for feedback!

How about the change in the latest commit?

garyrussell:
No; that won't work; if I enter "10-5", it will fail validation after resetting the current values, which is not a desired outcome. I think we need to perform validation first.

garyrussell:
Unfortunately, this is still not enough - bypassing the setters means `adjustConsumers()` will not be called.

artembilan:
@garyrussell ,

anything else I'm missing here?
Having the fixed problem about POM, I'd like to go ahead with 2.1.10 release. 
Would be great to have this included as well.

Thanks

garyrussell:
LGTM now

artembilan:
... And cherry-picked to `2.1.x`.
The `setConcurrency()` is there `@since 2.0`, so, nothing to back-port into `1.7.x`

PR #1097, opened 2019-09-16
GH-1093: Remove redundant isInterface() call

garyrussell:
- redundant since interfaces are always abstract


artembilan:

PR #1100, opened 2019-09-19
Update setUsePublisherConnection javadocs

kjastrzebski:
Docs in setUsePublisherConnection say this defaults to true since version 2.1 what, in my opinion, is not true

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
@@ -245,7 +245,7 @@ public class RabbitTemplate extends RabbitAccessor // NOSONAR type line count
 
 	private boolean userCorrelationId;
 
-	private boolean usePublisherConnection = true;
+	private boolean usePublisherConnection;
 
 	private boolean noLocalReplyConsumer;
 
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@kjastrzebski Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-amqp&pullRequestId=1100)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-amqp&pullRequestId=1100) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@kjastrzebski Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-amqp&pullRequestId=1100)!

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -245,7 +245,7 @@
 
 	private boolean userCorrelationId;
 
-	private boolean usePublisherConnection;
+	private boolean usePublisherConnection = true;
```
garyrussell: We need to change the javadoc instead. It turned out that `true` is not a safe default and will break many existing applications. 

The problem is that any auto-delete, exclusive `Queue` beans in the application context will be declared on the producer connection and any listener containers that consume from those queues will fail to start because they can only be accessed on the connection on which they were declared.

kjastrzebski:
Ok, so let's change the javadocs as users may be confused by them

PR #1102, opened 2019-09-23
GH-1101: Message Listener Micrometer Timers

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1101


garyrussell:
BTW, I originally had it working as an advice - but it was a bit involved and I wasn't happy with the implementation; this ended up simpler. 

I can push that branch if you want to look at it.

artembilan:
Well, I'm fine with simpler solution 😄 

Thanks

Issue #1085, opened 2019-09-10
Consider to add spring-rabbit-junit as a dependency for the spring-rabbit-test to let end-users to poll only the last and benefitve from both


artembilan:
Even if there is no hard dependencies between `spring-rabbit-test` and `spring-rabbit-junit`, it is much obvious that end-user is going to poll the first one as dependency for his/her project.
At the same time there are a lot of useful tools in the `spring-rabbit-junit` for testing, but it could be missed because of artifact name.
So, for better end-user experience we should consider to include `spring-rabbit-junit` as a compile dependency for the `spring-rabbit-test`. 

This way it is going to bring the same impression as `spring-integration-test` or `spring-kafka-test`, or just very same `spring-test`.

Issue #1084, opened 2019-09-08
Consider to move What's New chapter to the top.


artembilan:
Right now it is hidden under `Introduction`:

![изображение](https://user-images.githubusercontent.com/1071590/64491685-586cb980-d239-11e9-9813-b6b297082b28.png)

This is different from what we have in Spring Kafka or Spring Integration:

![изображение](https://user-images.githubusercontent.com/1071590/64491705-894cee80-d239-11e9-9791-e2e5e0d4c138.png)

![изображение](https://user-images.githubusercontent.com/1071590/64491750-de890000-d239-11e9-82fb-14240cd7f7f9.png)

PR #1104, opened 2019-10-01
GH-1085: spring-rabbit-junit -> spring-rabbit-test

artembilan:
Fixes: https://github.com/spring-projects/spring-amqp/issues/1085
Fixes: https://github.com/spring-projects/spring-amqp/issues/1084

* Make a `spring-rabbit-junit` as compile dependency for the
`spring-rabbit-test` for better end-user experience
* Also move `What's New` chapter to the top level as it is with other
projects

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -383,13 +383,11 @@ project('spring-rabbit-test') {
 	dependencies {
 
 		compile project(':spring-rabbit')
+		compile project(':spring-rabbit-junit')
 		compile "org.hamcrest:hamcrest-library:$hamcrestVersion"
 		compile "org.hamcrest:hamcrest-core:$hamcrestVersion"
-		compile ("org.mockito:mockito-core:$mockitoVersion") {
-			exclude group: 'org.hamcrest', module: 'hamcrest-core'
-		}
+		compile "org.mockito:mockito-core:$mockitoVersion"
 		testCompile project(':spring-rabbit').sourceSets.test.output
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -383,13 +383,11 @@ project('spring-rabbit-test') {
...
+		compile project(':spring-rabbit-junit')
 		compile "org.hamcrest:hamcrest-library:$hamcrestVersion"
 		compile "org.hamcrest:hamcrest-core:$hamcrestVersion"
-		compile ("org.mockito:mockito-core:$mockitoVersion") {
-			exclude group: 'org.hamcrest', module: 'hamcrest-core'
-		}
+		compile "org.mockito:mockito-core:$mockitoVersion"
 		testCompile project(':spring-rabbit').sourceSets.test.output
-		testCompile project(':spring-rabbit-junit')
 	}
```
garyrussell: We should add something to the What's New - people who are using TestNG will now have to exclude it.

artembilan: Well, not sure how does that affect.
A `junit` and `junit-jupiter-api` are *optional* in the `spring-rabbit-junit`, but we indeed can add a note in What's New.

garyrussell: Oh, of course; never mind.

garyrussell:
Merged as 34e423f8eeae06a57390306dfe52cc0912599542 with minor polishing.

Issue #891, opened 2019-01-26
SpringMultirabbit for auto-configuration of multiple ConnectionFactories


rwanderc:
## Enhancement

Recently I created a [blog post](https://inside.mytaxi.com/springboot-with-multiple-rabbitmq-brokers-cec203c3f77) about **SpringMultirabbit**. This is a solution to provide multiple ConnectionFactories, auto-configured, extending `RabbitProperties`, so as to have a simpler way to introduce different brokers in applications (e.g.: one for configuration hot-reloading and another for business events). This feature is available at http://github.com/mytaxi/spring-multirabbit

Out of the comments, @garyrussell suggested to contribute this work to the framework.
I have already followed the instructions at [Contributor Guidelines](https://github.com/spring-projects/spring-amqp/blob/master/CONTRIBUTING.adoc). However, couldn't find how to create the issue in Jira [AMQP project](https://jira.spring.io/secure/RapidBoard.jspa?rapidView=3&projectKey=AMQP&view=planning.nodetail), thus, I'm creating here in GitHub.

Before submitting the PRs, I wanted to know if there are any further instructions or concerns.

garyrussell:
@rwanderc Sorry about that - we recently switched from JIRA to GitHub issues; there's a note on the JIRA home page, but we haven't updated the contributor guidelines yet.

My only thoughts/concerns is that some of the classes in the boot autoconfiguration part might belong here instead (`ConnectionFactoryContextWrapper` might be an example), so that non-boot users can use this functionality.

Obviously, the boot-specific parts will need to be a separate contribution over there, once this is merged. I haven't talked to the boot team yet, but I don't anticipate any problems with them accepting a PR.

I like the idea of auto-determining the admins but we might be able to make it simpler than using naming conventions, perhaps by adding a getter on the admin so we can match admin and container factory connection factories. That functionality can be merged into the main BPP rather than subclassing it.

Anyway, welcome; we look forward to working with you.

rwanderc:
PR created https://github.com/spring-projects/spring-amqp/pull/1111

garyrussell:
Nice PR# 😄

PR #1111, opened 2019-10-25
GH-891 Introduced MultiRabbit to handle multiple brokers

rwanderc:
Hi @garyrussell and Spring team, it took a long time but finally happened. =)

I have tried to make it simpler avoiding name conventions, but didn't manage to follow your advice. Maybe you could also give me some direction on what to change and how to really test it since the instances are provided in the `boot`.

I created the test class `MultiRabbitListenerAnnotationBeanPostProcessorCompatibilityTests` as a copy of `RabbitListenerAnnotationBeanPostProcessorCompatibilityTests` to make sure nothing is broken from the default behavior if eventually the multirabbit BPP is loaded.

Resolves https://github.com/spring-projects/spring-amqp/issues/891


garyrussell:
@rwanderc Thanks!

Will take a look ASAP.

garyrussell:
It will probably be next week before I can look at this in detail.

We will need an update to the reference manual (src/reference/asciidoc) before we merge, but it can wait until after the review if you prefer.

garyrussell:
Instead of duplicating that test, I think you can simply subclass it and subclass `Config` overloading the `postProcessor` bean.

rwanderc:
Thank you so much @garyrussell for your comments.
I will probably start working on that next week. Will also try to come with the documentation so you can share your thoughts on that.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitBootstrapConfiguration.java
```diff
@@ -0,0 +1,49 @@
...
+import org.springframework.amqp.rabbit.listener.RabbitListenerEndpointRegistry;
+import org.springframework.beans.factory.support.BeanDefinitionRegistry;
+import org.springframework.beans.factory.support.RootBeanDefinition;
+import org.springframework.context.annotation.ImportBeanDefinitionRegistrar;
+import org.springframework.core.type.AnnotationMetadata;
+
+/**
+ * An {@link ImportBeanDefinitionRegistrar} class that registers a
+ * {@link MultiRabbitListenerAnnotationBeanPostProcessor} (overwriting the regular
+ * {@link RabbitListenerAnnotationBeanPostProcessor}) bean capable of processing Spring's @{@link RabbitListener}
```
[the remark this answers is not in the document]
rwanderc: Done.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitConstants.java
```diff
@@ -0,0 +1,36 @@
...
+
+	public static final String CONNECTION_FACTORY_BEAN_NAME = "multiRabbitConnectionFactory";
+	public static final String CONNECTION_FACTORY_CREATOR_BEAN_NAME = "rabbitConnectionFactoryCreator";
+    static final String DEFAULT_RABBIT_ADMIN_BEAN_NAME = "defaultRabbitAdmin";
+    static final String RABBIT_ADMIN_SUFFIX = "-admin";
+
+
+    private MultiRabbitConstants() {
+		throw new UnsupportedOperationException("Construction not supported.");
+	}
```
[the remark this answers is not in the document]
rwanderc: Fixed.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,114 @@
...
+ * to the beans of Exchanges, Queues, and Bindings after they are created.
+ * <p>
+ * This processing restricts the {@link RabbitAdmin} according to the related configuration, preventing the server
+ * from automatic binding non-related structures.
+ *
+ * @author Wander Costa
+ * @see RabbitListenerAnnotationBeanPostProcessor
+ */
+public class MultiRabbitListenerAnnotationBeanPostProcessor extends RabbitListenerAnnotationBeanPostProcessor
+		implements ApplicationContextAware, BeanFactoryAware {
```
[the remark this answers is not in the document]
rwanderc: Removed.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,114 @@
...
+
+	private ApplicationContext applicationContext;
+	private BeanFactory beanFactory;
```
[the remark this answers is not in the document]
rwanderc: Done.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,114 @@
...
+
+	/**
+	 * Enhance beans with related RabbitAdmin, so as to be filtered when being processed by the RabbitAdmin.
+	 *
+	 * @param rabbitListener the RabbitListener to enhance its bean.
+	 */
+	private void enhanceBeansWithReferenceToRabbitAdmin(RabbitListener rabbitListener) {
+		RabbitAdmin rabbitAdmin = getRabbitAdminBean(rabbitListener);
+		// Enhance Exchanges
+		this.applicationContext.getBeansOfType(AbstractExchange.class).values().stream().filter(this::isNotProcessed)
```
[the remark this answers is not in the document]
rwanderc: Changed to `getBeansOfType(clazz, false, false)`.
Thanks for the advice.

About plugging in BPP, the reason is that BPP runs for each `@RabbitListener` and creates the beans of the respective declarables (`AbstractExchange`, `Queue` and `Binding`) during this process.
Therefore, plugging to BPP makes it possible to associate the proper `RabbitAdmin` to the declarables just created. After they are enhanced with the `RabbitAdmin`, they are filtered out from next `@RabbitListener` processed (that's the reason of method `isNotProcessed()`).
Indeed the complexity of each of these statements is very bad (`O(n2)` for `n` `@RabbitListeners`), however I could not find a better solution to automate this process without the user's input (e.g. defining the `RabbitAdmin`'s bean name in each declarable).

As far as I understand, implementing `SmartInitializingSingleton` will provide a processor after all beans are instantiated, which is much better in terms of complexity. But how could I know which declarable relates to each `RabbitAdmin` to make the association?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,114 @@
...
+				.forEach(exchange -> exchange.setAdminsThatShouldDeclare(rabbitAdmin));
```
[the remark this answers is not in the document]
rwanderc: No. The declarables (`AbstractExchange`, `Queue` and `Binding`) are not declared on all servers.

That's the reason for the `isNotProcessed()` method.
Once the declarable beans are created by BPP for one `@RabbitListener`, they will not have any `RabbitAdmin` yet associated.
Then, they are enhanced with the corresponding `RabbitAdmin`.

Then, when BPP processes the 2nd, 3rd, etc `@RabbitListener`s, the previous declarables will be filtered out and not enhanced, because they already have a `RabbitAdmin` associated to them.

This way, only the last declarables created are associated to the current `RabbitAdmin`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,114 @@
...
+
+	/**
+	 * Verifies the presence of an instance of RabbitAdmin or this object, as fallback.
+	 *
+	 * @param declarable the declarable to be verified.
+	 * @return true is the declarable was not processed.
+	 */
+	private boolean isNotProcessed(Declarable declarable) {
+		return declarable.getDeclaringAdmins() == null || (
+				declarable.getDeclaringAdmins().stream().noneMatch(item -> item == this) && declarable
```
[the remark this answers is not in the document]
rwanderc: Indeed. Fixed line break.
The `item == this` was a mistake. Thanks for the heads-up. Removed.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	@Override
+	public void setBeanFactory(@NonNull BeanFactory beanFactory) {
+		this.beanFactory = beanFactory;
+		super.setBeanFactory(beanFactory);
+	}
+
+	@Override
+	public void setApplicationContext(@NonNull ApplicationContext applicationContext) {
```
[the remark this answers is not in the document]
rwanderc: Done.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/MultiRabbitAdminNameResolverTest.java
```diff
@@ -0,0 +1,76 @@
...
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+/**
+ * @author Wander Costa
+ */
+class MultiRabbitAdminNameResolverTest {
```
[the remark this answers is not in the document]
rwanderc: Changed.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitBootstrapConfiguration.java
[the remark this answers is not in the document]
artembilan: @rwanderc,

Please, stop sending `Done` and `Fixed` message.
We definitely get a notification about new commits in the PR.
Me personally follow the rule: if I don't agree or have some doubts, I comment on the review. Otherwise I just push the fix silently.

It really has just sounded as a spam on my phone 😄 

Thanks for understanding

rwanderc:
Sorry spamming you @garyrussell and @artembilan previously ...

About the tests, I followed the advice on subclassing and providing an overloading, but it didn't replace the bean. I guess it's because the context is initialized always with a specific class `Config.class`. I worked around it by changing the current test to provide the configuration class as a parameter.

About the integration tests, I created a couple of tests to make sure the declarables are declared by the proper `RabbitAdmin`, which is pretty much the main concern on the amqp side of the feature. I didn't find a meaningful way to test with the connection factories. Would you have a suggestion?
In the other hand, I also thought that would introduce a lot of complexity to test with `@RabbitAvailable` since it would require at least 2 servers running. Therefore, I didn't provide such test.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitListenerAnnotationBeanPostProcessor.java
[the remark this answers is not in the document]
garyrussell: Oh; now I see the reason for my confusion; this is a bit brittle - you are using the fact that the BPP "just" added the beans  for the current `@RabbitListener.queueBindings()` and the beans for other `@QueueBinding`s don't exist yet.

The problem with this approach (using `getBeansOfType()`) is that all stand-alone queue/exchange/binding `@Bean`s will be declared by the first admin.

We could use the names from the `@QueueBinding`s and do `getBean()` for each individually to avoid picking up other unrelated beans.

However, this is complicated by the fact that the bean names are `foo.n` (where n is incremented with a common integer) so there is no way to know which beans were registered for the current annotation.

I can see a couple of possible solutions - 

- change `processAmqpListener()` to return `Collection<Declarable>` containing any beans that were registered for this listener in the super class
- add a new `@RabbitListener` field to `AbstractDeclarable` and use that as the `.filter`.

rwanderc:
Hi @garyrussell. As you recommended, I implemented the approach of returning the collection of `Declarables` in the BPP. Also, I created some integration tests that make sure the proper CF is touched as you advised.
I'm still lacking behind with the documentation, which I hope to work on soon.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * An extension of {@link RabbitListenerAnnotationBeanPostProcessor} that associates the
+ * proper RabbitAdmin to the beans of Exchanges, Queues, and Bindings after they are
+ * created.
+ * <p>
+ * This processing restricts the {@link RabbitAdmin} according to the related
+ * configuration, preventing the server from automatic binding non-related structures.
+ *
+ * @author Wander Costa
+ * @see RabbitListenerAnnotationBeanPostProcessor
```
artembilan: There is no reason in `@see`. You mention it in the JavaDocs and we have an extension from it anyway.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,72 @@
...
+		this.beanFactory = beanFactory;
+		super.setBeanFactory(beanFactory);
+	}
+
+	@Override
+	protected Collection<Declarable> processAmqpListener(RabbitListener rabbitListener, Method method,
+			Object bean, String beanName) {
+		final Collection<Declarable> declarables = super.processAmqpListener(rabbitListener, method, bean, beanName);
+		final RabbitAdmin rabbitAdmin = resolveRabbitAdminBean(rabbitListener);
+		declarables.stream()
```
artembilan: I would prefer regular loop with ifs instead of `Stream` objects.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/MultiRabbitIntegrationTests.java
```diff
@@ -0,0 +1,332 @@
...
+		}
+	}
+
+	@Configuration
+	@PropertySource("classpath:/org/springframework/amqp/rabbit/annotation/queue-annotation.properties")
+	static class MultiConfig {
+
+		static final SimpleRoutingConnectionFactory ROUTING_CONNECTION_FACTORY = new SimpleRoutingConnectionFactory();
+		static final ConnectionFactory DEFAULT_CONNECTION_FACTORY = Mockito.mock(ConnectionFactory.class);
+		static final ConnectionFactory CONNECTION_FACTORY_BROKER_B = Mockito.spy(ConnectionFactory.class);
```
artembilan: Typo? There is nothing to `spy` on the interface.

Also: doesn't look like an `integration` test if you used mock.
Maybe better to call it `MockMultiRabbitTests`?

rwanderc:
Hi @garyrussell and @artembilan. First of all, thank you for the previous review you've done and comments. I worked on them as advised.
I think the implementation in `spring-amqp` is probably done (missing docs, as I mention below). However, I have some follow-up questions:
- A big concern about this PR is the change in the BPP so as to retrieve the `Declarable`s returned from the processing. As @artembilan mentioned, this is possibly a breaking change (https://github.com/spring-projects/spring-amqp/pull/1111#pullrequestreview-346107472). How should I proceed? Should I write any sort of documentation about it (e.g. https://docs.spring.io/spring-amqp/reference/html/#_breaking_api_changes)?
- Most of the implementation of Spring MultiRabbit will be in the `spring-boot` project. What sort of documentation do you recommend writing in `spring-amqp` then? I thought about introducing in `What's new` something with the title`Support for multiple connection factories` as well in the `Reference` a topic about it. Could you pls give some direction?
- There are many commits. Before finishing, I suppose I'd rather squash them. We might come here later, of course.

garyrussell:
Thanks @rwanderc - we will try to review later this week. We have a big release day tomorrow (this will go into 2.3 later this year; there is no new Spring AMQP for this week's Boot 2.3).

We'll build Milestone 1 soon after this is merged.

Don't worry about squashing - we always do that before merging to master.

Discrete commits makes it easier to review deltas during the review phase.

garyrussell:
Finally getting around to looking at this; sorry for the delay...

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitBootstrapConfiguration.java
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright 2002-2020 the original author or authors.
```
garyrussell, line 2: Just 2020 for new files.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,99 @@
+/*
+ * Copyright 2002-2020 the original author or authors.
```
garyrussell, line 2: Just 2020.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,99 @@
...
+		super.setBeanFactory(beanFactory);
+	}
+
+	@Override
+	protected Collection<Declarable> processAmqpListener(RabbitListener rabbitListener, Method method,
+			Object bean, String beanName) {
+		final Collection<Declarable> declarables = super.processAmqpListener(rabbitListener, method, bean, beanName);
+		final RabbitAdmin rabbitAdmin = resolveRabbitAdminBean(rabbitListener);
+		for (final Declarable declarable : declarables) {
+			if (declarable.getDeclaringAdmins().isEmpty() && declarable instanceof AbstractDeclarable) {
```
garyrussell, line 63: I think we can add `setAdminsThatShouldDeclare` as a `default` method on `Declarable` and avoid the cast.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,99 @@
...
+
+	/**
+	 * Returns the RabbitAdmin bean of the requested name or the default one.
+	 *
+	 * @param rabbitListener the RabbitListener to retrieve its bean.
+	 * @return the bean found.
+	 */
+	private RabbitAdmin resolveRabbitAdminBean(RabbitListener rabbitListener) {
+		final String name = resolveAdminName(rabbitListener);
+		return this.beanFactory.getBean(name, RabbitAdmin.class);
```
garyrussell, line 79: We shouldn't call `getBean()` from a BPP - causes early instantiation.

That's why `setAdminsThatShouldDeclare` takes `Object...` - so we can just pass in the bean name.

Although I see that `super.resolveAdmin()` does that too; I need to fix that.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	/**
+	 * Resolves the name of the RabbitAdmin bean based on the RabbitListener.
+	 *
+	 * @param rabbitListener The RabbitListener to process the name from.
+	 * @return The name of the RabbitAdmin bean.
+	 */
+	public static String resolveAdminName(RabbitListener rabbitListener) {
+		String admin = rabbitListener.admin();
```
garyrussell, line 89: Need to call `resolveExpressionAsString()` - see `super.resolveAdmin()` - we should make all those `resolve*` methods protected.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/ConnectionFactoryContextWrapper.java
```diff
@@ -0,0 +1,116 @@
+/*
+ * Copyright 2002-2019 the original author or authors.
```
garyrussell, line 2: Just 2020

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/MockMultiRabbitTests.java
```diff
@@ -0,0 +1,332 @@
+/*
+ * Copyright 2002-2020 the original author or authors.
```
garyrussell, line 2: 2020

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/MultiRabbitBootstrapConfigurationTests.java
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright 2002-2019 the original author or authors.
```
garyrussell, line 2: 2020

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/MultiRabbitListenerAnnotationBeanPostProcessorCompatibilityTests.java
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright 2002-2019 the original author or authors.
```
garyrussell, line 2: 2020

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/ConnectionFactoryContextWrapperTests.java
```diff
@@ -0,0 +1,75 @@
+/*
+ * Copyright 2002-2020 the original author or authors.
```
garyrussell, line 2: 2020

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitBootstrapConfiguration.java
```diff
@@ -0,0 +1,49 @@
...
+ * <p>
+ * This configuration must be imported manually.
+ *
+ * @author Wander Costa
+ * @see MultiRabbitListenerAnnotationBeanPostProcessor
+ * @see RabbitListenerAnnotationBeanPostProcessor
+ * @see RabbitListenerEndpointRegistry
+ * @see EnableRabbit
+ */
+public class MultiRabbitBootstrapConfiguration implements ImportBeanDefinitionRegistrar {
```
garyrussell, line 40: Don't we need an annotation or `DeferredImportSelector` to trigger this to ensure it's loaded before the `RabbitBootstrapConfiguration`?

cc/ @artembilan

garyrussell: Can't the user just declare a bean with the subclass and `RABBIT_LISTENER_ANNOTATION_PROCESSOR_BEAN_NAME` since the `RabbitListenerConfigurationSelector` won't run until all `@Configuration` classes have been processed; removing the need for this class altogether.

artembilan: Well, need to understand how it is tested over here though.
Probably just `EnableRabbit(multi = true)` would be enough...

rwanderc:
> Looks good! Just some minor comments.
> 
> @rwanderc - I have scheduled the first milestone of 2.3 next Wednesday; it would be great to get this into that release (even without documentation).
> 
> If you don't think you can address these concerns by then, I can address them.
> 
> Thanks

@garyrussell I will look into the comments tomorrow, or later on Sunday. If none of them require other rounds, my guess is that we get this into 2.3. :)

Thanks

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitListenerAnnotationBeanPostProcessor.java
rwanderc: I changed according to the suggestion.
Also, I suppose you also mean removing the condition `if (... && declarable instanceof AbstractDeclarable) {...}` to rely directly on the interface.

If that's the case, having a `default` method on `Declarable` might introduce space for a behavioral side-effect in MultiRabbit (declarables being processed by all admins, which I guess is not the main use-case). At `spring-amqp`, I see no impact, since all concrete classes eventually also extend `AbstractDeclarable`.

My thinking is that since these are [breaking changes](https://github.com/spring-projects/spring-amqp/pull/1111#pullrequestreview-346107472), why not introduce  `setAdminsThatShouldDeclare()` as a regular signature (non-`default`) and force concrete classes to implement it?
`Declarable`'s concretes must also implement `getDeclaringAdmins()` which makes me understand it's in the scope of its responsibilities.

What do you think about it?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitListenerAnnotationBeanPostProcessor.java
rwanderc: That's a big point! I'm changing it to provide only the name, and not prematurely instantiate the bean.

Also, regarding the logic here (and [here](https://github.com/spring-projects/spring-amqp/pull/1111#discussion_r442488232)), as I understand, it's intended to provide a fallback `RabbitAdmin` name.

At this moment, `super.processAmqpListener()` was already executed and resolved the name provided (if any) through the regular flow. Then, it's only important to assign a fallback bean name (the `RabbitAdmin` generated from MultiRabbit) if none was specified.

Therefore, I guess `resolveExpressionAsString()` wouldn't be necessary. Is that right?
Nevertheless, I provided it as suggested.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitBootstrapConfiguration.java
rwanderc: I'm not too sure how to proceed here. No changes.

rwanderc:
@garyrussell & @artembilan, just (rebased and) pushed the changes for your recent comments.
I have a couple of follow-up questions, and couldn't understand how to proceed [here](https://github.com/spring-projects/spring-amqp/pull/1111#discussion_r442494318).

Pls let me know what you find, and I will of course jump in.
But to avoid another round and to be able to add it to the [release](https://github.com/spring-projects/spring-amqp/pull/1111#pullrequestreview-433622715) (if you see no blockers), you can also make any changes you find necessary.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitBootstrapConfiguration.java
garyrussell: I think we can just delete this; in order to override the default BPP, simply define the subclass as a bean with the default name, and the default one will be skipped here https://github.com/spring-projects/spring-amqp/blob/715f39ff3baab340e5d681d74697140c88a7874e/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitBootstrapConfiguration.java#L51-L56

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitListenerAnnotationBeanPostProcessor.java
garyrussell: Yes; that would be ok (non-default) - I doubt users implement their own `Declarable` s anyway.

KamilWitkowski7:
> Great work; thanks for the contribution.
> 
> Next up: docs (if you have time - otherwise I'll take care of it).

Where can I find docs?

garyrussell:
@KamilWitkowski7 Thanks for the ping, I had forgotten about this; and timely too, we are releasing 2.3.0 tomorrow.

Pull request here: https://github.com/spring-projects/spring-amqp/pull/1262

@rwanderc I would appreciate if you can give it a quick review, in case I missed anything.

Thanks.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitBootstrapConfiguration.java
rwanderc: Hi @garyrussell & @artembilan,
I'm writing this here to connect to the context of the issue.

I'm working on the boot part, and I tried to implement the instantiation of MultRabbitBPP the way suggested. But as overriding of beans is now disabled by default in Spring, it fails to register a bean with the same name as `RabbitListenerConfigUtils.RABBIT_LISTENER_ANNOTATION_PROCESSOR_BEAN_NAME`.

I'm trying to work around it by providing a MultiRabbit BPP bean with another name, via configuration. But when processing listeners, it still picks up the original one BPP and ALSO tries to process MultiRabbitBPP, leading to an error of multiple registrations of the same declarables.

Maybe it should never instantiate `RabbitListenerAnnotationBeanPostProcessor` since `MultiRabbitListenerAnnotationBeanPostProcessor` is provided, but I'm not figuring out how to make it work, or if that's related to some order of processing.

I got to make this work by directly implementing `RabbitBootstrapConfiguration` to instantiate `new RootBeanDefinition(MultiRabbitListenerAnnotationBeanPostProcessor.class)` whenever `EnableRabbit(multi = true)` (as per @artembilan's [suggestion](https://github.com/spring-projects/spring-amqp/pull/1111#discussion_r442497506)), but I feel this is not the best solution (very intrusive and not SOLID).

Would you have any suggestions on how to proceed with this?

garyrussell: @rwanderc I am not sure what you mean by SOLID but another, perhaps less intrusive, option would be to add a new Boot property `multiCluster`. While we can't reference the `RabbitProperties` we can get a reference to the environment:

```java
public class RabbitBootstrapConfiguration implements ImportBeanDefinitionRegistrar, EnvironmentAware {

	private Environment environment;

	@Override
	public void registerBeanDefinitions(@Nullable AnnotationMetadata importingClassMetadata,
			BeanDefinitionRegistry registry) {

		String multiS =  this.environment.getProperty("spring.rabbitmq.multi-cluster");
		boolean multi = multiS != null && Boolean.parseBoolean(multiS);
		if (multi) {
			...
		}
		else if (!registry.containsBeanDefinition(
				RabbitListenerConfigUtils.RABBIT_LISTENER_ANNOTATION_PROCESSOR_BEAN_NAME)) {

			registry.registerBeanDefinition(RabbitListenerConfigUtils.RABBIT_LISTENER_ANNOTATION_PROCESSOR_BEAN_NAME,
					new RootBeanDefinition(RabbitListenerAnnotationBeanPostProcessor.class));
		}

		if (!registry.containsBeanDefinition(RabbitListenerConfigUtils.RABBIT_LISTENER_ENDPOINT_REGISTRY_BEAN_NAME)) {
			registry.registerBeanDefinition(RabbitListenerConfigUtils.RABBIT_LISTENER_ENDPOINT_REGISTRY_BEAN_NAME,
					new RootBeanDefinition(RabbitListenerEndpointRegistry.class));
		}
	}

	@Override
	public void setEnvironment(Environment environment) {
		this.environment = environment;
	}

}
```

Issue #1113, opened 2019-10-28
java.io.IOException: Unknown consumerTag


bojanv55:
Not sure if this is related to core rabbit java lib or spring, but this is the issue I'm having:

```
java.io.IOException: Unknown consumerTag
at com.rabbitmq.client.impl.ChannelN.basicCancel(ChannelN.java:1476)
at org.springframework.amqp.rabbit.connection.PublisherCallbackChannelImpl.basicCancel(PublisherCallbackChannelImpl.java:617)
at sun.reflect.GeneratedMethodAccessor98.invoke(Unknown Source)
at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
at java.lang.reflect.Method.invoke(Method.java:498)
at org.springframework.amqp.rabbit.connection.CachingConnectionFactory$CachedChannelInvocationHandler.invoke(CachingConnectionFactory.java:1140)
at com.sun.proxy.$Proxy187.basicCancel(Unknown Source)
at org.springframework.amqp.rabbit.listener.BlockingQueueConsumer$InternalConsumer.handleDelivery(BlockingQueueConsumer.java:914)
at com.rabbitmq.client.impl.ConsumerDispatcher$5.run(ConsumerDispatcher.java:149)
at com.rabbitmq.client.impl.ConsumerWorkService$WorkPoolRunnable.run(ConsumerWorkService.java:104)
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
at java.lang.Thread.run(Thread.java:748)","exception_class":"java.io.IOException","exception_message":"Unknown consumerTag"
```

Received this exception while I was using rabbit 2.0.10, so I tried upgrading to 2.1.10 (just to see if it helps) - but seems that issue is the same. From time to time this thing pops up and I need to restart the service.

This is just graphical view of stacktrace with 2 different versions of rabbit lib.
![image](https://user-images.githubusercontent.com/1422698/67667563-6227b900-f96e-11e9-9975-6065a54ea092.png)

Do you maybe know what could be the problem here?

artembilan:
I think it would be great to explain what the whole situation is around to let us to determine the cause.
Also some simple project with instructions how to reproduce would be really helpful.

[...]

PR #1114, opened 2019-10-28
GH-1113: Clean up cals to basicCancel

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1113

Use `RabbitUtils` for all `basicCancel` calls, with try/catch for
all exceptions.

**cherry-pick to 2.1.x**

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitUtils.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitUtils.java
@@ -81,7 +81,7 @@ public abstract class RabbitUtils {
 	 */
 	public static final int CHANNEL_PROTOCOL_CLASS_ID_20 = 20;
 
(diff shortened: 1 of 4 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitUtils.java
```diff
@@ -182,18 +182,18 @@ public static void closeMessageConsumer(Channel channel, Collection<String> cons
...
-		}
 		catch (AlreadyClosedException e) {
 			if (logger.isTraceEnabled()) {
-				logger.trace(channel + " is already closed");
+				logger.trace(channel + " is already closed", e);
+			}
+		}
+		catch (Exception e) {
+			if (logger.isDebugEnabled()) {
+				logger.debug("Error performing 'basicCancel'", e);
```
artembilan: We don't need a wrapping `if` - it is redundant by the reason of a constant for a message.

artembilan:

PR #1117, opened 2019-11-06
Support quorum queues in QueueBuilder

garyrussell:


artembilan:
We fail on CI server somehow: https://build.spring.io/browse/AMQP-MASTER-1215/

Sorry that I have merged this too early

garyrussell:
Interesting - I tested it against a 3.7 broker .

garyrussell:
The 'quorum_queue' feature flag is disabled, class-id=50, method-id=10)

Issue #1125, opened 2019-11-24
Add DeclarableCustomizer or similar


garyrussell:
Add some mechanism to allow the admin the pre-process `Declarable` objects (e.g. add arguments) before declaring.

Supports https://github.com/spring-cloud/spring-cloud-stream-binder-rabbit/issues/274

PR #1126, opened 2019-11-25
GH-1125: Add DeclarableCustomizer

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1125

- pull up the common `arguments` field to `AbstractDeclarable`
- add `x-single-active-consumer` as a first class `QueueBuilder` argumemt

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/core/AbstractDeclarable.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/core/AbstractDeclarable.java
@@ -20,7 +20,10 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
 
+import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 
 /**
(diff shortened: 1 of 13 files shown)


spring-amqp/src/main/java/org/springframework/amqp/core/Declarable.java
```diff
@@ -50,4 +52,24 @@
...
+	}
+
+	/**
+	 * Remove an argument from the declarable.
+	 * @param name the argument name.
+	 * @return the argument value or null if not present.
+	 * @since 2.2.2
+	 */
+	default @Nullable Object removeArgument(String name) {
+		return false;
```
artembilan: `false`? Why not `null`?

spring-amqp/src/main/java/org/springframework/amqp/core/DeclarableCustomizer.java
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Beans of this type are invoked by the {@link AmqpAdmin} before declaring the
+ * {@link Declarable}, allowing customization thereof.
+ *
+ * @author Gary Russell
+ * @since 2.2.2
+ *
+ */
+@FunctionalInterface
+public interface DeclarableCustomizer extends Function<Declarable, Declarable> {
```
artembilan: Why `function`, but not `consumer`?
Are you intended to replace property with result of this customizer?

spring-amqp/src/main/java/org/springframework/amqp/core/AbstractDeclarable.java
```diff
@@ -38,6 +41,22 @@
...
 	private boolean ignoreDeclarationExceptions;
 
+	private final Map<String, Object> arguments;
+
+	/**
+	 * Construct an instance with the supplied arguments, or an empty map if null.
+	 * @param arguments the arguments.
+	 * @since 2.2.2
+	 */
+	public AbstractDeclarable(@Nullable Map<String, Object> arguments) {
```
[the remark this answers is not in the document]
garyrussell: I seriously doubt there are custom subclasses but, yes, I'll add back the default ctor.

spring-amqp/src/main/java/org/springframework/amqp/core/Declarable.java
garyrussell: Oops

spring-amqp/src/main/java/org/springframework/amqp/core/DeclarableCustomizer.java
garyrussell: Yes; they can replace the object if they need to for some reason (aside from the arguments, they are mostly immutable).

Issue #1130, opened 2019-12-06
Missing maxFrameHeadroom check on cause.getMessage()


thpeng:
https://github.com/spring-projects/spring-amqp/blob/c57fd6434ffbba00a9f14afb4ce25135330e8437/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/retry/RepublishMessageRecoverer.java#L155

The RepublishMessageRecoverer truncates the stacktrace if its length exceeds the maxFrameHeadroom configured. This behavior should also apply with cause.getMessage() to allow republishing a failed message with a long message.

garyrussell:
Interesting; it's rather unusual to have a large message but I agree.

I wonder, though, which is more important - keeping as much message as possible or as much stack trace? i.e. what to do if both are too big.

thpeng:
I would suggest prioritizing the stacktrace. The cause was calling toString on a big object graph and inserting it in the cause message, which is not optimal.

PR #1131, opened 2019-12-06
GH-1130: Repub Recoverer include ex. message size

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1130

`RepublishingMessageRecoverer` - include the exception message length in the
truncation algorithm.

Note that the message is included in the stack trace.

If the stack trace + exception message would exceed the limit
- first truncate the message within the stack trace to 100 bytes
-- if the stack trace and original message still exceed the limit
-- also truncate the `X_EXCEPTION_MESSAGE` header to 100 bytes and
   use the remaing space for stack trace

If, after truncating the message in the stack trace, there is room remaining
the full stack trace as well as the truncated message, re-truncate the
`X_EXCEPTION_MESSAGE` header to use the remaining available bytes.

examples:
  message 150 bytes, stack trace 350 bytes, available 300 bytes,
  stack trace after message truncation 300 bytes
   - truncate message to 100, trace to 200

  message 200 bytes, stack trace 250 bytes, available 300 bytes,
  stack trace after message truncation 150 bytes
   - truncate message to 100 bytes, trace to 200

  message 200 bytes, stack trace 250 bytes, available 300 bytes,
  stack trace after message truncation 150 bytes
   - truncate message to 150 bytes, trace remains at 150

These are for illustration only, the available bytes is generally must larger.

**cherry-pick to 2.1.x**

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/retry/RepublishMessageRecoverer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/retry/RepublishMessageRecoverer.java
@@ -49,6 +49,8 @@ import org.springframework.util.Assert;
  */
 public class RepublishMessageRecoverer implements MessageRecoverer {
 
+	private static final int ELIPSIS_LENGTH = 3;
+
 	public static final String X_EXCEPTION_STACKTRACE = "x-exception-stacktrace";
 
 	public static final String X_EXCEPTION_MESSAGE = "x-exception-message";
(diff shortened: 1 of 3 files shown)


artembilan:
Cherry-picked to `2.1.x` as https://github.com/spring-projects/spring-amqp/commit/5cadcc3f79852ddd7e527258d1eec361cca006ec after fixing conflicts in the `RepublishMessageRecovererIntegrationTests`.

Issue #1136, opened 2019-12-31
Support Meta Annotations


garyrussell:
See https://stackoverflow.com/questions/59546604/spring-aliasfor-is-not-working-with-annotation-rabbitlistener/59547017#59547017

We currently use `AnnotationUtils.find...` instead of `MergedAnnotations.from()` which will property populate the aliased attributes.

garyrussell:
Work around (for a single container)

```java
	@Bean
	public ApplicationRunner runner(RabbitListenerEndpointRegistry registry) {
		return args -> {
			Method method = So59546604Application.class.getDeclaredMethod("listen");
			MergedAnnotation<RabbitListener> mergedAnnotation = MergedAnnotations.from(method)
					.get(RabbitListener.class);
			AnnotationAttributes attrs = mergedAnnotation.asAnnotationAttributes();
			String[] queues = (String[]) attrs.get("queues");
			AbstractMessageListenerContainer container = (AbstractMessageListenerContainer) registry
					.getListenerContainers().iterator().next();
			container.addQueueNames(queues);
		};
	}
```

PR #1137, opened 2019-12-31
GH-1136: Support Meta Annotations with AliasFor

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1136

Note that this is a minimal implementation that avoids major refactoring
of the BPP.

Repeatable user meta-annotatiions are not supported.

In a future release we should perform that refactoring as well as supporting
`@Repeatable` user annotations.

Also restore the `stop()` with finally to fix `EnableRabbitTests.testProperShutdownOnException()`
which took 30 seconds to run after the finally was removed by
c1e31798961262795c4aa2b4bab3f0f67fedda26


garyrussell:
Do not merge.

PR #1144, opened 2020-01-16
Use of PluginBuilderFactory on AmqpAppender (Log4j2)

### LINKED ISSUE #1142: PluginBuilderFactory instead of PluginFactory for Log4j2 Appender
**Affects Version(s):** v2.2.2.RELEASE

---

## Use of PluginBuilderFactory instead of PluginFactory for Log4j2 AmqpAppender 

I'm tryin to extends **AmqpAppender** to customize method **postProcessMessageBeforeSend**.
I'm tryin to avoid **copy & paste** of static method **createAppender** (annotated with **PluginFactory**) but it's not possibile.

Is there another solution or simple code to create a method annotated with **PluginFactory** ?

I suggest to use **Builder** and Annotation **PluginBuilderFactory**

[Extend Log4j2 Appenders](https://logging.apache.org/log4j/log4j-2.2/manual/extending.html#Appenders)
[Log4j2 Plugin Builder](https://logging.apache.org/log4j/2.x/manual/extending.html#Plugin_Builders)

Scip88:
resolves #1142 

Refactor "@ PluginFactory" and use of "@ PluginBuilderFactory"

Use of PluginBuilderFactory simplifies subclasses' implementations of AmqpAppender (Log4j2)


pivotal-issuemaster:
@Scip88 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-amqp&pullRequestId=1144)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-amqp&pullRequestId=1144) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@Scip88 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-amqp&pullRequestId=1144)!

artembilan:
@Scip88 ,

Thank you for contribution!

> Use of PluginBuilderFactory simplifies subclasses' implementations of AmqpAppender

Would you mind to demonstrate such a feature in the tests? For example, `AmqpAppenderTests` is aimed for an `AmqpAppender` coverage. 

Also, it sounds like Docs (`logging.adoc`) deserves some attention with this change...

Scip88:
@artembilan,
no new test are needed, this is a refactor and tests from the same `AmqpAppenderTest` class from package `org.springframework.amqp.rabbit.log4j2` are sufficient.

As indicated in [Plugin Builder](https://logging.apache.org/log4j/2.x/manual/extending.html#Plugin_Builders), the *Log4j2* core initializes an _Appender_ (as other log4j plugin) via two possible static methods: `@PluginBuilderFactory` or `@PluginFactory`

I'm working on logging.adoc

Scip88:
I've update `logging.adoc`, please take a look

artembilan:
> no new test are needed

Then why do we do this change at all?
Seems for me you have claimed that with a `@PluginBuilderFactory` it is going to be easier to extend an `AmqpAppender` functionality.
So, confirm with a test, please, that we are good to go with the fix.

Thanks

src/reference/asciidoc/logging.adoc
```diff
@@ -270,6 +270,42 @@ public class MyEnhancedAppender extends AmqpAppender {
...
 
+Starting with <BACKLOG VERSION>, the log4j2 `AmqpAppender` can be extended using `@PluginBuilderFactory` and extending also
+`AmqpAppender.Builder`
+
+
+.log4j2
+====
+[source, java]
+----
+public class MyEnhancedAppender extends AmqpAppender {
```
artembilan: I think such a sample would be great candidate to have in our tests. 
Also consider to show how to use it in the config - I believe the test is not going to work without a proper logging config.

src/reference/asciidoc/logging.adoc
```diff
@@ -270,6 +270,42 @@ public class MyEnhancedAppender extends AmqpAppender {
 ----
 ====
 
+Starting with <BACKLOG VERSION>, the log4j2 `AmqpAppender` can be extended using `@PluginBuilderFactory` and extending also
```
artembilan: There is no such a `<BACKLOG VERSION>` placeholder. Please, consider to use an explicit current version.
You can find it in the `gradle.properties`.

Also, please, follow the rule *One sentence per line*: https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

src/reference/asciidoc/logging.adoc
```diff
@@ -270,6 +270,42 @@ public class MyEnhancedAppender extends AmqpAppender {
...
+`AmqpAppender.Builder`
+
+
+.log4j2
+====
+[source, java]
+----
+public class MyEnhancedAppender extends AmqpAppender {
+
+    @Override
```
artembilan: It's not clear what kind of indents are here, but please, be sure that whitespaces (not tabs) are used in the docs for indents.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
```diff
@@ -144,7 +146,7 @@ public AmqpAppender(String name, Filter filter, Layout<? extends Serializable> l
 		this.events = eventQueue;
 	}
 
-	@PluginFactory
```
artembilan: So, if we are going to keep this annotation, a conflict will happen and Log4J won't know which option to choose for configuration?

I don't mind to remove it since a builder approach is fully compatible with Log4J inversion of control; just want some clarification for myself.

Thanks

src/reference/asciidoc/logging.adoc
Scip88: already fixed

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
Scip88: "When plugins are being constructed after a configuration has been parsed, a plugin builder will be used if available, otherwise a plugin factory method will be used as a fallback. If a plugin contains neither factory, then it cannot be used from a configuration file (it can still be used programmatically of course)." 

I will restore it

src/reference/asciidoc/logging.adoc
Scip88: OK, i will check my IDE

src/reference/asciidoc/logging.adoc
Scip88: It will be fixed

src/reference/asciidoc/logging.adoc
Scip88: I have done it in the next commit

artembilan: There is no need in making excuses - just push a commit.
Please, consider to add comments into PR only if you are not agree with me or I really ask about some clarification.
If we are on the same page, you just push commit, I get a (single!) notification and we are good to go.
Right now I have at about 10 emails from you and only with an info that you are going to fix it.

Now imaging that I have like at about 10 PRs in parallel and everybody is going to ask excuse for each my comment...
I feel so counterproductive! 😄 

Thank you for understanding!

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/log4j2/ExtendAmqpAppenderTest.java
```diff
@@ -0,0 +1,252 @@
+/*
+ * Copyright 2016-2019 the original author or authors.
```
artembilan: Same here.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/log4j2/ExtendAmqpAppenderTest.java
```diff
@@ -0,0 +1,252 @@
...
+import org.springframework.core.io.ClassPathResource;
+import org.springframework.test.util.ReflectionTestUtils;
+
+/**
+ * @author Francesco Scipioni
+ *
+ * @since 2.2.4
+ */
+@RabbitAvailable
+public class ExtendAmqpAppenderTest {
```
artembilan: We tend to use only plural for suffix (`*Tests`), even if you have only one test method inside.

src/reference/asciidoc/logging.adoc
artembilan: I'm not sure why you have resolved this conversation, but I also asked about an XML config sample.
And I with the test I understand what to do - the `@Plugin` is the trick! But you don't show it in this sample for the doc.
Or do you mean mean that `@Plugin` is optional and the one on the super class has an effect?

Scip88: Annotation `@Plugin`  will not be inherited and is not optional. Added into the doc

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/log4j2/ExtendAmqpAppenderTests.java
```diff
@@ -0,0 +1,252 @@
...
+	public static void teardown() {
+		LOGGER_CONTEXT.setConfigLocation(ORIGINAL_LOGGER_CONFIG);
+		LOGGER_CONTEXT.reconfigure();
+		RabbitAvailableCondition.getBrokerRunning().deleteQueues("log4jTest", "log4j2Test");
+		RabbitAvailableCondition.getBrokerRunning().deleteExchanges("log4j2Test", "log4j2Test_uri");
+	}
+
+	@Test
+	public void test() {
+		CachingConnectionFactory ccf = new CachingConnectionFactory("localhost");
```
artembilan: Would you mind to `destroy()` this factory in the end of test to avoid resources leaking?

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/log4j2/ExtendAmqpAppenderTests.java
```diff
@@ -0,0 +1,252 @@
...
+		RabbitTemplate template = new RabbitTemplate(ccf);
+		RabbitAdmin admin = new RabbitAdmin(ccf);
+		FanoutExchange fanout = new FanoutExchange("log4j2Test");
+		admin.declareExchange(fanout);
+		Queue queue = new Queue("log4j2Test");
+		admin.declareQueue(queue);
+		admin.declareBinding(BindingBuilder.bind(queue).to(fanout));
```
artembilan: It would be great to make all these RabbitMQ objects as `autoDelete`, so they won't stay on the broker after finishing these tests.
Or you can remove them manually using the same `RabbitAdmin`!

Scip88: Annotated method `teardown()` (`@AfterAll`) should do that

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/log4j2/ExtendAmqpAppenderTests.java
Scip88: Should I also fix this in `spring-rabbit/src/test/java/org/springframework/amqp/rabbit/log4j2/AmqpAppenderTests.java` ?

artembilan: Yes, please. I believe not closed `ConnectionFactory` leads to resource leaking.

Thanks

garyrussell:
@Scip88 Thanks for the contribution! Looking forward to more.

@artembilan Thanks for the review.

Issue #1148, opened 2020-01-27
Defer initialization of RabbitTemplate until the LogAdapter.Log4jLog is initialized


garyrussell:
See https://github.com/spring-projects/spring-amqp-samples/issues/59

garyrussell:
We will be able to revert this change with a change to Spring Framework. See https://github.com/spring-projects/spring-framework/issues/24440

PR #1149, opened 2020-01-27
GH-1148: Log4jAppender - fix initialization cycle

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1148

Defer creation of all Spring static loggers from appender until the
appender is initialized.

The first logger (e.g. Spring Boot) initializes the static `LoggerContext`
field in `LogAdapter.Log4jLog`; this, in turn initializes the appender.

However, the appender references Spring objects that have static loggers.
This causes the `Log4jLog` CTOR to be called before the field is initialized,
causing an NPE.

Defer creation of all static loggers until the appender is fully initialized.

Tested with the log4j2 sample app.

Also add a test for async sends.

**cherry-pick to 2.1.x**

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
@@ -362,7 +362,7 @@ public class AmqpAppender extends AbstractAppender {
 				message = new Message(msgBody.toString().getBytes(), amqpProps); //NOSONAR (default charset)
 			}
 			message = postProcessMessageBeforeSend(message, event);
-			this.rabbitTemplate.send(this.manager.exchangeName, routingKey, message);
+			this.rabbitTemplate.send(this.manager.exchangeName, routingKey, message); // NOSONAR (sync)
 		}
 		catch (AmqpException e) {
 			int retries = event.incrementRetries();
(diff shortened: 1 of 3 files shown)


artembilan:
Please, consider to re-push with the commit message about `Fixes/Resolves https://github.com/spring-projects/spring-amqp/issues/1148` as well.

Thanks

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/log4j2/AmqpAppender.java
```diff
@@ -126,7 +126,7 @@
 	/**
 	 * The template.
 	 */
-	private final RabbitTemplate rabbitTemplate = new RabbitTemplate();
+	private RabbitTemplate rabbitTemplate;
```
artembilan: Has to be `volatile` then?

garyrussell: Actually, no, because the `synchronized` block will invalidate the cache.

I re-pushed with the proper `resolves`; another change is needed for the sample when this is merged.

garyrussell: Sonar will bark about inconsistent sync, so I added a NOSONAR there.

artembilan: Sorry, I don't see `NOSONAR` in your changes

garyrussell: oops (was still local).

artembilan:
@garyrussell ,

It doesn't cherry-pick properly, would you mind to take a look?

Thanks

garyrussell:
Cherry-picked to 2.1.x as 028550cf7fcf6eacf3063b746414151c1ccb2117

Issue #1152, opened 2020-01-31
Incorrect content encoding in Message's toString


shark300:
Version Spring AMQP 2.2.2.RELEASE

I would like to log incoming JSON type message but Message's toString() uses default charset. It causes an encoding issue in the log messages in a dockerized environment. 

In my humble opinion it should use the contentEncoding of the message in this (and similar) cases.

To reproduce: https://github.com/shark300/amqpMessageToStringBug

garyrussell:
I agree it should be changed, but why can't you do your own `byte[] -> String` when logging?

shark300:
Hi @garyrussell, yes, I did it, but to log the whole message is the most elegant way instead of logging messageProperties and payload separately.

```java
@RabbitListener(queues = INCOMING_QUEUE)
public void receive(Message message, @Payload Content content) {
  logger.info("Received message {}", message);
  messageProcessorStuff.process(content);
}
```

PR #1153, opened 2020-01-31
GH-1152: Use contentEncoding in Message.toString()

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1152

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/core/Message.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/core/Message.java
@@ -115,12 +115,13 @@ public class Message implements Serializable {
 			return null;
 		}
 		try {
-			String contentType = (this.messageProperties != null) ? this.messageProperties.getContentType() : null;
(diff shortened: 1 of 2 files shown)


spring-amqp/src/main/java/org/springframework/amqp/core/Message.java
```diff
@@ -107,11 +120,15 @@ private String getBodyContentAsString() {
 				return SerializationUtils.deserialize(new ByteArrayInputStream(this.body), whiteListPatterns,
 						ClassUtils.getDefaultClassLoader()).toString();
 			}
+			String encoding = this.messageProperties.getContentEncoding();
```
shark300: As I see, messageProperties could be null here.

garyrussell: Probably not in practice, but good catch; thanks.

shark300:
Guys, thank you for your quick help.

PR #1158, opened 2020-02-06
GH-1156: Fixes javadoc, overridable logging for fatal exceptions

### LINKED ISSUE #1156: Javadoc for AbstractMessageListenerContainer::setErrorHandler does not reflect current implementation
According to the line below, the default is to have no error handler and log the exception at `error` level: https://github.com/spring-projects/spring-amqp/blob/3793fc1d612cccf1d2f77a14dd6cd26c960b6294/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java#L452

however, the current default now is to have a `ConditionalRejectingErrorHandler` (which logs at `warning` level, making it also misleading in that way).

See https://github.com/spring-projects/spring-amqp/blob/3793fc1d612cccf1d2f77a14dd6cd26c960b6294/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java#L190

Ideally, such logging level could be made customizable, as logging as an `error` there is very interesting for a number of applications (though this is just an idea).

I didn't send this as a PR right away because I wanted to check first if that's really the case.

andrebrait:
Fixes #1156

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/ConditionalRejectingErrorHandler.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/ConditionalRejectingErrorHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2019 the original author or authors.
+ * Copyright 2014-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/ConditionalRejectingErrorHandler.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/ConditionalRejectingErrorHandler.java
@@ -167,12 +167,7 @@ public class ConditionalRejectingErrorHandler implements ErrorHandler {
 				cause = cause.getCause();
 			}
 			if (t instanceof ListenerExecutionFailedException && isCauseFatal(cause)) {
-				if (this.logger.isWarnEnabled()) {
-					this.logger.warn(
-							"Fatal message conversion error; message rejected; "
-									+ "it will be dropped or routed to a dead letter exchange, if so configured: "
-									+ ((ListenerExecutionFailedException) t).getFailedMessage());
-				}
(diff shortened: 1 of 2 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/ConditionalRejectingErrorHandler.java
```diff
@@ -187,6 +182,22 @@ private boolean isCauseFatal(Throwable cause) {
...
 		}
 
+		/**
+		 * Log the fatal ListenerExecutionFailedException at WARN level, excluding stack trace.
+		 * Subclasses can override this behavior.
+		 * @param t the {@link ListenerExecutionFailedException}.
+		 * @param cause the cause
+		 * @since 2.2.4
+		 */
+		protected void logFatalException(ListenerExecutionFailedException t, Throwable cause) {
```
andrebrait: I added the `cause` parameter here so the subclass can algo use the (already properly determined) cause if they so desire.

Not sure about the method name here, though I think this one is good enough.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/ConditionalRejectingErrorHandler.java
```diff
@@ -187,6 +182,22 @@ private boolean isCauseFatal(Throwable cause) {
 					|| isUserCauseFatal(cause);
 		}
 
+		/**
+		 * Log the fatal ListenerExecutionFailedException at WARN level, excluding stack trace.
+		 * Subclasses can override this behavior.
+		 * @param t the {@link ListenerExecutionFailedException}.
+		 * @param cause the cause
+		 * @since 2.2.4
```
andrebrait: I hope this version is correct. Let me know if it's not.

garyrussell:
@andrebrait Thanks!

Merged as 5e9ec034f4dd744140a165772352d038d88ff241 with minor polishing.

In future, please note that `./gradlew clean check` will automatically update the copyright of any changed (unstaged) files.

Also, we limit javadoc line lengths to 90 chars (code 120).

Thanks again; looking forward to more contributions.

PR #1165, opened 2020-02-16
Update Message.java 

weihongbin1:
Modify getBodyContentAsString method and make method getBodyContent public

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/core/Message.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/core/Message.java
@@ -111,43 +111,42 @@ public class Message implements Serializable {
 	}
 
 	private String getBodyContentAsString() {
-		Object content = getBodyContent();
-		if (content == null) {
-			return null;
-		}
-		if (content instanceof String) {
-			return (String)content;
-		}else if(content == body){
-			return this.body.toString() + "(byte[" + this.body.length + "])";
-		}else {
-			return content.toString();
-		}
-	}
-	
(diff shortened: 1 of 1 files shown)


weihongbin1:
> Can you explain the reason for this change?
> 
> The proper way to access the message body is using a `MessageConverter` (e.g. the `SimpleMessageConverter` for java `Serialized` objects).
> 
> Also, `public` methods require javadocs.
The getBodyContentAsString method of Message class itself is to convert the object to the corresponding object first and then toString. Why not divide it into two methods? And public converts the object method, so that you can get the object directly in the Message class.

weihongbin1:
> Can you explain the reason for this change?
> 
> The proper way to access the message body is using a `MessageConverter` (e.g. the `SimpleMessageConverter` for java `Serialized` objects).
> 
> Also, `public` methods require javadocs.




The getBodyContentAsString method of Message class itself is to convert the object to the corresponding object first and then toString. Why not divide it into two methods? And public converts the object method, so that you can get the object directly in the Message class.

artembilan:
> And public converts the object method, so that you can get the object directly in the Message class.

To be honest that's not a message object responsibility. It just has to contain a body as it is provided for a message and some headers to let us know what is the content about.

I would even think now that it was a mistake to do that conversion in a `toString()` implementation at all: if you want to have some meaningful string representation of the message content, do that yourself externally. 
The `toString()` should do just whatever is present in the message. Not all `byte[]` can be converted to string so easy as we have in this solution.

So, 👎 from me for this PR.

PR #1170, opened 2020-02-21
Default to TLS 1.2

carlhoerberg:
TLS 1.1 is deprecated but most software today

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
@@ -93,7 +93,7 @@ public class RabbitConnectionFactoryBean extends AbstractFactoryBean<ConnectionF
 
 	private static final String TRUST_STORE_TYPE = "trustStore.type";
 
-	private static final String TLS_V1_1 = "TLSv1.1";
+	private static final String TLS_V1_2 = "TLSv1.2";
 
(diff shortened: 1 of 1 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
```diff
@@ -125,7 +125,7 @@
 
 	private String trustStoreType;
 
-	private String sslAlgorithm = TLS_V1_1;
+	private String sslAlgorithm = TLS_V1_2;
```
garyrussell: We should simply remove the default; and remove the `sslAlgorithmSet` boolean; then use `useSSlProtocol()` if `this.sslAgorithm` is not null.

https://github.com/spring-projects/spring-amqp/blob/f7b7bedefc3de6149d5d1222c3c30d6eac40d6e2/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java#L732-L737

However, unfortunately, we can't merge this until we switch the master branch to 2.3.

garyrussell:
Actually, no, this change is not really needed because the default here is never used. The client picks the default. We should clean it up though to remove the confusion.

garyrussell:
After further review, this is indeed required because this default is used in some cases.

Sorry for the noise and confusion.

garyrussell:
@carlhoerberg Thanks for the contribution!

Replacing this with https://github.com/spring-projects/spring-amqp/pull/1171 which provides a more consistent solution and adds a test.

Issue #1169, opened 2020-02-21
Enable TLS v.1.2.


johanrhodin:
No reason to force users to use v.1.1 anymore: https://github.com/spring-projects/spring-amqp/blob/f7b7bedefc3de6149d5d1222c3c30d6eac40d6e2/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java#L128

garyrussell:
?

You can already set your own version.

https://github.com/spring-projects/spring-amqp/blob/f7b7bedefc3de6149d5d1222c3c30d6eac40d6e2/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java#L184-L191

If you mean change the default, we can do that in a future release.

garyrussell:
We can simply remove the default and let the amqp-client select it's own preferred default.

https://github.com/spring-projects/spring-amqp/blob/f7b7bedefc3de6149d5d1222c3c30d6eac40d6e2/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java#L732-L737

johanrhodin:
Letting the client select sounds good 👍 
I assume spring-amqp is by default using a newer than https://github.com/rabbitmq/rabbitmq-java-client/pull/145 client?

garyrussell:
That is correct; but that logic is already in place due to the `sslAlgorithmSet` being true only if the user actually calls the setter - the default on the field in this class is never used.

johanrhodin:
Thank you very much! 🙏

PR #1171, opened 2020-02-24
Always default to TLSv1.2

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1169

Previously, the default was TLSv1.2 if the context was created by the
amqp-client, but 1.1 if the context was created by the factory bean.


artembilan:
@garyrussell ,

does this fix issue https://github.com/spring-projects/spring-amqp/issues/1169 ?

garyrussell:
Yes, sorry - re-pushed.

Issue #1175, opened 2020-03-11
Consider adding @SpringRabbitTest 


garyrussell:
Configure common infrastructure for integration tests that are not using Spring Boot.

https://pivotal.slack.com/archives/C055EM4JJ/p1583961606004700?thread_ts=1583958288.001700&cid=C055EM4JJ

PR #1176, opened 2020-03-12
GH-1175: Add @SpringRabbitTest

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1175

- provision boilerplate infrastructure


spring-rabbit-test/src/main/java/org/springframework/amqp/rabbit/test/junit/SpringRabbitTest.java
```diff
@@ -0,0 +1,135 @@
...
+
+/**
+ * Adds infrastructure beans to a Spring test context; do not use with Spring Boot since
+ * it has its own auto configuration mechanism.
+ *
+ * @author Gary Russell
+ * @since 2.3
+ *
+ */
+@SpringJUnitConfig
```
artembilan: `@SpringIntegrationTest` doesn't do this because you can't guarantee that end-user always going to use JUnit 5.
Also I'm not sure why have you made it based on the JUnit 5 `Condition`, nut not `ContextCustomizer` and `ContextCustomizerFactory` from Spring Test...

garyrussell: Ah, right; makes sense.

I was thinking that it's not unreasonable to require JUnit5 starting with this version. But I will rework it.

garyrussell:
Reworked as `ContextCustomizer`.

spring-rabbit-test/src/main/java/org/springframework/amqp/rabbit/test/context/SpringRabbitContextCustomizer.java
```diff
@@ -0,0 +1,82 @@
...
+import org.springframework.util.Assert;
+
+/**
+ * Adds infrastructure beans to the Spring test context.
+ *
+ * @author Gary Russell
+ * @since 2.3
+ *
+ */
+class SpringRabbitContextCustomizer implements ContextCustomizer {
```
artembilan: Shouldn't it be SpringRabbit**Test**ContextCustomizer ?

src/reference/asciidoc/testing.adoc
```diff
@@ -13,6 +13,56 @@ Spring AMQP version 1.6 introduced the `spring-rabbit-test` jar, which provides
...
+
+        ...
+
+	}
+
+}
+----
+====
+
+With JUnit4, replace `@SpringJunitConfig` with `@RunWith(SpringRunnner.class)` and `@ContextConfiguration`.
```
artembilan: The `@ContextConfiguration` is not a requirement since some Spring version.
Now we need it only for fine-grained context configuration: for standard one the `SpringRunnner` is enough.

So, I mean let's remove the `@ContextConfiguration` note!

spring-rabbit-test/src/main/java/org/springframework/amqp/rabbit/test/context/SpringRabbitContextCustomizer.java
garyrussell: Why? It's already in a test jar and package.

I will change it if you insist.

garyrussell: ...and `ContextCustomizer` has no `Test`.

artembilan: Ah! OK. I can live with that. 😄

src/reference/asciidoc/testing.adoc
garyrussell: How soon we forget 😄

PR #1177, opened 2020-03-18
Rename schema; remove duplicates from zip

garyrussell:


spring-rabbit/src/main/resources/META-INF/spring.schemas
```diff
@@ -1,4 +1,4 @@
-http\://www.springframework.org/schema/rabbit/spring-rabbit-2.2.xsd=org/springframework/amqp/rabbit/config/spring-rabbit-2.2.xsd
-http\://www.springframework.org/schema/rabbit/spring-rabbit.xsd=org/springframework/amqp/rabbit/config/spring-rabbit-2.2.xsd
-https\://www.springframework.org/schema/rabbit/spring-rabbit-2.2.xsd=org/springframework/amqp/rabbit/config/spring-rabbit-2.2.xsd
-https\://www.springframework.org/schema/rabbit/spring-rabbit.xsd=org/springframework/amqp/rabbit/config/spring-rabbit-2.2.xsd
+http\://www.springframework.org/schema/rabbit/spring-rabbit-2.3.xsd=org/springframework/amqp/rabbit/config/spring-rabbit-2.3.xsd
+http\://www.springframework.org/schema/rabbit/spring-rabbit.xsd=org/springframework/amqp/rabbit/config/spring-rabbit-2.3.xsd
+https\://www.springframework.org/schema/rabbit/spring-rabbit-2.3.xsd=org/springframework/amqp/rabbit/config/spring-rabbit-2.3.xsd
+https\://www.springframework.org/schema/rabbit/spring-rabbit.xsd=org/springframework/amqp/rabbit/config/spring-rabbit-2.3.xsd
```
artembilan: Same here, please.
Also see how we have extra entries in similar file in SI
I mean we have to map all the version to the one XSD we have in classpath.
I mean the version should not matter any more: we always rely on whatever we have in classpath.

So, we also don't need a `checkTestConfigs` Gradle task any more, too.

Issue #1181, opened 2020-04-08
Memory Leak in RabbitTemplate when using custom correlationId


harald-raffelt:
**Affects Version(s):** \2.2.3.RELEASE
Since https://jira.spring.io/browse/AMQP-779
---

When using custom correlationId there is a memory leak in the following line:

https://github.com/spring-projects/spring-amqp/blob/838b88aa87a8d4557651123af9d37b1b0d02f6ac/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java#L1967

The field replyHolder of class RabbitTemplate is always cleaned up using the `messageTag` variable in the end of method doSendAndReceiveAsListener. 
Even if the correlatrionId is used to put a pending reply in that replyHolder map.

---
In my humble opinion the method  should ony use one of the local variables   `messageTag` and  `correlationId`.

garyrussell:
Good catch; thanks.

This has been there a long time; I guess most people don't use a custom correlation id.

PR #1182, opened 2020-04-09
GH-1181: Fix memory leak with user correlation

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1181


artembilan:
... and cherry-picked to `2.2.x`.
Back-ported to `2.1.x` after fixing conflicts in test class.

Issue #1190, opened 2020-05-04
ExampleRabbitListenerSpyAndCaptureTest  fails


winster:
Depdencies: spring-boot-starter-test:2.2.5 & spring-rabbit-test:2.2.5
I had the impression that, I can use testcases in spring-rabbit-test in my application. But unfortunately, it does not run.
Major issue with org/springframework/amqp/rabbit/test/ExampleRabbitListenerSpyAndCaptureTest & ExampleRabbitListenerCaptureTest.

Test fails with error
```
Test ignored.
java.lang.NoClassDefFoundError: org/junit/Assume
```
2. RabbitListenerTestHarness is not imported (because the same package). After import in my testcase, IntelliJ shows warning, (but allow me to run test - but test throws error mentioned above)
```
Could not autowire. No beans of 'RabbitListenerTestHarness' type found. 
```

artembilan:
1. `java.lang.NoClassDefFoundError: org/junit/Assume` - good catch! We have to fix `BrokerRunningSupport` to remove that JUnit 4 class.

2. RabbitListenerTestHarness is not imported - right because those sample classes are indeed in the same package. We can fix that moving them to a different one, e.g. `test.sample`

The IntelliJ warning is not relevant. We don't create those beans explicitly so IDEA doesn't know about their configuration. Those beans are created by the `@RabbitListenerTest` processor.

So, two bugs over here.

Please, don't hesitate to fix them via contribution:  https://github.com/spring-projects/spring-amqp/blob/master/CONTRIBUTING.adoc !

garyrussell:
FYI; this only happens when RabbitMQ is not running locally; we only reference the `Assume` class if we get an exception when connecting to Rabbit. When RabbitMQ is running, the test passes.

garyrussell:
I am going to take it; it's a bit more complicated then just removing the references to `Assume`.

winster:
Thanks for the quick fix. How do I get the fix? I tried with 2.2.6.RELEASE (for spring-boot-starter-test & rabbit-test), but no luck.

garyrussell:
It will be in 2.2.7 which is not currently scheduled (we just released 2.2.6 last week).

You can work around it by adding the JUnit4 jar to the class path. If there is some valid reason that you can't do that, I can look at accelerating 2.2.7.

artembilan:
[...]

PR #1191, opened 2020-05-04
GH-1190: Remove reference to Junit4 Assume

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1190

- Remove reference to `Assume` in `BrokerRunningSupport`
- Move example test cases to a new package so they can be easily copied/pasted
- Remove `assumeOnline` field - it looks like it was intended to support running
  tests only if RabbitMQ is NOT running; but there was never any way to set it
  to false

**cherry-pick to 2.2.x**


artembilan:
... and cherry-picked to `2.2.x` .

Issue #1157, opened 2020-02-05
RabbitTestHarness on proxied Beans


Iggle:
Currently, if a Listener Bean contains transactional methods it is wrapped in a cglib proxy, as expected, to ensure transactionality. RabbitTestHarness can thus not produce functioning spies of the Listener as all the proxy's methods are final.

Possible solutions would be to either:

* Return a Mock with AdditionalAnswers.delegatesTo() instead of spy()
* Unwrapping the proxy before spying with Advised.getTargetSource().getTarget()

While only functionally a spy in behaviour, having a look at [SpyDefinition.createSpy](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/mock/mockito/SpyDefinition.java#L88) for @SpyBean, the first option seems to be an effective way to go.

If desired, I could draft up a merge request.

garyrussell:
>If desired, I could draft up a merge request.

Yes, please; thanks in advance.

PR #1199, opened 2020-05-14
GH-1157: Defer spy stubs to original bean

Iggle:
Resolves https://github.com/spring-projects/spring-amqp/issues/1157

CGLib proxies due to e.g. declarative transactional semantics couldn't
be spied upon due to being final. Replacing the spy, which tries to hold
the state itself, with a mock that delegates execution fixes this.


garyrussell:
Good work; thanks! Merged as d3af2095 with minor formatting changes; cherry-picking to 2.2.x, 2.1.x.

Issue #1201, opened 2020-05-15
Support DeBatching Producer Batches in the DMLC


garyrussell:
Also with the SMLC, even when consumer-side batching is not enabled.

https://stackoverflow.com/questions/61822464/directmessagelistenercontainer-is-it-possible-to-use-with-batchmessagelistener/61823993#comment109352702_61823993

PR #1202, opened 2020-05-15
GH-1201: Native BatchMessageListener Support

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1201

Support de-batching producer batches to a `List<Meessage>` with native
`BatchMessageListener`s; previously this was only possible with
`@RabbitListener` which does the debatching itself.

**cherry-pick to 2.2.x**

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
@@ -42,6 +42,7 @@ import org.springframework.amqp.AmqpRejectAndDontRequeueException;
 import org.springframework.amqp.ImmediateAcknowledgeAmqpException;
 import org.springframework.amqp.core.AcknowledgeMode;
 import org.springframework.amqp.core.AmqpAdmin;
+import org.springframework.amqp.core.BatchMessageListener;
 import org.springframework.amqp.core.Message;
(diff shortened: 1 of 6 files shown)


artembilan:
Not sure if related, but posting anyway since it has failed:
```
RabbitTemplateDirectReplyToContainerIntegrationTests > testReceiveTimeoutRequeue() FAILED

    java.lang.AssertionError at RabbitTemplateIntegrationTests.java:396
```

garyrussell:
Yeah - looks like a race in that test; rebuilding "solved" it 😄

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -1920,6 +1921,17 @@ private void checkPossibleAuthenticationFailureFatalFromProperty() {
...
 
+	@Nullable
+	protected List<Message> debatch(Message message) {
+		if (isDeBatchingEnabled() && getBatchingStrategy().canDebatch(message.getMessageProperties())
+				&& getMessageListener() instanceof BatchMessageListener) {
+			final List<Message> messageList = new ArrayList<>();
+			getBatchingStrategy().deBatch(message, fragment -> messageList.add(fragment));
+			return messageList;
+		}
+		return null;
```
kurenchuksergey, line 1932: empty list?

artembilan:

Issue #1215, opened 2020-06-23
Jackson Message Conversion - Allow Abstract Classes


ryandanielspmc:
I am using Kotlin, and I have a sealed class as follows:
```kotlin
@JsonDeserialize(using = FooDeserializer::class)
sealed class Foo
data class FooA(val s: String) : Foo()
data class FooB(val l: Long) : Foo()
```
Then `FooDeserializer` handles determining whether to instantiate `FooA` or `FooB`. The Jackson `ObjectMapper` handles this situation no problem. But, the message conversion functionality around `Jackson2JsonMessageConverter` does not allow abstract classes, which is what `Foo` gets compiled to.

Please allow abstract classes and interfaces in the Jackson message conversion functionality. Let it be the responsibility of the Jackson `ObjectMapper` to handle this situation instead of preventing it.

garyrussell:
We will have to give this some thought; simply relaxing the the check could break many applications; currently, if the `@RabbitListener` parameter is abstract, we fall back to using type information in headers. so we probably need some option.

garyrussell:
Can you provide more information? I just tested it with this code and it worked fine...

```java
	@Test
	void customAbstractClass() {
		byte[] bytes = "{\"field\" : \"foo\" }".getBytes();
		MessageProperties messageProperties = new MessageProperties();
		messageProperties.setInferredArgumentType(Baz.class);
		Message message = new Message(bytes, messageProperties);
		ObjectMapper mapper = new ObjectMapper();
		mapper.registerModule(new BazModule());
		Jackson2JsonMessageConverter j2Converter = new Jackson2JsonMessageConverter(mapper);
		Baz baz = (Baz) j2Converter.fromMessage(message);
		assertThat(((Qux) baz).getField()).isEqualTo("foo");
	}

	public interface Baz {

	}

	public static class Qux implements Baz {

		private String field;

		public Qux(String field) {
			this.field = field;
		}

		public String getField() {
			return this.field;
		}

		public void setField(String field) {
			this.field = field;
		}

	}

	public static class BazDeserializer extends StdDeserializer {

		public BazDeserializer() {
			super(Baz.class);
		}

		@Override
		public Object deserialize(JsonParser p, DeserializationContext ctxt)
				throws IOException, JsonProcessingException {

			p.nextFieldName();
			return new Qux(p.nextTextValue());

		}

	}

	public static class BazModule extends SimpleModule {

		public BazModule() {
[...]

PR #1216, opened 2020-06-23
GH-1215: Allow Abstract Class Deserialization

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1215

Previously, the message converter would fall back to header type info
if the inferred type was abstract.

Furthermore, we did not examine container type content being abstract.

With a custom deserializer, abstract classes can be deserialized.


artembilan:
... and cherry-picked to `2.2.x` after resolving conflicts in `What's New`

Issue #1226, opened 2020-07-21
Spied on RabbitListener not working as expected


liljeholm:
**Affects Version(s):** v2.2.8.RELEASE

---
## Bug report

In the latest version, v2.2.8.RELEASE, there seems to be an issue with `RabbitListenerTestHarness`.
If a `RabbitListener` has an injected Spring bean it is not injected anymore, 
resulting in a NullPointerException when using `LatchCountDownAndCallRealMethodAnswer`.
Rolling back to v2.2.7.RELEASE seems to resolve the problem.

I believe this issue is related to this change: https://github.com/spring-projects/spring-amqp/issues/1157

I've created a demo project to reproduce the problem.
https://github.com/liljeholm/spring-amqp-npe

garyrussell:
Thanks for the example; that change went into 2.2.7
```
* 2a86003fbc6b5fb2e8d6befd7776bb7f62163200 (tag: v2.2.7.RELEASE) [artifactory-release] Release version 2.2.7.RELEASE
* a2119b3c420365b56fbe229663ffa3b60b98d21f GH-1201: Fix for ChannelAwareBatchMessageListener
* 6abdd27e279557fa3e3ff358516c22f1aa099e51 GH-1201: Native BatchMessageListener Support
* 664f470eefb41c288ef54c0157d806ac8779cd36 GH-1198: Support AddressResolver
* bb3b8c89fa0b81db240f3a982df117c280d9ee13 GH-1157: Defer spy stubs to original bean <<<<<<<<<<<<<<<<<<
* a0cc7b42b16ee422de572fa344be286b9fe72425 GH-1194: Address new Sonar issues
* ab7f08f3cff6c3f472530bc8eddac2f0555c0217 GH-1194: Fix cache limit with Pub Confirms channel
* 568854b22a99430b692eb1f99d901f84904daff6 GH-1190: Remove reference to Junit4 Assume
* 9ccd52ac2970b528553bd1aaf64e419ea6eaa0e5 [artifactory-release] Next development version
* 5c0943cb6e57bcb6729e70083c4636ea734c199a (tag: v2.2.6.RELEASE) [artifactory-release] Release version 2.2.6.RELEASE
```

Strangely, for me; I get a different problem - when running against 2.2.7. the first call to the listener occurs before the spy is applied; a second call fails with an NPE - we hit the spy, which has no demoService - there is a race.

Hmmm - just dropped back to 2.2.6 and it still fails the same way - investigating...

[...]

PR #1227, opened 2020-07-21
GH-1226: Fix Test Harness

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1226

https://github.com/spring-projects/spring-amqp/issues/1157 changed the way we spy
listeners - to support CGLIB proxies (e.g. `@Transactional`).

Instead of spying the listener, it mocks the listener and sets a default answer
to call the real method on the delegate.

This broke when users used other answers, such as those provided by the framework.

Change the provided answers to subclass `ForwardsInvocation`.

Also fix `ConcurrentModificationException` in `getExceptions()`.


artembilan:
... and cherry-picked to `2.2.x`.

Back-ported to `2.1.x` as https://github.com/spring-projects/spring-amqp/commit/a141ede7f176d1d2fa195546616913535730b943 after resolving conflicts.

PR #1225, opened 2020-07-15
Update AmqpAppender.java

yhyecho:
if stopped is true,  this.manager.stop(timeout, timeUnit)  is not call, and the rabbit connecting not destory

PR #1229, opened 2020-07-22
GH-1225: Fix Log4j2 Appender Termination

garyrussell:
Replaces https://github.com/spring-projects/spring-amqp/pull/1225

`manager.stop()` was never called to destroy the connection factory, preventing
JVM exit.

Also protect for re-connecting after stop (both appenders).

Tested with a Spring Boot application.

**cherry-pick to 2.2.x, 2.1.x, 1.7.x**


artembilan:
Cherry-picked to `2.2.x` & `2.1.x`.

Looks like we need a separate fix for `1.7.x`: there is no `ApplicaitonContextAware` impl on the `AbstractConnectionFactory` yet.
Thanks

garyrussell:
ok; thanks.

Issue #1230, opened 2020-07-28
Race to get confirm Vs. return


garyrussell:
The `channelsExecutor` is a `ThreadPoolExecutor`, possible to lose the natural ordering of receiving a return after a confirm.

There is a `returnLatch` that is supposed to prevent this, but it is final, so subsequent confirms/returns can come out of order.

Caused by https://github.com/spring-projects/spring-amqp/issues/1055

Issue #1055, opened 2019-07-27
Run the return/confirm callbacks on an executor thread


garyrussell:
https://stackoverflow.com/questions/57234770/with-spring-amqp-whats-the-best-way-to-send-a-message-to-rabbitmq-from-inside/57235777#57235777

garyrussell:
See latest comment on SO answer.

PR #1231, opened 2020-07-28
GH-1230: Fix return/confirm delivery order

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1230

We must guarantee that returns are delivered before confirms.
https://github.com/spring-projects/spring-amqp/issues/1055 changed the logic
to deliver on a different thread to avoid a deadlock in the amqp-client.

Since the executor service might have multiple threads, we added a latch to
ensure the return is delivered first.

Unfortunately, this was a one-shot latch and subsequent deliveries using the
same channel might arrive in the wrong order.

Move the latch to the `PendingConfirm` to reinstate in-order delivery.

**cherry-pick to 2.2.x**


artembilan:

PR #1234, opened 2020-07-30
Add CorrelationData spring-messaging header

garyrussell:
In preparation for https://github.com/spring-cloud/spring-cloud-stream-binder-rabbit/issues/303

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitMessagingTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitMessagingTemplate.java
@@ -197,9 +197,9 @@ public class RabbitMessagingTemplate extends AbstractMessagingTemplate<String>
 	@Override
 	protected void doSend(String destination, Message<?> message) {
 		try {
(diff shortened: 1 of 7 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitMessagingTemplate.java
```diff
@@ -195,7 +197,13 @@ public void convertAndSend(String exchange, String routingKey, Object payload,
 	@Override
 	protected void doSend(String destination, Message<?> message) {
 		try {
-			this.rabbitTemplate.send(destination, createMessage(message));
+			if (message.getHeaders().get(AmqpHeaders.PUBLISH_CONFIRM_CORRELATION) instanceof CorrelationData) {
```
artembilan, line 200: May we perform this `get()` only once?

garyrussell: I was trying to avoid a cast, but I suppose it's better to do that.

garyrussell:
Pushed another commit to close the test CCFs.

Issue #1236, opened 2020-08-05
Extract contentType from MimeType in MessagingMessageConverter


diderash:
When using RabbitMessagingTemplate with ProtobufMessageConverter the contentType header is a MimeType object instead of a String.
I suppose that one change introduced in #1221 is the cause of this issue.

In particular only a String contentType is handled in:
https://github.com/spring-projects/spring-amqp/blob/5b8b3dc8b5cca9d34a5430c5aa38d4498d216390/spring-amqp/src/main/java/org/springframework/amqp/support/converter/MessagingMessageConverter.java#L116-L121

I think that MimeType should be handled too as already done in:
https://github.com/spring-projects/spring-amqp/blob/5b8b3dc8b5cca9d34a5430c5aa38d4498d216390/spring-amqp/src/main/java/org/springframework/amqp/support/SimpleAmqpHeaderMapper.java#L188-L208

PR #1239, opened 2020-08-05
GH-1236: Handle non-String contentType

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1236

The user may set the contentType to a `MimeType`.

Always use `toString()` when replacing the `contentType`.

**cherry-pick to 2.2.x, 2.1.x, 1.7.x**


artembilan:
... and cherry-picked to `2.2.x`, `.2.1.x` & `1.7.x`.

Issue #1251, opened 2020-09-25
Clarification of content encoding with multiple values


bjmi:
**Affects Version(s):** 2.2.5.RELEASE
---
Can you please clarify why a colon **:** was used as separator if multiple content-encoding values are processed.
This happens if e.g. a message body is compressed with `GZipPostProcessor`. A previous `UTF-8` value becomes `gzip:UTF-8` when Spring AMQP was used.

Where does this **:** originates from? Are there any specification documents for that?
Following references advocate a comma **,**
* https://www.rabbitmq.com/publishers.html / https://www.rabbitmq.com/consumers.html
> For example, messages with JSON payload should use application/json. If the payload is compressed with the LZ77 (GZip) algorithm, its content encoding should be gzip.
> 
> Multiple encodings can be specified by separating them with **commas**.
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding

> Content-Encoding: gzip, identity
* (Maybe also interesting https://docs.oasis-open.org/amqp/core/v1.0/amqp-core-messaging-v1.0.html)

> Implementations SHOULD NOT specify multiple content-encoding values except as to be compatible with messages originally sent with other protocols, e.g. HTTP or SMTP.

Our existing AMQP implementation uses comma **,** as separator and this breaks interoperability with newly adopted spring applications that uses Spring AMQP in conjunction with compressed message bodies. What would the right fix for that?
It would be really useful if Spring PostProcessors could be configured with right separator.
Additionally if whitespaces (`gzip : UTF-8` or `gzip, UTF-8`) are present then current implementation is broken too.

Thanks in advance for your thoughts.

### Affected classes
* `AbstractCompressingPostProcessor`
* `AbstractDecompressingPostProcessor`
* `DelegatingDecompressingPostProcessor`

garyrussell:
It's a bug, but we'll have to make it a property to avoid a breaking change. We can make comma the default in 2.3.

bjmi:
Thank you for your fast response.

One thing what was also observed. Some applications leave _contentEncoding_ empty for "raw" data and only fill _contentEncoding_ property if e.g. compression was applied. Media types can contain used charset.
### Uncompressed
```
contentType: application/json    or    application/json; charset=utf-8
contentEncoding: UTF-8
```
vs.
```
contentType: application/json    or    application/json; charset=utf-8
contentEncoding: 
```

### Compressed with gzip
```
[...]

PR #1252, opened 2020-09-25
Fix Compressed contentEncoding Delimiter

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1251

Delimiter should be a comma and whitespace trimmeed.

Handle both delimiters in decompressors and add property for backwards compatibility.

**I will backport to 2.2.x with default `:`**


garyrussell:
Back ported to 2.2.x with flipped default as d73e6aeb0

Issue #1258, opened 2020-10-16
Process not terminated on OOM error


bojanv55:
How come this happened - that is - is this expected behavior? My application was not terminated and started working pretty strange - seems like it could not recover from this java.lang.OutOfMemoryError - but still, spring did not terminate the APP but it continue to run.

```
2020-10-16 16:57:46 (ERROR): ForgivingExceptionHandler Consumer SimpleConsumer [queue=my.partitioned.32.8.user, consumerTag=phoenix.lcoo.clientodds.7b892e20-480a-420a-
java.lang.OutOfMemoryError: Java heap space
        at java.base/java.util.Arrays.copyOf(Arrays.java:3745)
        at java.base/java.io.ByteArrayOutputStream.grow(ByteArrayOutputStream.java:120)
        at java.base/java.io.ByteArrayOutputStream.ensureCapacity(ByteArrayOutputStream.java:95)
        at java.base/java.io.ByteArrayOutputStream.write(ByteArrayOutputStream.java:156)
        at org.springframework.util.StreamUtils.copy(StreamUtils.java:143)
        at org.springframework.util.FileCopyUtils.copy(FileCopyUtils.java:110)
        at org.springframework.amqp.support.postprocessor.AbstractDecompressingPostProcessor.postProcessMessage(AbstractDecompressingPostProcessor.java:91)
        at org.springframework.amqp.rabbit.listener.AbstractMessageListenerContainer.doExecuteListener(AbstractMessageListenerContainer.java:1423)
        at org.springframework.amqp.rabbit.listener.AbstractMessageListenerContainer.executeListener(AbstractMessageListenerContainer.java:1400)
        at org.springframework.amqp.rabbit.listener.DirectMessageListenerContainer$SimpleConsumer.callExecuteListener(DirectMessageListenerContainer.java:1036)
        at org.springframework.amqp.rabbit.listener.DirectMessageListenerContainer$SimpleConsumer.handleDelivery(DirectMessageListenerContainer.java:996)
        at com.rabbitmq.client.impl.ConsumerDispatcher$5.run(ConsumerDispatcher.java:149)
        at com.rabbitmq.client.impl.ConsumerWorkService$WorkPoolRunnable.run(ConsumerWorkService.java:104)
        at com.codahale.metrics.InstrumentedExecutorService$InstrumentedRunnable.run(InstrumentedExecutorService.java:181)
[...]

PR #1259, opened 2020-10-20
GH-1258: Add OOMHandler to the listener containers

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1258#issuecomment-711670729

**I will backport to 2.2.x with a no-op default**


garyrussell:
Backported as 387949691 with no-op default for backwards compatibility.

PR #1262, opened 2020-10-27
GH-891: Docs for Multi RabbitMQ Support

garyrussell:
Supplement to https://github.com/spring-projects/spring-amqp/pull/1111


rwanderc:
Thanks @garyrussell for making it.

PR #1265, opened 2020-11-04
Getter For MessageListener Tuned With Its Setter

ankushgats:
Getter For MessageListener Tuned With Its Setter

Now that we have removed deprecated setMessageListener which used to accept an argument of type 'Object' and now that we have only one setter which accepts an argument of type MessageListener, it now makes sense to return MessageListener from getter method. This not only makes getter and setter tuned but would fix the problem wherein the getter and setter required to be working on the same parameter type. e.g. in jxm library.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
@@ -455,7 +455,7 @@ public abstract class AbstractMessageListenerContainer extends RabbitAccessor
 	/**
 	 * @return The message listener object to register.
 	 */
-	public Object getMessageListener() {
+	public MessageListener getMessageListener() {
 		return this.messageListener;
 	}
 


artembilan (approved):
👍  from me.

Even if it is bytecode incompatible change, it is unlikely that users are going to upgrade their solutions without recompiling. More over it is really fixing that JMX expectation for getter and setter (weird though).

We probably can't make it `protected` - this is going to be much bigger breaking change.

@ankushgats ,

thanks for reporting this and fixing!

@garyrussell ,

WDYT?

garyrussell:
I am a bit concerned about making a breaking API change in a point release; it's unfortunate this wasn't caught before last week's 2.3.0 GA.

@ankushgats It's not clear to me why you would expose this attribute over JMX - it's a complex object that most likely cannot be rendered by a JMX client.

ankushgats:
One simple case is when we register SimpleMessageListenerContainer. This is just my case, we could have many such cases from many people when they start consuming new version with this broken functionality and to make it work we would have to ask clients to change their code.

artembilan:
Right. The question was why would one tried to expose the `MessageListener` property over JMX since it is not that primitive which could be modified from there. Plus it is not clear how it is rendered in the JMX tree. 
So, probably another workaround is to exclude `MessageListener` from MBean info.

ankushgats:
Yes, we always have a workaround. But remember we are asking clients to adjust their code due to a miss in our code.

artembilan:
Right. See @garyrussell 's argument about breaking change in the point release.
We definitely going to accept your change in the next `3.0` version, but the current one probably is going to suffer for this incompatibility. 

We will discuss with him personally what's the problem making getter to return `MessageListener`...

bsalotti:
> Right. The question was why would one tried to expose the `MessageListener` property over JMX since it is not that primitive which could be modified from there. Plus it is not clear how it is rendered in the JMX tree.
> So, probably another workaround is to exclude `MessageListener` from MBean info.

Hi @artembilan,
Just to be sure, the workaround you mention involves for example wiring an instance of `MethodExclusionMBeanInfoAssembler` (configured to ignore the method `setMessageListener`) to the `MBeanExporter` ?

artembilan:
Yes, sounds like that, @bsalotti .

Thanks for the pointer!
You need to inject it into the `MBeanExporter` bean.

garyrussell:
Thanks; merged as e25600b89

Issue #1269, opened 2020-11-18
Question: Limitations of argument type inferral from target method signature


KaiStapel:
This is just a question: In #390 you created the mechanism for inferring the argument type from the target method signature. You introduced the following limitation:

> Methods with a single unannotated parameter or one with a @Payload annotation
> are considered for this feature.

What is the reasoning for limiting the feature for "single unannotated parameter" methods as opposed to simply "single parameter" methods?

I constantly run into the problem of breaking my listeners when I add a `@Valid` annotation to the input parameter. Then I have to analyze for hours what the problem is until I find the solution of also having to add a `@Payload` annotation to make it work again.

This is why I came to this question. Is this limitation really necessary? If so, why?

artembilan:
Would you mind to show the code produces the problem for you?

I'm not sure what drove you to use `@Valid` since `@RabbitListener` doesn't do any validation...

The reason for the limitation is really about a multi-parameter methods when we can't decide which one is intended for payload injection.

KaiStapel:
Env: Spring Boot 2.3.4.RELEASE with spring-boot-starter-amqp

Before (working):
```
  @RabbitListener(queues = "some-queue")
  public void listen(@Valid @Payload SomeDto someDto) {
    [...]
  }
``` 

Then (broken):
```
  @RabbitListener(queues = "some-queue")
  public void listen(@Valid SomeDto someDto) {
    [...]
  }
``` 

Finally (working again):
```
  @RabbitListener(queues = "some-queue")
  public void listen(@Valid @Payload SomeDto someDto) {
    [...]
  }
``` 

The `@Valid` annotation triggers Dto validation. I believe this is outside of `@RabbitListener` and has nothing to do with this question besides being an annotation that breakes the documented limitation "with a single **unannotated** parameter".

Exactly, if the reason is about multi-parameter methods, then why are you introducing the requirement for "no annotation" for single parameter methods? This is exactly my question.

artembilan:
Well, we probably didn't think about any other possible annotation on the single parameter.
That's why my question to you: why do you use that `@Valid` annotation since it doesn't trigger anything when this method is called by the Listener Container? How that DTO validation is triggered? Are you sure that you can confirm the validation is working for you at this point?

We indeed going to look into the solution and take away that limitation, so you still can use `@Valid` without `@Payload`.

KaiStapel:
That would be great.

[...]

PR #1270, opened 2020-11-18
WIP: Proposal how to solve #1269

KaiStapel:
WIP #1269

Tests missing


artembilan (changes requested):
Looks good!

Please, consider to modify some existing `@RabbitListener` to satisfy your expectations for a `@Validated` annotation and type inference.

We probably also need to adjust the mentioned docs since there is no limitation anymore.

Thank you for your effort!

garyrussell:
Closing; see https://github.com/spring-projects/spring-amqp/issues/1269#issuecomment-733843525

garyrussell:
Thanks for the contribution.

garyrussell:
By the way, the proposal doesn't work; this one now fails....

```java
@RabbitListener(queues = "test.converted.args1")
public String foo2(Foo2 foo2, @Header("amqp_consumerQueue") String queue) {
	return foo2 + queue;
}
```

KaiStapel:
Yes, sorry. The proposal was only to highlight the position in the code base which I thought had to be adjusted. The change itself was never tested by me (due to lack of time).

PR #1275, opened 2020-11-25
GH-1269: @RabbitListener: Allow other Annotations

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1269

Previously, a parameter annotated with a "foreign" annoation (e.g. `@Validated`)
would not be considered as an eligible payload conversion target; it must also
have been annotated with `@Payload`.

- remove the check for zero annotations
- add a check that the method is not annotated with both `@Payload` and `@Header`
  - ignore if it does, with a warn log, to be consistent with previous behavior.
  - in a future release we might consider this to be fatal.

**cherry-pick to 2.2.x**


artembilan:

Issue #1279, opened 2020-12-07
"Foo" deserialized as "LinkedHashMap" if listener method argument is "List<Foo>"


andrebrait:
**Affects Version(s):** 2.2.12 (but probably there since 2.2.9)

---

## Context

I have an application that publishes a list of objects (as a JSON array) into an exchange, and a service that consumes those lists of objects. This service has a method similar to this:

```java
@RabbitListener(queue="myqueue")
public void readList(List<Foo> foos) {
    doSomething(foo);
}
```

I attempted upgrading from Spring Boot 2.3.1 to 2.3.3 and then to 2.3.6. With both of them, I started seeing error messages related to the objects inside `List<Foo> foos` being instances of `LikedHashMap` instead of `Foo`. I confirmed that this is the case by remotely debugging. For some reason, I think the generic type is not being used correctly. Reverting back to Spring-AMQP 2.2.7 makes it work again and the objects are indeed instances of `Foo`.

garyrussell:
Are these discrete messages containing a `List<Foo>`?

Or are using using batching (either producer- or consumer-created batches - the latter being added in 2.2)?

Can you show your application.yml/properties?

garyrussell:
Also show your message converter bean - we did add support for deserializing Abstract classes, which had some side effects in the message converter if incorrectly configured.

andrebrait:
Each message is a single `List<Foo>`. There's no batching involved in this particular listener.

I do have batching in another place and that is working fine (argument is a `List<Foo>` and each message is a `Foo`).
I do use a few different message listeners in this application, but the one used in this listener is the following bean:

```java
    public SimpleRabbitListenerContainerFactory defaultRabbitListenerContainerFactory(
            ConnectionFactory connectionFactory,
            SimpleRabbitListenerContainerFactoryConfigurer configurer,
            MessageRecoverer retryRecoverer) {
        var factory = new SimpleRabbitListenerContainerFactory();
        configurer.configure(factory, connectionFactory);
        factory.setErrorHandler(new LoggingConditionalRejectingErrorHandler());
        factory.setAdviceChain(RetryInterceptorBuilder.stateless()
                .recoverer(retryRecoverer)
                .maxAttempts(1)
                .build());
        return factory;
    }
```

Everything else is very boring (mostly Spring Boot autoconfiguration) :smile: 

```java
    @Bean
    public MessageConverter jsonMessageConverter(ObjectMapper objectMapper) {
[...]

PR #1280, opened 2020-12-07
GH-1729: Fix JSON Regression

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1279

Regression: #1215

Test for abstract class should not be applied to container types, which can
be abstract.

**cherry-pick to 2.2.x**


artembilan:

PR #1283, opened 2020-12-08
Use management tag for RabbitMQ image on GH actions

artembilan:

### DIFF
--- a/.github/workflows/pr-build-workflow.yml
+++ b/.github/workflows/pr-build-workflow.yml
@@ -14,6 +14,7 @@ jobs:
         image: rabbitmq:management
         ports:
           - 5672:5672
+          - 15762:15762
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/pr-build-workflow.yml
+++ b/.github/workflows/pr-build-workflow.yml
@@ -11,7 +11,7 @@ jobs:
 
     services:
       rabbitmq:
-        image: rabbitmq
+        image: rabbitmq:management
         ports:
           - 5672:5672
 


.github/workflows/pr-build-workflow.yml
```diff
@@ -11,7 +11,7 @@ jobs:
 
     services:
       rabbitmq:
-        image: rabbitmq
+        image: rabbitmq:management
         ports:
           - 5672:5672
```
garyrussell, line 16: Don't we need to map the management port (15762)?

artembilan: I just copy/paste all of this stuff blindly 😄 
Is that just enough to have another entry like `- 15762:15762` ?
Do we have some tests for that to be sure that they run properly?

garyrussell: Hmmm - now I don't know how the previous PR builds were good; for example `RabbitAdminIntegrationTests` has `@RabbitAvailable(management = true)`

But we have other tests (e.g. `EnableRabbitIntegrationTests`) that don't check for the management plugin is available, but require it.

garyrussell: Oh; it has a try/catch

```java
	@Test
	public void deadLetterOnDefaultExchange() {
		this.rabbitTemplate.convertAndSend("amqp656", "foo");
		assertThat(this.rabbitTemplate.receiveAndConvert("amqp656dlq", 10000)).isEqualTo("foo");
		try {
			Client rabbitRestClient = new Client("http://localhost:15672/api/", "guest", "guest");
			QueueInfo amqp656 = rabbitRestClient.getQueue("/", "amqp656");
			if (amqp656 != null) {
				assertThat(amqp656.getArguments().get("test-empty")).isEqualTo("");
				assertThat(amqp656.getArguments().get("test-null")).isEqualTo("undefined");
			}
		}
		catch (Exception e) {
			// empty
		}
	}
```

Issue #1289, opened 2020-12-23
routingConnectionFactory does not support setting publisher-returns, publisher-confirm-type for child objects


weilingfeng98:
I use the RoutingConnectionFactory setting as my connectionFactory, by setting the defaultTargetConnectionFactory and targetConnectionFactories I give it to fill in the correct values, and at the same time I set the publisher-returns, publisher-confirm-type attribute values for them,
Receive ack information by customizing new RabbitTemplate(routingConnectionFactory).setConfirmCallback(function()), and report an error after running:
java.lang.IllegalArgumentException: Listener not registered: com.mytaxi.spring.multirabbit.example.Application$1@33af2dfd []
	at org.springframework.util.Assert.notNull(Assert.java:198) ~[spring-core-5.2.0.RELEASE.jar:5.2.0.RELEASE]
	at org.springframework.amqp.rabbit.connection.PublisherCallbackChannelImpl.addPendingConfirm(PublisherCallbackChannelImpl.java:1049) ~[spring-rabbit-2.2.0.RELEASE.jar:2.2.0.RELEASE]
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[na:1.8.0_191]
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[na:1.8.0_191]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[na:1.8.0_191]
	at java.lang.reflect.Method.invoke(Method.java:498) ~[na:1.8.0_191]
	at org.springframework.amqp.rabbit.connection.CachingConnectionFactory$CachedChannelInvocationHandler.invoke(CachingConnectionFactory.java:1190) ~[spring-rabbit-2.2.0.RELEASE.jar:2.2.0.RELEASE]
	at com.sun.proxy.$Proxy76.addPendingConfirm(Unknown Source) ~[na:na]
	at org.springframework.amqp.rabbit.core.RabbitTemplate.setupConfirm(RabbitTemplate.java:2344) ~[spring-rabbit-2.2.0.RELEASE.jar:2.2.0.RELEASE]
	at org.springframework.amqp.rabbit.core.RabbitTemplate.doSend(RabbitTemplate.java:2309) ~[spring-rabbit-2.2.0.RELEASE.jar:2.2.0.RELEASE]
[...]

PR #1290, opened 2020-12-23
GH-1289: Confirms and Returns with Routing CF

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1289

`RoutingConnectionFactory` did not support correlated confirms or returns.

Target factories (and default) must have the same settings.


artembilan:
Looks like after cherry-picking we have started to fail: https://build.spring.io/chain/result/viewChainResult.action?planKey=AMQP-R22X&buildNumber=58.

Probably some `returnedMessage` logic had not been back-ported before...

artembilan:
It fails same way on `2.1.x` branch. But there are more work to do for back porting: it doesn't support JUnit 5 yet.

Thanks

garyrussell:
Backported to 2.1.x as ad612a8ae - and to 2.2.x as c72298f475bec7e691f7776ba5c1558e4696cbdf and 73e18747ff8fb1c45a6ac202bd3545e5ed6cb133

PR #1292, opened 2020-12-23
@RabbitListener property improvements

garyrussell:
For bean name properties that can have an expression, allow
the expression to evaluate to an instance of the desired
type instead of a bean name.

### DIFF
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitReturnTypesTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitReturnTypesTests.java
@@ -146,9 +146,9 @@ public class EnableRabbitReturnTypesTests {
 			return (amqpMessage, message, exception) -> null;
 		}
 
-		@RabbitListener(queues = "EnableRabbitReturnTypesTests.1", admin = "#{admin}",
+		@RabbitListener(queues = "EnableRabbitReturnTypesTests.1", admin = "#{@admin}",
 				containerFactory = "#{@rabbitListenerContainerFactory}",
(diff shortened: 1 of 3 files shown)


spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitReturnTypesTests.java
```diff
@@ -117,12 +121,35 @@ public CachingConnectionFactory cf() {
...
+
+		@Bean
+		public RabbitListenerErrorHandler rleh() {
+			return (amqpMessage, message, exception) -> null;
+		}
+
+		@RabbitListener(queues = "EnableRabbitReturnTypesTests.1", admin = "#{admin}",
+				containerFactory = "#{@rabbitListenerContainerFactory}",
+				executor = "#{@exec}", replyPostProcessor = "#{@rpp}", messageConverter = "#{converter}",
+				errorHandler = "#{@rleh}")
```
artembilan, line 152: Why difference in these expression? Why some of them with the `@`, but other not?
Are they evaluated to a different result? Like some to bean references, but other to their bean name?
Thanks

artembilan (changes requested):

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitReturnTypesTests.java
garyrussell: Interesting; looks like the `@` is optional here. I will investigate.

garyrussell: Yes, the `@` is optional here (resolved by the `BeanExpressionContextAccessor` if not there and `BeanFactoryResolver` if present).

Issue #1296, opened 2021-01-12
Starting DirectMessageListenerContainer does not addConsumerToRestart when consumer failed to start


davsclaus:
**Affects Version(s):** 2.3.2

## Bug report

I am using a `DirectMessageListenerContainer` to consume from RabbitMQ via spring-rabbit. 

When I start the listener container via its `start` method, and the consumer fail to start due to for example queue does not exists. The spring-rabbit logs at DEBUG level 

    Queue not present, scheduling consumer SimpleConsumer [queue=myqueue, consumerTag=null identity=68fe1f37] for restart`

But unfortunately the consumer is not added to be restarted as the error happens during startup, and the `started` flag is not yet set to true, as this happens AFTER the consumer is attempted to be started.

See the 3 screenshots from source code debugging.

davsclaus:
<img width="1030" alt="rabbit-start-3" src="https://user-images.githubusercontent.com/477100/104298566-6a610b00-54c4-11eb-8447-083095c819bd.png">
<img width="1100" alt="rabbit-start-2" src="https://user-images.githubusercontent.com/477100/104298569-6af9a180-54c4-11eb-8a67-2681fec953cd.png">
<img width="1022" alt="rabbit-start-1" src="https://user-images.githubusercontent.com/477100/104298571-6b923800-54c4-11eb-87d8-eec3fe0d47ed.png">

davsclaus:
Maybe add `starting` flag to the listener

boolean starting = false

And then set 

`DirectMessageListenerContainer.this.starting = true;` in the top of the `startConsumers` method, so we can check for both starting and started flag in the `addConsumerToRestart` method

```
	private void addConsumerToRestart(SimpleConsumer consumer) {
		if (this.starting || this.started) {
			this.consumersToRestart.add(consumer);
			if (this.logger.isTraceEnabled()) {
				this.logger.trace("Consumers to restart now: " + this.consumersToRestart);
			}
		}
	}
```

artembilan:
Thank you for reporting this, Claus!
This is a good catch: how we have missed it with our tests...

With a quick look into the code I would say that we don't need to check that `this.started` in the `addConsumerToRestart()` and add consumer into that list unconditionally.

Then we need just to adjust a task in the `startMonitor(long idleEventInterval, final Map<String, Queue> namesToQueues)` method to not clean `this.consumersToRestart`  if we are not started yet.

WDYT?

/CC @garyrussell

[...]

PR #1297, opened 2021-01-12
GH-1296: Fix DMLC Recovery: Missing Queue at Start

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1296

- Add `MissingQueueEvent`
- Fix detection of a deleted queue in recovery - previously incorrectly used the
  absense of the queue in `consumersByQueue`, which can be empty if missing during
  start
- Add an index to `SimpleConsumer`
- When adjusting consumer counts, look for gaps in the index sequence because reducing
  the consumer count can remove any idle consumer.
- Change consumers to restart to a `Set` to avoid OOM when no broker
  (see https://github.com/spring-projects/spring-amqp/pull/642)
- Unconditionally add consumers to `consumersToRestart`

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


artembilan (approved):
So smart!

I also thought about moving to `Set` for consumers to avoid OOM, but then it was a problem for me to figure out how to make them unique.
The `index` is a good solution although it is hard to digest it at a glance why would one be so strict trying to fill the gaps 😄 

I mean we just could use unique ids for consumers to reuse them on restart.
However I find your solution less stressful than those UUIDs.

artembilan:
Cherry-picked to `2.2.x` as https://github.com/spring-projects/spring-amqp/commit/b52a549d4af5835bb412a0d80c76411051daff5c after fixing conflicts.

To `2.1.x` as https://github.com/spring-projects/spring-amqp/commit/69ec27847d30a3c05bd5d8c33d91a843dcf4192e after more conflicts fixing.

PR #1301, opened 2021-01-30
Log error when direct message listener fails

arnis:
Currently when `Error` is encountered while invoking direct message listener, default error handler terminates JVM without any information about the issue. My Spring application was shutdown and I had no clue why that happened. During debugging I discovered an exception which was caused by incorrect project dependencies:
`java.lang.NoSuchMethodError: 'com.fasterxml.jackson.core.util.JacksonFeatureSet com.fasterxml.jackson.core.JsonParser.getReadCapabilities()'`

I'm suggesting to log an error message before JVM is terminated.

@pivotal-issuemaster This is an Obvious Fix


artembilan:
@arnis ,

thank you very much for the contribution; looking forward for more!

PR #1303, opened 2021-01-31
Introducing MultiRabbit Bootstrap

rwanderc:
- Creation of the **MultiRabbitBootstrapConfiguration**, activated via properties, since the default **RabbitListenerBPP** bean cannot be overridden with auto-config.
- Injection of the MultiRabbit **admin** to the **RabbitListener**, before it's processed by **RabbitListenerBPP**, so as to have the bean resolved and assigned to the **MethodRabbitListenerEndpoint**. It fixes an issue lately found on the order of loading where the **MultiRabbitAutoConfig** (at boot) was not being processed before **MultiRabbitBPP** starts the **postProcessAfterInitialization()**, which leads to bean not found.
- Move some configs to **RabbitListenerConfigUtils** as they are shared among Rabbit and MultiRabbit, as well as used in the boot.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitBootstrapConfiguration.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitBootstrapConfiguration.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2002-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.amqp.rabbit.annotation;
+
+import org.springframework.amqp.rabbit.config.RabbitListenerConfigUtils;
+import org.springframework.amqp.rabbit.listener.RabbitListenerEndpointRegistry;
+import org.springframework.beans.factory.support.BeanDefinitionRegistry;
+import org.springframework.beans.factory.support.RootBeanDefinition;
+import org.springframework.context.EnvironmentAware;
+import org.springframework.context.annotation.ImportBeanDefinitionRegistrar;
+import org.springframework.core.env.Environment;
+import org.springframework.core.type.AnnotationMetadata;
+import org.springframework.lang.Nullable;
+
+/**
+ * An {@link ImportBeanDefinitionRegistrar} class that registers
+ * a {@link MultiRabbitListenerAnnotationBeanPostProcessor} bean, if MultiRabbit
+ * is enabled.
+ *
+ * @author Wander Costa
+ *
+ * @since 1.4
+ *
+ * @see RabbitListenerAnnotationBeanPostProcessor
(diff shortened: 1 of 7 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitBootstrapConfiguration.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+public class MultiRabbitBootstrapConfiguration implements ImportBeanDefinitionRegistrar, EnvironmentAware {
+
+	private Environment environment;
+
+	@Override
+	public void registerBeanDefinitions(@Nullable AnnotationMetadata importingClassMetadata,
+			BeanDefinitionRegistry registry) {
+
+		if (isMultiRabbitEnabled() && !registry.containsBeanDefinition(
```
rwanderc, line 51: Hi @garyrussell, to give you some background on this class, I initially implemented a slightly different version of [your suggestion](https://github.com/spring-projects/spring-amqp/pull/1111#discussion_r563866329), centralizing the condition for both AutoConfig (at boot) and for the MultiRabbitBPP.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/RabbitListenerConfigUtils.java
```diff
@@ -36,4 +36,29 @@
...
+
+	/**
+	 * The bean name of the default ConnectionFactory.
+	 */
+	public static final String RABBIT_CONNECTION_FACTORY_BEAN_NAME = "rabbitConnectionFactory";
+
+	/**
+	 * The default property to enable/disable MultiRabbit processing.
+	 */
+	public static final String MULTI_RABBIT_ENABLED_PROPERTY = "spring.multirabbitmq.enabled";
```
garyrussell, line 52: Shouldn't this be aligned with boot property naming? `spring.rabbitmq.multi-cluster-enabled` or similar, and added to `RabbitProperties` (`multiClusterEnabled`)?

Otherwise users will get warnings in their IDE about an unknown property.

garyrussell (commented):
Looks pretty clean to me after a quick review. Just one comment.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/RabbitListenerConfigUtils.java
rwanderc: I will provide boot part with that.
The reason I initially didn't add it under `spring.rabbitmq` (and therefore at `RabbitProperties`), is because the attribute would also be replicated to all brokers under `spring.multirabbitmq.connections.[]` which might cause confusion.
E.g.: `spring.multirabbitmq.connections.brokerA.multi-cluster-enabled`

What would you think about `spring.multirabbitmq.multi-cluster-enabled` (and therefore at `MultiRabbitProperties`)?

```
@ConfigurationProperties(prefix = "spring.multirabbitmq")
public class MultiRabbitProperties {

	/**
	 * A flag to enable/disable MultiRabbit processing.
	 */
	@Value("${" + RabbitListenerConfigUtils.MULTI_RABBIT_ENABLED_PROPERTY + "}")
	private boolean multiClusterEnabled;

	/**
	 * A map representing the RabbitProperties of all available brokers.
	 */
	private Map<String, RabbitProperties> connections = new HashMap<>();

	public Map<String, RabbitProperties> getConnections() {
		return this.connections;
	}

	public void setConnections(@Nullable final Map<String, RabbitProperties> connections) {
		this.connections = Optional.ofNullable(connections).orElse(new HashMap<>());
	}

}
```

garyrussell: Ah, yes; makes sense. 

Maybe just `spring.multirabbitmq.enabled`?

Thanks.

garyrussell (approved):

PR #1304, opened 2021-02-08
Fix PublishCallCh.afterAckCallback race condition

artembilan:
Related to https://stackoverflow.com/questions/66071476/impossible-nullpointerexception-springframework-rabbitmq-failed-to-invoke

When we have several pending confirms, we handle them concurrently
in the `doHandleConfirm()`.
The `afterAckCallback` can be set `null` in one thread and
another will fail with NPE after acquiring a monitor from the
`getPendingConfirmsCount()`.

* Extract a local variable for the `callback` inside an `if` block
* Initialize it with `this.afterAckCallback` only if
`getPendingConfirmsCount() == 0` is true,
and set `this.afterAckCallback` to null inside `synchronized (this)`
block
* Call `callback.accept(this)` when local variable is not `null`
and outside of `synchronized` for better performance on end-user callback

**Cherry-pick to 2.2.x**

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/PublisherCallbackChannelImpl.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/PublisherCallbackChannelImpl.java
@@ -1026,9 +1026,18 @@ public class PublisherCallbackChannelImpl
 			}
 			finally {
 				try {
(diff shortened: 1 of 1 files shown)


garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1305

garyrussell:
Cherry-picked as 67e587061

Issue #1309, opened 2021-03-14
Message are not commited (loss) when using @TransactionalEventListener to send a message in a JPA Transaction


kuldeepkalassts:
**Background of the code:** 

In order to replicate a production scenario, I have created a dummy app that will basically save something in DB in a transaction, and in the same method, it publishEvent and publishEvent send a message to rabbitMQ.

**Classes and usages**

Transaction Starts from this method.:
```
@Override
	@Transactional
	public EmpDTO createEmployeeInTrans(EmpDTO empDto) {
		return createEmployee(empDto);
	}
```
This method saves the record in DB and also triggers publishEvent

```
@Override
	public EmpDTO createEmployee(EmpDTO empDTO) {
		
		EmpEntity empEntity = new EmpEntity();
		BeanUtils.copyProperties(empDTO, empEntity);

		System.out.println("<< In Transaction : "+TransactionSynchronizationManager.getCurrentTransactionName()+" >>  Saving data for employee " + empDTO.getEmpCode());

		// Record data into a database
		empEntity = empRepository.save(empEntity);	
		
		// Sending event , this will send the message.
		eventPublisher.publishEvent(new ActivityEvent(empDTO));
		
		return createResponse(empDTO, empEntity);
	}
```

This is ActivityEvent

```
import org.springframework.context.ApplicationEvent;
import com.kuldeep.rabbitMQProducer.dto.EmpDTO;
public class ActivityEvent extends ApplicationEvent {
	public ActivityEvent(EmpDTO source) {
		super(source);
	}
}
```
And this is TransactionalEventListener for the above Event.

```
	//@Transactional(propagation = Propagation.REQUIRES_NEW)
	@TransactionalEventListener(phase = TransactionPhase.AFTER_COMMIT)
	public void onActivitySave(ActivityEvent activityEvent) {
		System.out.println("Activity got event ... Sending message .. ");
		kRabbitTemplate.convertAndSend(exchange, routingkey, empDTO);		
	}
```
This is kRabbitTemplate is a bean config like this :

```
@Bean
	public RabbitTemplate kRabbitTemplate(ConnectionFactory connectionFactory) {
		final RabbitTemplate kRabbitTemplate = new RabbitTemplate(connectionFactory);
		kRabbitTemplate.setChannelTransacted(true);
		kRabbitTemplate.setMessageConverter(kJsonMessageConverter());
		return kRabbitTemplate;
	}
```

**Problem Definition**

[...]

PR #1310, opened 2021-03-15
GH-1309: Don't Bind Resource When Sync Not Active

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1309

If the `RabbitTemplate` was called from a `@TransactionalEventListener`,
with phase `AFTER_COMMIT`, a transaction is "active", but synchronizations
are already cleared. It is too late to synchronize this transaction.

We end up with an orphaned `ResourceHolder` with pending transaction commits.

Don't bind the resource holder if synchronization is not active.

However, the proper solution, if users want to synchronize the rabbit transaction
with the global transaction, is to use the `BEFORE_COMMIT` phase.

See the discussion on the Github issue for more information.

**cherry-pick to 2.2.x**

### DIFF
--- a/.github/workflows/pr-build-workflow.yml
+++ b/.github/workflows/pr-build-workflow.yml
@@ -30,8 +30,9 @@ jobs:
           arguments: check
 
       - name: Capture Test Results
+        if: ${{ failure() }}
         uses: actions/upload-artifact@v2
         with:
           name: test-results
-          path: spring-rabbit/build/reports/tests
+          path: 'spring-rabbit/build/reports/tests/**/*.*'
           retention-days: 3
--- a/.github/workflows/pr-build-workflow.yml
+++ b/.github/workflows/pr-build-workflow.yml
@@ -30,9 +30,9 @@ jobs:
           arguments: check
 
       - name: Capture Test Results
-        if: ${{ failure() }}
+        if: failure()
         uses: actions/upload-artifact@v2
         with:
           name: test-results
-          path: 'spring-rabbit/build/reports/tests/**/*.*'
+          path: '*/build/reports/tests/**/*.*'
           retention-days: 3
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/TransactionalEventListenerTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/TransactionalEventListenerTests.java
@@ -48,7 +48,7 @@ import com.rabbitmq.client.Channel;
 
 /**
  * @author Gary Russell
- * @since 2.3.6
+ * @since 2.2.16
  *
  */
 @SpringJUnitConfig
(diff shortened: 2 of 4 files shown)


spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/TransactionalEventListenerTests.java
```diff
@@ -0,0 +1,141 @@
...
+import org.springframework.transaction.event.TransactionPhase;
+import org.springframework.transaction.event.TransactionalEventListener;
+import org.springframework.transaction.support.AbstractPlatformTransactionManager;
+import org.springframework.transaction.support.DefaultTransactionStatus;
+
+import com.rabbitmq.client.Channel;
+
+/**
+ * @author Gary Russell
+ * @since 2.3.6
```
artembilan, line 51:
```suggestion
 * @since 2.2.16
```

artembilan (changes requested):
Does error on GH Actions report say anything:
```
org.springframework.amqp.rabbit.connection.ClientRecoveryCompatibilityTests > testDefeatRecovery() FAILED
    java.lang.AssertionError at ClientRecoveryCompatibilityTests.java:90
```
?

garyrussell:
It's unrelated; hard to tell what went wrong without a stack trace.

garyrussell:
Strange; same failure; passes locally; investigating.

artembilan:
So, looks like channel proxy has been refreshed in between somehow:
```
proxy = (ChannelProxy) channel;
assertThat(proxy.getTargetChannel()).isSameAs(channel1);
```

See the latest GH Actions report.

artembilan:
Let’s try this instead : https://github.com/marketplace/actions/publish-unit-test-results ?

artembilan:
So, `**/build/test-results/tests/**.xml` ?

garyrussell:
With the artifacts on Bamboo, we capture the HTML: https://build.spring.io/build/admin/edit/defaultBuildArtifact.action?buildKey=AMQP-MASTER-JOB1

garyrussell:
Success! https://github.com/spring-projects/spring-amqp/actions/runs/656070418

garyrussell:
@artembilan Fixed - I need to figure out which test is leaving that flag set, but that's a task for another day 😄 

Sorry, I squashed one too many commits - the fix is in the last one.

.github/workflows/pr-build-workflow.yml
```diff
@@ -30,8 +30,9 @@ jobs:
           arguments: check
 
       - name: Capture Test Results
+        if: ${{ failure() }}
         uses: actions/upload-artifact@v2
         with:
           name: test-results
-          path: spring-rabbit/build/reports/tests
+          path: 'spring-rabbit/build/reports/tests/**/*.*'
```
artembilan, line 37: Don't we need to gather all the test reports?
From all the modules in the project?

artembilan (changes requested):

.github/workflows/pr-build-workflow.yml
garyrussell: Yes; I am testing that now on Kafka.

garyrussell:
Sorry 😦

artembilan:
... and cherry-picked to `2.2.x` without `pr-build-workflow.yml` file.

Issue #1331, opened 2021-04-29
Context Switching with ThreadChannelConnectionFactory


garyrussell:
Allow a thread to hand over its channel(s) to another thread.

PR #1332, opened 2021-04-29
GH-1331: ThreadChannelConnFactory Improvements

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1331

Provide a mechanism so that a thread can hand off its channel(s) to another thread.
Add protection to close a channel that would be orphaned if a second channel is transferred
to the thread and the thread failed to close its channel beforehand.

Other improvements:

- only call the channel listener when a channel is actually created
- physically close a transactional channel that is no longer in the thread local because
  `closeThreadChannel` was called
- move reset of physical close flag to the actual close

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/ThreadChannelConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/ThreadChannelConnectionFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 the original author or authors.
+ * Copyright 2020-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,6 +17,9 @@
 package org.springframework.amqp.rabbit.connection;
 
 import java.io.IOException;
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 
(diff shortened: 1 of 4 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/ThreadChannelConnectionFactory.java
```diff
@@ -144,6 +150,40 @@ public synchronized void destroy() {
...
+	public Object prepareSwitchContext() {
+		return prepareSwitchContext(UUID.randomUUID());
+	}
+
+	Object prepareSwitchContext(UUID uuid) {
+		if (getPublisherConnectionFactory() instanceof ThreadChannelConnectionFactory) {
+			((ThreadChannelConnectionFactory) getPublisherConnectionFactory()).prepareSwitchContext(uuid);
+		}
+		this.contextSwitches.put(uuid, ((ConnectionWrapper) createConnection()).prepareSwitchContext());
+		return uuid;
```
artembilan, line 168: This all looks cool: very interesting and smart approach.

I can merge it as when you answer some my questions.

How about just to return a `Context` as is without storing in between into that `contextSwitches` map?
I wonder if some one will call `prepareSwitchContext()`, but won't call `switchContext()`... So, feels like a memory leak.

At least we should consider to empty that map on CF destroy.

Also it looks like when we call `prepareSwitchContext()` several times, we end up with empty `Context`, just because we called `this.channels.remove()` for the first one. Probably better to protect ourselves just only for a one call.

And then `switchContext(Context context)` against an empty one will finish just with void.

Does it ring a bell?

Thanks

artembilan (changes requested):

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/ThreadChannelConnectionFactory.java
garyrussell: I thought about just returning a `Context` but the problem is that we have a nested `publisherConnectionFactory` which could be (and is by default) another TCCF, and the `RabbitTemplate` uses it by default.

That means we need a `Context` for each factory bound to the same key.

I also didn't want to make it easy for them to reflect into the `Context` and access the channels directly.

I will push another commit with some of your other suggestions to add some more protection, but if a user does something like never call  `switchContext` there's not really much we can do about that.

Issue #1295, opened 2021-01-11
Payload Validation on @RabbitHandler methods.


garyrussell:
See https://github.com/spring-projects/spring-kafka/issues/1675

Also needs the equivalent of https://github.com/spring-projects/spring-kafka/pull/824

PR #1333, opened 2021-05-04
GH-1295: Simplify Validator configuration; add payload Validation on @RabbitHandler methods

bdshadow:
based on links to spring-kafka here in the comment: https://github.com/spring-projects/spring-amqp/issues/1295#issue-783683976

While developing, i found 3 things which i didn't like or which confused me. Could you please help me with them?

1. Why  doesn't  `PayloadMethodArgumentResolver.validate` do "validator.supports()" check? I added it to `DelegatingInvocableHandler.invoke` method, which is not in spring-kafka corresponding implementation, but i'm not sure if it's correct
2. Also `@Max` annotation [here](https://github.com/spring-projects/spring-kafka/pull/824/files#diff-7a504b2b073e72009bb3d490a22504d3ff2eeb53e14ab50c0fa847ed98a88f44R1781) in tests is not needed at all. It just confused me for some time, and i didn't put it to the current MR.
3. I don't like my solution with `@DirtiesContext` in the second commit. I spent quite much time to use  `@RabbitErrorHandler` with rethrowing error and `convertSendAndRecieve` with configuring `this.rabbitTemplate.setMessageConverter(new RemoteInvocationAwareMessageConverterAdapter(new Jackson2JsonMessageConverter()));` , but didn't succeed to make it work, when validation fails for json. `toMessage` was called fine, however `fromMessage` wasn't and exception couldn't be converted to be recieved. `@DirtiesContext` in my mr helps to cleanup `errorHandlerError`, used in my new `testRabbitHandlerValidation` test and `testInvalidPojoConversion`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -46,6 +46,7 @@ ext {
 	commonsPoolVersion = '2.9.0'
 	googleJsr305Version = '3.0.2'
 	hamcrestVersion = '2.2'
+	hibernateValidationVersion = '6.2.0.Final'
 	jacksonBomVersion = '2.11.3'
 	jaywayJsonPathVersion = '2.4.0'
 	junit4Version = '4.13.1'
@@ -362,6 +363,7 @@ project('spring-rabbit') {
 
 		testApi project(':spring-rabbit-junit')
 		testImplementation("com.willowtreeapps.assertk:assertk-jvm:$assertkVersion")
+		testImplementation "org.hibernate.validator:hibernate-validator:$hibernateValidationVersion"
 		testRuntimeOnly 'org.springframework:spring-web'
 		testRuntimeOnly "org.apache.httpcomponents:httpclient:$commonsHttpClientVersion"
 		testRuntimeOnly 'com.fasterxml.jackson.core:jackson-core'
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
@@ -92,6 +92,7 @@ import org.springframework.util.ClassUtils;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.ReflectionUtils;
 import org.springframework.util.StringUtils;
+import org.springframework.validation.Validator;
 
 /**
  * Bean post-processor that registers methods annotated with {@link RabbitListener}
@@ -960,6 +961,10 @@ public class RabbitListenerAnnotationBeanPostProcessor
 
 		private MessageHandlerMethodFactory createDefaultMessageHandlerMethodFactory() {
 			DefaultMessageHandlerMethodFactory defaultFactory = new DefaultMessageHandlerMethodFactory();
+			Validator validator = RabbitListenerAnnotationBeanPostProcessor.this.registrar.getValidator();
+			if (validator != null) {
+				defaultFactory.setValidator(validator);
+			}
 			defaultFactory.setBeanFactory(RabbitListenerAnnotationBeanPostProcessor.this.beanFactory);
 			DefaultConversionService conversionService = new DefaultConversionService();
 			conversionService.addConverter(
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2020 the original author or authors.
+ * Copyright 2014-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -85,6 +85,7 @@ import org.springframework.core.task.TaskExecutor;
 import org.springframework.lang.Nullable;
 import org.springframework.messaging.handler.annotation.support.DefaultMessageHandlerMethodFactory;
 import org.springframework.messaging.handler.annotation.support.MessageHandlerMethodFactory;
+import org.springframework.messaging.handler.invocation.HandlerMethodArgumentResolver;
 import org.springframework.messaging.handler.invocation.InvocableHandlerMethod;
 import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
@@ -964,6 +965,10 @@ public class RabbitListenerAnnotationBeanPostProcessor
 			conversionService.addConverter(
 					new BytesToStringConverter(RabbitListenerAnnotationBeanPostProcessor.this.charset));
 			defaultFactory.setConversionService(conversionService);
+
+			List<HandlerMethodArgumentResolver> customArgumentsResolver =
+					new ArrayList<>(RabbitListenerAnnotationBeanPostProcessor.this.registrar.getCustomMethodArgumentResolvers());
+			defaultFactory.setCustomArgumentResolvers(customArgumentsResolver);
 			defaultFactory.afterPropertiesSet();
 			return defaultFactory;
 		}
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/MultiMethodRabbitListenerEndpoint.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/MultiMethodRabbitListenerEndpoint.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2019 the original author or authors.
+ * Copyright 2015-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -25,6 +25,7 @@ import org.springframework.amqp.rabbit.listener.adapter.HandlerAdapter;
 import org.springframework.amqp.rabbit.listener.adapter.MessagingMessageListenerAdapter;
 import org.springframework.lang.Nullable;
 import org.springframework.messaging.handler.invocation.InvocableHandlerMethod;
+import org.springframework.validation.Validator;
 
 /**
  * @author Gary Russell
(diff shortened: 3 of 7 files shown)


garyrussell (changes requested):
Thanks for the contribution!

For 1, I suggest you open an issue and/or a PR against spring-framework; we don't own that code.

Here is one solution to the `@DirtiesContext` problem:

```
$ git diff
diff --git a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java
index 0fd744314..6e64450a6 100644
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java
@@ -189,7 +189,10 @@ public class EnableRabbitIntegrationTests {
        private RabbitAdmin rabbitAdmin;
 
        @Autowired
-       private CountDownLatch errorHandlerLatch;
+       private CountDownLatch errorHandlerLatch1;
+
+       @Autowired
+       private CountDownLatch errorHandlerLatch2;
 
        @Autowired
        private AtomicReference<Throwable> errorHandlerError;
@@ -503,11 +506,10 @@ public class EnableRabbitIntegrationTests {
        }
 
        @Test
-       @DirtiesContext
        public void testInvalidPojoConversion() throws InterruptedException {
                this.rabbitTemplate.convertAndSend("test.invalidPojo", "bar");
 
-               assertThat(this.errorHandlerLatch.await(10, TimeUnit.SECONDS)).isTrue();
+               assertThat(this.errorHandlerLatch1.await(10, TimeUnit.SECONDS)).isTrue();
                Throwable throwable = this.errorHandlerError.get();
                assertThat(throwable).isNotNull();
                assertThat(throwable).isInstanceOf(AmqpRejectAndDontRequeueException.class);
@@ -526,12 +528,11 @@ public class EnableRabbitIntegrationTests {
        }
 
        @Test
-       @DirtiesContext
        public void testRabbitHandlerValidation() throws InterruptedException {
                ValidatedClass validatedObject = new ValidatedClass();
                validatedObject.setBar(42);
                this.jsonRabbitTemplate.convertAndSend("multi.json.exch", "multi.json.valid.rk", validatedObject);
-               assertThat(this.errorHandlerLatch.await(10, TimeUnit.SECONDS)).isTrue();
+               assertThat(this.errorHandlerLatch2.await(10, TimeUnit.SECONDS)).isTrue();
                Throwable throwable = this.errorHandlerError.get();
                assertThat(throwable).isNotNull();
                assertThat(throwable).isInstanceOf(AmqpRejectAndDontRequeueException.class);
@@ -1786,7 +1787,12 @@ public class EnableRabbitIntegrationTests {
                }
 
                @Bean
-               public CountDownLatch errorHandlerLatch() {
+               public CountDownLatch errorHandlerLatch1() {
+                       return new CountDownLatch(1);
+               }
+
+               @Bean
+               public CountDownLatch errorHandlerLatch2() {
                        return new CountDownLatch(1);
                }
 
@@ -1804,7 +1810,13 @@ public class EnableRabbitIntegrationTests {
                                }
                                catch (Throwable e) {
                                        errorHandlerError().set(e);
-                                       errorHandlerLatch().countDown();
+                                       if (e.getCause().getCause() instanceof
+                                                       org.springframework.messaging.converter.MessageConversionException) {
+                                               errorHandlerLatch1().countDown();
+                                       }
+                                       else {
+                                               errorHandlerLatch2().countDown();
+                                       }
                                        throw e;
                                }
                        }).given(handler).handleError(Mockito.any(Throwable.class));
```

artembilan (changes requested):
See `RabbitListenerAnnotationBeanPostProcessor.setMessageHandlerMethodFactory()`:
```
	/**
	 * Set the {@link MessageHandlerMethodFactory} to use to configure the message
	 * listener responsible to serve an endpoint detected by this processor.
	 * <p>By default, {@link DefaultMessageHandlerMethodFactory} is used and it
	 * can be configured further to support additional method arguments
	 * or to customize conversion and validation support. See
	 * {@link DefaultMessageHandlerMethodFactory} Javadoc for more details.
	 * @param messageHandlerMethodFactory the {@link MessageHandlerMethodFactory} instance.
	 */
	public void setMessageHandlerMethodFactory(MessageHandlerMethodFactory messageHandlerMethodFactory) {
```

So, why do we need this new complex logic if it is already covered by existing structures and we have a hook to configure it. And actually from the same `RabbitListenerAnnotationBeanPostProcessor` injection point.

Thank for understanding!

bdshadow:
@garyrussell thank you! I'll fix the MR.

@artembilan thank you for your remark too. I can remove it, no problem. I used the implementation from the spring-kafka project as an example. It was [referenced in the issue](https://github.com/spring-projects/spring-amqp/issues/1295#issue-783683976) by @garyrussell . [Here](https://docs.spring.io/spring-kafka/reference/html/#kafka-validation) in the docs it is said:
> Starting with version 2.2, it is now easier to add a Validator to validate @KafkaListener @Payload arguments. Previously, you had to configure a custom DefaultMessageHandlerMethodFactory and add it to the registrar. Now, you can add the validator to the registrar itself.

So i believe the motivation behind this solution is that it is easier for users to configure it. In my opinion, it is easier, but not so much, the old configuration option was also good. Maybe @garyrussell can advise something about it. As a new contributor, i'm open to your suggestions.

artembilan:
Sure! That's also possible to expose that `Validator` option to align it with what we have so far in Spring for Apache Kafka as you point out. Although I still think that we just need to propagate `Validator` into an internal `DefaultMessageHandlerMethodFactory` - nothing more (unless some unit tests to verify).

garyrussell:
I don't remember the motivation, but user applications typically don't interact with the BPP directly, whereas customizing the registrar is a well-known extension point via the `RabbitListenerConfigurer`.

I believe that was why I did it that way.

artembilan:
So, probably the solution should be like expose a `Validator` option on the `RabbitListenerEndpointRegistrar` and propagate it into an internal `MessageHandlerMethodFactory`. If a `MessageHandlerMethodFactory` provided, it is already its responsibility to supply a `Validator` properly. 

I'm still not sure that we have to do some validation logic manually. Why that `DefaultMessageHandlerMethodFactory` and its `Validator` behavior is not enough for us here?

garyrussell:
>like expose a `Validator` option 

?

This PR does that.

>I'm still not sure that we have to do some validation logic manually. 

Where is that?

artembilan:
> This PR does that.

Sorry, I was confused that `validator` option is also exposed on the `BPP` in this PR. But now I see that changes to `BPP` is the only way to propagate a `Validator` from the end-user code via `RabbitListenerConfigurer` to the `DefaultMessageHandlerMethodFactory`.

> Where is that?

See `DelegatingInvocableHandler` changes. My point is that it is still delegating to some of the `InvocableHandlerMethod` which is going to call its `MethodArgumentResolver`s. So, the question generally why that custom inner `PayloadValidator` which essentially a `PayloadMethodArgumentResolver` without any propagation somewhere? I see it is stored locally as `this.validator` and used directly in the code. Therefore the question if we really can't live without this one and why? Maybe the factory for handlers in that  `DelegatingInvocableHandler` is not supplied with a proper `PayloadMethodArgumentResolver` which performs a validation logic?..

garyrussell:
Thanks for making me revisit this.

The reason we need to do validation here is because we only invoke the payload resolver if the payload (in `providedArgs`) is not already the required type...

```java
	/**
	 * Get the method argument values for the current message, checking the provided
	 * argument values and falling back to the configured argument resolvers.
	 * <p>The resulting array will be passed into {@link #doInvoke}.
	 * @since 5.1.2
	 */
	protected Object[] getMethodArgumentValues(Message<?> message, Object... providedArgs) throws Exception {
		MethodParameter[] parameters = getMethodParameters();
		if (ObjectUtils.isEmpty(parameters)) {
			return EMPTY_ARGS;
		}

		Object[] args = new Object[parameters.length];
		for (int i = 0; i < parameters.length; i++) {
			MethodParameter parameter = parameters[i];
			parameter.initParameterNameDiscovery(this.parameterNameDiscoverer);
			args[i] = findProvidedArgument(parameter, providedArgs);
			if (args[i] != null) {
				continue;  <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<< WE EXIT HERE
			}
			if (!this.resolvers.supportsParameter(parameter)) {
				throw new MethodArgumentResolutionException(
						message, parameter, formatArgumentError(parameter, "No suitable resolver"));
			}
			try {
				args[i] = this.resolvers.resolveArgument(parameter, message);
			}
			catch (Exception ex) {
				// Leave stack trace for later, exception may actually be resolved and handled...
					String exMsg = ex.getMessage();
					if (exMsg != null && !exMsg.contains(parameter.getExecutable().toGenericString())) {
						logger.debug(formatArgumentError(parameter, exMsg));
					}
				}
				throw ex;
			}
		}
		return args;
	}
```

With `@RabbitHandler` - we must have already converted to the required type because that's how we route to that handler.

With `@RabbitListener` we go through the resolver to extract the payload from the message (and the validation is done there). With `@RabbitHandler`, we have to add the payload to the provided args (hence the early exit).

However, I noticed that we only do that (add the payload to the provided args) when there is a default handler. So it does mean we'll do a double validation if there is no default handler.

https://github.com/spring-projects/spring-amqp/blob/1d56763bdaee527f9b8e078528001bcc2a165890/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/HandlerAdapter.java#L73-L79

So, the validation here should be skipped if the `defaultHandler` is `null`.

garyrussell:
Fix for Kafka: https://github.com/spring-projects/spring-kafka/pull/1788

artembilan (approved):
LGTM

garyrussell:
@bdshadow Thanks for the contribution!

In future, please don't force push to the PR (except after rebasing). It's easier for us to review if responses to review comments are in a discrete commit; we always squash before the final merge.

Waiting for the PR build and will then merge.

Thanks again.

garyrussell (approved):

garyrussell:
Merged as c577fbbf311b7ab8fcfc9274724f1d37d6929742 

Thanks again!

Issue #1312, opened 2021-03-19
Direct reply-to still needs correlation ID to work!?


igor-susic:
When communicating between two different `RabbitMQ` clients and using direct reply-to feature, `spring-amqp` project requests message correlation ID to be set. This is not quite how this feature should behave if I understood the documentation well enough.

So per [direct reply-to documentation](https://www.rabbitmq.com/direct-reply-to.html) when you send request (message) that has property `repy-to` set to `amq.rabbitmq.reply-to` then Rabbit will generate some unique string in a way that will look like this:
`amq.rabbitmq.reply-to.g1hkABlyYWJiaXRAcmFiYml0LXNlY29uZC1ub2RlAAAEBQAAAABgVO6L.dRFik1oCgCYm5rUQycVZ3Q==`
which then should be used by the server that will respond by sending it's response to the `""` exchange and with the mentioned unique string as routing key.

Nowhere does it asks for correlation ID, it should not be requested as the Rabbit itself can manage pairing request with response with this approach.

Also in [spring-amqp documentation for direct repy-to](https://docs.spring.io/spring-amqp/reference/html/#direct-reply-to) it stands:

> Starting with version 3.4.0, the RabbitMQ server supports direct reply-to. This eliminates the main reason for a fixed reply queue (to avoid the need to create a temporary queue for each request). Starting with Spring AMQP version 1.4.1 direct reply-to is used by default (if supported by the server) instead of creating temporary reply queues.

and also

> When using a fixed reply queue (other than amq.rabbitmq.reply-to), you must provide correlation data so that replies can be correlated to requests.

This leads me to believe that correlation ID should not come in question in this use case. Spring on the other hand behaves differently.
While sending message it will set the correlation ID to the value of `message tag` this happens at the [line 2033 inside RabbitTemplate](https://github.com/spring-projects/spring-amqp/blob/03fca1d6a3f8aff2e46dbfd1eff2bebbc3169673/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java#L2033) and when receiving reply it fails on [line 615 in RabbitTemplate](https://github.com/spring-projects/spring-amqp/blob/03fca1d6a3f8aff2e46dbfd1eff2bebbc3169673/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java#L2033) where it needs correlation ID to proceed.

Some code to reproduce this would be:

[...]

PR #1334, opened 2021-05-05
GH-1312: Correlation on Channel with Direct Reply

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1312

With Direct Reply-To we can correlate the reply using the channel
instead of requiring the server to echo back the correlation id.


src/reference/asciidoc/whats-new.adoc
```diff
@@ -34,6 +34,9 @@ See <<template-confirms>> for more information.
 When using returns and correlated confirms, the `CorrelationData` now requires a unique `id` property.
 See <<template-confirms>> for more information.
 
+When using direct reply-to, you can now configure the template such that the server does not need to return correlation data with the reply.
+See <<>> for more information.
```
artembilan, line 38: Missed link.

artembilan (changes requested):

Issue #1335, opened 2021-05-05
Improve RejectAndDontRequeueRecoverer


garyrussell:
Allow customization of the exception message.

PR #1336, opened 2021-05-05
GH-1335: RejectAndDontRequeueRecoverer Improvement

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1335

Allow customization of exception message.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/retry/RejectAndDontRequeueRecoverer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/retry/RejectAndDontRequeueRecoverer.java
@@ -51,6 +51,7 @@ public class RejectAndDontRequeueRecoverer implements MessageRecoverer {
 
 	/**
 	 * Construct an instance with the provided exception message.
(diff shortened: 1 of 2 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/retry/RejectAndDontRequeueRecoverer.java
```diff
@@ -35,14 +38,33 @@
...
+	public RejectAndDontRequeueRecoverer() {
+		this(() -> "Retry Policy Exhausted");
+	}
+
+	/**
+	 * Construct an instance with the provided exception message supplier.
+	 * @param messageSupplier the message supplier.
+	 * @since 2.3.7
+	 */
+	public RejectAndDontRequeueRecoverer(Supplier<String> messageSupplier) {
```
artembilan, line 57: Can't we also have a plain String-based ctor?
Just in case for typical consistency with what we have so far in many other exceptions...

artembilan (changes requested):

Issue #1339, opened 2021-05-20
MessagingMessageListenerAdapter Skips Error Handler with MessageConversionException


nkvaratskhelia:
I'm using spring-amqp's (v 2.3.6) `rabbitTemplate.sendAndReceive(exchange, routingKey, message)` method for sending messages. `RabbitTemplate` is configured with `Jackson2JsonMessageConverter`.

If I send a malformed json in the message body, I can see from the log that message conversion fails with `org.springframework.amqp.support.converter.MessageConversionException: Failed to convert Message content`, as expected. However, the `sendAndReceive` method doesn't abort and continues execution until a `replyTimeout` is hit. This exception occurs before execution reaches my `@RabbitListener` annotated method. It seems that `sendAndReceive` doesn't react to this exception and continues waiting for a reply from the consumer, which never arrives, so it just keeps on waiting until replyTimeout passes and then returns null.

As a comparison, if there is an exception inside my `@RabbitListener` annotated method, then `sendAndReceive` aborts instantly and returns.

I have the same situation as described in [this issue](https://github.com/spring-projects/spring-amqp-samples/issues/58). But in my case, I'm using rpc pattern and I need to somehow signal `sendAndReceive` to return immediately if an exception occurs and not wait for the full replyTimeout duration to elapse.

Is there any way to abort `sendAndReceive` in case of the `MessageConversionException` scenario described above?

Here's a [minimal reproducer](https://github.com/nkvaratskhelia/rabbit-demo) of the issue.

nkvaratskhelia:
I first asked this as a [question on stackoverflow](https://stackoverflow.com/questions/67590935/how-to-abort-sendandreceive-on-messageconversionexception) and Artem suggested to move it here as an issue. As a possible solution he also suggested implementing a custom `ConditionalRejectingErrorHandler` and overriding it's `protected void handleDiscarded(Message failed)` method to send back a reply with an error message. I'm not quite sure how this can be done, so any suggestions are welcome.

garyrussell:
The problem is we only catch LEFEs

https://github.com/spring-projects/spring-amqp/blob/cc0323a8c40509986f1d3a182fdefe98c40756e5/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapter.java#L158-L164

We should catch all exceptions and allow the user to configure an error handler to return some kind of error result.

artembilan:
[...]

PR #1346, opened 2021-06-07
GH-1339: Fix RLErrorHandler with Conversion Ex.

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1339

Error handler was not called for conversion exceptions, preventing
the application frm returning some error to the caller for request/reply
processing.

**cherry-pick to 2.2.x**


artembilan:

ThomHurks:
We just updated to Spring Boot 2.4.7 and tests started to fail related to AMQP. It seems some tests where we expect ListenerExecutionFailedException to be thrown it now throws for MessageConversionException. Looking at the changelog it’s likely caused by this change. Doesn’t that mean this is kind of a breaking change in a patch release? It surprised us at least. We can change the tests to expect a different exception, but didn’t yet verify if this affects runtime behavior in any way.

Issue #1347, opened 2021-06-07
DirectMessageListenerContainer Does Not Consider Authentication Failures as Fatal


garyrussell:
`possibleAuthenticationFailureFatal` is only used by the SMLC; docs say it applies to both.

We might need to change it to false by default for the DMLC to avoid a behavior change.

PR #1349, opened 2021-06-08
GH-1347: DMLC Ignored Authentication Failures

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1347

Container starts and keeps trying to connect instead of honoring
the property.

**cherry-pick to 2.2.x**

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
@@ -41,6 +41,7 @@ import java.util.stream.Stream;
 import org.apache.commons.logging.Log;
 
 import org.springframework.amqp.AmqpApplicationContextClosedException;
+import org.springframework.amqp.AmqpAuthenticationException;
 import org.springframework.amqp.AmqpConnectException;
 import org.springframework.amqp.AmqpException;
 import org.springframework.amqp.AmqpIOException;
@@ -145,6 +146,7 @@ public class DirectMessageListenerContainer extends AbstractMessageListenerConta
 	 */
 	public DirectMessageListenerContainer() {
 		setMissingQueuesFatal(false);
+		doSetPossibleAuthenticationFailureFatal(false);
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -422,6 +425,22 @@ protected void actualStart() {
 		super.doStart();
 		final String[] queueNames = getQueueNames();
 		checkMissingQueues(queueNames);
+		if (isPossibleAuthenticationFailureFatal()) {
+			Connection connection = null;
+			try {
+				getConnectionFactory().createConnection();
+			}
+			catch (AmqpAuthenticationException ex) {
```
artembilan, line 433: In docs you say `PossibleAuthenticationFailureException` and I see that exactly this `PossibleAuthenticationFailureException` is caught in the `SimpleMessageListenerContainer.AsyncMessageProcessingConsumer`.

Shouldn't the same one be caught here as well?

artembilan (changes requested):

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
garyrussell: I think that is dead code; I just ran a test and it is handled by this

https://github.com/spring-projects/spring-amqp/blob/8e8dda721f9f7a0bd2a2373ccf4c40103f3ffbdd/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java#L601-L603

and

https://github.com/spring-projects/spring-amqp/blob/8e8dda721f9f7a0bd2a2373ccf4c40103f3ffbdd/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java#L1361-L1364

I'll create a new issue to clean that up.

artembilan: OK. Should we merge then this as is?
I see that `RabbitExceptionTranslator` does the proper conversion:
```
if (ex instanceof PossibleAuthenticationFailureException) {
	return new AmqpAuthenticationException(ex);
}
```

garyrussell: I would say yes, if you don't see any problems with it.

artembilan:
... and cherry-picked to `2.2.x` after fixing conflicts.

PR #1353, opened 2021-06-11
GH-1339: Fix NPEs with returns after conversion ex

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1339

Previous fix calls the error handler after a conversion exception, but
sending a reply returned by the eh (or returning the exception) caused
NPEs because the return logic referenced the inbound converted message.

**cherry-pick to 2.2.x**


artembilan:

Issue #1359, opened 2021-06-24
Duplicate afterReceivePostProcessors execution with rabbitTemplate.convertSendAndReceiveAsType


bastiat:
## Bug report

Version:
[INFO] |  \- org.springframework.amqp:spring-rabbit:jar:2.2.15.RELEASE:compile
[INFO] |     +- com.rabbitmq:amqp-client:jar:5.9.0:compile
[INFO] |     +- org.springframework.amqp:spring-amqp:jar:2.2.15.RELEASE:compile

Problem:

// given
rabbitTemplate.addAfterReceivePostProcessors(someLoggingPostProcessor)

// when
rabbitTemplate.convertSendAndReceiveAsType(final String exchange, final String routingKey, final Object message,
			@Nullable final MessagePostProcessor messagePostProcessor, @Nullable final CorrelationData correlationData,
			ParameterizedTypeReference<T> responseType)

// then
**someLoggingPostProcessor is executed twice.**

Possible source of problem:

afterReceivePostProcessors from rabbitTemplate are set to DirectReplyToMessageListenerContainer
https://github.com/spring-projects/spring-amqp/blame/main/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java#L1984

and now afterReceivePostProcessors are executed twice:

first in 
https://github.com/spring-projects/spring-amqp/blob/main/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java#L1543

and next in 
https://github.com/spring-projects/spring-amqp/blame/main/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java#L2031

PR #1360, opened 2021-06-25
GH-1359: After Receive Post Processor Invoked x2

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1359

When the `RabbitTemplate` is configured with `afterReceivePostProcessors`
and uses the default internal `DirectReplyToMessageListenerContainer`, the
postprocessors are applied twice, once by the container and once by the
template.

The template should not propagate the post processors into the container.


artembilan:

Issue #1362, opened 2021-07-16
Callback handling in case of exception in Spring Transaction (synchronization).


kuldeepkalassts:
## Enhancement

This is in continuation of https://github.com/spring-projects/spring-amqp/issues/1341,

We have seen that in case exception occurs in Spring Transaction After Completion Phase, the exception is thrown back to the caller method of Spring transaction . So is there some callback handling available to caller code in case the error is thrown ?

Also we have seen that in this case the resource is left over from previous transaction in case of exception thrown while commit and this corrupts the thread for subsequent flows. Can this be handled?

garyrussell:
Errors in `TransactionSynchronization.afterCompletion()` are simply logged in `TransactionSynchronizationUtils`. There is no way to throw an exception to the calling thread. I don't see an easy way to add a callback; that code is all in core spring-transaction.

As discussed on that issue; I can't see any way that a resource can be "left over" from one operation to another; this code has been around for over a decade.  Please provide a [MCRE](https://stackoverflow.com/help/minimal-reproducible-example) that exhibits this behavior.

garyrussell:
On second thought - I **do** see a way to provide a callback, we could add it to the `RabbitResourceSynchronization`,

PR #1363, opened 2021-07-16
GH-1362: Enable Capture of Tx Synchronization Fail

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1362

**cherry-pick to 2.3.x**

### DIFF
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitTemplateTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitTemplateTests.java
@@ -17,6 +17,7 @@
 package org.springframework.amqp.rabbit.core;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 import static org.assertj.core.api.Assertions.assertThatIllegalStateException;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.assertj.core.api.Assertions.fail;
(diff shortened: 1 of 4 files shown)


spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitTemplateTests.java
```diff
@@ -638,6 +642,9 @@ void resourcesClearedAfterTxFailsWithSync() throws IOException, TimeoutException
 		assertThatIllegalStateException()
 			.isThrownBy(() -> (TransactionSynchronizationManager.getSynchronizations()).isEmpty())
 			.withMessage("Transaction synchronization is not active");
+		assertThatExceptionOfType(AfterCompletionFailedException.class)
+			.isThrownBy(() -> ConnectionFactoryUtils.checkAfterCompletion());
+		ConnectionFactoryUtils.enableAfterCompletionFailureCapture(true);
```
artembilan, line 647: You probably mean `false` in the end of test.
Plus I believe you realize that this is global static flag, which can be changed in one thread, but checked in the other.
Is it really intentional?
Also: you probably have missed to mention this flag in the doc.

artembilan (changes requested):

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitTemplateTests.java
garyrussell: Oops; yes; it was intentionally global - they shouldn't have to set it on each thread.

Doh - I had an unsaved change to the doc file.

artembilan: Good, but why this is still `true` in the end of this test ? 😄

artembilan (changes requested):

artembilan:
... and cherry-picked to `2.3.x`

Issue #1352, opened 2021-06-10
Add Support for RabbitMQ Stream Java Client


garyrussell:
New listener container for the stream API to support invoking `@RabbitListener` etc.

https://github.com/rabbitmq/rabbitmq-stream-java-client

welsh:
Is there a tentative date for 2.4 being release with this being present? I see the last Milestone was 25 days ago.

garyrussell:
Next milestone is Sept 20; with release candidate in October and GA in November, aligned with the Boot 2.6 RC and GA.

https://calendar.spring.io/

PR #1364, opened 2021-07-22
GH-1352: ConsumerCustomizer Improvements

garyrussell:
- pass in the listener id, if available
- narrow to a specific interface to aid Boot auto configuration
- add null check

### DIFF
--- a/spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/listener/ConsumerCustomizer.java
+++ b/spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/listener/ConsumerCustomizer.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/listener/ConsumerCustomizer.java
```diff
@@ -0,0 +1,31 @@
...
+import com.rabbitmq.stream.ConsumerBuilder;
+
+/**
+ * Customizer for {@link ConsumerBuilder}.
+ *
+ * @author Gary Russell
+ * @since 2.4
+ *
+ */
+public interface ConsumerCustomizer extends BiConsumer<String, ConsumerBuilder> {
```
artembilan, line 30:
```suggestion
@FunctionalInterface
public interface ConsumerCustomizer extends BiConsumer<String, ConsumerBuilder> {
```

?

artembilan (changes requested):
Just one minor consistency concern.

PR #1366, opened 2021-09-07
GH-1352: Add RabbitStreamTemplate

garyrussell:
See https://github.com/spring-projects/spring-amqp/issues/1352

Initial commit of Stream Producer support.

### DIFF
--- a/spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/producer/ProducerCustomizer.java
+++ b/spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/producer/ProducerCustomizer.java
@@ -22,7 +22,8 @@ import com.rabbitmq.stream.ProducerBuilder;
 
 /**
  * Called to enable customization of the {@link ProducerBuilder} when a new producer is
- * created. The first parameter is the creator's bean name.
+ * created. The first parameter should be the bean name of the component that calls this
+ * customizer. Refer to the RabbitMQ Stream Java Client for customization options.
  *
  * @author Gary Russell
  * @since 2.4
--- a/spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/producer/ProducerCustomizer.java
+++ b/spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/producer/ProducerCustomizer.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.rabbit.stream.producer;
+
+import java.util.function.BiConsumer;
+
+import com.rabbitmq.stream.ProducerBuilder;
+
+/**
+ * Called to enable customization of the {@link ProducerBuilder} when a new producer is
+ * created. The first parameter is the creator's bean name.
+ *
+ * @author Gary Russell
+ * @since 2.4
+ *
+ */
+@FunctionalInterface
+public interface ProducerCustomizer extends BiConsumer<String, ProducerBuilder> {
+}
--- a/spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/producer/RabbitStreamTemplate.java
+++ b/spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/producer/RabbitStreamTemplate.java
@@ -21,7 +21,7 @@ import org.springframework.amqp.core.MessagePostProcessor;
 import org.springframework.amqp.support.converter.MessageConverter;
 import org.springframework.amqp.support.converter.SimpleMessageConverter;
 import org.springframework.beans.factory.BeanNameAware;
-import org.springframework.beans.factory.DisposableBean;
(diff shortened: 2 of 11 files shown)


spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/producer/ProducerCustomizer.java
```diff
@@ -0,0 +1,33 @@
...
+
+package org.springframework.rabbit.stream.producer;
+
+import java.util.function.BiConsumer;
+
+import com.rabbitmq.stream.ProducerBuilder;
+
+/**
+ * Called to enable customization of the {@link ProducerBuilder} when a new producer is
+ * created. The first parameter is the creator's bean name.
```
artembilan, line 25: "Creator"?
Many you mean a `ProducerBuilder` bean name?
Otherwise it is confusing who is "creator"...

spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/producer/RabbitStreamTemplate.java
```diff
@@ -0,0 +1,191 @@
...
+import com.rabbitmq.stream.ConfirmationHandler;
+import com.rabbitmq.stream.Constants;
+import com.rabbitmq.stream.Environment;
+import com.rabbitmq.stream.MessageBuilder;
+import com.rabbitmq.stream.Producer;
+import com.rabbitmq.stream.ProducerBuilder;
+
+/**
+ * @author Gary Russell
+ * @since 2.8
```
artembilan, line 42: Probably `2.4` and some JavaDocs should be there anyway. I guess.
At least some general one.

spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/producer/RabbitStreamTemplate.java
```diff
@@ -0,0 +1,191 @@
...
+	private ProducerCustomizer producerCustomizer = (name, builder) -> { };
+
+	/**
+	 * Construct an instance with the provided {@link Environment}.
+	 * @param environment the environment.
+	 * @param streamName the stream name.
+	 */
+	public RabbitStreamTemplate(Environment environment, String streamName) {
+		this.environment = environment;
+		this.streamName = streamName;
```
artembilan, line 68: `Assert.notNull()`? For both?

spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/producer/RabbitStreamTemplate.java
```diff
@@ -0,0 +1,191 @@
...
+	@Override
+	public ListenableFuture<Boolean> convertAndSend(Object message) {
+		return convertAndSend(message, null);
+	}
+
+	@Override
+	public ListenableFuture<Boolean> convertAndSend(Object message, @Nullable MessagePostProcessor mpp) {
+		Message message2 = this.messageConverter.toMessage(message, new StreamMessageProperties());
+		if (mpp != null) {
+			message2 = mpp.postProcessMessage(message2);
```
artembilan, line 131: What if `postProcessMessage()` returns `null`?

spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/producer/RabbitStreamTemplate.java
```diff
@@ -0,0 +1,191 @@
...
+	 * @param environment the environment.
+	 * @param streamName the stream name.
+	 */
+	public RabbitStreamTemplate(Environment environment, String streamName) {
+		this.environment = environment;
+		this.streamName = streamName;
+	}
+
+
+	private synchronized Producer createOrGetProducer() {
```
artembilan, line 72: Why can't we do this in the `afterPropertiesSet()`?
I can guess because this template might be used outside of Spring management, but then that `DisposableBean` contract is still not going to be satisfied...

artembilan (changes requested):

spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/producer/ProducerCustomizer.java
garyrussell: I was trying not to refer to the template since this can be reused by any component that creates a producer from the environment.

Maybe "The first parameter should be the bean name of the component that calls the customizer." ??

spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/producer/RabbitStreamTemplate.java
garyrussell: Hmmm - checkstyle didn't detect this.

artembilan: Right; because it probably does not parse a content of such a JavaDoc comment. So, even that `@author` might be enough to pass the rule 😄

spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/producer/ProducerCustomizer.java
artembilan: Yes, this makes more sense.
Thanks

spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/producer/RabbitStreamTemplate.java
garyrussell: 1. Because it will cause a connection to the broker and `InitializingBean` is too soon.
2. I added `reset()` for programmatic - I will change it to `AutoCloseable`.

artembilan (approved):
LGTM.

Let me know if I can merge it already or you have something else in mind to add, e.g. docs 😄

garyrussell:
I just pushed one more fix for the test; this can be merged after the action completes.

Docs will be in a separate PR.

Issue #1365, opened 2021-08-19
Auto recovery mechanism to support Declareables declared by RabbitAdmin.declaredXXX methods


jhassony:
Some applications require dynamic queue creation and deletion without needing or knowing their names in advance for creating Spring Beans.

Since the framework disables the client's auto-recovery feature (and the documentation specifically says that Spring's auto-recovery replaces it completely), one would expect that every exchange, queue or binding that is declared using the framework would really be recoverable.

I suggest to have `RabbitAdmin`'s auto recovery mechanism recover all `Declarable` elements (exchanges/queues/bindings) even if they are not Spring Beans, as long as they were declared by `RabbitAdmin.declareXXX` methods.

PR #1367, opened 2021-09-09
GH-1365: Recover Manual Declarations

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1365

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitAdmin.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitAdmin.java
@@ -19,10 +19,14 @@ package org.springframework.amqp.rabbit.core;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Properties;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -63,6 +67,7 @@ import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 
 import com.rabbitmq.client.AMQP.Queue.DeclareOk;
+import com.rabbitmq.client.AMQP.Queue.DeleteOk;
 import com.rabbitmq.client.AMQP.Queue.PurgeOk;
 import com.rabbitmq.client.Channel;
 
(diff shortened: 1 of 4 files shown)


src/reference/asciidoc/amqp.adoc
```diff
@@ -5371,6 +5371,20 @@ Starting with version 2.1, anonymous queues are declared with argument `Queue.X_
...
 You can revert to the previous behavior by calling `queue.setLeaderLocator(null)` after constructing the instance.
 
+[[declarable-recovery]]
+===== Recovering Auto-Delete Declarations
+
+Normally, the `RabbitAdmin` (s) only recover queues/exchanges/bindings that are declared as beans in the application context; if any such declarations are auto-delete, they will be removed by the broker if the connection is lost.
+When the connection is re-established, the admin will redeclare the entities.
+Normally, entities created by calling `admin.declareQueue(...)`, `admin.declareExchange(...)` and `admin.declareBinding(...)` will not be recovered.
+
+Starting with version 2.4, the admin has a new property `redeclareManualDeclarations`; when true, the admin will recover these entities in addition to the beans in the application context.
```
artembilan, line 5381: The Spring Boot configuration property candidate?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitAdmin.java
```diff
@@ -220,6 +229,9 @@ public void declareExchange(final Exchange exchange) {
 		try {
 			this.rabbitTemplate.execute(channel -> {
 				declareExchanges(channel, exchange);
+				if (this.redeclareManualDeclarations) {
+					this.manualDeclarables.put(exchange.getName(), exchange);
```
artembilan, line 233: Doesn't this look like even beans are added to the `this.manualDeclarables` ?
I mean I don't see a difference in the logic when admin declares a bean or this is a manual call of this method...

artembilan (approved):
LGTM.

I can merge after answers to my questions.

Thanks

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitAdmin.java
garyrussell: Oops; good catch.

src/reference/asciidoc/amqp.adoc
garyrussell: I am not so sure it will commonly be used; this is the first request in over 10 years.

Probably best to wait for on-demand.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitAdmin.java
garyrussell: Actually, its ok; the admin doesn't call these methods internally - it uses `declareQueues()` etc.

But I will add a test to ensure that doesn't change in the future.

PR #1369, opened 2021-09-16
GH-1352: Docs for RabbitMQ Stream Plugin Support

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1352

### DIFF
--- a/src/reference/asciidoc/stream.adoc
+++ b/src/reference/asciidoc/stream.adoc
@@ -1,0 +1,138 @@
+[[stream-support]]
+=== Using the RabbitMQ Stream Plugin
+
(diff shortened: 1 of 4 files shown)


src/reference/asciidoc/stream.adoc
```diff
@@ -0,0 +1,138 @@
+[[stream-support]]
+=== Using the RabbitMQ Stream Plugin
+
+Version 2.4 introduces initial support for the https://github.com/rabbitmq/rabbitmq-stream-java-client[RabbitMQ Stream Plugin Java Client] for the https://rabbitmq.com/stream.html[RabbitMQ Stream Plugin].
```
artembilan, line 4: Why "initial"? 
Is there anything else we could add on top of existing API or something else?

artembilan (changes requested):
Just one question.

src/reference/asciidoc/stream.adoc
garyrussell: The current implementation (client/plugin) doesn't support concurrency - it's coming in a future release.

Issue #1372, opened 2021-09-25
"queue-names" attribute of <rabbit:listener/> element can not cooperates with property value correctly


EddieChoCho:
## Bug report
* The version of Spring AMQP: 2.3.10
* The version of spring-core: 5.2.14.RELEASE

### Description
Config with XML
    ```
    <rabbit:listener ref="message" queue-names="${system.queues}" />
    ```

And add a property with names of multiple queue
    ```
    system.queues=queue1,queue2
    ```
    
When I start the service, the value of the property will not be convert to an array of String.
    ```
    Caused by: com.rabbitmq.client.ShutdownSignalException: channel error; protocol method: #method<channel.close>(reply-code=404, reply-text=NOT_FOUND - no queue 'queue1,queue2' in vhost...
    ```
    
It works fine if I set the value directly into the xml file
    ```
     <rabbit:listener ref="message" queue-names="queue1,queue2" />
    ```
    
I found out that in the method:"applyPropertyValues" of [AbstractAutowireCapableBeanFactory.java#L1674](https://github.com/spring-projects/spring-framework/blob/50f20162934c1d3282615dce6696e3b56141f7ac/spring-beans/src/main/java/org/springframework/beans/factory/support/AbstractAutowireCapableBeanFactory.java#L1674), the value of the property is an instance of ManagedList which has a TypedStringValue element: "queue1,queue2".
Maybe we should change it to an instance of String, so it can be converted successfully through [TypeConverterDelegate.java#L385](https://github.com/spring-projects/spring-framework/blob/50f20162934c1d3282615dce6696e3b56141f7ac/spring-beans/src/main/java/org/springframework/beans/TypeConverterDelegate.java#L385).

PR #1373, opened 2021-09-27
GH-1732: Fix Listener Container Parser

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1372

Do not split `queue-names` in the parser - leave it to Spring's type converter.
Also support SpEL in `queues`; a more elegant solution would require major
refactoring of the parser; so this is just a compromise.


artembilan:

PR #1385, opened 2021-11-03
Initial Commit for Spring Framework 6.0/JDK 17

garyrussell:


build.gradle
```diff
@@ -242,6 +242,7 @@ subprojects { subproject ->
     compileKotlin.dependsOn updateCopyrights
 
 	test {
+        jvmArgs '--add-opens java.base/java.util.zip=ALL-UNNAMED'
```
artembilan, line 245: Does it still work if I run tests from IDE?

build.gradle
```diff
@@ -549,6 +550,7 @@ task api(type: Javadoc) {
 	options.overview = 'src/api/overview.html'
 	options.stylesheetFile = file('src/api/stylesheet.css')
 	options.links(rootProject.ext.javadocLinks)
+	options.addStringOption('Xdoclint:none', '-quiet')
```
artembilan, line 553: Do you still see JavaDocs warning when you run this task?

gradle.properties
```diff
@@ -1,5 +1,6 @@
-version=2.4.0-SNAPSHOT
+version=3.0.0-SNAPSHOT
 org.gradlee.caching=true
 org.gradle.daemon=true
 org.gradle.parallel=true
+org.gradle.jvmargs=--add-opens java.base/java.util.zip=ALL-UNNAMED
```
artembilan, line 5: Is it really necessary?
I believe it does not help for consumers of our library: they still have to add this `--add-opens` when they run their applications. 
No?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.8.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.2-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionSha256Sum=8de6efc274ab52332a9c820366dd5cf5fc9d35ec7078fd70c8ec6913431ee610
```
artembilan, line 6: We need to add respective checksum to let the Gradle be cached on CI: https://gradle.org/release-checksums/

artembilan (changes requested):

gradle/wrapper/gradle-wrapper.properties
garyrussell: Strange; I added it when I updated the version; I guess `./gradlew wrapper` removed it 😦

build.gradle
garyrussell: Yes.

build.gradle
garyrussell: You have to add it to the run configuration for `BatchingRabbitTemplateTests`.

gradle/wrapper/gradle-wrapper.properties
artembilan: Correct. It does remove it. We have to add it manually afterwards.
Probably question to Gradle team 😄

build.gradle
artembilan: Yeah... That's not I do on regular basis...
We might need to revise the logic to avoid reflection over there (if any).

PR #1389, opened 2021-11-10
Increase Xmx - GH PR Action Failures

garyrussell:
See if this fixes it.


artembilan (changes requested):
Still failing...
I see in SI we have this in `build.gradle`:
```
	test {
		maxHeapSize = '2g'
		jvmArgs '-XX:+HeapDumpOnOutOfMemoryError'
```
Might be the case here  as well?

PR #1390, opened 2021-11-10
Message toString() Improvement

garyrussell:
Don't convert large message bodies to a `String` in `toString()`.

Set a limit (50) that can be modified by users.

Avoid possible OOM Errors.


artembilan:
Still on GH Action.
In SI we use these options for Gradle: `org.gradle.jvmargs=-Xms512m -Xmx4g -XX:MaxPermSize=1024m -XX:MaxMetaspaceSize=1g`

artembilan:

PR #1391, opened 2021-11-10
Fix DEBUG Logging for Projection

garyrussell:
Don't Log Converted Message With Projection

`toString()` cannot be invoked on the payload proxy created by
the projection factory.

Increase memory for builds.


x
```diff
@@ -0,0 +1,3 @@
+https://github.com/spring-projects/spring-kafka/blob/bd154520a5cc7454748b4ca24e692d82594edd6a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L2743-L2754
```
artembilan, line 1: This file is junk

artembilan (changes requested):

artembilan:
... and back-ported to `2.3.x` & `2.2.x` after some conflicts fixing in the Gradle config.

Issue #1401, opened 2021-12-12
SimpleMessageListenerContainer#setConcurrency issue


zysaaa:
Hi, I have some questions about the ```SimpleMessageListenerContainer#setConcurrency``` method.

In this [pr](https://github.com/spring-projects/spring-amqp/pull/1081), ```SimpleMessageListenerContainer#setConcurrency``` logic is adjusted:
```
int consumersToSet = Integer.parseInt(concurrency.substring(0, separatorIndex));
int maxConsumersToSet = Integer.parseInt(concurrency.substring(separatorIndex + 1));
Assert.isTrue(maxConsumersToSet >= consumersToSet,
		"'maxConcurrentConsumers' value must be at least 'concurrentConsumers'");
this.concurrentConsumers = 1;
this.maxConcurrentConsumers = null;
setConcurrentConsumers(consumersToSet);
setMaxConcurrentConsumers(maxConsumersToSet);                              
```

So, if I call  ```setConcurrency``` twice:
```
simpleMessageListenerContainer.setConcurrency("5-5");
simpleMessageListenerContainer.setConcurrency("10-10");
```
1. ```Log``` is confusing, will always be ```from 1 to xxx```:

```
2021-12-12 15:18:38.319 DEBUG [,,] 30246 --- [       Thread-3] o.s.a.r.l.SimpleMessageListenerContainer : Changing consumers from 1 to 5
2021-12-12 15:18:38.380 DEBUG [,,] 30246 --- [       Thread-3] o.s.a.r.l.SimpleMessageListenerContainer : Changing consumers from 1 to 10
```

2. The number of consumers will grow to ```14 = (1 + (5-1) + (10-1))```, and then slowly decrease to 10 (if no new messages received).

I think we should add a parameter to ```public void setConcurrentConsumers(final int concurrentConsumers)```, such as``` boolean check```.  If we call  ```setConcurrency("xx")``` or ```setConcurrentConsumers(xx)```, we have to check  current ```maxConcurrentConsumers``` and this newly set ```concurrentConsumers```.

When ```setConcurrency("xx-xx")``` is called, validation of current ```maxConcurrentConsumers``` and newly set ```concurrentConsumers ``` is unnecessary.

(```setMaxConcurrentConsumers``` also make the same change)

So we can remove these two lines in ```setConcurrency```, these two values ​​should not be changed ​​before ```setConcurrentConsumers``` called:
```
this.concurrentConsumers = 1;
[...]

PR #1403, opened 2021-12-13
GH-1401: SMLC: Fix setConcurrency bug

zysaaa:
Fix https://github.com/spring-projects/spring-amqp/issues/1401


garyrussell:
Thanks; merged as 07443ceb77272c165c12dc8154cac22da41dd10b after adding a test.

Issue #1406, opened 2021-12-15
SimpleMessageListenerContainer: repeat ack when batch listen enabled


zysaaa:
https://github.com/spring-projects/spring-amqp/blob/cd429dcf9e068ca2c4728d65dc96e6163b9d6e0b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java#L932-L948

When message is  ```null``` after ```MessagePostProcessor#process```, original message will be acked directly. But subsequent ```consumer.commitIfNecessary(isChannelLocallyTransacted(), getMessageAckListener())```  method will also perform an multiple ack, may resulting in repeat ack and raise some exception:
```
DEBUG org.springframework.amqp.rabbit.listener.BlockingQueueConsumer - Received shutdown signal for consumer tag=amq.ctag-X4tiGLDLmy4f0BH8pJfINg
com.rabbitmq.client.ShutdownSignalException: channel error; protocol method: #method<channel.close>(reply-code=406, reply-text=PRECONDITION_FAILED - unknown delivery tag 2, class-id=60, method-id=80)
	at com.rabbitmq.client.impl.ChannelN.asyncShutdown(ChannelN.java:517)

```

It doesn't seem necessary to do an ACK here, ```consumer.commitIfNecessary``` will do a multiple ack anyway.

Test code:

```
        SimpleMessageListenerContainer smlc = new SimpleMessageListenerContainer(connectionFactory);
        smlc.addQueueNames("testq");
        smlc.setAfterReceivePostProcessors(new MessagePostProcessor() {
            int index = 0;
            @Override
            public org.springframework.amqp.core.Message postProcessMessage(org.springframework.amqp.core.Message message) throws AmqpException {
                if (index++ >= 1) {
                    return null;
                } else {
                    return message;
                }
            }
        });
        smlc.setBatchSize(10);
        smlc.setConsumerBatchEnabled(true);
        smlc.setMessageListener(new BatchMessageListener() {
            @Override
            public void onMessageBatch(List<Message> messages) {
                System.out.println("I got: " + messages.size());
            }
        });
```

garyrussell:
Good catch - it looks like this will only happen if the last record in the batch is skipped because `commitIfNecessary` uses `multiple=true` (only acks the last message in the batch).

PR #1407, opened 2021-12-15
GH-1406: Fix Possible Double Ack in Consumer Batch

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1406

Previously, if an MPP caused the last record in an ack to be skipped
two acks for the same deliveryTag would be sent causine a channel
shutdown.

There is no need to ack skipped records because the entire batch is
acked (with basicAck multiple=true).


artembilan:
... and cherry-picked to `2.3.x` and `2.2.x` as https://github.com/spring-projects/spring-amqp/commit/c2b81248181c25adaad3ed8e7f7dc6c262b0f511 after fixing conflicts in the test class.

Issue #1338, opened 2021-05-12
i can't find anything about callback after acked. why there's no before and after method in MessageListener?


bthulu:
**Affects Version(s):** \<Spring AMQP version>
all
---
## Enhancement
sometimes we used some state in memory map. needs to get it in `listener#onMessage`, and remove it after acked.
but now i can only remove it in onMessage, if the method exits and ack failed, the state lost.
next time, the same message comes to onMessage, but the state is gone.
so, i was wondering why there's no interceptor like before and after method in MessageListener, then we can do some inilialize before invoke onMessage and release some resources after acked ok or not.

`MessagePostProcessor` seems to be act as before interceptor, what about create a new `MessageAckProcessor`?

artembilan:
Let's see if `adviceChain` option on the Listener Container may help you!

https://docs.spring.io/spring-amqp/docs/current/reference/html/#containerAttributes

> An array of AOP Advice to apply to the listener execution. This can be used to apply additional cross-cutting concerns, such as automatic retry in the event of broker death. Note that simple re-connection after an AMQP error is handled by the `CachingConnectionFactory`, as long as the broker is still alive.

So, you write your own `MethodInterceptor` to deal with these args before and after calling `proceed()`: `Channel channel, Object data`.

bthulu:
it works, thanks

bthulu:
the aop advice works on target method `void onMessage(Message message)`, not after acked.
in some condition, i need to cancel consume queue after consume message acked.
the aop advice can not do this.

garyrussell:
You would need to add a reference to the listener container to the advice and stop the container (use the async `stop(() -> {})` to avoid a deadlock pause).

bthulu:
> You would need to add a reference to the listener container to the advice and stop the container (use the async `stop(() -> {})` to avoid a deadlock pause).

why stop container, i only need to cancel some not all.
i used one container to consume from thousands queues. but not all queues have the same behavior, some needs to cancel when run in some condition, the rest doesn't.
if every cancel by a stop, then nothing to do but stop, start, stop, start.

bthulu:
right now i used manual ack, then remove queueNames from container.
i wonder why not support ack callback. in a product environment, many things need to do after ack.
the advice chain works around on the method `void onMessage(Message message)`, but the method is defined by developer himself, and only one MessageListener per container, we can just write it in the method, the advice chain seems to worth nothing.
```
[...]

PR #1408, opened 2021-12-16
GH-1338: Add MessageAckListener

zysaaa:
See: https://github.com/spring-projects/spring-amqp/issues/1338.

A listener for message ack when using {@link AcknowledgeMode#AUTO} mode.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
@@ -87,6 +87,7 @@ import com.rabbitmq.utility.Utility;
  * @author Alex Panchenko
  * @author Johno Crawford
  * @author Ian Roberts
+ * @author Cao Weibo
  */
 public class BlockingQueueConsumer {
 
@@ -813,11 +814,11 @@ public class BlockingQueueConsumer {
 	/**
 	 * Perform a commit or message acknowledgement, as appropriate.
 	 * @param localTx Whether the channel is locally transacted.
+	 * @param messageAckListener MessageAckListener set on the message listener.
 	 * @return true if at least one delivery tag exists.
 	 * @throws IOException Any IOException.
 	 */
-	public boolean commitIfNecessary(boolean localTx) throws IOException {
-
+	public boolean commitIfNecessary(boolean localTx, MessageAckListener messageAckListener) throws IOException {
 		if (this.deliveryTags.isEmpty()) {
 			return false;
 		}
@@ -834,7 +835,14 @@ public class BlockingQueueConsumer {
 
 			if (ackRequired && (!this.transactional || isLocallyTransacted)) {
 				long deliveryTag = new ArrayList<Long>(this.deliveryTags).get(this.deliveryTags.size() - 1);
-				this.channel.basicAck(deliveryTag, true);
+				try {
+					this.channel.basicAck(deliveryTag, true);
+					notifyMessageAckListener(messageAckListener, true, deliveryTag, null);
+				}
+				catch (Exception e) {
+					logger.error("Error acking.", e);
+					notifyMessageAckListener(messageAckListener, false, deliveryTag, e);
+				}
 			}
 
 			if (isLocallyTransacted) {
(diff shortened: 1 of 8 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -1226,12 +1227,51 @@ private void rollback(long deliveryTag, Exception e) {
...
+		 * @param deliveryTag DeliveryTag of this ack.
+		 * @param multiple Whether multiple ack.
+		 * @throws Exception Occured when ack.
+		 */
+		private void sendAckWithNotify(long deliveryTag, boolean multiple) throws Exception {
+			try {
+				getChannel().basicAck(deliveryTag, multiple);
+				notifyMessageAckListener(getMessageAckListener(), true, deliveryTag, null);
+			}
+			catch (Exception e) {
```
garyrussell, line 1247: This can be `IOException` since `notifyMessageAckListener` does not throw any exception - we can then remove all those `throws ...` and `catch (Exception...` changes.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -1226,12 +1227,51 @@ private void rollback(long deliveryTag, Exception e) {
...
+		 * @param messageAckListener MessageAckListener set on the message listener.
+		 * @param success Whether ack succeeded.
+		 * @param deliveryTag The deliveryTag of ack.
+		 * @param cause If an exception occurs.
+		 */
+		private void notifyMessageAckListener(@Nullable MessageAckListener messageAckListener,
+											  boolean success,
+											  long deliveryTag,
+											  @Nullable Throwable cause) {
+			if (messageAckListener == null) {
```
garyrussell, line 1264: We can avoid this null check by making the default `(success, tag, ex) -> { }` and `Assert.notNull(...)` in the setter.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/MessageAckListener.java
```diff
@@ -0,0 +1,41 @@
...
+
+package org.springframework.amqp.rabbit.listener;
+
+import org.springframework.amqp.core.AcknowledgeMode;
+import org.springframework.lang.Nullable;
+
+/**
+ * A listener for message ack when using {@link AcknowledgeMode#AUTO}.
+ *
+ * @author Cao Weibo
```
garyrussell, line 25: `@since 2.4.3`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
```diff
@@ -813,11 +814,11 @@ public void rollbackOnExceptionIfNecessary(Throwable ex) {
 	/**
 	 * Perform a commit or message acknowledgement, as appropriate.
 	 * @param localTx Whether the channel is locally transacted.
+	 * @param messageAckListener MessageAckListener set on the message listener.
 	 * @return true if at least one delivery tag exists.
 	 * @throws IOException Any IOException.
 	 */
-	public boolean commitIfNecessary(boolean localTx) throws IOException {
-
+	public boolean commitIfNecessary(boolean localTx, MessageAckListener messageAckListener) throws IOException {
```
garyrussell, line 821: This is a public API; we can't change it in a point release; instead, inject the listener after creating the consumer.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
```diff
@@ -851,6 +859,28 @@ public boolean commitIfNecessary(boolean localTx) throws IOException {
...
+	 * @param messageAckListener MessageAckListener set on the message listener.
+	 * @param success Whether ack succeeded.
+	 * @param deliveryTag The deliveryTag of ack.
+	 * @param cause If an exception occurs.
+	 */
+	private void notifyMessageAckListener(@Nullable MessageAckListener messageAckListener,
+										  boolean success,
+										  long deliveryTag,
+										  @Nullable Throwable cause) {
+		if (messageAckListener == null) {
```
garyrussell, line 873: We can avoid this null check by making the default `(success, tag, ex) -> { }` and `Assert.notNull(...)` in the setter (which you have in the AMLC).

garyrussell (changes requested):
Thanks for the contribution! Sorry for the delay. Please address review comments and rebase to main.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
zysaaa: The reason why it was changed to ```Exception``` at that time was because ```channel#ack``` might throw other types of exceptions (not ```IOexception```) during the test.

garyrussell:
@zysaaa Thanks; sorry for the delay; merged as 76348cb584f0449448995e9b01777f9f748404ba with polishing 247c45ee7e47009a474819407c643f617c1fae7f - both back-ported to 2.4.x

Issue #1409, opened 2021-12-16
MessageConversionException causes all unacked messages to be nacked


peterlgh7:
## Bug report

When using asynchronous `@RabbitListener`(with a `ListenableFuture` as return type), a `MessageConversionException` causes all unacked messages to be nacked.

It can be reproduced with the following:
<details>
<summary>Listener.java</summary>

```java
package com.example.demo;

import org.springframework.amqp.core.MessageBuilder;
import org.springframework.amqp.core.QueueBuilder;
import org.springframework.amqp.rabbit.annotation.RabbitListener;
import org.springframework.amqp.rabbit.connection.CachingConnectionFactory;
import org.springframework.amqp.rabbit.core.RabbitAdmin;
import org.springframework.amqp.rabbit.core.RabbitTemplate;
import org.springframework.beans.factory.annotation.Autowired;
import org.springframework.boot.CommandLineRunner;
import org.springframework.context.ConfigurableApplicationContext;
import org.springframework.stereotype.Component;
import org.springframework.util.concurrent.ListenableFuture;
import org.springframework.util.concurrent.SettableListenableFuture;

import java.util.concurrent.Executors;
import java.util.concurrent.ScheduledExecutorService;
import java.util.concurrent.TimeUnit;

import javax.annotation.PostConstruct;

@Component
public class Listener implements CommandLineRunner {
  
  @Autowired
  private CachingConnectionFactory connectionFactory;
  
  @Autowired
  private RabbitTemplate rabbitTemplate;
  
  @Autowired
  private ConfigurableApplicationContext context;
  
  private ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor();
  
  @PostConstruct
  public void init() {
    RabbitAdmin rabbitAdmin = new RabbitAdmin(connectionFactory);
    connectionFactory.addConnectionListener(connection -> {
      rabbitAdmin.declareQueue(QueueBuilder.durable("test.dlq").build());
      rabbitAdmin.declareQueue(QueueBuilder.durable("test")
          .withArgument("x-dead-letter-exchange", "")
          .withArgument("x-dead-letter-routing-key", "test.dlq")
          .build());
    });
  }

  @Override
  public void run(String... args) throws Exception {
    rabbitTemplate.send("", "test", MessageBuilder.withBody("{\"text\": \"Test1\"}".getBytes()).build());
    rabbitTemplate.send("", "test", MessageBuilder.withBody("invalidJson".getBytes()).build());
[...]

PR #1411, opened 2021-12-19
GH-1409: Fix Nacks for Async Replies

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1409

Normally, when message has a fatal exception (such as message conversion)
`basicNack` with `multiple` true is used, to nack any previously unacked
messages (e.g. when using batch size to limit the ack traffic).
Even when using manual acks, fatal exceptions are nacked by the container
because the user does not have access to the message.

However, when using async replies, this has the side effect of nacking
unprocessed messages.

Detect whether async replies are being used and only nack individual
records that cause fatal exceptions.

Also, coerce the `AcknowledgeMode` to `MANUAL` for such listners.

Add a test for both containers; send a good message followed by a
bad one without actually completing the reply future.
After the exception occurs and the container is stopped, there should
be one messag in the queue.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
@@ -255,6 +255,8 @@ public abstract class AbstractMessageListenerContainer extends RabbitAccessor
 
 	private volatile boolean lazyLoad;
 
+	private boolean asyncReplies;
+
 	@Override
 	public void setApplicationEventPublisher(ApplicationEventPublisher applicationEventPublisher) {
 		this.applicationEventPublisher = applicationEventPublisher;
@@ -439,6 +441,7 @@ public abstract class AbstractMessageListenerContainer extends RabbitAccessor
 		this.messageListener = messageListener;
 		this.isBatchListener = messageListener instanceof BatchMessageListener
 				|| messageListener instanceof ChannelAwareBatchMessageListener;
+		this.asyncReplies = messageListener.isAsyncReplies();
 	}
 
 	/**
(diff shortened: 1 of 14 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -1220,6 +1225,10 @@ public void afterPropertiesSet() {
 		catch (IllegalStateException e) {
 			this.logger.debug("Could not enable micrometer timers", e);
 		}
+		if (this.isAsyncReplies() && !AcknowledgeMode.MANUAL.equals(this.acknowledgeMode)) {
+			this.acknowledgeMode = AcknowledgeMode.MANUAL;
+			logger.warn("Async replies require AcknowledgeMode.MANUAL, forced");
```
artembilan, line 1230: We don't do `this.` for method and I think this message should go to `info()`.
Doesn't look like it cause some problem for a target solution having it be reconfigured properly.
I would say it even easier for me just have one option to set - `asyncReplies` and have everything else reconfigured properly.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/BlockingQueueConsumer.java
```diff
@@ -777,8 +777,21 @@ public synchronized void stop() {
 	/**
 	 * Perform a rollback, handling rollback exceptions properly.
 	 * @param ex the thrown application exception or error
+	 * @deprecated in favor of {@link #rollbackOnExceptionIfNecessary(Throwable, long)}
```
artembilan, line 780: Why deprecate? Looks like you really have a heavy logic based on the `tag < 0`. So, a dedicated API with a default `-1` for `tag` doesn't look bad....

artembilan (changes requested):
I believe there are must be some docs for this feature.

Thanks

artembilan (changes requested):
LGTM, but:

>  add docs.

Please, push those docs for review 😄

artembilan (approved):

artembilan:
... and cherry-picked to `2.3.x` as: https://github.com/spring-projects/spring-amqp/commit/94fff33adfdf4969e8a90c171383ae6b481e4aaf after some minor conflict fixing.
And to `2.2.x` as https://github.com/spring-projects/spring-amqp/commit/aa3589c2d63b65afce358f16268e97660ef1d4b1 after more conflicts resolution.

Issue #1412, opened 2021-12-27
Receive without destination in RabbitMessagingTemplate tries to receive from defaultDestination which is a routing key


aspan:
When receiving messages with RabbitMessagingTemplate.receive() method it tries to receive messages from the default destination set in the RabbitMessagingTemplate which in my case is a routing key since I also use it for sending. I solved the issue by subclassing the RabbitMessagingTemplate with the following implementation of receive which uses the receive queue configured in the RabbitTemplate.

```
    @Override
    public Message<?> receive() {
        try {
            var amqpMessage = getRabbitTemplate().receive();
            return convertAmqpMessage(amqpMessage);
        } catch (RuntimeException ex) {
            throw convertAmqpException(ex);
        }
    }
```

garyrussell:
@aspan I resolved it but, unfortunately, I had to add an opt-in option to avoid it being a breaking change.

PR #1416, opened 2022-01-03
GH-1412: Fix Messaging Template

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1412

The use of the `defaultDestination` is not correct when the template is
used for both sends and receives because it is a queue name for receives
and a routing key for sends.

Add a new option to use the template's configured default receive queue
for receive only methods. False by default to avoid a breaking change;
it should be true by default in a future release.


artembilan:

Issue #1422, opened 2022-02-10
Broker-Named Queue Declaration is Broken


garyrussell:
Somehow, the container is getting a different `Queue` object to the one that is modified (with the actual name) by the `RabbitAdmin`.

https://stackoverflow.com/questions/71033589/spring-amqp-anonymous-queue-declaration/71035913#comment125628703_71035913

garyrussell:
The issue is the BPP never injects `Queue` objects into the endpoints, it grabs the queue name and always injects the name - this is too early since the admin updates the actual name after declaration.

PR #1423, opened 2022-02-10
GH-1422: @RabbitListener: Fix Broker-Named Queues

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1422

Broker-named queues did not work with `@RabbitListener` because the BPP
only passed the name of the queue bean, not the bean itself, into the
endpoint.

Support bean injection; fall back to the previous behavior if a mixture
of beans and names are encountered (the containers don't support both
types of configuration).

Add a note to the javadoc to indicate that broker-named queues are not
supported via `queuesToDeclare` and `bindings` properties; such queues
must be declared as discrete beans.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListener.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListener.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2020 the original author or authors.
+ * Copyright 2014-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -148,6 +148,8 @@ public @interface RabbitListener {
 	 * application context, the queue will be declared on the broker with default
 	 * binding (default exchange with the queue name as the routing key).
 	 * Mutually exclusive with {@link #bindings()} and {@link #queues()}.
+	 * NOTE: Broker-named queues cannot be declared this way, they must be defined
+	 * as beans (with an empty string for the name).
 	 * @return the queue(s) to declare.
 	 * @see org.springframework.amqp.rabbit.listener.MessageListenerContainer
 	 * @since 2.0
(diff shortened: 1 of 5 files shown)


garyrussell:
Forgot to indicate **cherry-pick to 2.4.x, 2.3.x**

artembilan (changes requested):
I see a restriction logic here in the `AbstractRabbitListenerEndpoint`:
```
if (queuesEmpty) {
				Collection<String> names = qNames;
				container.setQueueNames(names.toArray(new String[0]));
			}
			else {
				Collection<Queue> instances = getQueues();
				container.setQueues(instances.toArray(new Queue[0]));
			}
```

But at the same time I see that listener container has also this contract:
```
	/**
	 * Add queue(s) to this container's list of queues.
	 * @param queueNames The queue(s) to add.
	 */
	public void addQueueNames(String... queueNames) {
		Assert.notNull(queueNames, "'queueNames' cannot be null");
		Assert.noNullElements(queueNames, "'queueNames' cannot contain null elements");
		addQueues(Arrays.stream(queueNames)
				.map(Queue::new)
				.toArray(Queue[]::new));
	}

	/**
	 * Add queue(s) to this container's list of queues.
	 * @param queues The queue(s) to add.
	 */
	public void addQueues(Queue... queues) {
		Assert.notNull(queues, "'queues' cannot be null");
		Assert.noNullElements(queues, "'queues' cannot contain null elements");
		if (isRunning()) {
			for (Queue queue : queues) {
				Assert.hasText(queue.getName(), "Cannot add broker-named queues dynamically");
			}
		}
		this.queues.addAll(Arrays.asList(queues));
	}
```
Doesn't that mean that we can change `AbstractRabbitListenerEndpoint` and have mixed configuration with those annotations?

Thanks

garyrussell:
We could do that, but it would break a lot of tests; since this is a bit of a corner case, I wanted to make only minimal changes to support it. I doubt that someone will define a listener that listens to a broker-named queue as well as others. We can always revisit in future.

garyrussell:
I wasn't planning on doing so (aside from the comments I added to the javadocs); it's a bug. What do you recommend?

garyrussell:
https://docs.spring.io/spring-amqp/docs/current/reference/html/#containers-and-broker-named-queues

garyrussell:
Good point; will push a commit.

artembilan:

PR #1428, opened 2022-03-09
Fix PooledChannelConnectionFactory#publisherConnectionFactory configuration

LeonardoFerreiraa:
There is a little bug in `PooledChannelConnectionFactory` that this PR aims to fix. 

When we create a new instance of `PooledChannelConnectionFactory` it's also creates an another instance to populate `publisherConnectionFactory`, but to do this the method `PooledChannelConnectionFactory#setPublisherConnectionFactory`  is used and it sets `defaultPublisherFactory = false`, so all configurations after that will not affect the `publisherConnectionFactory` causing a misconfiguration for those that wants to use the `publisherConnectionFactory`;

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/PooledChannelConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/PooledChannelConnectionFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 the original author or authors.
+ * Copyright 2020-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


garyrussell:
Thanks; if you had run `./gradlew check` with before committing, the `updateCopyrights` task would have fixed the dates; please change to `2020-2022` and consider adding your name to the author list. You can just push another commit and we'll squash during the merge.

garyrussell (changes requested):
.

garyrussell (approved):

garyrussell:
Thanks for the contribution; merged as 909ba571bb096ed8acf77fed43e944f7b8ebb471 and cherry-picked to supported branches.

In future, please limit commit comment first line to 50 chars and code lines to 120 chars.

Thanks again.

LeonardoFerreiraa:
Thanks @garyrussell 🙂

Issue #1420, opened 2022-02-01
Determine charset from contentType if any


bjmi:
> There is no guarantee that the consumer uses Jackson to decode the content.
> 
> We can consider moving the charset to the contentType; please open a new issue for that.

_Originally posted by @garyrussell in https://github.com/spring-projects/spring-amqp/issues/1251#issuecomment-700028191_

Affected version: spring-amqp 2.4.2
Recently we were affected by this as the *charset* was provided in *contentType* header like e.g.  
`content-type: application/xml; charset=utf-16` and causes deserialization problems.
For the sending part: non spring-amqp consumers don't expect to find *charset* at *contentEncoding* header.

Please consider moving the *charset* to the *contentType*.

garyrussell:
This might also impact the compression post processors, including the `DelegatingDecompressingPostProcessor`.

PR #1429, opened 2022-03-09
GH-1251: Jackson2JsonMessageConverter Improvements

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1420

- detect and use `charset` in `contentType` when present
- allow Jackson to determine the decode `charset` via `ByteSourceJsonBootstrapper.detectEncoding()`
- allow configuration of the `MimeType` to use, which can include a `charset` parameter

**cherry-pick to main - will require what's new fix**

### DIFF
--- a/src/reference/asciidoc/amqp.adoc
+++ b/src/reference/asciidoc/amqp.adoc
@@ -3938,10 +3938,33 @@ public DefaultClassMapper classMapper() {
 Now, if the sending system sets the header to `thing1`, the converter creates a `Thing1` object, and so on.
 See the <<spring-rabbit-json>> sample application for a complete discussion about converting messages from non-Spring applications.
 
+Starting with version 2.4.3, the converter will not add a `contentEncoding` message property if the `supportedMediaType` has a `charset` parameter; this is also used for the encoding.
+A new method `setSupportedMediaType` has been added:
+
+====
+[source, java]
+----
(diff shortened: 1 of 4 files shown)


src/reference/asciidoc/amqp.adoc
```diff
@@ -3938,11 +3938,34 @@ public DefaultClassMapper classMapper() {
...
+converter.setSupportedContentType(MimeTypeUtils.parseMimeType(utf16));
+----
+====
+
 [[Jackson2JsonMessageConverter-from-message]]
 ====== Converting from a `Message`
 
 Inbound messages are converted to objects according to the type information added to headers by the sending system.
 
+Starting with version 2.4.3, if there is no `contentEncoding` message property, the converter will attempt to detect a `charset` parameter in the `contentEncoding` message property and use that.
```
artembilan, line 3957:
```suggestion
Starting with version 2.4.3, if there is no `contentEncoding` message property, the converter will attempt to detect a `charset` parameter in the `contentType` message property and use that.
```
?

artembilan (changes requested):

artembilan:
... and cherry-picked to `main` after fixing `What's New` doc conflict.

garyrussell:
Moved the what's new to the appendix and fixed a missing entry there.

cdf1a9cda

PR #1430, opened 2022-03-11
add checkConfirmsAndReturns in addTargetConnectionFactory

LeonardoFerreiraa:
There is a little inconsistency in `AbstractRoutingConnectionFactory` that this PR aims to fix.

```
@Bean
ConnectionFactory connectionFactory() {
    return new MyAbstractRoutingConnectionFactory(new HashMap<>() {{
        put("first", newPooledChannelConnectionFactory());
        put("second", newCachingConnectionFactory());
    }});
}

static class MyAbstractRoutingConnectionFactory extends AbstractRoutingConnectionFactory {

    MyAbstractRoutingConnectionFactory(final Map<Object, ConnectionFactory> factories) {
        factories.forEach(this::addTargetConnectionFactory);
    }

    @Override
    protected Object determineCurrentLookupKey() {
        return null;
    }

}
```

in this case, `AbstractRoutingConnectionFactory#afterPropertiesSet` throw a `java.lang.IllegalArgumentException: At least one target factory (or default) is required` because the propety `confirms` is null (the method `addTargetConnectionFactory` doesn't call `checkConfirmsAndReturns` that fill this field);

### DIFF
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/RoutingConnectionFactoryTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/RoutingConnectionFactoryTests.java
@@ -17,7 +17,7 @@
 package org.springframework.amqp.rabbit.connection;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assertions.fail;
+import static org.assertj.core.api.Assertions.assertThatNoException;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyBoolean;
 import static org.mockito.BDDMockito.given;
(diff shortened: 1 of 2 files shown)


spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/RoutingConnectionFactoryTests.java
```diff
@@ -323,4 +324,23 @@ protected synchronized void redeclareElementsIfNecessary() {
...
+			}
+		};
+
+		routingFactory.addTargetConnectionFactory("1", Mockito.mock(ConnectionFactory.class));
+
+		try {
+			routingFactory.afterPropertiesSet();
+		}
+		catch (Exception e) {
+			fail("afterPropertiesSet should not throw any exception after addTargetConnectionFactory", e);
```
artembilan, line 342: We prefer to use `Assertions.assertThatNoException()` for this type of assertions instead.

artembilan (changes requested):
Good catch!

Please, consider also to add your name to the `@author` list of the affected classes.

Thank you!

Issue #1433, opened 2022-03-15
DMLC - null in Monitor Thread Name


garyrussell:
We are using `getBean()` instead of `getListenerId()`.

Work around is to manually set a scheduler or copy the listener id into the bean name property.

https://stackoverflow.com/questions/71468360/which-component-is-creating-null-consumermonitor-threads/71483827#71483827

PR #1435, opened 2022-03-15
GH-1433: Fix DMLC MonitorThreadName

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1433

Used `beanName` instead of `listenerId` (which falls back to `beanName` if `null`).
Containers for annotations are not beans per se.


artembilan:

Issue #1436, opened 2022-03-16
SimpleMessageListenerContainer takes a long time to shutdown


mjaggard:
**In what version(s) of Spring AMQP are you seeing this issue?**
2.3.14

**Describe the bug**
When I have a lot of listeners, shutdown takes a long time because each listener is shut down and then awaited before starting the shutdown of the next one. These should be shut down asynchronously as `RabbitListenerEndpointRegistry` already implements `SmartLifecycle` to handle this style of working.

**To Reproduce**
Create a container with a large number of listeners which take some time to close, and then shut them down.

**Expected behavior**
The shutdown should be completed in the time taken for the longest wait.

A PR to fix this will be produced shortly.

bjmi:
> **In what version(s) of Spring AMQP are you seeing this issue?**
>  2.4.8

Spring AMQP 2.4.**3** was released since a few days. Do you actually mean 2.**3**.8?

mjaggard:
I have no clue where that version came from now. I've updated to the one we're using in our current project.

PR #1437, opened 2022-03-16
Allow shutdown to be started but waiting to be completed asynchronously

mjaggard:
A fix for #1436 by allowing the shutdown to be started in all listeners but the waiting to be done asynchronously.
Currently this starts a thread per wait which might not be optimal but represents and improvement on the current behaviour in most, if not all, situations. It would be better to use a single-thread executor but I wasn't sure about when to create that.


artembilan:
Please, sing the CLA, then we can start reviewing your contribution.

Thanks

mjaggard:
> Please, sing the CLA, then we can start reviewing your contribution.
> 
> Thanks

Thank you. I'm just getting the company version of the CLA through legal, hopefully they'll come back quickly.

mjaggard:
> Please, sing the CLA, then we can start reviewing your contribution.
> 
> Thanks

@artembilan CLA has been signed.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -605,59 +606,79 @@ private void waitForConsumersToStart(Set<AsyncMessageProcessingConsumer> process
...
 
+			if (callback != null) {
+				callback.run();
+			}
+		};
+		if (callback == null) {
+			awaitShutdown.run();
+		}
+		else {
+			new Thread(awaitShutdown).start();
```
artembilan, line 680: There is a `getTaskExecutor()` instead of this uncontrolled thread.

artembilan (changes requested):

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -16,46 +16,19 @@
...
-import org.springframework.amqp.AmqpConnectException;
-import org.springframework.amqp.AmqpException;
-import org.springframework.amqp.AmqpIOException;
-import org.springframework.amqp.AmqpIllegalStateException;
-import org.springframework.amqp.AmqpRejectAndDontRequeueException;
-import org.springframework.amqp.ImmediateAcknowledgeAmqpException;
+import com.rabbitmq.client.Channel;
+import com.rabbitmq.client.PossibleAuthenticationFailureException;
+import com.rabbitmq.client.ShutdownSignalException;
+import org.springframework.amqp.*;
```
artembilan, line 22: We don't allow asterisk imports at all.
Please, revert such a change.
You can configure your IDE to switch to asterisk somewhere after big enough number.

artembilan (changes requested):

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
mjaggard: @artembilan Fixed

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -605,59 +606,79 @@ private void waitForConsumersToStart(Set<AsyncMessageProcessingConsumer> process
...
-			this.consumers = null;
-			this.cancellationLock.deactivate();
-		}
+		Runnable awaitShutdown = () -> {
+			logger.info("Waiting for workers to finish.");
+			try {
+				boolean finished = this.cancellationLock.await(getShutdownTimeout(), TimeUnit.MILLISECONDS);
+				if (finished) {
+					logger.info("Successfully waited for workers to finish.");
+				} else {
```
artembilan, line 650: The `else` has to be from a new line.

artembilan (changes requested):

garyrussell:
Unfortunately, this missed todays 2.3.x and 2.4.x releases (it will be in the 3.0.0-M2 milestone today).

However, does it really need to be back-ported? A subclass could call `super.doShutdown()` async.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
mjaggard: Fixed

mjaggard:
> Reverted; this breaks `EnableRabbitTests.testProperShutdownOnException()` - the `running` boolean is no longer reset.

Fixed

artembilan (approved):

garyrussell:
@mjaggard Thanks for the contribution; merged as ff67612ae after rebase; cherry picked to 2.4.x as 55d1289a3 and 2.3.x as 3cb1a744e

Issue #1434, opened 2022-03-15
Multiples connection factories with unequal isPublisherConfirms in RabbitTemplate


garyrussell:
### Discussed in https://github.com/spring-projects/spring-amqp/discussions/1431

<div type='discussions-op-text'>

<sup>Originally posted by **LeonardoFerreiraa** March 11, 2022</sup>
Hello guys, I have a case here that I want to use 2 connection factories in the same `RabbitTemplate`. One `CachingConnectionFactory` with `ConfirmType = Correlated` and another with type `PooledCachingConnectionFactory`. I need this because in some cases I want to be sure of the message delivery and in others, I don't need this.

I found in the `RabbitTemplate` the property `sendConnectionFactorySelectorExpression` that allows me to have more than one `ConnectionFactory` based on a SPEL, but for this, I need an instance of `AbstractRoutingConnectionFactory` that has a validation avoiding different `isPublisherConfirms` and `isPublisherReturns` values between the factories.

If `RabbitTemplate` didn't depend on `AbstractRoutingConnectionFactory` but `RoutingConnectionFactory`, and some more small changes in `confirmsOrReturnsCapable` property, it would be possible;

What do you think about it?
</div>

PR #1438, opened 2022-03-19
GH-1434 allowing to rabbit template have multiple connection factories with not same confirms and returns flags

LeonardoFerreiraa:
This PR aims to add an ability to `RabbitTemplate` use multiple connections not necessarily with the same `confirms` and `returns` values;

It will be useful in cases where some messages need to be extremely reliable and others don't.

Discussed in #1431;
Solves #1434;


garyrussell:
I was not planning to include this in today's release(s).

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/PooledChannelConnectionFactory.java
```diff
@@ -221,6 +221,8 @@ private Channel createProxy(Channel channel, boolean transacted) {
 							return channel.confirmSelect();
 						case "isConfirmSelected":
 							return confirmSelected.get();
+						case "isPublisherConfirms":
+							return false;
```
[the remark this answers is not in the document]
LeonardoFerreiraa: @artembilan it was. there is no way to change the value of `isPublisherConfirms` in the `PooledChannelConnectionFactory`;
But seeing it now, I guess it's better to use the same logic of `simplePublisherConfirms`. 

I will change it

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/DelegatingConnectionFactory.java
```diff
@@ -0,0 +1,96 @@
...
+
+import org.springframework.amqp.AmqpException;
+import org.springframework.lang.NonNull;
+import org.springframework.util.Assert;
+
+/**
+ * a {@link ConnectionFactory} that delegates all calls to another ConnectionFactory.
+ * @author Leonardo Ferreira
+ */
+public class DelegatingConnectionFactory implements ConnectionFactory {
```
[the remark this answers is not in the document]
LeonardoFerreiraa: Yes, it's just a void delegate that anyone can extend and override just the methods that need to be changed.
I meant to use this in production just like `FlexibleRoutingConnectionFactory` is used in tests here.

but if you think that it's not useful to the framework to have this, I don't see any problem removing it.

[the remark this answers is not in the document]
artembilan: No problem. Let's see what @garyrussell thinks on the matter when he's done with the releases this week!

I have some tomorrow as well. So, bear with us how slow we review your PRs.

Thank you anyway!

garyrussell:
Just a heads-up; this might (will) be complicated by https://github.com/spring-projects/spring-amqp/issues/1439

Currently, if there is a confirm callback in the template and the channel is a `PublisherCallbackChannel`, it causes a memory leak if the factory has returns, but not confirms, enabled because we call `publisherCallbackChannel.addPendingConfirm()` even though the factory doesn't have confirms enabled.

In this hybrid environment, we might have some factories with confirms and returns, some with none (which is ok) and some with just returns (not ok).

We probably need to, somehow, re-check the factory config before calling `publisherCallbackChannel.addPendingConfirm()` or add some logic `publisherCallbackChannel.addPendingConfirm()` to make it a no-op (e.g. if the `nextPublishSeqNo` is zero - which is the case when confirms are not selected).

garyrussell:
I found a solution that will not affect this PR https://github.com/spring-projects/spring-amqp/pull/1440

LeonardoFerreiraa:
nice.
great solution, btw

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/DelegatingConnectionFactory.java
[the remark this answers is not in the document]
LeonardoFerreiraa: @artembilan which solution? #1440?
I guess it's not related.

garyrussell (changes requested):

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/PooledChannelConnectionFactory.java
```diff
@@ -137,7 +137,8 @@ public synchronized Connection createConnection() throws AmqpException {
 		if (this.connection == null || !this.connection.isOpen()) {
 			Connection bareConnection = createBareConnection(); // NOSONAR - see destroy()
 			this.connection = new ConnectionWrapper(bareConnection.getDelegate(), getCloseTimeout(), // NOSONAR
-					this.simplePublisherConfirms, this.poolConfigurer, getChannelListener()); // NOSONAR
+					this.simplePublisherConfirms, this.poolConfigurer, getChannelListener(), // NOSONAR
+					this.isPublisherConfirms()); // NOSONAR
```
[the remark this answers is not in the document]
LeonardoFerreiraa: @artembilan are you ok with this?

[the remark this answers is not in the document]
artembilan: Gary knows better. So, let's make it `false` by default!

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractRoutingConnectionFactory.java
```diff
@@ -229,6 +234,10 @@ public ConnectionFactory getTargetConnectionFactory(Object key) {
 		return this.targetConnectionFactories.get(key);
 	}
 
+	public void setConsistentConfirmsReturns(boolean consistentConfirmsReturns) {
```
artembilan, line 237: Such a `public` API deserves a JavaDoc and `@since`

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitTemplateRoutingConnectionFactoryIntegrationTests.java
```diff
@@ -0,0 +1,135 @@
...
+import org.springframework.amqp.rabbit.connection.ConnectionFactory;
+import org.springframework.amqp.rabbit.connection.CorrelationData;
+import org.springframework.amqp.rabbit.connection.PooledChannelConnectionFactory;
+import org.springframework.amqp.rabbit.junit.BrokerTestUtils;
+import org.springframework.amqp.rabbit.junit.RabbitAvailable;
+import org.springframework.expression.Expression;
+import org.springframework.expression.spel.standard.SpelExpressionParser;
+
+/**
+ * @author Leonardo Ferreira
```
artembilan, line 46: `@since 2.4.4`

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitTemplateRoutingConnectionFactoryIntegrationTests.java
```diff
@@ -0,0 +1,135 @@
...
+ */
+@RabbitAvailable(queues = RabbitTemplateRoutingConnectionFactoryIntegrationTests.ROUTE)
+class RabbitTemplateRoutingConnectionFactoryIntegrationTests {
+
+	public static final String ROUTE = "test.queue.RabbitTemplateRoutingConnectionFactoryIntegrationTests";
+
+	private RabbitTemplate rabbitTemplate;
+
+	@BeforeEach
```
artembilan, line 55: Cannot this setting be done only once for all the tests below via `@BeforeAll`?

artembilan (changes requested):
I wonder if we also need something to mention in the docs for these changes...

artembilan (changes requested):
I'm OK with this so far.
Unless you have a thought adding something to the docs about `RabbitTemplate`...

LeonardoFerreiraa:
@artembilan I can add, no problems.
do you have any suggestions? 
like, adding an example of how to use or just mentioning that this feature exists.

artembilan:
I would that sample is too simple to worth it adding to the docs, but this new option on the routing CF might be good to mention.
See this docs https://github.com/spring-projects/spring-amqp/blob/main/src/reference/asciidoc/amqp.adoc#routing-connection-factory.
That's an `amqp.adoc` file in the project. Looks like it has something about publisher confirms and returns.
Won't hurt to revise it if your change affects it somehow.

Thanks

src/reference/asciidoc/amqp.adoc
```diff
@@ -696,6 +696,48 @@ For example, with lookup key qualifier `thing1` and a container listening to que
...
+
+	final Expression sendExpression = new SpelExpressionParser().parseExpression(
+			"messageProperties.headers['x-use-publisher-confirms'] ?: false");
+	rabbitTemplate.setSendConnectionFactorySelectorExpression(sendExpression);
+}
+----
+====
+
+This way messages with the header `x-use-publisher-confirms: true` will be sent through the caching connection and you can ensure the message delivery.
+See <<cf-pub-conf-ret>>
```
artembilan, line 739: Looks like the sentence is not finished: `for more information.` ?

artembilan (changes requested):

src/reference/asciidoc/amqp.adoc
```diff
@@ -696,6 +696,48 @@ For example, with lookup key qualifier `thing1` and a container listening to que
 IMPORTANT: The target (and default, if provided) connection factories must have the same settings for publisher confirms and returns.
 See <<cf-pub-conf-ret>>.
 
+Starting with version 2.4.4, this validation can be disabled. If you have a case that the values between confirms and returns need to be unequal,
+you can use `AbstractRoutingConnectionFactory#setConsistentConfirmsReturns` to turn of the validation.
```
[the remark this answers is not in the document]
LeonardoFerreiraa: hahah, no problem

artembilan (approved):
LGTM

@garyrussell ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/PooledChannelConnectionFactory.java
```diff
@@ -221,6 +221,8 @@ private Channel createProxy(Channel channel, boolean transacted) {
 							return channel.confirmSelect();
 						case "isConfirmSelected":
 							return confirmSelected.get();
+						case "isPublisherConfirms":
```
garyrussell, line 224: I don't think we need changes in this factory at all; the interface `default` method returns `false`.

Same thing with the `ThreadChannel...`

garyrussell (changes requested):
Just one (two) nitpick.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/PooledChannelConnectionFactory.java
LeonardoFerreiraa: @garyrussell I guessed the same.

but when I remove it, I get an `org.springframework.aop.AopInvocationException: AOP configuration seems to be invalid: tried calling method [public default boolean org.springframework.amqp.rabbit.connection.ChannelProxy.isPublisherConfirms()] on target [AMQChannel(amqp://guest@127.0.0.1:5672/,1)]; nested exception is java.lang.IllegalArgumentException: object is not an instance of declaring class`

or if I put only the `advisor.addMethodName("isPublisherConfirms");` and remove the `case "isPlublisherConfirms"`, I get a `org.springframework.aop.AopInvocationException: Null return value from advice does not match primitive return type for: public default boolean org.springframework.amqp.rabbit.connection.ChannelProxy.isPublisherConfirms()`

:/

LeonardoFerreiraa: there is some spring-aop configuration I'm missing?

garyrussell: Oh; ok; I guess it's a runtime JVM thing - the JVM has to check to see if the implementation has overridden the default method and, only then, call the default method if there is none; with a proxy he can't tell.

Sorry for the noise.

garyrussell (approved):

garyrussell:
@artembilan Do you still feel this should not be back-ported to 2.4.x?

If so, what changes do you think would be needed; the 3.0 timeline is still a long way out.

artembilan:
No, now it is OK for back-porting to `2.4.x`
Thanks

garyrussell:
Thanks; merged as eccdb47155413fa8800849086810f965a6836c0d and cherry-picked to 2.4.x as 7421de45a

Issue #1439, opened 2022-03-22
Detect Misconfiguration of Publisher Confirms


garyrussell:
See https://stackoverflow.com/questions/71482356/memory-leak-outgoing-messages-rabbitmq/71578881#71578881

If the `RabbitTemplate` is configured with a confirm listener, ensure that the connection factory is configured for confirms.

PR #1440, opened 2022-03-23
GH-1439: Fix Memory Leak with Misconfiguration

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1439

Do not store pending confirms/returns if `RabbitTemplate` has confirms
enabled but the factory does not support confirms.

Suggest review with `?w=1`


artembilan:

Issue #1441, opened 2022-03-23
RabbitListener does not include inferrer value when injecting header object


nightknight77:
Version:
Spring Boot 2.6 - spring starter

############

**Description:**

The documentation of the RabbitListener annotation suggests multiple ways on how to retrieve the message headers. If you use MessageHeaders and inject it as an argument for getting access to all headers, then the **inferrer** header (which in my understanding contains the consumer model) is **not** provided with a value anymore and message converter implementations like the Jackson2JsonMessageConverter are then using the __TYPE_ID__ property and convert it into the given domain model. That property is provided by the producer and usually contains the fully qualified package name, so the converter on the consumer side looks for the domain model of the producer. If the consumer has its model not in the exact same package structure, it throws an ClassNotFound exception. This is confusing because if you cut the MessageHeader object and instead inject a simple Map<String, Object> and annotate that with @Headers (also suggested in the docs) then it works again and the inferrer value is provided. So my guess is this behaviour is not intended. 

**Steps to reproduce the behavior:**
1.  Use two models for json conversion. The package structure must differ.
2.  Add next to your model an additional parameter of type MessageHeaders. 
3.  Receive a message on consumer side.

**Actual behavior**
ClassnotFound exception is thrown and converter cannot convert into inferrer type.

**Expected behavior**
Inferrer type is provided and converts successfully into consumers model.

**Sample**
A short sample will be provided soon.

garyrussell:
Thanks for reporting; reproduced:

```java
@SpringBootApplication
public class Rgh1441Application {

	public static void main(String[] args) {
		SpringApplication.run(Rgh1441Application.class, args);
	}

	@RabbitListener(queues = "foo")
	public void listen1(Foo foo, @Headers Map<String, Object> headers) {
		System.out.println("1: " + foo);
	}

	@RabbitListener(queues = "bar")
	public void listen2(Foo foo, MessageHeaders headers) {
		System.out.println("2: " + foo);
	}

	@Bean
	ApplicationRunner runner(RabbitTemplate template, AbstractRabbitListenerContainerFactory<?> factory) {
		factory.setMessageConverter(converter());
		return args -> {
			template.convertAndSend("foo", "{\"bar\":\"baz\"}", msg -> {
				msg.getMessageProperties().setContentType("application/json");
				return msg;
			});
			template.convertAndSend("bar", "{\"bar\":\"baz\"}", msg -> {
				msg.getMessageProperties().setContentType("application/json");
[...]

PR #1442, opened 2022-03-23
GH-1441: Fix Payload Detection with MessageHeaders

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1441

Previously, `MessageHeaders` had to be annotated with `@Headers` so that
it was ignored during payload parameter resolution; otherwise it caused
ambiguity.

Ignore `MessageHeaders` even when not so annotated.

Also fix some tests that were checking the same topic and count down latch so were
unconditionally passing.

Change one of those tests to verify the fix.


artembilan:

PR #1445, opened 2022-03-29
GH-1443: Pull CCF.resetConnection() to CF

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1443


artembilan:

Issue #1452, opened 2022-04-20
The BlockingQueueConsumer.checkMissingQueues() should take care properly about connection


artembilan:
There is a dead code in the method:
```
Connection connection = null; // NOSONAR - RabbitUtils                               
Channel channelForCheck = null;                                                      
try {                                                                                
	channelForCheck = this.connectionFactory.createConnection().createChannel(false);
	channelForCheck.queueDeclarePassive(queueToCheck);                               
	if (logger.isInfoEnabled()) {                                                    
		logger.info("Queue '" + queueToCheck + "' is now available");                
	}                                                                                
}                                                                                    
catch (IOException e) {                                                              
	available = false;                                                               
	if (logger.isWarnEnabled()) {                                                    
		logger.warn("Queue '" + queueToCheck + "' is still not available");          
	}                                                                                
}                                                                                    
finally {                                                                            
	RabbitUtils.closeChannel(channelForCheck);                                       
	RabbitUtils.closeConnection(connection);                                         
}
```
That `connection` is never initialized.
It looks like the `this.connectionFactory.createConnection()` result has to be stored in that variable.

PR #1453, opened 2022-04-20
GH-1452: Close Connection in checkMissingQueues

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1452

While not a problem with connection factories provided by the framework
(which all ignore `close()`) the code should properly close the connection
after use.


artembilan:

Issue #1455, opened 2022-04-26
Allow StreamRabbitListenerContainerFactory message listeners to retry messages in a standard manner


daniel-bray-sonalake:
**Expected Behavior**

I spotted this when looking into this: [Can StreamRabbitListenerContainerFactory message listeners retry messages?](https://github.com/spring-projects/spring-amqp/discussions/1454)

Retries for deliveries in the classic listener containers is implemented using the advice chain (with a retry interceptor https://docs.spring.io/spring-amqp/docs/current/reference/html/#retry).

So there's no standard way to handle retries as part of the stream container factory, that do exist on the [SimpleMessageListenerContainer](https://docs.spring.io/spring-amqp/api/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.html)

**Current Behavior**

There is currently no advice chain support in the StreamListenerContainer.

**Context**

There are workarounds - suggested in the [Can StreamRabbitListenerContainerFactory message listeners retry messages?](https://github.com/spring-projects/spring-amqp/discussions/1454) - but it pushes the retry logic out into all the `@RabbitListener`s, instead of keeping it central.

PR #1456, opened 2022-04-26
GH-1455: AdviceChain on Stream Listener Container

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1455

Add an advice chain to the stream listener container and its factory.
Add a `StreamMessageRecoverer` for native stream messages.
Add a retry interceptor to work with native stream messages.

**cherry-pick to 2.4.x**

(What's New? will conflict so I left it out - will polish it in both branches after merge).

### DIFF
--- a/spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/listener/StreamListenerContainer.java
+++ b/spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/listener/StreamListenerContainer.java
@@ -165,6 +165,7 @@ public class StreamListenerContainer implements MessageListenerContainer, BeanNa
 	/**
 	 * Set an advice chain to apply to the listener.
 	 * @param advices the advice chain.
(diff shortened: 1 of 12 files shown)


spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/listener/StreamListenerContainer.java
```diff
@@ -154,6 +161,17 @@ public void setAutoStartup(boolean autoStart) {
 	public boolean isAutoStartup() {
 		return this.autoStartup;
 	}
+
+	/**
+	 * Set an advice chain to apply to the listener.
+	 * @param advices the advice chain.
```
artembilan, line 167: `@since 2.4.5` ?

artembilan (changes requested):

artembilan:

Issue #1459, opened 2022-05-17
Enhanced configurability for micrometer utilization in AbstractMessageListenerContainer


jo-george:
**Expected Behavior**

It should be possible to configure the default behavior and disable utilization of micrometer in context of the AMQP integration, particularly the `AbstractMessageListenerContainer`, before it even attempts to establish the micrometer support via creating a `MicrometerHolder`. A convenient way would be the possibility to configure the attribute `AbstractMessageListenerContainer.micrometerEnabled` and set it to false during the initialization, if needed.

If the case, that multiple instances of `MeterRegistry`s are available in the application context, is actually relevant and not just a theoretical case, it should be possible to specify the particular `MeterRegistry` (instantiable) subclass via configuration.

**Current Behavior**

Currently, the utilization of micrometer by the `AbstractMessageListenerContainer` is enabled by default when micrometer (or rather `io.micrometer.core.instrument.MeterRegistry`) is found on the classpath. So, when micrometer is used by any part of the application, it is also enabled by default for the *MessageListenerContainers, even if not wanted or not needed.

Additionally, during instantiation of the `MicrometerHolder` there is an assumption that only a single instance of `MeterRegistry` can be present in the application context. Technically, it might be possible to have more than one `MeterRegistry` and in that case it is not possible to specify, which particular `MeterRegistry` should be used and the message "No micrometer registry present" would also be misleading in that case. 

**Context**

In our application, we're collecting some business related metrics using an `InfluxMeterRegistry` which should not automatically be used in context of our RabbitMQ integration via spring-amqp. The most convenient way, at least for us, to avoid this utilization would be to make `AbstractMessageListenerContainer.micrometerEnabled` configurable, so we could just configure it to be false.

Setting the `AbstractMessageListenerContainer.micrometerEnabled` attribute to false at the right moment during bean initialization via customized post-processing seems not to be possible, at least in our case, where the `SimpleMessageListenerContainer` is created and also initialized (including afterPropertiesSet()) as part of the call `invokeInitMethods(beanName, wrappedBean, mbd);` inside `AbstractAutowireCapableBeanFactory.initializeBean(...)` while the post-processings are applied right before and right after this whole step. Actually, a post-processing between creation and initialization, probably as a part of the `ListenerContainerFactoryBean` would be needed to programmatically set the `micrometerEnabled` flag which seems to be currently not in place for the instantiation of `SimpleMessageListenerContainer`s.

garyrussell:
[...]

PR #1460, opened 2022-05-17
GH-1459: Improve MeterRegistry Discovery

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1459

**cherry-pick to 2.4.x**

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/MicrometerHolder.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/MicrometerHolder.java
@@ -1,0 +1,142 @@
+/*
+ * Copyright 2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.amqp.rabbit.listener;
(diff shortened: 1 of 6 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/MicrometerHolder.java
```diff
@@ -0,0 +1,142 @@
...
+	private final Map<String, String> tags;
+
+	private final String listenerId;
+
+	MicrometerHolder(@Nullable ApplicationContext context, String listenerId, Map<String, String> tags) {
+		if (context == null) {
+			throw new IllegalStateException("No micrometer registry present");
+		}
+		Map<String, MeterRegistry> registries = context.getBeansOfType(MeterRegistry.class, false, false);
+		registries = filterRegistries(registries, context);
```
artembilan, line 58: I wonder why does `ObjectProvider.getIfUnique()` not work for us here:
```

	/**
	 * Return an instance (possibly shared or independent) of the object
	 * managed by this factory.
	 * @return an instance of the bean, or {@code null} if not available or
	 * not unique (i.e. multiple candidates found with none marked as primary)
	 * @throws BeansException in case of creation errors
	 * @see #getObject()
	 */
	@Nullable
	T getIfUnique() throws BeansException;
```
?

artembilan (changes requested):

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/MicrometerHolder.java
garyrussell: Where do we get one from?

artembilan: `ApplicationContext.getBeanProvider(Class<T> requiredType)`

garyrussell: Oh, I didn't know about that; I've only seen them in the context of auto wiring. Cool.

artembilan:

Issue #1465, opened 2022-06-16
RabbitMQ Super Streams with Single Active Consumer


garyrussell:
Add support for super streams.

artembilan:
I'm reopening this one since it fails on CI https://build.spring.io/chain/result/viewChainResult.action?planKey=AMQP-MAIN&buildNumber=1656 and locally for me as well.
Somehow this super stream is not consumed at all.
When I make this change to the test:
```
			container.setQueueNames("ss.sac.test-0");
//			container.superStream("ss.sac.test", "test");
```
I got some interactivity:
```
(Body:'foo' MessageProperties [headers={}, contentType=text/plain, contentEncoding=UTF-8, contentLength=0, deliveryMode=PERSISTENT, expiration=0, priority=0, deliveryTag=0])
(Body:'foo' MessageProperties [headers={}, contentType=text/plain, contentEncoding=UTF-8, contentLength=0, deliveryMode=PERSISTENT, expiration=0, priority=0, deliveryTag=0])
(Body:'foo' MessageProperties [headers={}, contentType=text/plain, contentEncoding=UTF-8, contentLength=0, deliveryMode=PERSISTENT, expiration=0, priority=0, deliveryTag=0])
```
Which means that setup is OK, but super stream feature doesn't work.
Perhaps something else has to be configured in the environment...

I'm disabling `SuperStreamSACTests` for time being.

Thank you!

acogoluegnes:
@artembilan The test passes on my machine with

```shell
./gradlew :spring-rabbit-stream:test --tests "org.springframework.rabbit.stream.listener.SuperStreamSACTests"
```

I suspect you're having an old Docker image in your cache, try to delete it and run the test again:

```shell
docker rmi pivotalrabbitmq/rabbitmq-stream
```

It may be the same on CI.

artembilan:
Thank you, @acogoluegnes !

That really works.

I had to do it manually though:
```
docker rmi pivotalrabbitmq/rabbitmq-stream
docker logout
docker pull pivotalrabbitmq/rabbitmq-stream
```

Because Testcontainers somehow cannot pull the image for me - `invalid credentials` 😢 .

Will ask Trevor to remove that image from our CI Docker.

eddumelendez:
@artembilan maybe time to upgrade from testcontainers `1.15.3` to `1.17.3`. IIRC the issue you are having should be solved by now.

artembilan:
Thank you, @eddumelendez !

[...]

PR #1466, opened 2022-06-16
GH-1465: Part 1: Provision Super Streams over AMQP

garyrussell:
See https://github.com/spring-projects/spring-amqp/issues/1465

Inspired by https://github.com/rabbitmq/rabbitmq-stream-java-client/blob/29b1a3eca72c8e3f719db4fd9fafdf503f37ea20/src/test/java/com/rabbitmq/stream/impl/TestUtils.java#L244-L276

cc/ @acogoluegnes

### DIFF
--- a/spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/config/SuperStream.java
+++ b/spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/config/SuperStream.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/config/SuperStream.java
```diff
@@ -0,0 +1,63 @@
...
+		String[] rks = IntStream.range(0, partitions).mapToObj(String::valueOf).toArray(String[]::new);
+		declarables.add(new DirectExchange(name, true, false, Map.of("x-super-stream", true)));
+		for (int i = 0; i < partitions; i++) {
+			String rk = rks[i];
+			Queue q = new Queue(name + "-" + rk, true, false, false, Map.of("x-queue-type", "stream"));
+			declarables.add(q);
+			declarables.add(new Binding(q.getName(), DestinationType.QUEUE, name, rk,
+					Map.of("x-stream-partition-order", i)));
+		}
+		return declarables.toArray(new Declarable[0]);
```
artembilan, line 60: Looks like there is `Declarables(Collection<? extends Declarable> declarables)` ctor to avoid this `toArray()` conversion.

artembilan (changes requested):
Any docs, please?

Issue #1467, opened 2022-06-28
Add native-image support for spring-rabbit


mhalbritter:
When running the `rabbitmq` sample from spring-native in the `sb-3.0.x` branch (don't let the name fool you, it's NOT using spring native, but uses the AOT support from Spring Framework 6 and Spring Boot 3), the native image fails with this exception:

```
com.oracle.svm.core.jdk.UnsupportedFeatureError: Proxy class defined by interfaces [interface org.springframework.amqp.rabbit.connection.ChannelProxy] not found. Generating proxy classes at runtime is not supported. Proxy classes need to be defined at image build time by specifying the list of interfaces that they implement. To define proxy classes use -H:DynamicProxyConfigurationFiles=<comma-separated-config-files> and -H:DynamicProxyConfigurationResources=<comma-separated-config-resources> options.
        at com.oracle.svm.core.util.VMError.unsupportedFeature(VMError.java:89) ~[na:na]
        at com.oracle.svm.reflect.proxy.DynamicProxySupport.getProxyClass(DynamicProxySupport.java:158) ~[na:na]
        at java.lang.reflect.Proxy.getProxyConstructor(Proxy.java:48) ~[rabbitmq:na]
        at java.lang.reflect.Proxy.newProxyInstance(Proxy.java:1037) ~[rabbitmq:na]
        at org.springframework.amqp.rabbit.connection.CachingConnectionFactory.getCachedChannelProxy(CachingConnectionFactory.java:662) ~[rabbitmq:3.0.0-M3]
        at org.springframework.amqp.rabbit.connection.CachingConnectionFactory.getChannel(CachingConnectionFactory.java:540) ~[rabbitmq:3.0.0-M3]
        at org.springframework.amqp.rabbit.connection.CachingConnectionFactory$ChannelCachingConnectionProxy.createChannel(CachingConnectionFactory.java:1418) ~[na:na]
        at org.springframework.amqp.rabbit.core.RabbitTemplate.doExecute(RabbitTemplate.java:2179) ~[rabbitmq:3.0.0-M3]
        at org.springframework.amqp.rabbit.core.RabbitTemplate.execute(RabbitTemplate.java:2146) ~[rabbitmq:3.0.0-M3]
        at org.springframework.amqp.rabbit.core.RabbitTemplate.execute(RabbitTemplate.java:2126) ~[rabbitmq:3.0.0-M3]
[...]

PR #1471, opened 2022-07-19
GH-1467: Support Spring AOT for Native Image

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1467

Migrate proxy hints from spring-native.

Also add a proxy hint for `ChannelProxy` when it is also a `PublisherCallbackChannel`.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/aot/RabbitRuntimeHintsRegistrar.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/aot/RabbitRuntimeHintsRegistrar.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.amqp.rabbit.aot;
+
+import org.springframework.amqp.rabbit.annotation.RabbitListener;
+import org.springframework.amqp.rabbit.connection.ChannelProxy;
+import org.springframework.amqp.rabbit.connection.PublisherCallbackChannel;
+import org.springframework.aop.SpringProxy;
+import org.springframework.aop.framework.Advised;
+import org.springframework.aot.hint.ProxyHints;
+import org.springframework.aot.hint.RuntimeHints;
+import org.springframework.aot.hint.RuntimeHintsRegistrar;
+import org.springframework.aot.hint.TypeReference;
+import org.springframework.aot.hint.support.RuntimeHintsUtils;
+import org.springframework.core.DecoratingProxy;
+import org.springframework.lang.Nullable;
+
+/**
+ * {@link RuntimeHintsRegistrar} for spring-rabbit.
+ *
+ * @author Gary Russell
+ * @since 3.0
+ *
+ */
+public class RabbitRuntimeHintsRegistrar implements RuntimeHintsRegistrar {
+
+	@Override
+	public void registerHints(RuntimeHints hints, @Nullable	ClassLoader classLoader) {
+		RuntimeHintsUtils.registerAnnotation(hints, RabbitListener.class);
+		ProxyHints proxyHints = hints.proxies();
+		proxyHints.registerJdkProxy(ChannelProxy.class);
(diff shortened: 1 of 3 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/aot/RabbitRuntimeHintsRegistrar.java
```diff
@@ -0,0 +1,51 @@
...
+ * @author Gary Russell
+ * @since 3.0
+ *
+ */
+public class RabbitRuntimeHintsRegistrar implements RuntimeHintsRegistrar {
+
+	@Override
+	public void registerHints(RuntimeHints hints, @Nullable	ClassLoader classLoader) {
+		ProxyHints proxyHints = hints.proxies();
+		proxyHints.registerJdkProxy(ChannelProxy.class);
```
artembilan, line 43: Gary, don't we need also these interfaces for each proxy: `SpringProxy.class, Advised.class, DecoratingProxy.class` ?
See my PR in SI: https://github.com/spring-projects/spring-integration/pull/3832/files#diff-b471878878aa8838786d4a93b19f8014d497db6ace582572f834da814720e01bR150.

Also see this new API in SF `6.0`: 
```
runtimeHints.proxies().registerJdkProxy(AopProxyUtils.completeJdkProxyInterfaces(proxyInterfaces))
```

artembilan (changes requested):

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/aot/RabbitRuntimeHintsRegistrar.java
garyrussell: I just replicated what we have in spring-native (which works).

https://github.com/spring-projects-experimental/spring-native/blob/2d1fc19c3833e6ce2bd4d342504b180a88407572/spring-native-configuration/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitHints.java#L28-L35

I will find out soon because the compile problem has now been resolved.

artembilan: I see. Our logic for this proxy is not about Spring. Agreed:
```
		if (ConfirmType.CORRELATED.equals(this.confirmType) || this.publisherReturns) {
			interfaces = new Class<?>[] { ChannelProxy.class, PublisherCallbackChannel.class };
		}
		else {
			interfaces = new Class<?>[] { ChannelProxy.class };
		}
		return (ChannelProxy) Proxy.newProxyInstance(ChannelProxy.class.getClassLoader(),
				interfaces, new CachedChannelInvocationHandler(connection, targetChannel, channelList,
						transactional));
```
However I I guess we need to register one more proxy for the `ChannelProxy.class` and `PublisherCallbackChannel.class` combination according the logic of that `CachingConnectionFactory` class.
The one for `RabbitListener` annotation is OK. However it might be something like:
```
			RuntimeHintsUtils.registerAnnotation(hints, Scheduled.class);
```
instead ?

artembilan (changes requested):

spring-rabbit/src/main/resources/META-INF/spring/aot.factories
```diff
@@ -0,0 +1 @@
+org.springframework.amqp.rabbit.aot.RabbitRuntimeHintsRegistrar=org.springframework.amqp.rabbit.aot.RabbitRuntimeHintsRegistrar
```
artembilan, line 1: The key in this file is wrong. Must be: `org.springframework.aot.hint.RuntimeHintsRegistrar`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/aot/RabbitRuntimeHintsRegistrar.java
```diff
@@ -0,0 +1,52 @@
...
+ * @since 3.0
+ *
+ */
+public class RabbitRuntimeHintsRegistrar implements RuntimeHintsRegistrar {
+
+	@Override
+	public void registerHints(RuntimeHints hints, @Nullable	ClassLoader classLoader) {
+		RuntimeHintsUtils.registerAnnotation(hints, RabbitListener.class);
+		ProxyHints proxyHints = hints.proxies();
+		proxyHints.registerJdkProxy(ChannelProxy.class, PublisherCallbackChannel.class);
```
artembilan, line 45: I believe we still need an individual proxy for just a `ChannelProxy.class` according to the usage logic we’ve discussed before.

artembilan (changes requested):

spring-rabbit/src/main/resources/META-INF/spring/aot.factories
garyrussell: Yes; I fixed that last night (but failed to re-push).

garyrussell:
This is good to go now; smoke test works.

artembilan (approved):
I can merge this if there is no need in anyone else approval.

garyrussell:
Please do; the smoke test runs fine with this.

Issue #1474, opened 2022-07-25
two problems with receiving-batch


Fenhy:
**In what version(s) of Spring AMQP are you seeing this issue?**

spring-rabbit-2.4.6

**Describe the bug**

I need to use batch send and batch receive.

1. When messages sent in batches are de-batch, the SimpleBatchingStrategy.deBatch() method shares the same instance of MessageProperties when processing AmqpHeaders.LAST_IN_BATCH, resulting in all messages in the batch becoming amqp_lastInBatch=true.

2. When SimpleRabbitListenerContainerFactory generates an instance, the batch strategy is only set to the endpoint, but not to the instance, so the default strategy is still used when batching: new SimpleBatchingStrategy(0, 0, 0L)

**To Reproduce**

1. Enable BatchingRabbitTemplate
2. Send multiple messages
3. Enable ConsumerBatchEnabled
4. Receive messages in batches:
- Can't tell if it's the last message in the batch
- The configured batch strategy cannot be used correctly

**Expected behavior**

1. Correctly set AmqpHeaders.LAST_IN_BATCH
2. Custom batch policies can take effect normally.

I provide override code samples at the end.

**Sample**

import dependencies
```xml
<dependency>
    <groupId>org.springframework.boot</groupId>
    <artifactId>spring-boot-dependencies</artifactId>
    <version>2.7.1</version>
    <type>pom</type>
    <scope>import</scope>
</dependency>
```

java ：
bind
```java
@Configuration
@Slf4j
public class RabbitConfigAutoByAnnotation  {

    @Autowired
    RabbitTemplate rabbitTemplate;

    @Bean
    public Queue queueAnnotation() {
        Map<String, Object> queueArguments = new HashMap<>(1);
        queueArguments.put("x-single-active-consumer", true);
        return new Queue(RabbitConstant.QUEUE_ANNOTATION, true, false, false, queueArguments);
    }

    @Bean
    public DirectExchange exchangeAnnotation() {
        return new DirectExchange(RabbitConstant.EXCHANGE_ANNOTATION, true, false, null);
    }

    @Bean
    public Binding workbenchDataBinding() {
        return BindingBuilder.bind(queueAnnotation()).to(exchangeAnnotation()).with(RabbitConstant.ROUTING_KEY_ANNOTATION);
    }
}
```

RabbitBatchConfig.java
```java
@Configuration
@Slf4j
public class RabbitBatchConfig {
    /**
     * 批量收集最大消息条数
     */
[...]

PR #1475, opened 2022-07-25
GH-1474: Fix BatchingStrategy Propagation

garyrussell:
See https://github.com/spring-projects/spring-amqp/issues/1474
(Does not resolve because 2 issues are reported there).

`BatchingStrategy` was not set by container factory.

**cherry-pick to 2.4.x**


artembilan:

PR #1476, opened 2022-07-25
GH-1474: Fix MessageProperties.lastInBatch

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1474

When consuming the whole debatched batch as a list, all messages
had the `lastInBatch` property set to true.

Clone the message properties for the last record.

**cherry-pick to 2.4.x**


artembilan:

PR #1478, opened 2022-07-27
GH-1473: Switch to CompletableFuture

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1473

Given the stability of the project, it was simplest to copy the `AsyncRabbitTemplate`
rather than adding a lot of conditional code.

**2.4.x only; I will submit a separate PR for main**

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/core/AsyncAmqpTemplate2.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/core/AsyncAmqpTemplate2.java
@@ -1,0 +1,205 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.amqp.core;
+
+import java.util.concurrent.CompletableFuture;
+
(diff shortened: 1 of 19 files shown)


spring-amqp/src/main/java/org/springframework/amqp/core/AsyncAmqpTemplate2.java
```diff
@@ -0,0 +1,205 @@
...
+
+/**
+ * Classes implementing this interface can perform asynchronous send and
+ * receive operations using {@link CompletableFuture}s.
+ *
+ * @author Gary Russell
+ * @since 2.4.7
+ *
+ */
+public interface AsyncAmqpTemplate2 {
```
artembilan, line 31: Let me guess: in `3.0` this is going to be like:
```
@Deprecated
public interface AsyncAmqpTemplate2 extends AsyncAmqpTemplate {
}
```
where all the `CompletableFuture` stuff will migrate over there which give us an opportunity to de-deprecate it 😄 .

This will give an opportunity for end-users to migrate their code to `CompletableFuture` only once: in this `2.4.x` version or directly in the `3.0.x`.
Remove `2` from type name of the variable is not so bad.

artembilan (changes requested):
Just a question to confirm (or not)

spring-amqp/src/main/java/org/springframework/amqp/core/AsyncAmqpTemplate2.java
garyrussell: Right that was my intention, or we could just pull it out altogether (like we decided to do in spring-kafka).

artembilan: Well, it depends how much breaking change we have already done in Spring AMQP `3.0`...
I believe not too much. So, let's downstream projects to take an chance to migrate easily!

artembilan: For example an `AsyncRabbitTemplate` is used in SI `AsyncAmqpOutboundGateway`...

garyrussell:
OK; makes sense.

PR #1479, opened 2022-07-28
GH-1473: Move RabbitFutures to Top Level Classes

garyrussell:
- to aid migration from 2.4.x to 3.0.x so that the return types will not change

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/RabbitConverterFuture.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/RabbitConverterFuture.java
@@ -28,6 +28,8 @@ import org.springframework.core.ParameterizedTypeReference;
  * A {@link RabbitFuture} with a return type of the template's
  * generic parameter.
  * @param <C> the type.
+ *
+ * @author Gary Russell
  * @since 2.4.7
  */
 public class RabbitConverterFuture<C> extends RabbitFuture<C> {
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/RabbitConverterFuture.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/RabbitConverterFuture.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2022 the original author or authors.
+ *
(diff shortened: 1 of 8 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/RabbitConverterFuture.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.springframework.amqp.core.Message;
+import org.springframework.amqp.rabbit.listener.DirectReplyToMessageListenerContainer.ChannelHolder;
+import org.springframework.core.ParameterizedTypeReference;
+
+/**
+ * A {@link RabbitFuture} with a return type of the template's
+ * generic parameter.
+ * @param <C> the type.
+ * @since 1.6
```
artembilan, line 31:
```suggestion
 * @since  2.4.7
```
and `@author` ?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/RabbitConverterFuture.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+public class RabbitConverterFuture<C> extends RabbitFuture<C> {
+
+	private volatile ParameterizedTypeReference<C> returnType;
+
+	RabbitConverterFuture(String correlationId, Message requestMessage,
```
artembilan, line 37: Looks like those classes in the `AsyncRabbitTemplate` come with a `public` ctor...
Is there a particular reason to change that matter now, since I believe in `3.0` we are going to use this top-level classes instead in a new version of the  `AsyncRabbitTemplate`?
Or was that just a mistake from day first when those inner classes were introduced since IDE just generates ctors with `public` by default?..

artembilan (changes requested):

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/RabbitConverterFuture.java
garyrussell: It was a mistake - they are not static classes so they can't be instantiated outside of the template scope anyway.

PR #1482, opened 2022-07-28
GH-1480: Switch to CompletableFuture in s-r-stream

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1480

Also reinstate deprecated `AsyncAmqpTemplate2`.

### DIFF
--- a/spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/producer/RabbitStreamTemplate.java
+++ b/spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/producer/RabbitStreamTemplate.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 the original author or authors.
+ * Copyright 2021-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/producer/RabbitStreamTemplate.java
```diff
@@ -138,27 +138,27 @@ public StreamMessageConverter streamMessageConverter() {
...
 		Message message2 = this.messageConverter.toMessage(message, new StreamMessageProperties());
 		Assert.notNull(message2, "The message converter returned null");
 		if (mpp != null) {
 			message2 = mpp.postProcessMessage(message2);
 			if (message2 == null) {
 				this.logger.debug("Message Post Processor returned null, message not sent");
-				SettableListenableFuture<Boolean> future = new SettableListenableFuture<>();
-				future.set(false);
+				CompletableFuture<Boolean> future = new CompletableFuture<>();
+				future.complete(false);
```
artembilan, line 161: Just FYI: `CompletableFuture.completedFuture(false);` can be used instead.

artembilan (approved):

Issue #1484, opened 2022-08-02
Add batch to `@RabbitListener`


garyrussell:
SImilar to `@KafkaListener` so that a single container factory can creatre containers for both types of listener.

PR #1486, opened 2022-08-02
GH-1484: RabbitListener.batch() Override

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1484

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListener.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListener.java
@@ -332,4 +332,18 @@ public @interface RabbitListener {
 	 */
 	String converterWinsContentType() default "true";
 
+	/**
+	 * Override the container factory's {@code batchListener} property. The listener
+	 * method signature should receive a {@code List<?>}; refer to the reference
+	 * documentation. This allows a single container factory to be used for both record
+	 * and batch listeners; previously separate container factories were required.
+	 * @return "true" for the annotated method to be a batch listener or "false" for a
+	 * single message listener. If not set, the container factory setting is used. SpEL and
+	 * property place holders are not supported because the listener type cannot be
+	 * variable.
+	 * @since 3.0
+	 * @see Boolean#parseBoolean(String)
+	 */
+	String batch() default "";
+
 }
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/RabbitListenerEndpoint.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/RabbitListenerEndpoint.java
@@ -117,6 +117,14 @@ public interface RabbitListenerEndpoint {
 	}
 
 	/**
+	 * Whether this endpoint is for a batch listener.
+	 * @return {@link Boolean#TRUE} if batch.
+	 * @since 3.0
+	 */
+	@Nullable
+	Boolean getBatchListener();
+
+	/**
 	 * Set a {@link BatchingStrategy} to use when debatching messages.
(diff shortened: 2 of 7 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListener.java
```diff
@@ -332,4 +332,18 @@
...
 	String converterWinsContentType() default "true";
 
+	/**
+	 * Override the container factory's {@code batchListener} property. The listener
+	 * method signature should receive a {@code List<?>}; refer to the reference
+	 * documentation. This allows a single container factory to be used for both record
+	 * and batch listeners; previously separate container factories were required.
+	 * @return "true" for the annotated method to be a batch listener or "false" for a
+	 * single message listener. If not set, the container factory setting is used. SpEL and
+	 * property place holders are not supported because the listener type cannot be
```
artembilan, line 342: So, if they are not supported, then why the property is `String` but not `boolean`?
Or the idea is because there is no way to have it set to something like `null`, so we can rely on the container factory option?
Then maybe `ValueConstants.DEFAULT_NONE` would be better?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/RabbitListenerEndpoint.java
```diff
@@ -116,6 +116,14 @@ default TaskExecutor getTaskExecutor() {
 	default void setBatchListener(boolean batchListener) {
 	}
 
+	/**
+	 * Whether this endpoint is for a batch listener.
+	 * @return {@link Boolean#TRUE} if batch.
+	 * @since 3.0
+	 */
+	@Nullable
+	Boolean getBatchListener();
```
artembilan, line 125:
```suggestion
	Boolean isBatchListener();
```
?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListener.java
```diff
@@ -332,4 +332,18 @@
 	 */
 	String converterWinsContentType() default "true";
 
+	/**
+	 * Override the container factory's {@code batchListener} property. The listener
+	 * method signature should receive a {@code List<?>}; refer to the reference
```
artembilan, line 337: Just for my info: do we validate somewhere that listener method has a `List` signature for a batch mode?

artembilan (changes requested):
Documentation?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/RabbitListenerEndpoint.java
garyrussell: No; javaBean spec says `get` for `Boolean` `is` for `boolean`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListener.java
garyrussell: We currently don't; I'll take a look to see if it's feasible.

artembilan: No problem! Perhaps other issue/PR.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListener.java
garyrussell: Well, if we change this one, we should change the others; I made a quick attempt at that and it is not trivial. I prefer to leave this as-is, at least for now.

Issue #1487, opened 2022-08-03
Consumer failed to start in XX milliseconds, SimpleMessageListenerContainer abnormal shut down 


renyansongno1:
**In what version(s) of Spring AMQP are you seeing this issue?**

every version 

**Describe the bug**

We use the SimpleMessageListenerContainer to dynamically create consumer(listen some queue) when we receive some signal, 
but If in the process of adding (addQueues(Queue... queue)), the spring context is destroyed(shutdown the program). so there will be some exception here: Consumer failed to start in XX milliseconds; does the task executor have enough threads to support the container concurrency?

This exception is very unclear, we find the SimpleMessageListenerContainer is getting the thread start result, but now thread get the spring context was destroyed, so return, the AsyncMessageProcessingConsumer is not called CountDownLatch.countdown.

so we think there's a bug here, we should call countdown in all cases

**To Reproduce**

1. thread a invoke SimpleMessageListenerContainer.addQueues
2. thread b close the spring context

**Expected behavior**

context close now, do not report "Consumer failed to start in XX milliseconds; does the task executor have enough threads to support the container concurrency?"

**Sample**

PR #1488, opened 2022-08-03
AsyncMessageProcessingConsumer countdown when the container is not active

renyansongno1:
Resolves https://github.com/spring-projects/spring-amqp/issues/1487

### DIFF
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerTests.java
@@ -107,6 +107,7 @@ import com.rabbitmq.client.PossibleAuthenticationFailureException;
  * @author Gary Russell
  * @author Artem Bilan
  * @author Mohammad Hewedy
+ * @author Yansong Ren
  */
 public class SimpleMessageListenerContainerTests {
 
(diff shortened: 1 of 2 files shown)


artembilan (approved):
LGTM, but is there a chance to add some unit test to confirm the problem and the fix?

Thanks

renyansongno1:
@artembilan Hi, I have submitted the test code, please review.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerTests.java
```diff
@@ -784,4 +809,33 @@ protected void doRollback(DefaultTransactionStatus status) throws TransactionExc
 
 	}
 
+	@SuppressWarnings("serial")
+	private class TestExecutor extends SimpleAsyncTaskExecutor {
```
artembilan, line 813:
```suggestion
	private static final class TestExecutor extends SimpleAsyncTaskExecutor {
```

Also, please, add your name to the `@author` list of affected classes.

artembilan (changes requested):

renyansongno1:
@artembilan Thank you for your time!

artembilan:

@renyansongno1 ,

thank you very much for the contribution; looking forward for more!

Issue #1497, opened 2022-08-26
AbstractConnectionFactory - addressShuffleMode RANDOM as default


rvervaek:
**Expected Behavior**

When using a RabbitMQ cluster, client applications using Spring's ConnectionFactory abstraction connect by default to the first host configured in the list of addresses provided. Using a RabbitMQ cluster provides not only fault tolerance in the case of node failures, but also resiliency when client connections are distributed among all nodes. I can't think of a single use-case where it is desired to let all client applications connect to the same node in a cluster. Therefor I would expect the default value of addressShuffleMode in AbstractConnectionFactory to be set to RANDOM so clients randomly choose a node to connect to, resulting in a somewhat even distribution of client connections among nodes in the cluster.

**Current Behavior**

The default addressShuffleMode is NONE, resulting in all client applications connecting to the first node defined in the addresses.

**Context**

In our cluster a single node handles almost all of the client connections, resulting in unnecessary load on the node. Additionally, performing an upgrade of that node or observing a crash would result in all client applications losing their connections instead of just a portion. 

Happy to make a PR if this sounds feasible.

garyrussell:
Seems reasonable; `NONE` was chosen because it was added in a patch release and we can't make behavior changes until minor/major releases.

Since the upcoming 3.0 release is a major release, we can make such a change; a PR would be appreciated; we would also need a change to the doc as well as an entry in the what's new section.

Thanks.

PR #1498, opened 2022-08-26
GH-1497: Use RANDOM as default addressShuffleMode

rvervaek:
Changed the default behavior of the address which is chosen during creation of a connection in AbstractConnectionFactory. The default addressShuffleMode is now RANDOM. 

See feature request: https://github.com/spring-projects/spring-amqp/issues/1497

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2021 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
```diff
@@ -525,17 +525,20 @@ protected String getBeanName() {
...
-	 * {@link #setAddressShuffleMode(AddressShuffleMode)}.
 	 */
 	@Deprecated
 	public void setShuffleAddresses(boolean shuffleAddresses) {
-		setAddressShuffleMode(AddressShuffleMode.RANDOM);
+		if (shuffleAddresses) {
+			setAddressShuffleMode(AddressShuffleMode.RANDOM);
+		} else {
+			setAddressShuffleMode(AddressShuffleMode.NONE);
+		}
```
garyrussell, line 541: This deprecated method can/should be removed in 3.0.

garyrussell (changes requested):

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/AbstractConnectionFactory.java
rvervaek: Done

garyrussell (approved):

garyrussell:
Thanks for the contribution; merged as c929cd5d3 with some doc polishing.

PR #1500, opened 2022-09-07
GH-1444: Listener Observability Initial Commit

garyrussell:
- expand scope to include error handler: https://github.com/spring-projects/spring-amqp/pull/1287
- tracing can't be used with a batch listener (multiple messages in listener call)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -399,7 +399,7 @@ project('spring-rabbit') {
 		testApi project(':spring-rabbit-junit')
 		testImplementation("com.willowtreeapps.assertk:assertk-jvm:$assertkVersion")
 		testImplementation "org.hibernate.validator:hibernate-validator:$hibernateValidationVersion"
-        testImplementation 'io.micrometer:micrometer-observation-test'
+		testImplementation 'io.micrometer:micrometer-observation-test'
 		testImplementation 'io.micrometer:micrometer-tracing-bridge-brave'
 		testImplementation 'io.micrometer:micrometer-tracing-test'
 		testImplementation 'io.micrometer:micrometer-tracing-integration-test'
--- a/build.gradle
+++ b/build.gradle
@@ -387,6 +387,7 @@ project('spring-rabbit') {
 		optionalApi "ch.qos.logback:logback-classic:$logbackVersion"
 		optionalApi 'org.apache.logging.log4j:log4j-core'
 		optionalApi 'io.micrometer:micrometer-core'
+		api 'io.micrometer:micrometer-observation'
 		optionalApi 'io.micrometer:micrometer-tracing'
 		// Spring Data projection message binding support
 		optionalApi ("org.springframework.data:spring-data-commons") {
@@ -398,6 +399,7 @@ project('spring-rabbit') {
 		testApi project(':spring-rabbit-junit')
 		testImplementation("com.willowtreeapps.assertk:assertk-jvm:$assertkVersion")
 		testImplementation "org.hibernate.validator:hibernate-validator:$hibernateValidationVersion"
+        testImplementation 'io.micrometer:micrometer-observation-test'
 		testImplementation 'io.micrometer:micrometer-tracing-bridge-brave'
 		testImplementation 'io.micrometer:micrometer-tracing-test'
 		testImplementation 'io.micrometer:micrometer-tracing-integration-test'
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
@@ -2439,7 +2439,7 @@ public class RabbitTemplate extends RabbitAccessor // NOSONAR type line count
 		}
 		else {
 			observation = RabbitTemplateObservation.TEMPLATE_OBSERVATION.observation(registry,
-						new RabbitMessageSenderContext(message, this.beanName))
+						new RabbitMessageSenderContext(message, this.beanName, exch + "/" + rKey))
 					.observationConvention(this.observationConvention);
 		}
 		observation.observe(() -> sendToRabbit(channel, exch, rKey, mandatory, message));
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
@@ -74,6 +74,9 @@ import org.springframework.amqp.rabbit.support.ListenerContainerAware;
 import org.springframework.amqp.rabbit.support.MessagePropertiesConverter;
 import org.springframework.amqp.rabbit.support.RabbitExceptionTranslator;
 import org.springframework.amqp.rabbit.support.ValueExpression;
+import org.springframework.amqp.rabbit.support.micrometer.MessageSenderContext;
+import org.springframework.amqp.rabbit.support.micrometer.RabbitTemplateObservation;
+import org.springframework.amqp.rabbit.support.micrometer.RabbitTemplateObservationConvention;
 import org.springframework.amqp.support.converter.MessageConverter;
 import org.springframework.amqp.support.converter.SimpleMessageConverter;
 import org.springframework.amqp.support.converter.SmartMessageConverter;
@@ -83,6 +86,8 @@ import org.springframework.beans.factory.BeanFactory;
 import org.springframework.beans.factory.BeanFactoryAware;
 import org.springframework.beans.factory.BeanNameAware;
 import org.springframework.beans.factory.DisposableBean;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
 import org.springframework.context.expression.BeanFactoryResolver;
 import org.springframework.context.expression.MapAccessor;
 import org.springframework.core.ParameterizedTypeReference;
@@ -108,6 +113,8 @@ import com.rabbitmq.client.GetResponse;
 import com.rabbitmq.client.Return;
 import com.rabbitmq.client.ShutdownListener;
 import com.rabbitmq.client.ShutdownSignalException;
+import io.micrometer.observation.Observation;
+import io.micrometer.observation.ObservationRegistry;
 
 /**
  * <p>
@@ -152,7 +159,7 @@ import com.rabbitmq.client.ShutdownSignalException;
  * @since 1.0
  */
 public class RabbitTemplate extends RabbitAccessor // NOSONAR type line count
-		implements BeanFactoryAware, RabbitOperations, ChannelAwareMessageListener,
+		implements BeanFactoryAware, RabbitOperations, ChannelAwareMessageListener, ApplicationContextAware,
 		ListenerContainerAware, PublisherCallbackChannel.Listener, BeanNameAware, DisposableBean {
 
 	private static final String UNCHECKED = "unchecked";
@@ -197,6 +204,8 @@ public class RabbitTemplate extends RabbitAccessor // NOSONAR type line count
 			new HashMap<>();
 
 	private final AtomicInteger containerInstance = new AtomicInteger();
+
+	private ApplicationContext applicationContext;
 
 	private String exchange = DEFAULT_EXCHANGE;
 
@@ -258,13 +267,19 @@ public class RabbitTemplate extends RabbitAccessor // NOSONAR type line count
 
 	private ErrorHandler replyErrorHandler;
 
+	private boolean useChannelForCorrelation;
+
+	private boolean observationEnabled;
+
+	private RabbitTemplateObservationConvention observationConvention;
+
 	private volatile boolean usingFastReplyTo;
 
 	private volatile boolean evaluatedFastReplyTo;
 
 	private volatile boolean isListener;
 
-	private boolean useChannelForCorrelation;
+	private volatile boolean observationRegistryObtained;
 
 	/**
 	 * Convenient constructor for use with setter injection. Don't forget to set the connection factory.
@@ -295,6 +310,29 @@ public class RabbitTemplate extends RabbitAccessor // NOSONAR type line count
 		if (connectionFactory instanceof ThreadChannelConnectionFactory) {
 			this.usePublisherConnection = true;
 		}
+	}
+
+	@Override
+	public void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
+		this.applicationContext = applicationContext;
+	}
+
+	/**
+	 * Enable observation via micrometer.
+	 * @param observationEnabled true to enable.
+	 * @since 3.0
+	 */
+	public void setObservationEnabled(boolean observationEnabled) {
+		this.observationEnabled = observationEnabled;
+	}
+
+	/**
+	 * Set an observation convention; used to add additional key/values to observations.
+	 * @param observationConvention the convention.
+	 * @since 3.0
+	 */
+	public void setObservationConvention(RabbitTemplateObservationConvention observationConvention) {
+		this.observationConvention = observationConvention;
 	}
 
 	/**
@@ -2348,7 +2386,7 @@ public class RabbitTemplate extends RabbitAccessor // NOSONAR type line count
 	 * @throws IOException If thrown by RabbitMQ API methods.
 	 */
 	public void doSend(Channel channel, String exchangeArg, String routingKeyArg, Message message,
-			boolean mandatory, @Nullable CorrelationData correlationData) throws IOException {
+			boolean mandatory, @Nullable CorrelationData correlationData) {
 
 		String exch = nullSafeExchange(exchangeArg);
 		String rKey = nullSafeRoutingKey(routingKeyArg);
@@ -2378,12 +2416,36 @@ public class RabbitTemplate extends RabbitAccessor // NOSONAR type line count
 			logger.debug("Publishing message [" + messageToUse
 					+ "] on exchange [" + exch + "], routingKey = [" + rKey + "]");
 		}
-		sendToRabbit(channel, exch, rKey, mandatory, messageToUse);
+		observeTheSend(channel, message, mandatory, exch, rKey, messageToUse);
 		// Check if commit needed
 		if (isChannelLocallyTransacted(channel)) {
 			// Transacted channel created by this template -> commit.
 			RabbitUtils.commitIfNecessary(channel);
 		}
+	}
+
+	protected void observeTheSend(Channel channel, Message message, boolean mandatory, String exch, String rKey,
+			Message messageToUse) {
+
+		if (!this.observationRegistryObtained) {
+			obtainObservationRegistry(this.applicationContext);
+			this.observationRegistryObtained = true;
+		}
+		Observation observation;
+		ObservationRegistry registry = getObservationRegistry();
+		if (!this.observationEnabled || registry == null) {
+			observation = Observation.NOOP;
+		}
+		else {
+			observation = RabbitTemplateObservation.TEMPLATE_OBSERVATION.observation(registry,
+						new MessageSenderContext(message))
+					.lowCardinalityKeyValue(RabbitTemplateObservation.TemplateLowCardinalityTags.BEAN_NAME.asString(),
+							this.beanName);
+			if (this.observationConvention != null) {
+				observation.observationConvention(this.observationConvention);
+			}
+		}
+		observation.observe(() -> sendToRabbit(channel, exch, rKey, mandatory, messageToUse));
 	}
 
 	/**
@@ -2407,10 +2469,16 @@ public class RabbitTemplate extends RabbitAccessor // NOSONAR type line count
 	}
 
 	protected void sendToRabbit(Channel channel, String exchange, String routingKey, boolean mandatory,
-			Message message) throws IOException {
+			Message message) {
+
 		BasicProperties convertedMessageProperties = this.messagePropertiesConverter
 				.fromMessageProperties(message.getMessageProperties(), this.encoding);
-		channel.basicPublish(exchange, routingKey, mandatory, convertedMessageProperties, message.getBody());
+		try {
+			channel.basicPublish(exchange, routingKey, mandatory, convertedMessageProperties, message.getBody());
+		}
+		catch (IOException ex) {
+			throw RabbitExceptionTranslator.convertRabbitAccessException(ex);
+		}
 	}
 
 	private void setupConfirm(Channel channel, Message message, @Nullable CorrelationData correlationDataArg) {
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
@@ -74,6 +74,7 @@ import org.springframework.amqp.rabbit.support.ListenerContainerAware;
 import org.springframework.amqp.rabbit.support.MessagePropertiesConverter;
 import org.springframework.amqp.rabbit.support.RabbitExceptionTranslator;
 import org.springframework.amqp.rabbit.support.ValueExpression;
+import org.springframework.amqp.rabbit.support.micrometer.DefaultRabbitTemplateObservationConvention;
 import org.springframework.amqp.rabbit.support.micrometer.RabbitMessageSenderContext;
 import org.springframework.amqp.rabbit.support.micrometer.RabbitTemplateObservation;
 import org.springframework.amqp.rabbit.support.micrometer.RabbitTemplateObservationConvention;
@@ -114,7 +115,6 @@ import com.rabbitmq.client.Return;
 import com.rabbitmq.client.ShutdownListener;
 import com.rabbitmq.client.ShutdownSignalException;
 import io.micrometer.observation.Observation;
-import io.micrometer.observation.ObservationConvention;
 import io.micrometer.observation.ObservationRegistry;
 
 /**
@@ -273,7 +273,7 @@ public class RabbitTemplate extends RabbitAccessor // NOSONAR type line count
 	private boolean observationEnabled;
 
 	@Nullable
-	private ObservationConvention<RabbitMessageSenderContext> observationConvention;
+	private RabbitTemplateObservationConvention observationConvention;
 
 	private volatile boolean usingFastReplyTo;
 
@@ -333,7 +333,7 @@ public class RabbitTemplate extends RabbitAccessor // NOSONAR type line count
 	 * @param observationConvention the convention.
 	 * @since 3.0
 	 */
-	public void setObservationConvention(ObservationConvention<RabbitMessageSenderContext> observationConvention) {
+	public void setObservationConvention(RabbitTemplateObservationConvention observationConvention) {
 		this.observationConvention = observationConvention;
 	}
 
@@ -2439,7 +2439,7 @@ public class RabbitTemplate extends RabbitAccessor // NOSONAR type line count
 		}
 		else {
 			observation = RabbitTemplateObservation.TEMPLATE_OBSERVATION.observation(this.observationConvention,
-					RabbitTemplateObservationConvention.INSTANCE,
+					DefaultRabbitTemplateObservationConvention.INSTANCE,
 						new RabbitMessageSenderContext(message, this.beanName, exch + "/" + rKey), registry);
 
 		}
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
@@ -74,7 +74,7 @@ import org.springframework.amqp.rabbit.support.ListenerContainerAware;
 import org.springframework.amqp.rabbit.support.MessagePropertiesConverter;
 import org.springframework.amqp.rabbit.support.RabbitExceptionTranslator;
 import org.springframework.amqp.rabbit.support.ValueExpression;
-import org.springframework.amqp.rabbit.support.micrometer.MessageSenderContext;
+import org.springframework.amqp.rabbit.support.micrometer.AmqpMessageSenderContext;
 import org.springframework.amqp.rabbit.support.micrometer.RabbitTemplateObservation;
 import org.springframework.amqp.rabbit.support.micrometer.RabbitTemplateObservationConvention;
 import org.springframework.amqp.support.converter.MessageConverter;
@@ -2416,7 +2416,7 @@ public class RabbitTemplate extends RabbitAccessor // NOSONAR type line count
 			logger.debug("Publishing message [" + messageToUse
 					+ "] on exchange [" + exch + "], routingKey = [" + rKey + "]");
 		}
-		observeTheSend(channel, message, mandatory, exch, rKey, messageToUse);
+		observeTheSend(channel, messageToUse, mandatory, exch, rKey);
 		// Check if commit needed
 		if (isChannelLocallyTransacted(channel)) {
 			// Transacted channel created by this template -> commit.
@@ -2424,8 +2424,7 @@ public class RabbitTemplate extends RabbitAccessor // NOSONAR type line count
 		}
 	}
 
-	protected void observeTheSend(Channel channel, Message message, boolean mandatory, String exch, String rKey,
-			Message messageToUse) {
+	protected void observeTheSend(Channel channel, Message message, boolean mandatory, String exch, String rKey) {
 
 		if (!this.observationRegistryObtained) {
 			obtainObservationRegistry(this.applicationContext);
@@ -2438,14 +2437,14 @@ public class RabbitTemplate extends RabbitAccessor // NOSONAR type line count
 		}
 		else {
 			observation = RabbitTemplateObservation.TEMPLATE_OBSERVATION.observation(registry,
-						new MessageSenderContext(message))
+						new AmqpMessageSenderContext(message))
 					.lowCardinalityKeyValue(RabbitTemplateObservation.TemplateLowCardinalityTags.BEAN_NAME.asString(),
 							this.beanName);
 			if (this.observationConvention != null) {
 				observation.observationConvention(this.observationConvention);
 			}
 		}
-		observation.observe(() -> sendToRabbit(channel, exch, rKey, mandatory, messageToUse));
+		observation.observe(() -> sendToRabbit(channel, exch, rKey, mandatory, message));
 	}
 
 	/**
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
@@ -114,6 +114,7 @@ import com.rabbitmq.client.Return;
 import com.rabbitmq.client.ShutdownListener;
 import com.rabbitmq.client.ShutdownSignalException;
 import io.micrometer.observation.Observation;
+import io.micrometer.observation.ObservationConvention;
 import io.micrometer.observation.ObservationRegistry;
 
 /**
@@ -271,7 +272,8 @@ public class RabbitTemplate extends RabbitAccessor // NOSONAR type line count
 
 	private boolean observationEnabled;
 
-	private RabbitTemplateObservationConvention observationConvention = new RabbitTemplateObservationConvention();
+	@Nullable
+	private ObservationConvention<RabbitMessageSenderContext> observationConvention;
 
 	private volatile boolean usingFastReplyTo;
 
@@ -331,8 +333,7 @@ public class RabbitTemplate extends RabbitAccessor // NOSONAR type line count
 	 * @param observationConvention the convention.
 	 * @since 3.0
 	 */
-	public void setObservationConvention(RabbitTemplateObservationConvention observationConvention) {
-		Assert.notNull(observationConvention, "'observationConvention' cannot be null");
+	public void setObservationConvention(ObservationConvention<RabbitMessageSenderContext> observationConvention) {
 		this.observationConvention = observationConvention;
 	}
 
@@ -2437,9 +2438,10 @@ public class RabbitTemplate extends RabbitAccessor // NOSONAR type line count
 			observation = Observation.NOOP;
 		}
 		else {
-			observation = RabbitTemplateObservation.TEMPLATE_OBSERVATION.observation(registry,
-						new RabbitMessageSenderContext(message, this.beanName, exch + "/" + rKey))
-					.observationConvention(this.observationConvention);
+			observation = RabbitTemplateObservation.TEMPLATE_OBSERVATION.observation(this.observationConvention,
+					RabbitTemplateObservationConvention.INSTANCE,
+						new RabbitMessageSenderContext(message, this.beanName, exch + "/" + rKey), registry);
+
 		}
 		observation.observe(() -> sendToRabbit(channel, exch, rKey, mandatory, message));
 	}
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
@@ -271,8 +271,7 @@ public class RabbitTemplate extends RabbitAccessor // NOSONAR type line count
 
 	private boolean observationEnabled;
 
-	private RabbitTemplateObservationConvention observationConvention =
-			new RabbitTemplateObservationConvention(null, null);
+	private RabbitTemplateObservationConvention observationConvention = new RabbitTemplateObservationConvention();
 
 	private volatile boolean usingFastReplyTo;
 
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
@@ -74,7 +74,7 @@ import org.springframework.amqp.rabbit.support.ListenerContainerAware;
 import org.springframework.amqp.rabbit.support.MessagePropertiesConverter;
 import org.springframework.amqp.rabbit.support.RabbitExceptionTranslator;
 import org.springframework.amqp.rabbit.support.ValueExpression;
-import org.springframework.amqp.rabbit.support.micrometer.AmqpMessageSenderContext;
+import org.springframework.amqp.rabbit.support.micrometer.RabbitMessageSenderContext;
 import org.springframework.amqp.rabbit.support.micrometer.RabbitTemplateObservation;
 import org.springframework.amqp.rabbit.support.micrometer.RabbitTemplateObservationConvention;
 import org.springframework.amqp.support.converter.MessageConverter;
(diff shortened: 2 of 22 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitListenerObservation.java
```diff
@@ -0,0 +1,72 @@
...
+	LISTENER_OBSERVATION {
+
+		@Override
+		public String getName() {
+			return "spring.rabbit.listener";
+		}
+
+		@Override
+		public String getContextualName() {
+			return "RabbitListener Observation";
```
artembilan, line 43: I see this in the `PropagatingSenderTracingObservationHandler`:
```
span.name(context.getContextualName() != null ? context.getContextualName() : context.getName());
```

so, the plain name (in your case `spring.rabbit.listener`) is not going to be visible on a span.
Or there is another logic in the super interface:
```
    /**
     * Get the span name.
     * @param context handler context
     * @return name for the span
     */
    default String getSpanName(T context) {
        String name = context.getName();
        if (StringUtils.isNotBlank(context.getContextualName())) {
            name = context.getContextualName();
        }
        return SpanNameUtil.toLowerHyphen(name);
    }
```

So, we need to consult with @marcingrzejszczak to be sure where to use one or another, or both.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/MessageSenderContext.java
```diff
@@ -0,0 +1,37 @@
...
+import io.micrometer.observation.transport.SenderContext;
+
+/**
+ * {@link SenderContext} for {@link Message}s.
+ *
+ * @author Gary Russell
+ * @since 3.0
+ *
+ */
+public class MessageSenderContext extends SenderContext<Message> {
```
artembilan, line 30: I call them the same way in Spring Integration as well.
I wonder if we should think about different names.
For example `RabbitSenderContext` 😄 or `RabbitMessageSenderContext`.
Just to avoid confusion on the end-user side.

Or... since it is apparently your Pr is going to be merged first, I probably need to rename them in Spring Integration... 😉

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitTemplateObservationConvention.java
```diff
@@ -0,0 +1,73 @@
...
+	public RabbitTemplateObservationConvention(@Nullable KeyValues lowCardinality,
+			@Nullable KeyValues highCardinality) {
+
+		this.lowCardinality = lowCardinality;
+		this.highCardinality = highCardinality;
+	}
+
+	@Override
+	public boolean supportsContext(Context context) {
+		return context instanceof MessageReceiverContext;
```
artembilan, line 56: Isn't this supposed to be a `SenderContext` instead?

spring-rabbit/src/test/resources/log4j2-test.xml
```diff
@@ -6,7 +6,7 @@
 		</Console>
 	</Appenders>
 	<Loggers>
-		<Logger name="org.springframework.amqp.rabbit" level="info"/>
+		<Logger name="org.springframework.amqp.rabbit" level="debug"/>
```
artembilan, line 9: Maybe we don't need this any more?
I mean you have checked logs locally - OK: no need to increase logging for `main`

build.gradle
```diff
@@ -398,6 +399,7 @@ project('spring-rabbit') {
 		testApi project(':spring-rabbit-junit')
 		testImplementation("com.willowtreeapps.assertk:assertk-jvm:$assertkVersion")
 		testImplementation "org.hibernate.validator:hibernate-validator:$hibernateValidationVersion"
+        testImplementation 'io.micrometer:micrometer-observation-test'
```
artembilan, line 402: Indention

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -2378,14 +2416,38 @@ public void doSend(Channel channel, String exchangeArg, String routingKeyArg, Me
...
+		observeTheSend(channel, message, mandatory, exch, rKey, messageToUse);
 		// Check if commit needed
 		if (isChannelLocallyTransacted(channel)) {
 			// Transacted channel created by this template -> commit.
 			RabbitUtils.commitIfNecessary(channel);
 		}
 	}
 
+	protected void observeTheSend(Channel channel, Message message, boolean mandatory, String exch, String rKey,
+			Message messageToUse) {
```
artembilan, line 2428: Why do we need two `Message` instances over here?
Why just one `messageToUse` from the previous modification is not enough?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -2378,14 +2416,38 @@ public void doSend(Channel channel, String exchangeArg, String routingKeyArg, Me
...
+		Observation observation;
+		ObservationRegistry registry = getObservationRegistry();
+		if (!this.observationEnabled || registry == null) {
+			observation = Observation.NOOP;
+		}
+		else {
+			observation = RabbitTemplateObservation.TEMPLATE_OBSERVATION.observation(registry,
+						new MessageSenderContext(message))
+					.lowCardinalityKeyValue(RabbitTemplateObservation.TemplateLowCardinalityTags.BEAN_NAME.asString(),
+							this.beanName);
```
artembilan, line 2443: Cannot this be a part of `MessageSenderContext` and populated transparently by that `RabbitTemplateObservation`?

artembilan (changes requested):

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
garyrussell: Refactor error.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
garyrussell: I don't see how to add key values to the context internally, `addLowCardinalityKeyValue()` is package-private.

garyrussell: I can do it via the ...Convention - stay tuned.

artembilan: Yeah... Probably.
See `DefaultApacheHttpClientObservationConvention` for example.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitListenerObservationConvention.java
```diff
@@ -0,0 +1,71 @@
...
+	public KeyValues getLowCardinalityKeyValues(RabbitMessageReceiverContext context) {
+		return this.lowCardinality
+				.and(RabbitListenerObservation.ListenerLowCardinalityTags.LISTENER_ID.asString(),
+						context.getListenerId());
+	}
+
+	@Override
+	public KeyValues getHighCardinalityKeyValues(RabbitMessageReceiverContext context) {
+		return this.highCardinality
+					.and(RabbitListenerObservation.ListenerLowCardinalityTags.LISTENER_ID.asString(),
```
artembilan, line 67: I don't think we need to add this tag into `high` as well

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitTemplateObservationConvention.java
```diff
@@ -0,0 +1,73 @@
...
+	public KeyValues getLowCardinalityKeyValues(RabbitMessageSenderContext context) {
+		return this.lowCardinality
+					.and(RabbitTemplateObservation.TemplateLowCardinalityTags.BEAN_NAME.asString(),
+							context.getBeanName());
+	}
+
+	@Override
+	public KeyValues getHighCardinalityKeyValues(RabbitMessageSenderContext context) {
+		return this.highCardinality
+					.and(RabbitTemplateObservation.TemplateLowCardinalityTags.BEAN_NAME.asString(),
```
artembilan, line 69: DITTO

artembilan (changes requested):

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitListenerObservationConvention.java
garyrussell: oops; brain fade.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitListenerObservation.java
marcingrzejszczak: The standard name is the name you'll eventually see as a metric name (e.g. `metrics.rabbitmq.listener` - if we want to align against sth like OpenTelemetry metrics https://opentelemetry.io/docs/reference/specification/metrics/semantic_conventions/instrumentation/kafka/ cc @shakuzen). The contextual name will be translated to a span name. That could be done also in accordance with OpenTelemetry trace semantic conventions (https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/messaging/#span-name) and could be set to `<destination name> <operation name>` (that's why you need to have information from the context to set the name - this is why we called it contextual name)

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
marcingrzejszczak: Yup, you should set all keyvalues via conventions.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitTemplateObservationConvention.java
artembilan: Missed this one, Gary.

artembilan (changes requested):

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitListenerObservationConvention.java
```diff
@@ -0,0 +1,69 @@
...
+	private final KeyValues lowCardinality;
+
+	private final KeyValues highCardinality;
+
+	/**
+	 * Create an instance with the provided {@link KeyValues}.
+	 * @param lowCardinality the low cardinality {@link KeyValues}.
+	 * @param highCardinality the high cardinality {@link KeyValues}.
+	 */
+	public RabbitListenerObservationConvention(@Nullable KeyValues lowCardinality,
```
marcingrzejszczak, line 45: Why do you want to allow users to add key values like this? If they want to set their own key values, they would provide their own custom convention that could simply append their key values to whatever you're setting in your default observation convention implementation

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitListenerObservationConvention.java
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	@Override
+	public boolean supportsContext(Context context) {
+		return context instanceof RabbitMessageReceiverContext;
+	}
+
+	@Override
+	public KeyValues getLowCardinalityKeyValues(RabbitMessageReceiverContext context) {
+		return this.lowCardinality
```
marcingrzejszczak, line 59: How many listener ids can we have in a single application?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitTemplateObservation.java
```diff
@@ -0,0 +1,72 @@
...
+	TEMPLATE_OBSERVATION {
+
+		@Override
+		public String getName() {
+			return "spring.rabbit.template";
+		}
+
+		@Override
+		public String getContextualName() {
+			return "RabbitTemplate Observation";
```
marcingrzejszczak, line 43: Same comment as above

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitTemplateObservationConvention.java
```diff
@@ -0,0 +1,71 @@
...
+
+	@Nullable
+	private final KeyValues highCardinality;
+
+	/**
+	 * Create an instance with the provided {@link KeyValues}.
+	 * @param lowCardinality the low cardinality {@link KeyValues}.
+	 * @param highCardinality the high cardinality {@link KeyValues}.
+	 */
+	public RabbitTemplateObservationConvention(@Nullable KeyValues lowCardinality,
```
marcingrzejszczak, line 47: Same comment as above in terms of additional keyvalues

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitListenerObservation.java
artembilan: Well, this sounds like that `listenerId` has to be present in this contextual name instead of the tag (or in addition to).

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/support/micrometer/ObservationTests.java
```diff
@@ -0,0 +1,195 @@
...
+import io.micrometer.tracing.test.simple.SimpleTracer;
+
+/**
+ * @author Gary Russell
+ * @since 3.0
+ *
+ */
+@SpringJUnitConfig
+@RabbitAvailable(queues = { "observation.testQ1", "observation.testQ2" })
+public class ObservationTests {
```
marcingrzejszczak, line 64: I suggest using the SampleTestRunner as presented here https://micrometer.io/docs/tracing#_running_integration_tests to have all the machinery setup for sending spans to latency visualization tools (like Zipkin). That way you can see what you're doing and whether it's useful.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitListenerObservation.java
marcingrzejszczak: Why? For the standard name that would be literally `metrics.rabbitmq.listener`. For the contextual name you'd need to pass to the context the `destination name` (queue from rabbit template?) and `operation name` like (send / receive)

artembilan: I also see that there is a `Description` in those metrics. And we indeed add something like that to the `Timer`.
But I couldn't find anything similar in the `Observation` API. 
I may guess that this `contextualName` may be mapper.
But all of these are just speculations from my ignorance.

(Sorry for hijacking this PR for not related discussion 😄 )

marcingrzejszczak: These are very good questions. Since @shakuzen was even thinking about removing `description` from metrics (that was before we decided to go with 1.10 instead of 2.0) this is why the Observation doesn't have a notion of a description and we don't set the description on a metric.

garyrussell: I have just noticed that as well as sender and receiver contexts, there are also `RequestReplySenderContext` and `RequestReplyReceiverContext`.

I can probably support the sender side (where we know whether we are doing request/reply), but on the receiver side, the place where we instrument the call has no idea whether or not the downstream code is request only or request/reply. That will not be easy to change.

artembilan: I see. So, the name probably has to be `messaging.rabbitmq.listener`.
And the contextual name as you said: a destination we are listening from and `receive` as an operation.
An this could be obtained via `ObservationConvention.getContextualName(T context)`.
We just have to populate that destination to the context, therefore.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitListenerObservationConvention.java
garyrussell: Ok.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitListenerObservationConvention.java
garyrussell: Usually only a few; probably low teens at the most.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitListenerObservation.java
marcingrzejszczak: Even if we do instrument at a different place I wonder if that will change anything for tools like Zipkin. What I mean is that if we have a `RequestReply` case, that implies the kind type to be `CLIENT` and `SERVER`. Zipkin never assumed that there will be a broker in between. For messaging theoretically you should always be using the `SenderContext` and `ReceiverContext` cause those use kind `PRODUCER` and `CONSUMER` that are desgined for messaging. I think we should NOT be using the `RequestReply...` versions cause we can break the latency visualization tools (I'm talking here about tracing only of course) cc @shakuzen @jonatan-ivanov

artembilan: Yes, I also wanted to say that request-reply is an artificial feature for Messaging per se.
It is always a produce on one side and consumer on another. And then we got opposite direction for a reply process.
It might look strange for the first producer side when we block its observation waiting for a reply where ,probably, its consumer span going to look as a child...

marcingrzejszczak: Ok so in fact the request-reply is a fire and forget but in the other direction. So indeed we shouldn't be using the `RequestReply...` context but the standard `Sender...` and `Receiver...` options

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitListenerObservationConvention.java
marcingrzejszczak: Ok so that's indeed low-cardinality :+1:

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -1540,8 +1540,10 @@ protected void executeListener(Channel channel, Object data) {
 			observation = Observation.NOOP;
 		}
 		else {
+			Message message = (Message) data;
 			observation = RabbitListenerObservation.LISTENER_OBSERVATION.observation(registry,
-						new RabbitMessageReceiverContext((Message) data, getListenerId()))
+						new RabbitMessageReceiverContext(message, getListenerId(),
+								message.getMessageProperties().getConsumerQueue()))
```
artembilan, line 1546: May we just do this in the target `RabbitMessageReceiverContext`?
We already provide a `Message` over there...

artembilan (changes requested):

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitListenerObservation.java
garyrussell: I am getting a bit confused - if the contextual name must be retrieved from the context, why is there a getter on the observation itself (where there is no context).

Does this commit get us closer to what you expect?

https://github.com/spring-projects/spring-amqp/pull/1500/commits/c7d1821a47feb842ed8a13817fee758d7eb840aa

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
garyrussell: But it's not available in `getContextualName()`.

garyrussell: Oh; sorry - I see what you mean.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitListenerObservation.java
marcingrzejszczak: The contextual name should be applied on the Convention, you should not override any methods on the context itself. The idea behind `ObservationConvention` is that it contains all the information about naming and tagging (keyvalues). So that if users want to change these attributes they inject their own convention instead of chanhing anything in the production code.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -258,13 +267,19 @@ public class RabbitTemplate extends RabbitAccessor // NOSONAR type line count
 
 	private ErrorHandler replyErrorHandler;
 
+	private boolean useChannelForCorrelation;
+
+	private boolean observationEnabled;
+
+	private RabbitTemplateObservationConvention observationConvention = new RabbitTemplateObservationConvention();
```
artembilan, line 274: @marcingrzejszczak ,

is this what you have expected?

I thought the idea was just about a general `ObservationConvention<RabbitMessageSenderContext>` and that's it...

artembilan (approved):
LGTM.
It feels a bit involved with all those extra layers of enums and conventions, but I'm still learning this stuff and cannot be treated as a genuine source of opinion. 😸

marcingrzejszczak:
>  It feels a bit involved with all those extra layers of enums and conventions, but I'm still learning this stuff and cannot be treated as a genuine source of opinion. smile_cat

Enums here are like `@Configuration` for Spring. They separate what you want to have as observations, how they should be named and tagged from their actual usage. An additional advantage is that due to using enums you can automatically generate documentation of  your observations https://micrometer.io/docs/tracing#_documentation_building

A convention is the way to actually apply the naming and tagging. The way we designed things allow you to easily locally or globally alter that behaviour.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitMessageReceiverContext.java
```diff
@@ -0,0 +1,52 @@
...
+		this.message = message;
+		this.listenerId = listenerId;
+	}
+
+	public String getListenerId() {
+		return this.listenerId;
+	}
+
+	@Override
+	public String getContextualName() {
```
marcingrzejszczak, line 48: You should not set it here. You should set the contextual name in the `RabbitListenerObservationConvention`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitTemplateObservation.java
```diff
@@ -0,0 +1,67 @@
...
+public enum RabbitTemplateObservation implements DocumentedObservation {
+
+	/**
+	 * {@link org.springframework.kafka.core.KafkaTemplate} observation.
+	 */
+	TEMPLATE_OBSERVATION {
+
+		@Override
+		public String getName() {
+			return "spring.rabbit.template";
```
marcingrzejszczak, line 38: You can use the `getDefaultConvention` method - https://github.com/spring-cloud/spring-cloud-task/blob/main/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/observation/TaskDocumentedObservation.java#L32 which gives you better customization options

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitMessageReceiverContext.java
marcingrzejszczak: I mean you can override everything but it was never intended to be done like that. That's because users won't be able to easily customize the contextual name.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitTemplateObservation.java
artembilan: Hold on. Then it means that my question about `RabbitTemplateObservationConvention` as a custom convention from end-user remains opened.
According that task feature we should have an extra interface and have it default impl for use-cases like this and contextual name.
The interface would be used as a point of customization from end-user, not a class as it is right now.

marcingrzejszczak: Yeah I modified my response here https://github.com/spring-projects/spring-amqp/pull/1500#discussion_r966029706

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitMessageReceiverContext.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.springframework.amqp.core.Message;
+
+import io.micrometer.observation.transport.ReceiverContext;
+
+/**
+ * {@link ReceiverContext} for {@link Message}s.
+ *
+ * @author Gary Russell
+ * @since 2.8
```
artembilan, line 27: `3.0`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitTemplateObservation.java
```diff
@@ -0,0 +1,74 @@
...
+import io.micrometer.observation.docs.DocumentedObservation;
+
+/**
+ * Spring RabbitMQ Observation for {@link org.springframework.amqp.rabbit.core.RabbitTemplate}.
+ *
+ * @author Gary Russell
+ * @since 3.0
+ *
+ */
+public enum RabbitTemplateObservation implements DocumentedObservation {
```
artembilan, line 31: You know, in Spring Integration I'm planning to have all the observation impls in one class.
WDYT?

artembilan (changes requested):

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitTemplateObservation.java
garyrussell: For what purpose?

artembilan: Just to have all the observations in one place. The `enum` is not restricted only to a single entry...
I'm not sure how the final doc is going to look, but that's my thought: have all the built-in observations in one place.

@marcingrzejszczak , what is your idea on the matter, please?

artembilan (approved):
LGTM, unless we really have to go separate convention instances for default and custom one like I did in my Spring Integration PR: https://github.com/spring-projects/spring-integration/pull/3879/files#diff-20b96284d0af70c1c3003cf5fcdad335d6f189cb3dbe3070f6c77fd7fa05937eR84.

I took the logic from Spring Cloud Task mentioned in this Spring AMQP PR.

garyrussell:
In your case, that makes sense because `this.observationConvention` can be null.

I don't allow it to be null; I think that makes for a cleaner API.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitTemplateObservation.java
marcingrzejszczak: You don't need to have them in one class, we will do the scan of all of your sources and merge the results together (https://micrometer.io/docs/tracing#_documentation_building)

marcingrzejszczak:
I left a comment here https://github.com/spring-projects/spring-amqp/pull/1500/files#r965219710 about testing against Zipkin to see how the whole trace looks like there

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitListenerObservationConvention.java
```diff
@@ -29,6 +29,11 @@
  */
 public class RabbitListenerObservationConvention implements ObservationConvention<RabbitMessageReceiverContext> {
 
+	/**
+	 * A singleton instance of the convention.
+	 */
+	public static RabbitListenerObservationConvention INSTANCE = new RabbitListenerObservationConvention();
```
artembilan, line 35:
```suggestion
	public static final RabbitListenerObservationConvention INSTANCE = new RabbitListenerObservationConvention();
```

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitTemplateObservationConvention.java
```diff
@@ -29,6 +29,11 @@
  */
 public class RabbitTemplateObservationConvention implements ObservationConvention<RabbitMessageSenderContext> {
 
+	/**
+	 * A singleton instance of the convention.
+	 */
+	public static RabbitTemplateObservationConvention INSTANCE = new RabbitTemplateObservationConvention();
```
artembilan, line 35:
```suggestion
	public static final RabbitTemplateObservationConvention INSTANCE = new RabbitTemplateObservationConvention();
```

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -1186,8 +1188,7 @@ public void setObservationEnabled(boolean observationEnabled) {
 	 * @param observationConvention the convention.
 	 * @since 3.0
 	 */
-	public void setObservationConvention(RabbitListenerObservationConvention observationConvention) {
-		Assert.notNull(observationConvention, "'observationConvention' cannot be null");
+	public void setObservationConvention(ObservationConvention<RabbitMessageReceiverContext> observationConvention) {
```
artembilan, line 1191: I'm not sure how is that harmful to expect a generic convention, but in Spring Integration I extract an interface like this: https://github.com/spring-projects/spring-integration/pull/3879/files#diff-ae850c6c6d3c04d681a2509ef10dae6386011a072e47e761be4e320b3245539aR32.
Pay attention to that `supportsContext()` method impl.

@marcingrzejszczak , what is the proper path to go?

artembilan (changes requested):

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
garyrussell: That makes sense.

artembilan (approved):
LGTM

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/support/micrometer/ObservationIntegrationTests.java
```diff
@@ -0,0 +1,159 @@
...
+ * @author Gary Russell
+ *
+ * @since 3.0
+ */
+@RabbitAvailable(queues = { "int.observation.testQ1", "int.observation.testQ2" })
+public class ObservationIntegrationTests extends SampleTestRunner {
+
+	@Override
+	public TracingSetup[] getTracingSetup() {
+		return new TracingSetup[]{ TracingSetup.IN_MEMORY_BRAVE };
```
marcingrzejszczak, line 58: You don't have to explicitly set this - if you don't have Zipkin running or Wavefront configured they will be ignored. Why have you ignored the OTel tests?

garyrussell: Thanks; that wasn't obvious from the docs.

marcingrzejszczak: https://micrometer.io/docs/tracing#_running_integration_tests

```
 -   by asserting spans that were stored without emitting them to a reporting system

-    against running Tanzu Observability by Wavefront instance (this option turns on when you have passed the Wavefront related configuration in the constructor - otherwise the test will be disabled)

-    against running Zipkin instance (this option turns on when Zipkin is running - otherwise the test will be disabled)

```

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitMessageSenderContext.java
```diff
@@ -0,0 +1,52 @@
...
+		this.destination = destination;
+	}
+
+	public String getBeanName() {
+		return this.beanName;
+	}
+
+	@Override
+	public String getContextualName() {
+		return this.destination + " send";
```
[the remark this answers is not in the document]
artembilan: Looks like this one is missed for a fix.

artembilan (changes requested):
We probably also need some docs for this new stuff.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitMessageSenderContext.java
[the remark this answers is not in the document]
garyrussell: ok

artembilan (approved):

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/support/micrometer/ObservationIntegrationTests.java
marcingrzejszczak: So `testq2 send` is a child of a `testq1 receive`  ? If that's the case then it looks good!

In some cases the name has `/` at the beginning - should it look like this?

garyrussell: Yes, the test1 receiver forwards the message to test2 which is received by the second receiver.

Yes, with RabbitMQ, the sender doesn't know the destination queue; it sends to `exchange/routingKey`. In this case, I am sending to the default exchange `""` with a routing key equal to the queue name (which is a default binding provided by rabbitmq).

The `/` is not present in the Kafka PR because the sender knows the destination topic name.

src/reference/asciidoc/amqp.adoc
```diff
@@ -3764,6 +3764,23 @@ The timers are named `spring.rabbitmq.listener` and have the following tags:
...
+Also see <<observation>>.
+
+[[observation]]
+===== Micrometer Observation
+
+Using Micrometer for observation is now supported, since version 3.0, for the `RabbitTemplate` and listener containers.
+
+Set `observationEnabled` on each component to enable observation; this will disable <<micrometer,Micrometer Timers>> because the timers will now be managed with each observation.
+
+Refer to https://micrometer.io/docs/tracing#_running_integration_tests[Micrometer Tracing] for more information.
```
artembilan, line 3776: Why do you point here to the testing docs?
Why just not use a root link instead - https://micrometer.io/docs/tracing ?
```suggestion
Refer to https://micrometer.io/docs/tracing[Micrometer Tracing] for more information.
```

artembilan (changes requested):

src/reference/asciidoc/amqp.adoc
garyrussell: Oops - had two pages open and copied the wrong link 😄

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/DefaultRabbitListenerObservationConvention.java
```diff
@@ -0,0 +1,52 @@
...
+
+	@Override
+	public KeyValues getLowCardinalityKeyValues(RabbitMessageReceiverContext context) {
+		return KeyValues.of(RabbitListenerObservation.ListenerLowCardinalityTags.LISTENER_ID.asString(),
+						context.getListenerId());
+	}
+
+	@Override
+	public KeyValues getHighCardinalityKeyValues(RabbitMessageReceiverContext context) {
+		return KeyValues.empty();
```
artembilan, line 44: Same here about `default` methods in the super interface.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/DefaultRabbitTemplateObservationConvention.java
```diff
@@ -0,0 +1,52 @@
...
+
+	@Override
+	public KeyValues getLowCardinalityKeyValues(RabbitMessageSenderContext context) {
+		return KeyValues.of(RabbitTemplateObservation.TemplateLowCardinalityTags.BEAN_NAME.asString(),
+						context.getBeanName());
+	}
+
+	@Override
+	public KeyValues getHighCardinalityKeyValues(RabbitMessageSenderContext context) {
+		return KeyValues.empty();
```
artembilan, line 44: DITTO

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitTemplateObservation.java
```diff
@@ -0,0 +1,74 @@
...
+	public enum TemplateLowCardinalityTags implements KeyName {
+
+		/**
+		 * Bean name of the template.
+		 */
+		BEAN_NAME {
+
+			@Override
+			public String asString() {
+				return "bean.name";
```
artembilan, line 67: Same here about prefixing tags: https://github.com/spring-projects/spring-integration/pull/3879#pullrequestreview-1104184375

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitMessageSenderContext.java
```diff
@@ -0,0 +1,60 @@
...
+		this.destination = destination;
+	}
+
+	public String getBeanName() {
+		return this.beanName;
+	}
+
+	@Override
+	public String getName() {
+		return "spring.rabbit.template";
```
artembilan, line 49: I used the name directly on the `DocumentedObservation` impl instead of context.
Not sure what is the proper way to go though...
@marcingrzejszczak ?

Thanks

artembilan (changes requested):

artembilan (approved):

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitMessageSenderContext.java
marcingrzejszczak: The rule of thumb should be

- always try to use the names from a convention
- if for some reason it doesn't make sense (you would have a lot of conventions cause you have a lot of contexts that don't differ too much) you can set one on a DocumentedObservation

marcingrzejszczak:
Also we should double check that if we have 2 separate processes then the trace gets continued. It means that on the `SenderContext` you would need to set the `remoteServiceName` to be the name of the broker (e.g. `rabbitmq`). Then on the recipient side on the `ReceiverContext` you would need to se the `remoteServiceName` to be `rabbitmq` too. That way you will see in Zipkin in the dependency graph that application `A` is talking to a `rabbitmq` service that then talks to service `B`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -2378,14 +2418,34 @@ public void doSend(Channel channel, String exchangeArg, String routingKeyArg, Me
...
+		}
+		Observation observation;
+		ObservationRegistry registry = getObservationRegistry();
+		if (!this.observationEnabled || registry == null) {
+			observation = Observation.NOOP;
+		}
+		else {
+			observation = RabbitTemplateObservation.TEMPLATE_OBSERVATION.observation(this.observationConvention,
+					DefaultRabbitTemplateObservationConvention.INSTANCE,
+						new RabbitMessageSenderContext(message, this.beanName, exch + "/" + rKey), registry);
```
shakuzen, line 2443: If we took a `Supplier<Context>` here instead of `Context` where the supplier was only called if `ObservationRegistry` wasn't a no-op, and therefore the `Context` object wouldn't be allocated in the no-op case, would that simplify things for you? I think then you wouldn't need the if-else this is contained in, right?

artembilan: +1 from me as well. And I probably agree with @marcingrzejszczak as well.
Only the problem that @garyrussell is off today and next Monday is release day.
I can fix this quickly myself, but we have to be sure that this one is good to go for merging.

Thank you!

garyrussell: Or, we could just use a singleton for the No-op case.

garyrussell:
@marcingrzejszczak I don't see a `remoteServiceName` on the `ReceiverContext`, only on the `SenderContext`.

Also, should the service name just be `rabbitmq` or should it include, for example, the ip address of the actual broker?

garyrussell:
Forget the second comment in my last post - it wouldn't work because, with a cluster, they might be connected to different nodes.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
artembilan: There is one, Gary, in the `DocumentedObservation`: 
```
    default <T extends Observation.Context> Observation createNotStarted(
            @Nullable ObservationConvention<T> customConvention, @NonNull ObservationConvention<T> defaultConvention,
            @NonNull Supplier<T> contextSupplier, @NonNull ObservationRegistry registry) {
        if (registry.isNoop()) {
            return Observation.NOOP;
        }
        return observation(customConvention, defaultConvention, contextSupplier.get(), registry);
    }
```
I'm not sure why name is like that, but looks like it does exactly what is expected from us here.

marcingrzejszczak: We're working on unifying the names. Sorry about the commotion cc @ttddyy

marcingrzejszczak:
> @marcingrzejszczak I don't see a remoteServiceName on the ReceiverContext, only on the SenderContext.

That must be our mistake - I'll fix it

PR #1501, opened 2022-09-07
GH-1465: Part II - Super Stream SAC

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1465

Add support for single active consumers on super streams.

Stop containers in test.

Use Snapshot Repo

Use snapshot repo; use TestContainers.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -58,7 +58,7 @@ ext {
 	micrometerVersion = '1.10.0-SNAPSHOT'
 	micrometerTracingVersion = '1.0.0-SNAPSHOT'
 	mockitoVersion = '4.5.1'
-	rabbitmqStreamVersion = '0.4.0'
+	rabbitmqStreamVersion = '0.7.0'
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -105,7 +105,7 @@ allprojects {
 		maven { url 'https://repo.spring.io/libs-milestone' }
 		if (version.endsWith('-SNAPSHOT')) {
 			maven { url 'https://repo.spring.io/libs-snapshot' }
-			maven { url 'https://oss.sonatype.org/content/repositories/snapshots' }
+			// maven { url 'https://oss.sonatype.org/content/repositories/snapshots' }
```
artembilan, line 108: May be there is just enough to remove this repo altogether?

artembilan (changes requested):

build.gradle
garyrussell: Problem is remembering where it is the next time I have to use a snap.

Issue #1507, opened 2022-09-28
Fix Class Tangles in 3.0


garyrussell:
<img width="328" alt="Screen Shot 2022-09-28 at 4 55 16 PM" src="https://user-images.githubusercontent.com/483832/192890996-29d2046a-22ee-4991-ac65-9aa1e268bbf7.png">
<img width="334" alt="Screen Shot 2022-09-28 at 4 55 09 PM" src="https://user-images.githubusercontent.com/483832/192890997-935b73a6-0503-4015-af8b-1f6c47818645.png">

PR #1508, opened 2022-10-03
GH-1444: Observablility Documentation, Polish

garyrussell:
Refactor for latest snapshots; add documentation generation. Use NOOP registry and supplier for context.

Also fix class tangles.

Resolves https://github.com/spring-projects/spring-amqp/issues/1507

Disable auto doc generation and polish manually.


spring-rabbit-stream/src/test/java/org/springframework/rabbit/stream/listener/RabbitListenerTests.java
```diff
@@ -99,6 +100,7 @@ void nativeMsg(@Autowired RabbitTemplate template) throws InterruptedException {
 	}
 
 	@Test
+	@Disabled("Temporary until SF uses Micrometer snaps")
```
artembilan, line 103: I don't think we need this since this PR is going to be merged already after respective change in the SF.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -2428,21 +2429,16 @@ public void doSend(Channel channel, String exchangeArg, String routingKeyArg, Me
...
 		ObservationRegistry registry = getObservationRegistry();
-		if (!this.observationEnabled || registry == null) {
-			observation = Observation.NOOP;
-		}
-		else {
-			observation = RabbitTemplateObservation.TEMPLATE_OBSERVATION.observation(this.observationConvention,
-					DefaultRabbitTemplateObservationConvention.INSTANCE,
+		Observation observation = RabbitTemplateObservation.TEMPLATE_OBSERVATION.observation(this.observationConvention,
+				DefaultRabbitTemplateObservationConvention.INSTANCE,
+					(Supplier<RabbitMessageSenderContext>) () ->
```
artembilan, line 2439: With the latest Observation SNAPSHOT we don't need an explicit cast - they just remove the method with `Context`. Just `Supplier` is possible from now on.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -1536,15 +1539,11 @@ protected void invokeErrorHandler(Throwable ex) {
...
-			observation = Observation.NOOP;
-		}
-		else {
-			Message message = (Message) data;
-			observation = RabbitListenerObservation.LISTENER_OBSERVATION.observation(this.observationConvention,
-					DefaultRabbitListenerObservationConvention.INSTANCE,
+		Message message = (Message) data;
+		observation = RabbitListenerObservation.LISTENER_OBSERVATION.observation(this.observationConvention,
+				DefaultRabbitListenerObservationConvention.INSTANCE,
+					(Supplier<RabbitMessageReceiverContext>) () ->
```
artembilan, line 1545: DITTO

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitListenerObservation.java
```diff
@@ -72,4 +73,28 @@ public String asString() {
...
+
+		/**
+		 * A singleton instance of the convention.
+		 */
+		public static final DefaultRabbitListenerObservationConvention INSTANCE =
+				new DefaultRabbitListenerObservationConvention();
+
+		@Override
+		public KeyValues getLowCardinalityKeyValues(RabbitMessageReceiverContext context) {
+			return KeyValues.of(RabbitListenerObservation.ListenerLowCardinalityTags.LISTENER_ID.asString(),
```
artembilan, line 89: FYI: I have fixed this tangle just using literals instead of enums: https://github.com/spring-projects/spring-integration/pull/3896/files#diff-1966bee1c00ff5bd693214849eb988472f9f86f986023f598cc8d344a7740afbR39

artembilan (changes requested):
Also you have many failed tests in GH actions report.

spring-rabbit-stream/src/test/java/org/springframework/rabbit/stream/listener/RabbitListenerTests.java
garyrussell: We need it (temporarily) so that the PR build will succeed.

artembilan (approved):
LGTM.
Will merge when greed GH actions build.

garyrussell:
Green finally; will address the logback and web client issues after merge.

Issue #1419, opened 2022-02-01
Remove HOP Dependency (2.4.x)


garyrussell:
Hop will no longer support JDK 8 before the end of life of 2.4.x; there is only one hard dependency on Hop (in `LocalizedQueueConnectionFactory` to determine the hosting node of a classic mirrored queue).

acogoluegnes:
`LocalizedQueueConnectionFactory` uses only the `/queues/{vhost}/{queue}` endpoint. The easiest would be to use the `RestTemplate` and Jackson, to keep the same dependencies as Hop.

PR #1511, opened 2022-10-10
GH-1419: Remove RabbitMQ http-client Usage

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1419

Use Spring WebFlux instead, while allowing the user to choose some other technology in the `LocalizedQueueConnectionFactory`.

**I will backport after review/merge**

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -44,7 +44,6 @@ ext {
 	assertkVersion = '0.24'
 	awaitilityVersion = '4.2.0'
 	commonsCompressVersion = '1.20'
-	commonsHttpClientVersion = '5.1.3'
 	commonsPoolVersion = '2.11.1'
 	googleJsr305Version = '3.0.2'
 	hamcrestVersion = '2.2'
@@ -387,7 +386,6 @@ project('spring-rabbit') {
 		api 'org.springframework:spring-context'
 		api 'org.springframework:spring-messaging'
 		api 'org.springframework:spring-tx'
-		optionalApi 'org.springframework:spring-web'
 		optionalApi 'org.springframework:spring-webflux'
 		optionalApi 'io.projectreactor:reactor-core'
 		optionalApi "ch.qos.logback:logback-classic:$logbackVersion"
@@ -409,8 +407,6 @@ project('spring-rabbit') {
 		testImplementation 'io.micrometer:micrometer-tracing-bridge-brave'
 		testImplementation 'io.micrometer:micrometer-tracing-test'
 		testImplementation 'io.micrometer:micrometer-tracing-integration-test'
-		testRuntimeOnly 'org.springframework:spring-web'
-		testRuntimeOnly "org.apache.httpcomponents.client5:httpclient5:$commonsHttpClientVersion"
 		testRuntimeOnly 'com.fasterxml.jackson.core:jackson-core'
 		testRuntimeOnly 'com.fasterxml.jackson.core:jackson-databind'
 		testRuntimeOnly 'com.fasterxml.jackson.dataformat:jackson-dataformat-xml'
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -384,11 +383,12 @@ project('spring-rabbit') {
 
 		api project(':spring-amqp')
 		api "com.rabbitmq:amqp-client:$rabbitmqVersion"
-		optionalApi "com.rabbitmq:http-client:$rabbitmqHttpClientVersion"
 		optionalApi 'org.springframework:spring-aop'
 		api 'org.springframework:spring-context'
 		api 'org.springframework:spring-messaging'
 		api 'org.springframework:spring-tx'
+		optionalApi 'org.springframework:spring-web'
```
artembilan, line 390: This one comes as a transitive dep from WebFlux: `api(project(":spring-web"))`.
So, why to use it explicitly?

build.gradle
```diff
@@ -410,7 +410,7 @@ project('spring-rabbit') {
 		testImplementation 'io.micrometer:micrometer-tracing-test'
 		testImplementation 'io.micrometer:micrometer-tracing-integration-test'
 		testRuntimeOnly 'org.springframework:spring-web'
-		testRuntimeOnly "org.apache.httpcomponents:httpclient:$commonsHttpClientVersion"
+		testRuntimeOnly "org.apache.httpcomponents.client5:httpclient5:$commonsHttpClientVersion"
```
artembilan, line 413: Why doesn't `reactor-netty-http` works here for us?
I mean it is still OK since it is a test dep, but probably with Reactor in hands we may avoid and extra dep to manage.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/LocalizedQueueConnectionFactory.java
```diff
@@ -372,4 +343,113 @@ public void destroy() {
 		resetConnection();
 	}
 
+	/**
+	 * Used to obtain a connection factory for the queue leader.
+	 *
+	 * @param <T> the client type.
+	 * @since 2.4.8
```
artembilan, line 350: Why have you decided to back-port this fix?
Do you have an evidence that Spring Framework 5.3.x has moved to Commons Http Client 5.0?

artembilan (changes requested):
Just a couple questions.
And most important the reason behind back-porting this fix...

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/LocalizedQueueConnectionFactory.java
garyrussell: My plan is to back port but with the default `NodeLocator` being a `RabbitMQHopNodeLocator` (immediately deprecated) and make the `WebFluxNodeLocator` an optional component.

See the issue; the problem is HOP will be out of support before 2.4.x is, so we will need an option in 2.4.x that doesn't depend on HOP, in case of future REST changes in RabbitMQ.

build.gradle
garyrussell: Not sure; there was some class in it that wasn't on the CP until I added this; will try to remove.

build.gradle
garyrussell: I will try with that.

acogoluegnes:
I'd recommend going with Reactor Netty, you'll pull fewer dependencies. And IIRC Spring AMQP uses only 1 REST endpoint, so its needs are limited.

Note Hop uses [Reactor Netty](https://github.com/rabbitmq/hop/blob/main/src/main/java/com/rabbitmq/http/client/ReactorNettyClient.java) as well if you want some inspiration.

We had [a discussion](https://github.com/rabbitmq/hop/issues/122) years ago and the gist was that using `WebClient` was not a good call for Hop. Maybe things have changed and Spring AMQP is part of the Spring portfolio, so that's not such a big deal.

PR #1512, opened 2022-10-11
GH-1419: Add RestTemplateNodeLocator

garyrussell:
- also remove hard dependency on `spring-webflux` from `spring-rabbit-junit`.


artembilan:
@garyrussell ,

do we back-port exactly this fix?
Or what is your plan?
Thanks

garyrussell:
Yes; as you have seen; the back port includes this change (but changes the broker condition to use RestTemplate instead of the JVM HttpClient.

PR #1516, opened 2022-10-12
GH-1419: Fix Early Exit in NodeLocator

garyrussell:
If a node was returned by the REST call and the node was not in the map of nodes to addreses, the loop exited early.

The incorrect variable was being tested (never null).

Also add a more sophisticated integration test - using 2 brokers, ensure that the correct broker is located for the queue.

**I will back port to 2.4.x after merge**

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/LocalizedQueueConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/LocalizedQueueConnectionFactory.java
@@ -398,7 +398,7 @@ public class LocalizedQueueConnectionFactory implements ConnectionFactory, Routi
 						String node = (String) queueInfo.get("node");
 						if (node != null) {
 							String nodeUri = nodeToAddress.get(node);
-							if (uri != null) {
+							if (nodeUri != null) {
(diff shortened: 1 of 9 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/LocalizedQueueConnectionFactory.java
```diff
@@ -398,7 +398,7 @@ default ConnectionFactory locate(String[] adminUris, Map<String, String> nodeToA
 						String node = (String) queueInfo.get("node");
 						if (node != null) {
 							String nodeUri = nodeToAddress.get(node);
-							if (uri != null) {
+							if (nodeUri != null) {
```
artembilan, line 401: I think only this line change is a back-port candidate. 😄

garyrussell: Agreed plus maybe the NodeLocatorTest.

garyrussell:
Back ported just the `NodeLocator` fix and test). 17134529a0ede41bc3a3767a56093ed773a10e96

PR #1523, opened 2022-10-20
GH-1477: Reduce Log Noise While Broker Down

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/1477

Only log the stack trace on the first failure.

Tested with a Boot app and stop/start/stop the broker.

**cherry-pick to 2.4.x**


artembilan:

PR #1529, opened 2022-10-28
GH-1528: Fix possible type pollution in RabbitListenerAnnotationBeanPostProcessor

### LINKED ISSUE #1528: Fix type pollution RabbitListenerAnnotationBeanPostProcessor
Hi 👋 

I've been recently playing with the https://github.com/RedHatPerf/type-pollution-agent from @franz1981 and found that `RabbitListenerAnnotationBeanPostProcessor` - more specifically `findListenerAnnotations` -  seems to be causing the TOP 3 for us during startup.

```
3:      java.util.ArrayList
Count:  16432
Types:
        java.util.List
        java.util.Collection
        java.lang.Iterable
Traces:
        org.springframework.amqp.rabbit.annotation.RabbitListenerAnnotationBeanPostProcessor.findListenerAnnotations(Rabbi
tListenerAnnotationBeanPostProcessor.java:345)
                class: java.util.Collection
                count: 8123
        org.springframework.core.annotation.TypeMappedAnnotations.getAggregates(TypeMappedAnnotations.java:226)
                class: java.util.List
                count: 7905
...
```

The method returns `Collection` (and has corresponding type cast internally), while it returns a  `List` really. Everyone else in the traces seems to use largely `List` as well (see the next trace e.g.). Even if this is a false-positive in terms of actual type pollution it seems like a simple enough fix to be on the safe side. I'll provide a PR shortly where further discussion can happen, but I needed this issue to stick to the commit convention ;-)

Cheers,
Christoph

dreis2211:
Hi,

this PR fixes #1528 by returning `List` rather than `Collection` from `findListenerAnnotations`.
Regardless of the mentioned issue imho that's the better return type in any case for this method.

If possible, a backport to 2.x would be highly appreciated

Cheers,
Christoph

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
@@ -323,12 +323,12 @@ public class RabbitListenerAnnotationBeanPostProcessor
 	}
 
 	private TypeMetadata buildMetadata(Class<?> targetClass) {
-		Collection<RabbitListener> classLevelListeners = findListenerAnnotations(targetClass);
+		List<RabbitListener> classLevelListeners = findListenerAnnotations(targetClass);
 		final boolean hasClassLevelListeners = classLevelListeners.size() > 0;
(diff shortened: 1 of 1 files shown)


garyrussell:
Thanks for the contribution; in future please limit commit headlines to 50 chars.

Merged as 41802039a and cherry-picked to 2.4.x

PR #1530, opened 2022-11-08
Add spring-amqp-bom

garyrussell:


build.gradle
```diff
@@ -717,7 +746,7 @@ task distZip(type: Zip, dependsOn: [docsZip, schemaZip]) {
 		into "${baseDir}/schema"
 	}
 
-	subprojects.each { subproject ->
+	javaProjects.each { subproject ->
```
garyrussell, line 749: @artembilan - copied from SI - is it correct that the bom is not included in `distZip.zip`?

artembilan: I don't think so.
Not sure, though, if anyone uses this zip, but for consistency it is probably better to include that BOM file as well.

garyrussell: It doesn't seem to get staged anywhere, just in the .m2 repo with `publishToMavenLocal` so I don't see where to copy it from.

Let's just stick with this.

artembilan: See my solution in SI: https://github.com/spring-projects/spring-integration/pull/3940

build.gradle
```diff
@@ -717,13 +746,19 @@ task distZip(type: Zip, dependsOn: [docsZip, schemaZip]) {
...
 		into ("${baseDir}/libs") {
 			from subproject.jar
 			from subproject.sourcesJar
 			from subproject.javadocJar
 		}
 	}
+
+	from(project(':spring-amqp-bom').generatePomFileForMavenJavaPublication) {
+		into "${baseDir}/libs"
+		rename 'pom-default.xml', "spring-integration-bom-${project.version}.xml"
```
artembilan, line 759:
```suggestion
		rename 'pom-default.xml', "spring-amqp-bom-${project.version}.xml"
```

artembilan (changes requested):

build.gradle
garyrussell: Doh

settings.gradle
```diff
@@ -1,6 +1,7 @@
 rootProject.name = 'spring-amqp-dist'
 
 include 'spring-amqp'
+include 'spring-amqp-bom'
```
artembilan, line 4: Not sure how that works, but I think this module's dir is empty, so that's why it does not appear in the PR content.

artembilan (changes requested):

settings.gradle
artembilan: So, looks like we do just this on GH action:
```
        uses: burrunan/gradle-cache-action@v1
        with:
          arguments: check
```
which really does not involve all the resources with have in the project.

Should we revise it to `build dist` instead to be sure in the docs and all the artifacts we may build in the end?

garyrussell: Strsnge - I have `spring-amqp-bom.txt` in the IDE (copied from SI), but I don't see it in the dir; investigating...

PR #1535, opened 2022-11-16
Fix Class Tangle

garyrussell:
Between `LocalizedQueueConnectionFactory` and `NodeLocator` implementations.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/ConnectionFactoryUtils.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/ConnectionFactoryUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2021 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -25,6 +25,7 @@ import org.springframework.transaction.support.ResourceHolderSynchronization;
 import org.springframework.transaction.support.TransactionSynchronization;
 import org.springframework.transaction.support.TransactionSynchronizationManager;
 import org.springframework.util.Assert;
+import org.springframework.util.ClassUtils;
 
 import com.rabbitmq.client.Channel;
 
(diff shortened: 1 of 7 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/ConnectionFactoryUtils.java
```diff
@@ -43,6 +44,13 @@
  */
 public final class ConnectionFactoryUtils {
 
+	private static final boolean USING_WEBFLUX;
+
+	static {
+		USING_WEBFLUX = ClassUtils.isPresent("org.springframework.web.reactive.function.client.WebClient",
```
artembilan, line 50: According to the AOT requirements and to avoid extra reflection info and execution ticks at runtime, this kind of constants should be named with a `_PRESENT` prefix.

I'm not sure also why you need this extra `static` block.
In my code I have it like this:
```
public static final boolean WEB_FLUX_PRESENT =
			ClassUtils.isPresent("org.springframework.web.reactive.DispatcherHandler", null);
```

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/NodeLocator.java
```diff
@@ -0,0 +1,121 @@
...
+import org.springframework.amqp.AmqpException;
+import org.springframework.core.log.LogAccessor;
+import org.springframework.lang.Nullable;
+
+/**
+ * Used to obtain a connection factory for the queue leader.
+ * @param <T> the client type.
+ *
+ * @author Gary Russell
+ * @since 2.4.8
```
artembilan, line 34: `3.0`?
Or do you really want a back-port for this change?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/FactoryFinder.java
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.amqp.rabbit.connection;
+
+/**
+ * Callback to determine the connection factory using the provided information.
+ *
+ * @author Gary Russell
+ * @since 2.4.8
```
artembilan, line 23: `3.0`?

artembilan (changes requested):

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/NodeLocator.java
garyrussell: The interfaces were only moved, so they should keep their `@since` - I am not fixing the tangle in 2.4.x.

Issue #1561, opened 2023-02-03
SimpleMessageListenerContainer: Delayed shutdown because callback is not run


timbq:
**In what version(s) of Spring AMQP are you seeing this issue?**

2.4.9

**Describe the bug**

When a SimpleMessageListenerContainer in a spring boot app with graceful shutdown should stop but is already stopped (no consumers active), the callback given by DefaultLifecycleProcessor is not run. This means the count of successfully shut down beans dependent of RabbitListenerEndpointRegistry does never reach 0 (if there were more than one) delaying the shutdown of the spring-container, eventually printing the following message:
Failed to shut down 1 bean with phase value 2147483647 within timeout of 30000ms: [org.springframework.amqp.rabbit.config.internalRabbitListenerEndpointRegistry]

**To Reproduce**

- Have RabbitMQ set up
- Create a default spring-boot project with spring-amqp and spring-web
- Configure graceful shutdown
- Configure two RabbitListeners for two queues
- Configure shutdown-actuator
- Set up a Rest-endpoint which stops one of the two listeners
- Call the shutdown actuator
- Observe the delayed shutdown and log message

See also sample project.

**Expected behavior**

Callback given to SimpleMessageListener.stop(Runnable) is called and thus shutdown completed as fast as possible.

**Sample**

https://github.com/timbq/spring-rabbitmq-sample

PR #1562, opened 2023-02-03
Run callback for shutdown also when container is already shut down

timbq:
see #1561

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
@@ -622,6 +622,7 @@ public class SimpleMessageListenerContainer extends AbstractMessageListenerConta
 		Thread thread = this.containerStoppingForAbort.get();
 		if (thread != null && !thread.equals(Thread.currentThread())) {
 			logger.info("Shutdown ignored - container is stopping due to an aborted consumer");
+			runCallbackIfNotNull(callback);
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -641,6 +641,9 @@ private void shutdownAndWaitOrCallback(@Nullable Runnable callback) {
 			}
 			else {
 				logger.info("Shutdown ignored - container is already stopped");
+				if (callback != null) {
+					getTaskExecutor().execute(callback);
```
artembilan, line 645: There is no reason to call this callback on a different thread.

Plus, take a look, please, there is one more `return` in this method where we have to take care about this callback, too.

artembilan (changes requested):
Please, consider also to add your name to the `@author` list of those changed classes.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
timbq: You are right! I will fix this and cover the other case as soon as I get around to work on this again.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -686,6 +687,12 @@ private void shutdownAndWaitOrCallback(@Nullable Runnable callback) {
 		}
 	}
 
+	private void runCallbackIfNotNull(@Nullable Runnable callback) {
```
artembilan, line 690:
```suggestion
	private static void runCallbackIfNotNull(@Nullable Runnable callback) {
```

artembilan (changes requested):

artembilan:

@timbq,

Issue #2428, opened 2023-03-18
After recovery: RabbitAdmin manual declarabes are not being declared


EldarErel:
**In what version(s) of Spring AMQP are you seeing this issue?**

3.0.2 (latest)

**Describe the bug**

After the connection to the broker has been recovered, the RabbitAdmin which I used to declare queues manually, doesn't declare the manual declarables (even when `reDaclareManualDeclaration` is set to true). It happens because there is no application context configured within the RabbitAdmin instance
```
public void initialize() {
		if (this.applicationContext == null) {
			this.logger.debug("no ApplicationContext has been set, cannot auto-declare Exchanges, Queues, and Bindings");
			return;
		}
```

**To Reproduce**

1) Create a connection to a RabbitMQ broker
2) initialize RabbitAdmin instance (new RabbitAdmin(connection))
3) set `reDaclareManualDeclaration` to true
3) declare queues manually
4) Drop the connection and wait for the connection to re-established

**Expected behavior**
RabbitAdmin should re-declare the manual declarables even when no application context,

PR #2429, opened 2023-03-18
fix: #2428 - recovery manual declarables without application context

EldarErel:
Issue link: [#2428](https://github.com/spring-projects/spring-amqp/issues/2428)

Fixes an issue when manual declarations are not being redeclared upon connection recovery, when application context has not been set

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitAdmin.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitAdmin.java
@@ -638,6 +638,8 @@ public class RabbitAdmin implements AmqpAdmin, ApplicationContextAware, Applicat
 	@Override // NOSONAR complexity
 	public void initialize() {
 
+		redeclareManualDeclarables();
+
 		if (this.applicationContext == null) {
 			this.logger.debug("no ApplicationContext has been set, cannot auto-declare Exchanges, Queues, and Bindings");
 			return;
@@ -690,6 +692,14 @@ public class RabbitAdmin implements AmqpAdmin, ApplicationContextAware, Applicat
 			declareBindings(channel, bindings.toArray(new Binding[bindings.size()]));
 			return null;
 		});
+		this.logger.debug("Declarations finished");
+
+	}
+
+	/**
+	 * Process manual declarables.
+	 */
+	private void redeclareManualDeclarables() {
 		if (this.manualDeclarables.size() > 0) {
 			synchronized (this.manualDeclarables) {
 				this.logger.debug("Redeclaring manually declared Declarables");
@@ -706,7 +716,6 @@ public class RabbitAdmin implements AmqpAdmin, ApplicationContextAware, Applicat
 				}
 			}
 		}
-		this.logger.debug("Declarations finished");
 
 	}
 
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitAdminTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitAdminTests.java
@@ -445,6 +445,52 @@ public class RabbitAdminTests extends NeedsManagementTests {
 		cf.destroy();
 	}
 
+	@Test
+	void manualDeclarationsWithoutApplicationContext() {
+		CachingConnectionFactory cf = new CachingConnectionFactory(
+				RabbitAvailableCondition.getBrokerRunning().getConnectionFactory());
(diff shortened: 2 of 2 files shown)


garyrussell:
@EldarErel Why is there no `ApplicationContext`?

This is a Spring project and is designed for use in a Spring application; there are probably many other features that won't work when used outside of Spring.

I am not inclined to accept a change that would support/promote such usage because it could be a slippery slope and become a support nightmare.

If you don't want to use Spring, consider using the amqp-client API directly.

If you are using it in a Spring app, then the admin must be declared as a bean so that all of its `*Aware` methods (and `InitializingBean`) interfaces are satisfied.

I'll leave this open for a short while, in case you can provide a compelling use case for us to accep this change.

EldarErel:
> @EldarErel Why is there no `ApplicationContext`?
> 
> This is a Spring project and is designed for use in a Spring application; there are probably many other features that won't work when used outside of Spring.
> 
> I am not inclined to accept a change that would support/promote such usage because it could be a slippery slope and become a support nightmare.
> 
> If you don't want to use Spring, consider using the amqp-client API directly.
> 
> If you are using it in a Spring app, then the admin must be declared as a bean so that all of its `*Aware` methods (and `InitializingBean`) interfaces are satisfied.
> 
> I'll leave this open for a short while, in case you can provide a compelling use case for us to accep this change.


Hi, thank you for your quick response.

I use spring, the use case is trying to create several listeners to several connections, I’ve created a connection, listener container and amqp admin for each connection.
I initialed them as spring beans.
I declare the queues once using @rabbitlistener
And using bean processor and some other parameters, I declare the queues and register listeners for the other connections as well.

As I created the beans myself, I skipped the setter of the application context in the amqp admin
RabbitAdmin.setApplicationContex(..)
 (as it is not required by default, and didn’t use it in my case as im declaring the queues manually and dynamically for for other connections.

As you added manual declaration for non bean declarables, I thought it will be reasonable to use an AmqpAdmin bean for manual declarations only.

If application context is needed to use the manual declarations recovery, then maybe it should be enforced on initialization of a new RabbitAdmin instance.
So it would be explicit that we must set application context for the RabbitAdmin to work properly with manual redeclarations

garyrussell:
`RabbitAdmin.setApplicationContext(..)`. You don’t need to call it at all. If the admin is declared as a `@Bean`, Spring will set everything up for you. You should never create Spring beans manually. Or, if you do, you need to register them in the context and tell the context to fully initialize it.

garyrussell:
It’s not a Spring bean if you just create it yourself, it’s just an instance of the class.

EldarErel:
I created a bean for mapping the rabbitAdmins by the connection name:
```
	@Bean
	public Map<String,AmqpAdmin> rabbitAdminsMap(Map<String,ConnectionFactory> connections) {
		Map<String, AmqpAdmin> rabbitAdmins = new HashMap<>();
		connections.forEach((key,connection) ->
		{
			RabbitAdmin admin = new RabbitAdmin(connection);
			admin.setRedeclareManualDeclarations(true);
			rabbitAdmins.put(key, admin);
		});
		return rabbitAdmins;
	}
```

EldarErel:
With the spring support, upon connection recovery the admin 
`initialze()` method  is being called, but the manual declarations are not being redeclared, because the `RabbitAdmin.applicationContext == null`.
This PR solves it , so when setRedeclareManualDeclarations(true) the functionally will work.

garyrussell:
Adding a `@Bean` for the Map of admins does not register each admin as a bean (which is required for them to be beans). Here is some similar code to what you need...

```java
@Bean
Map<String, RabbitAdmin> admins(ConfigurableApplicationContext context, ConnectionFactory cf) {
	RabbitAdmin admin = new RabbitAdmin(cf);
	context.getBeanFactory().registerSingleton("foo", admin);
	admin = (RabbitAdmin) context.getBeanFactory().initializeBean(admin, "foo");
	return Map.of("foo", admin);
}
```

EldarErel:
Thank you very much for your time and response!
Managing dynamic creation and eviction of beans is a mission that I prefer to avoid when possible, as it required more maintenance in certain situations.

If the admin cannot functional without it being a bean, so ok I understand.
But as the manual declaration methods works without the admin being a been, the re declaration can be something to complete it. (As it doesn’t require the context to work, only re ordering the initialize method).

garyrussell:
Ok; I can make an exception this time, but please be aware that other features won't work either (e.g. publication of `DeclarationExceptionEvent`s), and I can make no guarantees that something else might break in the future. The admin is designed to be a Spring bean.

EldarErel:
exactly, well understood and make sense.

Issue #2432, opened 2023-03-21
AbstractMessageListenerContainer.redeclareElementsIfNecessary does not take into account Declarables


mirkoluchi:
**In what version(s) of Spring AMQP are you seeing this issue?**

2.4.7

**Describe the bug**

The message listener container, when restarted because of a consumer's failure, attempts to redeclare the queues. The code doing that - however - does not take into account queues that are declared using a bean of type Declarables.

**To Reproduce**

- Declare a Queue using a Declarables bean
- Declare a consumer on that queue
- While the application is running, delete the queue
- The consumer fails, attempts to restart and redeclare missing queues, but our queue is not redeclared (if the queue was instead declared as a bean of type Queue, the redeclaration would succeed).

**Expected behavior**

The consumer should redeclare also queues provided through beans of type Declarables.

**Example**

Here you can find 2 tests:
- QueueRedeclarationTest ==> succeeds showing that deleted queues are redeclared, if they are provided as Queue bean
- DeclarablesRedeclarationTest ==> fails showing that the same test described above fails if queues are declared using Declarables
https://github.com/mirkoluchi/springboot-rabbitmq-issue-2432

garyrussell:
Queue deletion is not something that the framework can recover from.

Redeclaration occurs when a connection is lost; when the connection is re-established, queues are redeclared (in case they were auto delete queues).

When a queue is deleted via some other means (e.g. directly on the admin UI), there is no connection interruption, and therefore redeclaration cannot occur.

mirkoluchi:
Actually the current behaviour is that a queue (declared using a Queue bean) is redeclared if deleted from the rabbit server.
So I would expect the same to happen for queues that are declared using Declarables bean.
I'm attaching a test soon.

garyrussell:
Hmmm; reopened.

garyrussell:
Thanks; will take a look.

garyrussell:
Thanks for the repro; this (redeclaration) was an unintended side-effect of a fix but, since it has been there for 6+ years, I will change the code to deal with the newer `Declarables` beans.

PR #2434, opened 2023-03-22
GH-2432: Fix Redeclaration of Declarables

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/2432

When a container starts; it looks to see if the context contains a queue bean for any of its queues and requests the admin to redeclare the infrastructure.

However, if the queue is declared within a `Declarables`, the check is not performed and therefore the queue is not declared.

Add a check for `Declarables` beans.

**cherry-pick to 2.4.x**


artembilan:

Issue #2439, opened 2023-03-23
Spring amqp doesnt support more than 1 server-named/anonymous queues


emoen:
*In what version(s) of Spring AMQP?
2.3.10

*Describe the bug
2 anonymous queues - with different bindings - results in 1 queue with both bindings and 1 queue with no bindings

*To Reproduce
See stackexchange: https://stackoverflow.com/questions/75823022/spring-amqp-queuebuilder-should-hold-ref-to-queue-and-not-string-name

*Expected behavior
2 anonymous queues - with different bindings - should have a server-generated queue-name and each binding assign to it

garyrussell:
Here is a workaround until the release is available:

```java
	@Bean
	Queue q1() {
		return QueueBuilder.nonDurable("").build();
	}

	@Bean
	Queue q2() {
		return QueueBuilder.nonDurable("").build();
	}

	@Bean
	Binding b1(Queue q1, DirectExchange ex) {
		return new Binding(q1.getName(), DestinationType.QUEUE, "ex", "foo", Collections.emptyMap()) {

			@Override
			public String getDestination() {
				return q1.getActualName();
			}

		};
	}

	@Bean
	Binding b2(Queue q2, DirectExchange ex) {
		return new Binding(q2.getName(), DestinationType.QUEUE, "ex", "bar", Collections.emptyMap()) {

			@Override
			public String getDestination() {
				return q2.getActualName();
			}

		};
	}

	@Bean
	DirectExchange ex() {
		return new DirectExchange("ex");
	}
```

PR #2440, opened 2023-03-23
GH-2439: Fix Bindings with Broker Gen Queue Names

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/2439

To configure a broker-named queue, the `Queue` name is set to an empty String; the name is later populated when the queue is declared.

However, if such a queue is used in a `BindingBuilder`, the builder accesses the name before it is populated.

When such a queue is used in a binding, retain a copy of the queue object and retrieve its actual name when the binding is declared.

Add a unit test for all binding types; also tested with the following, with both queues being bound properly to the exchange.

```java
@Bean
Queue q1() {
	return QueueBuilder.nonDurable("").build();
}

@Bean
Queue q2() {
	return QueueBuilder.nonDurable("").build();
}

@Bean
Binding b1(Queue q1, DirectExchange ex) {
	return BindingBuilder.bind(q1).to(ex).with("foo");
}

@Bean
Binding b2(Queue q2, DirectExchange ex) {
	return BindingBuilder.bind(q2).to(ex).with("bar");
}

@Bean
DirectExchange ex() {
	return new DirectExchange("ex");
}
```

**cherry-pick to 2.4.x**

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/core/Binding.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/core/Binding.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


artembilan:

Issue #2437, opened 2023-03-22
Missing @RabbitHandler method triggers an infinite retry loop


bcalmac:
**In what version(s) of Spring AMQP are you seeing this issue?**

2.4.6

**Describe the bug**

If the spring-boot application does not have a `@RabbitHandler` matching an incoming event then the event is re-queued and retried immediately, triggering an infinite reprocessing loop.

**To Reproduce**

Create a binding that sends a service an unexpected message. Say for instance, if you currently have:

```java
@Component
@RabbitListener(queues = "foo-service")
public class FooEventListener {

    @RabbitHandler
    public void onOrderCreated(OrderCreated orderCreated) {
        ...
    }
}
```

then create a binding that sends some other event, say `OrderUpdated` to the `foo-service` queue.

Here is the exception as seen from our application:

```
org.springframework.amqp.rabbit.support.ListenerExecutionFailedException: Failed to convert message
        at org.springframework.amqp.rabbit.listener.adapter.MessagingMessageListenerAdapter.onMessage(MessagingMessageListenerAdapter.java:156)
        at org.springframework.amqp.rabbit.listener.AbstractMessageListenerContainer.doInvokeListener(AbstractMessageListenerContainer.java:1670)
        at org.springframework.amqp.rabbit.listener.AbstractMessageListenerContainer.actualInvokeListener(AbstractMessageListenerContainer.java:1589)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:568)
        at org.springframework.aop.support.AopUtils.invokeJoinpointUsingReflection(AopUtils.java:344)
        at org.springframework.aop.framework.ReflectiveMethodInvocation.invokeJoinpoint(ReflectiveMethodInvocation.java:198)
        at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:163)
        at io.opentracing.contrib.spring.rabbitmq.RabbitMqReceiveTracingInterceptor.invoke(RabbitMqReceiveTracingInterceptor.java:48)
[...]

PR #2443, opened 2023-03-23
GH-2437: Fix Fatal When No Matching RabbitHandler

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/2437

Failure to find a `@RabbitHandler` method for the payload must be treated as a fatal exception to avoid an infinite loop.

Also, fix CREH cause traversal to stop when **any** fatal exception is found. Previously, traversal only stopped for the original subset of exceptions.

**cherry-pick to 2.4.x**


garyrussell:
Also tested with

```java
@SpringBootApplication
public class Rgh2437Application {

	public static void main(String[] args) {
		SpringApplication.run(Rgh2437Application.class, args);
	}

	@Bean
	ApplicationRunner runner(RabbitTemplate template) {
		return args -> {
			template.convertAndSend("", "foo", "test");
		};
	}

}

@Component
@RabbitListener(queues = "foo")
class Listener {

	@RabbitHandler
	void listen(Integer in) {
	}

}
```

artembilan:

Issue #2456, opened 2023-05-19
AmqpException when testing @RabbitListener with @RabbitHandler 


kstojanovski-novatec:
**In what version(s) of Spring AMQP are you seeing this issue?**

3.0.6

**Describe the bug**

I am not sure if it is a bug but I can not explain the thrown exception `Caused by: org.springframework.amqp.AmqpException: Ambiguous methods for payload type` for the given situation.

On integration test with test containers and spying a class with Mockito annotated with @RabbitListener and its methods annotated with @RabbitHandler leads to AmqpException.

Obviously, the original @RabbitHanlder method is loaded and the mocked one at the same time.
MessagingMessageListenerAdapter::invokeHandlerAndProcessResult at line 140 (this.handlerAdapter.delegatingHanlder)
```
    if (this.messagingMessageConverter.method == null && amqpMessage != null) {
      amqpMessage.getMessageProperties().setTargetMethod(this.handlerAdapter.getMethodFor(message.getPayload()));
    }
```
It there anything I can do to avoid this situation?

**To Reproduce**

You need to start the tests which are located in the linked project.

**Expected behavior**

I would not expect the error since the methods are not loaded twice.

**Sample**

Starting the **DifferentTypesDistributorListenerIntegrationTest** or **ManyCustomersWorkersSpyListenerIntegrationTest** from the project led to the exception https://github.com/kstojanovski-novatec/spring-amqp-testing-issue.

kstojanovski-novatec:
Hello Artem, thanks for the fast answer.

I saw the linked documentation and have also downloaded the four files from the GitHub repo:
* org.springframework.amqp.rabbit.test.examples.ExampleRabbitListenerCaptureTest
* org.springframework.amqp.rabbit.test.examples.ExampleRabbitListenerSpyAndCaptureTest
* org.springframework.amqp.rabbit.test.examples.ExampleRabbitListenerSpyTest
* org.springframework.amqp.rabbit.test.examples.TestRabbitTemplateTests

Only TestRabbitTemplateTests works well, but this kind of test I already have. I am interested in the other test where `RabbitListenerTestHarness` is used.
Unfortunately the `private RabbitListenerTestHarness harness;` is marked red with the message "Could not auto-wire. No beans of 'RabbitListenerTestHarness' type found." for the other three classes as well.

[...]

PR #2457, opened 2023-05-22
GH-2456: Suppress Duplicate Annotations with Spy

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/2456

When spying a `@RabbitListener` bean, duplicate methods are resolved as well as duplicate class level `@RabbitListener` annotations.

**cherry-pick to 2.4.x** (will require instanceof polishing for Java 8)

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2022 the original author or authors.
+ * Copyright 2014-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -339,7 +339,8 @@ public class RabbitListenerAnnotationBeanPostProcessor
 					multiMethods.add(method);
 				}
 			}
-		}, ReflectionUtils.USER_DECLARED_METHODS);
+		}, ReflectionUtils.USER_DECLARED_METHODS
+				.and(meth -> !meth.getDeclaringClass().getName().contains("$MockitoMock$")));
 		if (methods.isEmpty() && multiMethods.isEmpty()) {
 			return TypeMetadata.EMPTY;
 		}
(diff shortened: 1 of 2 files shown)


artembilan:
@garyrussell ,

does this confirm that premise of the `spy()` works even if we skip those Mockito generated methods from candidates?
How does those `verify()` then pass?

Thanks

garyrussell:
Yes, that's why I added the `verify()` to ensure the spy still works as expected.

I couldn't detect any difference at runtime - the mock method is still in the stack trace (set a breakpoint in `MultiListenerBean.bar()` and you will see it.

This doesn't actually make any difference to the spy, it just avoids us finding both sets of annotations which causes duplicate containers/methods.

artembilan:
... and cherry-picked to `2.4.x` after fixing `instanceof pattern`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -339,7 +339,8 @@ private TypeMetadata buildMetadata(Class<?> targetClass) {
 					multiMethods.add(method);
 				}
 			}
-		}, ReflectionUtils.USER_DECLARED_METHODS);
+		}, ReflectionUtils.USER_DECLARED_METHODS
+				.and(meth -> !meth.getDeclaringClass().getName().contains("$MockitoMock$")));
```
ztyzbb, line 343: cherry-picked to 2.4.x breaks compatibility with Spring Framework 5.2, USER_DECLARED_METHODS comes with a `and` method only after 5.3.2

garyrussell: Spring Framework 5.2.x is no longer supported as OSS and goes out of commercial support at the end of the year; if you have a commercial support contract with VMware; please raise an issue against spring-amqp via normal procedures there.

https://spring.io/projects/spring-framework#support

ztyzbb: But the [reference](https://docs.spring.io/spring-amqp/docs/2.4.15/reference/html/#compatibility) said it works with 5.2.x.
And in fact, everything else looks good with 5.2.x

garyrussell: That's a bug in the documentation but, in any case, as I said 5.2.x is out of support anyway.

garyrussell: I made an exception and made the change https://github.com/spring-projects/spring-amqp/pull/2530/commits/6e18f87258764540ab3f88fbae2a882feb172dfb

The next release is mid-October.

But I still recommend that you move to a supported version of Spring Framework.

garyrussell: I made an exception and made the change dcc49ba1b203fa5c79b5a81962e0995725099e64

The next release is mid-October.

But I still recommend that you move to a supported version of Spring Framework.

ztyzbb: Thank you! I'd like to upgrade, but it's out of my control.

Issue #1210, opened 2020-06-18
Add Kotlin Coroutine support


davidb-e4s:
Improvement 

Please can you add Kotlin Coroutine support so we can use the suspend keyword on @RabbitListener functions. Currently it errors.

artembilan:
We need to investigate that feature: don't have experience with Kotlin Coroutines.
We have support for Reactor `Mono`, so probably should be something similar for `suspend`: https://docs.spring.io/spring-amqp/docs/2.2.7.RELEASE/reference/html/

Feel free to provide a Pull Request on the matter if you see how that could be implemented.
Thanks

garyrussell:
It's not clear what you are trying to achieve here.

My understanding is `suspend` functions can only be called from a coroutine; since `@RabbitListener` methods are called by the framework not user code we'd have to add a shim between the framework and the listener - but exactly how would that perform any useful function? The shim would still block the calling thread until the listener exits.

garyrussell:
Closing due to inactivity.

eduanb:
Can we reopen this? Ran into a weird JSON serialisation issue which was really hard to debug because of this. Spring webflux supports coroutines and spring-amqp already supports reactive so it is confusing to the user.
@garyrussell

artembilan:
Hi @eduanb !

We are fully open for contribution.
Not sure how JSON is relevant and why you use `suspend` on the `@RabbitListener`, but if you open Pull Request for possible fix, we are good to review it ASAP.

Thank you!

verissimor:
It would be very helpful to have support for `coroutines/suspend` method on the `@RabbitListener`.

artembilan:
In Spring Integration we have a special wrapper:
```
public class IntegrationInvocableHandlerMethod extends InvocableHandlerMethod {

	public IntegrationInvocableHandlerMethod(Object bean, Method method) {
		super(bean, method);
	}

	@Override
	protected Object doInvoke(Object... args) throws Exception {
		Method method = getBridgedMethod();
		if (KotlinDetector.isSuspendingFunction(method)) {
			return CoroutinesUtils.invokeSuspendingFunction(method, getBean(), args);
		}
		else {
			return super.doInvoke(args);
		}
	}

}
```
Which we set into a `IntegrationMessageHandlerMethodFactory`.
That one returns a `Mono` for `suspend` functions which we have already support for in the `AbstractAdaptableMessageListener`:
```
[...]

PR #2460, opened 2023-05-24
GH-1210: Add Kotlin suspend functions support

artembilan:
Fixes https://github.com/spring-projects/spring-amqp/issues/1210

Kotlin Coroutines are essentially `Future` wrapping. Therefore, it is natural to have `suspend` support on `@RabbitListener` methods as we do now for `CompletableFuture` and `Mono`

* Introduce some utilities since we cannot reuse existing from Spring Messaging: they are there about Kotlin Coroutines only for reactive handlers
* Some code clean up in the `RabbitListenerAnnotationBeanPostProcessor` for the latest Java
* Add optional dep for `kotlinx-coroutines-reactor` and document the feature

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -56,6 +56,7 @@ ext {
 	jaywayJsonPathVersion = '2.7.0'
 	junit4Version = '4.13.2'
 	junitJupiterVersion = '5.9.2'
+	kotlinCoroutinesVersion = '1.6.4'
 	log4jVersion = '2.19.0'
 	logbackVersion = '1.4.4'
 	lz4Version = '1.8.0'
@@ -436,6 +437,7 @@ project('spring-rabbit') {
 		}
 		optionalApi "com.jayway.jsonpath:json-path:$jaywayJsonPathVersion"
 		optionalApi "org.apache.commons:commons-pool2:$commonsPoolVersion"
+		optionalApi "org.jetbrains.kotlinx:kotlinx-coroutines-reactor:$kotlinCoroutinesVersion"
 
 		testApi project(':spring-rabbit-junit')
 		testImplementation("com.willowtreeapps.assertk:assertk-jvm:$assertkVersion")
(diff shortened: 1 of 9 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -440,14 +432,10 @@ protected Collection<Declarable> processListener(MethodRabbitListenerEndpoint en
 		List<Object> resolvedQueues = resolveQueues(rabbitListener, declarables);
 		if (!resolvedQueues.isEmpty()) {
 			if (resolvedQueues.get(0) instanceof String) {
-				endpoint.setQueueNames(resolvedQueues.stream()
-						.map(o -> (String) o)
-						.collect(Collectors.toList()).toArray(new String[0]));
+				endpoint.setQueueNames(resolvedQueues.stream().map(o -> (String) o).toArray(String[]::new));
```
garyrussell, line 435: Convention is new line per `.`.

artembilan: I know, but looks a bit awkward as a method argument.
I can bring it back if you wish or introduce a local variable.

build.gradle
```diff
@@ -56,6 +56,7 @@ ext {
 	jaywayJsonPathVersion = '2.7.0'
 	junit4Version = '4.13.2'
 	junitJupiterVersion = '5.9.2'
+	kotlinCoroutinesVersion = '1.6.4'
```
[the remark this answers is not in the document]
garyrussell: OK

jhoanmanuelms:
Hi there, quick question, will this feature be supported in version `2.7.14`? Thanks!

garyrussell:
@jhoanmanuelms We can take a look, but can provide no guarantee - it depends on how cleanly the code back ports; there are many changes in 3.0.x that might cause too many conflicts.

Please open a new issue so it's not forgotten.

garyrussell:
@jhoanmanuelms Turns out the back port was not too challenging; however, I had to pull in an older version of kotlin coroutines because 2.4.x is used with Boot 2.7.x and the versions need to be aligned.

https://github.com/spring-projects/spring-amqp/pull/2519

Issue #2467, opened 2023-05-30
Add Observation to Stream Components


garyrussell:
Similar to that provided by AMQP components.

PR #2468, opened 2023-05-31
GH-2467: Micrometer Observation for Streams

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/2467

**Needs docs and observation generation**

It looks like the generation only takes a single input dir so perhaps the task needs to be moved to the project level (and replicated in spring-rabbit-stream). I will probably need help with that.

### DIFF
--- a/spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/micrometer/RabbitStreamListenerObservation.java
+++ b/spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/micrometer/RabbitStreamListenerObservation.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright 2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.rabbit.stream.micrometer;
+
+import io.micrometer.common.KeyValues;
+import io.micrometer.common.docs.KeyName;
+import io.micrometer.observation.Observation.Context;
+import io.micrometer.observation.ObservationConvention;
+import io.micrometer.observation.docs.ObservationDocumentation;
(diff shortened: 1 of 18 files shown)


spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/micrometer/RabbitStreamMessageReceiverContext.java
```diff
@@ -0,0 +1,102 @@
...
+					return new String(bytes, StandardCharsets.UTF_8);
+				}
+			}
+			return null;
+		});
+		setCarrier(message);
+		this.message = message;
+		this.listenerId = listenerId;
+		this.stream = stream;
+		setRemoteServiceName("RabbitMQ");
```
artembilan, line 63: Don't we need to have it as more granual:  `RabbitMQ Streams`?
And the same for those ` "spring.rabbit.listener" ->  "spring.rabbit..stream.listener"` ?

spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/micrometer/RabbitStreamListenerObservation.java
```diff
@@ -0,0 +1,101 @@
...
+import io.micrometer.common.docs.KeyName;
+import io.micrometer.observation.Observation.Context;
+import io.micrometer.observation.ObservationConvention;
+import io.micrometer.observation.docs.ObservationDocumentation;
+
+/**
+ * Spring Rabbit Observation for stream listeners.
+ *
+ * @author Gary Russell
+ * @since 3.0
```
artembilan, line 29:
```suggestion
 * @since 3.0.5
```

spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/producer/RabbitStreamTemplate.java
```diff
@@ -188,19 +224,49 @@ public CompletableFuture<Boolean> convertAndSend(Object message, @Nullable Messa
...
+			observation.error(ex);
+			observation.stop();
+			future.completeExceptionally(ex);
+		}
+	}
+
+	@Nullable
+	private ObservationRegistry obtainObservationRegistry() {
+		if (!this.observationRegistryObtained && this.observationEnabled && this.applicationContext != null) {
+			if (this.applicationContext != null) {
```
artembilan, line 250: I guess we don't need the first one then.
Even if there is no ctx, we still need to set `observationRegistryObtained` to avoid extra steps into this `if` for subsequent calls.

artembilan (changes requested):
Look great!
Just a couple nit-picks.
Thanks

Issue #2482, opened 2023-07-10
Ensure Strict Ordering with Single Active and Exclusive Consumers


garyrussell:
When using exclusive or Single Active consumers, add an option to terminate the container immediately on stop.

Currently, the consumer is canceled and the remaining messages are processed; this allows another consumer to start consuming from beyond the prefetch.

When this option is set; defer the cancel until after the current record is processed (successfully or not) and requeue all prefetched messages.

See https://github.com/spring-projects/spring-integration/issues/8668

PR #2483, opened 2023-07-12
GH-2482: Option for Containers to Stop Immediately

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/2482

`forceStop` means stop after the current record and requeue all prefetched.

Just close the channel - canceling the consumer first causes a race condition which could allow another exclusive or single-active consumer to start processing while this container is still running.

Also support async stop on DMLC (previously only available on the SMLC).

**cherry-pick to 2.4.x**


artembilan:
... and cherry-picked to `2.4.x` after minor conflict resolution.

PR #2486, opened 2023-07-12
Upgrade amqp-client Version

garyrussell:


build.gradle
```diff
@@ -58,7 +58,7 @@ ext {
 	micrometerVersion = '1.8.13'
 	mockitoVersion = '3.11.2'
 	rabbitmqStreamVersion = '0.4.0'
-	rabbitmqVersion = project.hasProperty('rabbitmqVersion') ? project.rabbitmqVersion : '5.13.1'
+	rabbitmqVersion = project.hasProperty('rabbitmqVersion') ? project.rabbitmqVersion : '5.14.3'
```
artembilan, line 61: Is it really legit to do this move in our point version?

artembilan (changes requested):

build.gradle
garyrussell: Yes, see the discussion in Slack (#spring-integration). 

>Yes, that's fine; I will bump 2.4.x to 5.14.3 (a bit unusual for us, but most users use Boot anyway).

I decided it was ok given that Boot 2.7.x pulls in this version anyway; I didn't want to add more burden than necessary for Arnaud to backport to yet another branch.

PR #2487, opened 2023-07-13
GH-2482: Fix Async Container Stop

garyrussell:
- Stop will callback did not reset `active` flag (both containers)
- DMLC did not release the `cancellationLock`
- Also restore `@LongRunning`

**cherry-pick to 2.4.x** (No need for back port issue, reused 2482).


artembilan:

Issue #2493, opened 2023-07-21
Spring AMQP artifacts have not published associated `.module` files needed by Gradle.


thorntonrp:
None of the Spring AMQP modules are publishing their associated Gradle `.module` files to Maven Central. Gradle puts a comment at the top of Gradle-published pom files to indicate that there is additional associated metadata. When Gradle reads the pom file, it looks for that comment and if it finds it, it tries to load the `.module` file. If that file has not been published, it can cause problems against private artifact repositories that proxy from Maven Central.

**Short explanation**
If you publish `.pom` files produced by Gradle, you MUST publish the `.module` files too. It is an error NOT to publish the `.module` files.

**Long explanation**
When Maven Central artifacts are being proxied through private repositories (such as internal JFrog Artifactory), and when those private repositories have security on them (which they almost always do), the repository isn't always able to respond with a 404 error and may instead respond with a 403 error, which causes Gradle to fail the build. The reason for the 403 error is quite complicated, but it has to do with how repositories like Artifactory attempt combine results from multiple secured repositories into a single virtual repository.

I have reached out to both Gradle and JFrog Artifactory. Artifactory says it MUST respond with 403 because it will not even search a repository if a user doesn't have permission to see what's in it. When using virtual repositories that proxy dozens of other repositories, the only user that will not get a 403 is a superuser that has permission to search all of them. Likewise, Gradle says it MUST fail the build when it gets a 403 error status. Because both Gradle and JFrog cannot change their behavior, that means that if you publish `.pom` files produced by Gradle, you MUST publish the associated `.module` files too. It is an error NOT to publish them.

garyrussell:
Thanks for reporting; `.module` is published to artifactory, but not copied to central.

https://github.com/spring-projects/spring-amqp/blob/fa944397e702453156f057f3ccd72bbbc9a58778/.github/release-files-spec.json#L12-L21

artembilan:
Apparently Spring Framework has started to publish those Gradle `.module`  files into Maven Central: https://repo1.maven.org/maven2/org/springframework/spring-core/6.0.11/.

In Spring Integration we do this instead for a while:
```
tasks.withType(GenerateModuleMetadata) {
	enabled = false
}
```

This could be done here as an interim solution.
However the proper fix would definitely to sync those Gradle files into Maven Central as well.

garyrussell:
[...]

PR #2494, opened 2023-07-24
GH-2493: Sync Gradle .module File to Maven

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/2493


garyrussell:
Correction - no need to cherry-pick, we always use the main branch for actions.

Issue #2495, opened 2023-07-26
Add maxInboundMessageSize to RabbitConnectionFactoryBean


garyrussell:
See https://github.com/rabbitmq/rabbitmq-java-client/pull/1063

PR #2496, opened 2023-07-26
GH-2495: Add maxInboundMessageBodySize to RCFB

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/2495

Although the factory bean has...

```java
	/**
	 * Access the connection factory to set any other properties not supported by
	 * this factory bean.
	 * @return the connection factory.
	 * @since 1.7.14
	 */
	public ConnectionFactory getRabbitConnectionFactory() {
```

...this new CF property could be a breaking change for users that configure the factory bean using XML.

**cherry-pick to 2.4.x**


artembilan (approved):
I'm accepting this, but, please, answer the question how is this a breaking change for XML configuration?
IF you use a `<bean>` for `RabbitConnectionFactoryBean` you are very well benefiting with adding an extra `<property>`. Perhaps you mean a `<rabbit:connection-factory>`?

Thanks

artembilan:

Issue #2504, opened 2023-08-02
Implement Lifecycle in Connection Factories


garyrussell:
Close connection(s) when stopped.

PR #2507, opened 2023-08-04
GH-2504: Clear Deffered Channel Close Executor

garyrussell:
Addendum to #2504 - when the connection is reset, the `channelsExecutor` is shut down. It needs to be reset so that a new one is created, if necessary.

**cherry-pick to 2.4.x**


artembilan:

Issue #2479, opened 2023-07-05
Add Mechanism to set Container Thread Names


garyrussell:
See https://github.com/spring-projects/spring-kafka/issues/2501

bdshadow:
I would like to do this. Actually, I've already started, but it turned out to be a little more difficult than in Kafka. I will provide a merge request soon

artembilan:
I'm closing this one in favor of `setTaskExecutor(new VirtualThreadTaskExecutor("my-virtual-threads-"))` option.
Otherwise you would have a paradox of choice and might lead yourself into some trap where you would inject a custom executor and that thread naming supplier.

Thanks for understanding.

PR #2509, opened 2023-08-06
GH-2479: add option to set listener thread name

bdshadow:

### DIFF
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainerIntegrationTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainerIntegrationTests.java
@@ -33,6 +33,7 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.ConcurrentSkipListSet;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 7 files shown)


spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainerIntegrationTests.java
```diff
@@ -372,6 +382,7 @@ public void testAddRemoveConsumers() throws Exception {
 		assertThat(activeConsumerCount(container, 0)).isTrue();
 		assertThat(TestUtils.getPropertyValue(container, "consumersByQueue", MultiValueMap.class)).hasSize(0);
 		template.stop();
+		assertThat(listenerThreadNames).contains("myThread"); // BUT IT'S STILL "client-1"
```
bdshadow, line 385: Could you, please, advise me, on how to do it better here? When there is a CachingConnectionFactory with a ThreadPoolTaskExecutor which has a threadNamePrefix, it is used when creating a connection thread and then this thread is used for consuming the message. It would be ok to set threadNamePrefix in the DirectMessageListenerContainer the same way it is done here in tests, but ConnectionFactory interface doesn’t have getters\setters for the executer.
Here, I’ve added some things to the tests which I expect from it, but it doesn’t work.

bdshadow:
I’m not sure about my approach in general. I tried to do it similarly to spring-kafka. I’ve managed to add it to the SimpleMessageListenerContainer. I’ve even tried it with the CachingConnectionFactory which has a ThreadPoolTaskExecutor with a threadNamePrefix and it worked. But looks like I miss something in how to do it better for the DirectMessageListenerContainer

garyrussell:
Sorry, I have not yet been able to look at your work, yet.

I doubt that I will get to it tomorrow and then I'll be away (time off + Spring One at VMware Explore) until August 28.

I just wanted to let you know that we are not ignoring you.

Thanks for your patience.

artembilan:
Closed as commented on the issue: https://github.com/spring-projects/spring-amqp/issues/2479#issuecomment-1896667195

PR #2517, opened 2023-08-22
Call objectMapper.findAndRegisterModules() to find possibly available…

### LINKED ISSUE #2516: Jackson2JsonMessageConverter: New ObjectMapper instance not aware of jsr310 module

3.0.6

**Describe the bug**

We are transporting Pojos over AMQP. These Pojos also include Java8 Time objects like LocalDateTime. Our application ran in a ""java 8 date/time type `java.time.localdatetime` not supported by default" error although jackson-datatype-jsr310 was present in the classpath.

**To Reproduce**

Return a Pojo including Java8 Time objects in a @RabbitListener annotated method.

**Expected behavior**

The returned Pojo (including Java8 Time objects) should be converted to JSON without any error if  jackson-datatype-jsr310 is present in the classpath.

***Solution**

The problem itself was that we provied a Jackson2JsonMessageConverter in our project using its default constructor, see:

https://github.com/spring-projects/spring-amqp/blob/c845cb7cc5d8d39224e3ea219fb3b005f35963f4/spring-amqp/src/main/java/org/springframework/amqp/support/converter/Jackson2JsonMessageConverter.java#L45C9-L45C37

This calls the construcor which creates a new ObjectMapper instance, see:

https://github.com/spring-projects/spring-amqp/blob/c845cb7cc5d8d39224e3ea219fb3b005f35963f4/spring-amqp/src/main/java/org/springframework/amqp/support/converter/Jackson2JsonMessageConverter.java#L58

This new ObjectMapper is not aware of any additional Jackson modules in the classpath. Since the mapper receives already some configuration in this constructor, I suggest to add a this.objectMapper.findAndRegisterModules() there.

I'll provide a pull request for this.

MichaelRisthausEviden:
… Java8 time modules (fixes #2516)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -367,6 +367,7 @@ project('spring-amqp') {
 		optionalApi 'com.fasterxml.jackson.core:jackson-databind'
 		optionalApi 'com.fasterxml.jackson.core:jackson-annotations'
 		optionalApi 'com.fasterxml.jackson.dataformat:jackson-dataformat-xml'
+		optionalApi 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310'
 
 		// Spring Data projection message binding support
 		optionalApi ('org.springframework.data:spring-data-commons') {
--- a/spring-amqp/src/main/java/org/springframework/amqp/support/converter/Jackson2JsonMessageConverter.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/support/converter/Jackson2JsonMessageConverter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -57,6 +57,7 @@ public class Jackson2JsonMessageConverter extends AbstractJackson2MessageConvert
 	public Jackson2JsonMessageConverter(String... trustedPackages) {
 		this(new ObjectMapper(), trustedPackages);
 		this.objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+		this.objectMapper.findAndRegisterModules();
 	}
 
 	/**
--- a/spring-amqp/src/test/java/org/springframework/amqp/support/converter/Jackson2JsonMessageConverterTests.java
+++ b/spring-amqp/src/test/java/org/springframework/amqp/support/converter/Jackson2JsonMessageConverterTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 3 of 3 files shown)


spring-amqp/src/main/java/org/springframework/amqp/support/converter/Jackson2JsonMessageConverter.java
```diff
@@ -57,6 +57,7 @@ public Jackson2JsonMessageConverter() {
 	public Jackson2JsonMessageConverter(String... trustedPackages) {
 		this(new ObjectMapper(), trustedPackages);
 		this.objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+		this.objectMapper.findAndRegisterModules();
```
artembilan, line 60: We prefer to do this without a `ServiceLoader` and reflection.
In Spring Integration we do this: https://github.com/spring-projects/spring-integration/blob/main/spring-integration-core/src/main/java/org/springframework/integration/support/json/Jackson2JsonObjectMapper.java#L87

We do something similar in Spring for Apache Kafka: https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/support/JacksonUtils.java#L59

build.gradle
```diff
@@ -367,6 +367,7 @@ project('spring-amqp') {
 		optionalApi 'com.fasterxml.jackson.core:jackson-databind'
 		optionalApi 'com.fasterxml.jackson.core:jackson-annotations'
 		optionalApi 'com.fasterxml.jackson.dataformat:jackson-dataformat-xml'
+		optionalApi 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310'
```
artembilan, line 370: Let's have more of them then!
```
        optionalApi 'com.fasterxml.jackson.datatype:jackson-datatype-jdk8'
        optionalApi 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310'
        optionalApi 'com.fasterxml.jackson.datatype:jackson-datatype-joda'
        optionalApi('com.fasterxml.jackson.module:jackson-module-kotlin') {
            exclude group: 'org.jetbrains.kotlin'
        }
```

spring-amqp/src/test/java/org/springframework/amqp/support/converter/Jackson2JsonMessageConverterTests.java
```diff
@@ -21,10 +21,8 @@
 
 import java.io.IOException;
 import java.math.BigDecimal;
-import java.util.Hashtable;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
+import java.time.LocalDateTime;
+import java.util.*;
```
artembilan, line 25: No asterisk imports, please. At all.

spring-amqp/src/test/java/org/springframework/amqp/support/converter/Jackson2JsonMessageConverterTests.java
```diff
@@ -478,6 +490,31 @@ public List<Fiz> fizLister() {
 		return null;
 	}
 
+	public static class JodasRevenge {
```
artembilan, line 493: I'm not sure that this is OK name for professional code.
Something like `TestJsonData` should be enough.
Doesn't look like this class really performs some revenge from Master Joda 😉 .

Plus it feels like this class could be just a `record`.

artembilan:
Superseded with https://github.com/spring-projects/spring-amqp/commit/b7e77fdf1aab1b6301e733f173ff359b4632b110.

@MichaelRisthausEviden ,

thank you for your effort!

PR #2521, opened 2023-09-06
Main to 3.1; Upgrade Dependency Versions

garyrussell:
Also compatibility changes for Mockito 5.


build.gradle
```diff
@@ -51,26 +51,26 @@ ext {
...
-	jaywayJsonPathVersion = '2.7.0'
+	jacksonBomVersion = '2.15.2'
+	jaywayJsonPathVersion = '2.8.0'
 	junit4Version = '4.13.2'
-	junitJupiterVersion = '5.9.2'
-	kotlinCoroutinesVersion = '1.6.4'
-	log4jVersion = '2.19.0'
-	logbackVersion = '1.4.4'
+	junitJupiterVersion = '5.10.0'
+	kotlinCoroutinesVersion = '1.7.3'
```
artembilan, line 58: There is also `ext.kotlinVersion = '1.7.0'` in the beginning of this file.
It has to go to: `1.9.10`

artembilan (changes requested):

artembilan:
There are also a couple tests failing.
Probably they also need an explicit `@Qualifier` for their `autowire` props...

Issue #2532, opened 2023-09-27
Ignore Kotlin Continuation Parameter While Inferring Conversion Type


garyrussell:
See https://github.com/spring-projects/spring-amqp/issues/1210#issuecomment-1737595939

Issue #2534, opened 2023-09-27
Ignore Kotlin Continuation Parameter While Inferring Conversion Type


garyrussell:
Back port of issue #2532

artembilan:
Fixed via: https://github.com/spring-projects/spring-amqp/commit/fceb0ec14949322ee9a60f68ffc8fb2484fc34ab

Issue #2533, opened 2023-09-27
Ignore Kotlin Continuation Parameter While Inferring Conversion Type


garyrussell:
Back port of issue #2532

PR #2535, opened 2023-09-27
GH-2532: Ignore Kotlin Continuation Parameter

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/2532

The presence of the parameter caused the inferred type to be set to `null` due to ambiguous parameters present.

It must be considered an ineligible type when inferring conversion types.

**cherry-pick to 3.0.x (#2533), 2.4.x (#2534)**


artembilan:
... and cherry-picked to `3.0.x` & `2.4.x`

Issue #2536, opened 2023-09-28
The log level should be adjusted


zhangheng0027:
When the consumer is set to exclusive and multiple instances are started, the console continuously outputs info level restart consumers log.

When setting consumers as exclusive and bound, the restart log level for other consumers should be TRACE or DEBUG.

https://github.com/zhangheng0027/testRabbitMQ.git

```java

@RabbitListener(
            bindings = @QueueBinding(
                    value = @Queue(value = "test2Queue"),
                    exchange = @Exchange(value = "testExchange"),
                    key = "test2"
            ),
            concurrency = "1",
            exclusive = true,
            ackMode = "MANUAL",
            containerFactory = "batchListenerContainerFactory"
    )
    @SneakyThrows
    public void test(@Payload List<Message<EquipmentStatusChangeEvent>> event, Channel channel) {
    }

    @RabbitListener(
            bindings = @QueueBinding(
                    value = @Queue(value = "test2Queue"),
                    exchange = @Exchange(value = "testExchange"),
                    key = "test2"
            ),
            concurrency = "1",
            exclusive = true,
            ackMode = "MANUAL",
            containerFactory = "batchListenerContainerFactory"
    )
    @SneakyThrows
    public void test2(@Payload List<Message<EquipmentStatusChangeEvent>> event, Channel channel) {
    }

```

```
2023-09-28T17:11:57.492+08:00  WARN 21016 --- [ntContainer#1-1] o.s.a.r.l.SimpleMessageListenerContainer : Exclusive consumer failure: com.rabbitmq.client.ShutdownSignalException: channel error; protocol method: #method<channel.close>(reply-code=403, reply-text=ACCESS_REFUSED - queue 'test2Queue' in vhost '/test' in exclusive use, class-id=60, method-id=20)
2023-09-28T17:11:57.498+08:00  INFO 21016 --- [ntContainer#1-1] o.s.a.r.l.SimpleMessageListenerContainer : Restarting Consumer@43982337: tags=[[]], channel=Cached Rabbit Channel: AMQChannel(amqp://oee@192.168.115.248:4626//test,2), conn: Proxy@42fb8c87 Shared Rabbit Connection: SimpleConnection@5a50d9fc [delegate=amqp://oee@192.168.115.248:4626//test, localPort=9562], acknowledgeMode=MANUAL local queue size=0
[...]

PR #2537, opened 2023-09-28
GH-2536: Exclusive Consumer Logging Improvements

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/2536

Log messages due to access refused due to exclusive consumers at DEBUG level instead of WARN and INFO.


spring-amqp/src/main/java/org/springframework/amqp/support/ConditionalExceptionLogger.java
```diff
@@ -37,4 +39,16 @@ public interface ConditionalExceptionLogger {
 	 */
 	void log(Log logger, String message, Throwable t);
 
+	/**
+	 * Log a consumer restart; debug by default.
+	 * @param logger the logger.
+	 * @param message the message.
+	 * @since 3.1
+	 */
+	default void logRestart(Log logger, Supplier<String> message) {
```
artembilan, line 48: Maybe we can use `org.springframework.core.log.LogMessage` instead?
And then we won't need the next `isDebugEnabled()`

artembilan (changes requested):

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -1460,7 +1462,13 @@ private void killOrRestart(boolean aborted) {
 				}
 			}
 			else {
-				logger.info("Restarting " + this.consumer);
+				if (this.failedExclusive) {
+					getExclusiveConsumerExceptionLogger().logRestart(logger,
+							LogMessage.of(() -> "Restarting " + this.consumer));
+				}
+				else {
+					logger.info("Restarting " + this.consumer);
```
artembilan, line 1470: Maybe this is moment when we can change that `logger` to the `LogAccessor` and use here a supplier variant?
Otherwise, consider to wrap it into `isInfoEnabled()`

artembilan (changes requested):

PR #2539, opened 2023-10-02
Change Defaults

garyrussell:
**cherry-pick to 3.0.x, 2.4.x**

No back port issues needed.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -313,6 +313,8 @@ configure(javaProjects) { subproject ->
 		if (name ==~ /(testAll)/) {
 			systemProperty 'RUN_LONG_INTEGRATION_TESTS', 'true'
 		}
+		environment "SPRING_AMQP_DESERIALIZATION_TRUST_ALL", "true"
+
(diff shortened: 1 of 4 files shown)


artembilan:
... and cherry-picked to `3.0.x` & `2.4.x`

ko-at-work:
@artembilan @garyrussell Are there any more details around why this change was made? We upgraded from spring-amqp 2.4.16 to 2.4.17, and it broke things because it was working based on the prior default explanation. This does not seem like it belongs in minor or patch version update. Is there an associated CVE this is addressing?

garyrussell:
I have hidden your comment; it is not appropriate to identify code that fixes CVEs. https://spring.io/security/cve-2023-34050

garyrussell:
I have hidden your comment; it is not appropriate to identify code that fixes CVEs. https://spring.io/security/cve-2023-34050

Our security team felt that it was better to protect the whole community, even while inconveniencing existing users.

Issue #2552, opened 2023-10-31
Stream/Queue size in bytes (x-max-length-bytes)


garyrussell:
### Discussed in https://github.com/spring-projects/spring-amqp/discussions/2551

Change `SuperStreamBuilder` `maxLength` to `long` and add overloaded method to `QueueBuilder`.

Also `maxSegmentSize`.

Erlang int is 8 bytes.

PR #2553, opened 2023-10-31
GH-2552: Fix Builder Argument Types maxLength etc

garyrussell:
Resolves https://github.com/spring-projects/spring-amqp/issues/2552

Variables map to Erlang ints which are 8 bytes, so need to be long in Java.


artembilan:
@garyrussell ,

is this supposed to be back-ported to `3.0.x` ?

garyrussell:
No; `SuperStreamBuilder` doesn't exist there.

I suppose we could back port the `QueueBuilder` part, but nobody has complained about it so we can defer that, I think.

garyrussell:
Also, `maxLength` is messages there (Vs. bytes with streams) so it's probably less of an issue for queues.

PR #2564, opened 2023-12-05
Antora

rwinch:


src/reference/antora/.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,33 @@
+name: Deploy Docs
```
artembilan, line 1: I think this GHA workflow file is some kind of duplication and out of use.
Or is that a new approach how things go from now on?
Then why do we need it then in the top-level `.github/workflows`?..

build.gradle
```diff
@@ -37,8 +39,8 @@ ext {
 	linkScmDevConnection = 'git@github.com:spring-projects/spring-amqp.git'
 	springAsciidoctorBackendsVersion = '0.0.7'
 
-	modifiedFiles =
-			files(grgit.status().unstaged.modified).filter{ f -> f.name.endsWith('.java') || f.name.endsWith('.kt') }
+	modifiedFiles = providers.provider {
+			files(grgit.status().unstaged.modified).filter{ f -> f.name.endsWith('.java') || f.name.endsWith('.kt') }}
```
artembilan, line 43: In my experience in other projects this has to be like this:
```
    modifiedFiles =
            files()
                    .from {
                        files(grgit.status().unstaged.modified)
                                .filter { f -> f.name.endsWith('.java') || f.name.endsWith('.kt') }
                    }
    modifiedFiles.finalizeValueOnRead()
```

Lazy, once calculated on demand and cached.

src/reference/antora/antora-playbook.yml
```diff
@@ -0,0 +1,44 @@
...
+  sourcemap: true
+urls:
+  latest_version_segment: ''
+runtime:
+  log:
+    failure_level: warn
+    format: pretty
+ui:
+  bundle:
+    url: https://github.com/spring-io/antora-ui-spring/releases/download/v0.3.4/ui-bundle.zip
```
artembilan, line 44: Looks like we have already much fresh version for this: https://github.com/spring-io/antora-ui-spring/releases.
Any particular reason to stay on the old one, please?

artembilan (changes requested):

build.gradle
```diff
@@ -223,7 +261,7 @@ configure(javaProjects) { subproject ->
 
 	task updateCopyrights {
 		onlyIf { !isCI }
-		inputs.files(modifiedFiles.filter { f -> f.path.contains(subproject.name) })
+		inputs.files(modifiedFiles.map(files -> files.filter { f -> f.path.contains(subproject.name) }))
```
artembilan, line 264:
```suggestion
		inputs.files(modifiedFiles.filter { f -> f.path.contains(subproject.name) })
```

Issue #2601, opened 2024-01-17
Add batchReceiveTimeout in SimpleMessageListenerContainer


laststem:
**Expected Behavior**

`receiveTimeout` is The maximum time to wait for each message.
If there is `batchReceiveTimeout`, the time while waiting for the entire batchSize can be applied not each message.

**Current Behavior**

There is no batchReceiveTimeout.
so, receiveTimeout applies to individual messages.. wait for `batchSize*receiveTimeout`  amount of time in worst case.

artembilan:
Sorry, the info you provide is not enough.
See `doReceiveAndExecute(BlockingQueueConsumer consumer)` logic in the `SimpleMessageListenerContainer`.
It really means that we would like to gather the whole batch before processing if you enable `consumerBatchEnabled`.
Therefore it feel natural to wait for each message for that `receiveTimeout` independently.
Otherwise we try to debatch a single message and break the loop.

I don't see how that extra new property can be helpful.

Please, elaborate.

laststem:
@artembilan Thanks for reply. and sorry for not enough explanation.

If it is not a batchlistener or batchSize is 1, it is natural to wait for each message for receiveTimeout.

But if batchSize is 1000 and receiveTimeout is 1000ms and `consumer.nextMessage(this.receiveTimeout)` waits for (receiveTimeout-1)ms to retrieve the message, the total waiting time(1000 * 1000ms) is too harsh.

This is that if messages are provided to BlockingQueueConsumer.queue every (receiveTimeout-1)ms.
As a result, in this situation, it waits for receiveTimeout to fetch one message, and this is repeated for as many times as batchSize, so the previously fetched messages have no choice but to wait a long time.

So, batchReceiveTimeout it can limit the time it waits to fill batchSize.
while accumulating waiting time through `consumer.nextMessage(this.receiveTimeout)`, if batchReceiveTimeout is exceeded, the loop is stopped and executeWithList is called.

PR #2605, opened 2024-01-22
GH-2601: Add a batchReceiveTimeout

laststem:
Fixes https://github.com/spring-projects/spring-amqp/issues/2601

stop to waiting next message and execute listener when batchReceiveTimeout is timed out.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/ListenerContainerFactoryBean.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/ListenerContainerFactoryBean.java
@@ -392,6 +392,14 @@ public class ListenerContainerFactoryBean extends AbstractFactoryBean<AbstractMe
 		this.receiveTimeout = receiveTimeout;
 	}
 
+	/**
+	 * The number of milliseconds of timeout for gathering batch messages.
+	 * It limits the time to wait to fill batchSize.
+	 * Default is 0 (no timeout).
+	 * @param batchReceiveTimeout the timeout for gathering batch messages.
+	 * @since 3.1.2
+	 * @see #setBatchSize(int)
+	 */
 	public void setBatchReceiveTimeout(long batchReceiveTimeout) {
 		this.batchReceiveTimeout = batchReceiveTimeout;
 	}
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/ListenerContainerFactoryBean.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/ListenerContainerFactoryBean.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2022 the original author or authors.
+ * Copyright 2016-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -55,6 +55,7 @@ import org.springframework.util.backoff.BackOff;
  * @author Gary Russell
  * @author Artem Bilan
  * @author Johno Crawford
+ * @author Jeonggi Kim
  *
  * @since 2.0
  *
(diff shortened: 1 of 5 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -121,6 +124,8 @@ public class SimpleMessageListenerContainer extends AbstractMessageListenerConta
 
 	private long receiveTimeout = DEFAULT_RECEIVE_TIMEOUT;
 
+	private long batchReceiveTimeout = DEFAULT_BATCH_RECEIVE_TIMEOUT;
```
artembilan, line 127: I don't think we can assume some default.
This has to be an `opt-in` property.
Otherwise it is a breaking change for the current patch version.
Therefore I suggest to have a default as it is right now: no any batch timeouts.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -330,6 +335,16 @@ public void setReceiveTimeout(long receiveTimeout) {
 		this.receiveTimeout = receiveTimeout;
 	}
 
+	/**
+	 * The time (in milliseconds) that a waiting time to fill batch size. Default
+	 * 3000 (3 second).
+	 * @param batchReceiveTimeout the timeout
```
artembilan, line 341: `@since 3.1.2`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -996,8 +1011,14 @@ private boolean doReceiveAndExecute(BlockingQueueConsumer consumer) throws Excep
 
 		List<Message> messages = null;
 		long deliveryTag = 0;
+		long startTime = System.currentTimeMillis();
 
 		for (int i = 0; i < this.batchSize; i++) {
+			boolean batchTimedOut = (System.currentTimeMillis() - startTime) > this.batchReceiveTimeout;
+			if (batchTimedOut) {
+				logger.trace("Timed out for gather batch messages.");
```
artembilan, line 1019: It might be useful to mention also how many messages are gathered for this batch during this interval.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/SimpleRabbitListenerContainerFactory.java
```diff
@@ -121,6 +124,14 @@ public void setReceiveTimeout(Long receiveTimeout) {
 		this.receiveTimeout = receiveTimeout;
 	}
 
+	/**
+	 * @param batchReceiveTimeout the timeout for gather batch messages.
+	 * @see SimpleMessageListenerContainer#setBatchReceiveTimeout
```
artembilan, line 129: Some short description in the Javadoc required.
Plus `@since 3.1.2`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/ListenerContainerFactoryBean.java
```diff
@@ -389,6 +392,10 @@ public void setReceiveTimeout(long receiveTimeout) {
 		this.receiveTimeout = receiveTimeout;
 	}
 
+	public void setBatchReceiveTimeout(long batchReceiveTimeout) {
```
artembilan, line 395: Add Javadoc here, please.
To be short: every public API must be with Javadocs.
I know, we have many missed, but let's cover at least whatever new we add so far!

artembilan (changes requested):
Please, also add this new property into the `containerAttributes.adoc`.
Just a new line in the table after `receiveTimeout`!

laststem:
@artembilan Thanks for review. please again.

artembilan:
@laststem ,

PR #2646, opened 2024-03-06
Ability to provide an implementation of ConnectionFactory

william00179:
Currently it's not possible to provide a custom implementation of a ConnectionFactory to this class. 

I need to be able to override some methods in the factory and being able to simply pass the implementation will prevent me from having to use hacky workarounds.

This is required in order to make modifications to allow use of an explicit proxy.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
@@ -121,7 +121,7 @@ public class RabbitConnectionFactoryBean extends AbstractFactoryBean<ConnectionF
 
 	private static final String TRUST_STORE_DEFAULT_TYPE = "JKS";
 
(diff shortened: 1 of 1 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
```diff
@@ -162,6 +162,11 @@ public class RabbitConnectionFactoryBean extends AbstractFactoryBean<ConnectionF
 	private String trustStoreAlgorithm = SUN_X509;
 
 	public RabbitConnectionFactoryBean() {
+		this(new ConnectionFactory());
+	}
+
+	public RabbitConnectionFactoryBean(ConnectionFactory connectionFactory) {
```
artembilan, line 168: This is not correct usage of the factory pattern: if you have `com.rabbitmq.client.ConnectionFactory` instance, then you don't need `RabbitConnectionFactoryBean` at all.
Also: we follow the rule do not mutate externally provided objects.

So, I cannot accept this PR altogether.

Will keep it open until your answer.

Thanks for understanding.

Issue #2572, opened 2023-12-15
RabbitTemplate with SIMPLE correlation causes uncached channels


artembilan:
### Discussed in https://github.com/spring-projects/spring-amqp/discussions/2571

<div type='discussions-op-text'>

<sup>Originally posted by **bjoernhaeuser** December 15, 2023</sup>
Hello everyone,

I hope this discussion is at the right place. 

We are using a CachingConnectionFactory with `CachingConnectionFactory.ConfirmType.SIMPLE` as publisher confirm type setting.

Then we are using the RabbitTemplate something like this:
```
this.amqpTemplate.invoke(operations -> {
    // send messages

    // operations.waitForConfirmsOrDie(10_000);
}   
```

I thought this reuses channels, but after looking at the RabbitMQ stats I see a relatively high channel churn (in our case 50-100 channels/sec) and digging into this it looks like the RabbitTemplate is creating a new channel for every `.invoke()` call. 

After a debugging session through the code I saw that the `RabbitTemplate::invoke` method calls `RabbitUtils.setPhysicalCloseRequired(channel, true);` [Reference](https://github.com/spring-projects/spring-amqp/blob/v3.1.0/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java#L2323), when the connectionFactory says that publisher confirms are deactivated. 

Right now the `connectionFactory.isPublisherConfirms()` would only return true when the `ConfirmType` is set to `CORRELATED` [Reference](https://github.com/spring-projects/spring-amqp/blob/v3.1.0/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java#L388).

I am not familiar enough to judge if this is the right condition, but it looks to be that it should be changed to say `return !ConfirmType.NONE.equals(this.confirmType)`.

Changing the value to be of confirmType `CORRELATED` changes the behaviour and channels are properly cached. Can someone please advise if my findings are correct and/or if I am using it simply wrong?

Thank you very much!

</div>

bjoernhaeuser:
I want ahead and started to work on this in #2717

PR #2651, opened 2024-03-11
Change the method return condition. connectionFactory.isPublisherConfirms()

esperar:
I fix this issue https://github.com/spring-projects/spring-amqp/issues/2572

`connectionFactory.isPublisherConfirms()` return condition.
I believe this condition is appropriate, **but some people may not think so, as mentioned in the issue above.** 
Please confirm.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
@@ -386,7 +386,7 @@ public class CachingConnectionFactory extends AbstractConnectionFactory
 
 	@Override
 	public boolean isPublisherConfirms() {
-		return ConfirmType.CORRELATED.equals(this.confirmType);
+		return !ConfirmType.NONE.equals(this.confirmType);
 	}
 
 	@Override


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
```diff
@@ -386,7 +386,7 @@ public void setConnectionLimit(int connectionLimit) {
 
 	@Override
 	public boolean isPublisherConfirms() {
-		return ConfirmType.CORRELATED.equals(this.confirmType);
+		return !ConfirmType.NONE.equals(this.confirmType);
```
artembilan, line 389: This is a behavior change in the patch version.
So, we are good with the fix, but only when we start a new minor `3.2` version in May.

Might be grate if you prove the fix with some tests meanwhile.

Thanks

artembilan (changes requested):

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
esperar: OK

artembilan:
Superseded with #2717.

Thank you for contribution anyway!

Issue #2653, opened 2024-03-13
Deadlock when reaching channel limit in DirectMessageListenerContainer


utikeev:
**In what version(s) of Spring AMQP are you seeing this issue?**

2.4.11 (Reproduction example is present for 3.1.2)

**Describe the bug**

We use `convertSendAndReceive` for asynchronous RPC in our project. Invoking such a method requires getting a new consumer (and thus requiring a channel to communicate with the broker).
In case when the load is high and all the channels are occupied, the consumers are tried to be created under `consumersLock` in the while loop, always failing to create a new one (`DirectMessageListenerContainer.adjustConsumers`). With the lock being acquired by the `adjustConsumers`, the timed out consumers can't be cancelled in `DirectMessageListenerContainer.checkConsumers` thus never releasing any channels.

**To Reproduce**

Run the example `RabbitmqDeadlockApplication` and observe that `TaskSubmitter` never has its futures completed. `main` thread is blocking `asyncRabbitTemplate-1` thread that tries to check the consumers and probably cancel some of them.

**Expected behavior**

Timed out requests release underlying channels leading to eventually completing all futures.

**Sample**

https://github.com/utikeev/spring-amqp-directmessagelistenercontainer-deadlock
The sample has a limit for channels set to 2 to make it easier to reproduce (with just 5 `convertSendAndReceive`). In our case, we hit the broker limit of 2047 channels per connection.
We also set the `MessageListener` programmatically, but that shouldn't be the issue as far as I understand. 

----

The problem might be about the wrong usage of Spring AMQP API. In that case, if you could recommend a better way to handle that use-case, I'd consider that a solution as well.

artembilan:
Thank you for the report!
Modified your application to this simple unit test:
```
	@Test
	public void limitedChannelsAreReleasedOnTimeout() {
		CachingConnectionFactory connectionFactory = new CachingConnectionFactory("localhost");
		connectionFactory.setChannelCacheSize(1);
		connectionFactory.setChannelCheckoutTimeout(500L);
		RabbitTemplate rabbitTemplate = new RabbitTemplate(connectionFactory);
		rabbitTemplate.setReceiveTimeout(100L);
		AsyncRabbitTemplate asyncRabbitTemplate = new AsyncRabbitTemplate(rabbitTemplate);
		asyncRabbitTemplate.start();

		RabbitConverterFuture<String> replyFuture1 = asyncRabbitTemplate.convertSendAndReceive("noReply1");
[...]

PR #2655, opened 2024-03-14
GH-2653: Fix deadlock in the DirectMessageListenerContainer

artembilan:
Fixes: #2653

When not enough channel in the cache, the `DirectMessageListenerContainer.consume()` returns null and `adjustConsumers()` goes into an infinite loop, since already active consumer does not release its channel.

* Fix `DirectMessageListenerContainer.consume()` to re-throw an `AmqpTimeoutException` which is thrown when no available channels in the cache
* Catch `AmqpTimeoutException` in the `DirectReplyToMessageListenerContainer.getChannelHolder()` and reset `this.consumerCount--` to allow to try existing consumer until it is available, e.g. when this one receives a reply or times out.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2023 the original author or authors.
+ * Copyright 2016-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


sobychacko (approved):

utikeev:
Is it safe to update `consumerCount` without a lock (or at least non-atomically)?
I mean, don't `processMonitorTask` and `getChannelHolder` execute on different threads?

artembilan:
@utikeev ,

thank you for the pointer!
So, changed that `consumerCount` to an `AtomicInteger`.
The point of the fix is to keep that `super.setConsumersPerQueue()` out of the lock.

artembilan:
Merging this into `main` for today's release.
At least this fixes an immediate dead-lock problem.
Anything else we can revise later when it arises.

Thanks

PR #2671, opened 2024-03-31
more concise and intuitive.

Java4ye:
just a little change


artembilan (changes requested):
Please, add your name to the `@author` list of the affected class.

PR #2677, opened 2024-04-06
Refactor getOrSetCorrelationIdAndSetReplyTo method

Java4ye:
Refactored the getOrSetCorrelationIdAndSetReplyTo method for improved readability and code organization. Merged the declaration and initialization of 'correlationId' into a single line. Ensured the logic remains unchanged. No functional changes introduced.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/AsyncRabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/AsyncRabbitTemplate.java
@@ -87,6 +87,7 @@ import com.rabbitmq.client.Channel;
  *
  * @author Gary Russell
  * @author Artem Bilan
+ * @author Java4ye
  *
  * @since 1.6
  */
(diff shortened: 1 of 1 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/AsyncRabbitTemplate.java
```diff
@@ -87,6 +87,7 @@
  *
  * @author Gary Russell
  * @author Artem Bilan
+ * @author Java4ye
```
artembilan, line 90: Sorry, that's not the name.
We can accept here only your official name.

artembilan (changes requested):
The change is OK, but we need your real name to add into `@author` list.
Thanks

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/AsyncRabbitTemplate.java
Java4ye: ok，I fixed it.
Thanks.

PR #2689, opened 2024-04-23
Restrict `getDeclarablesByType()` to Declarable

bjmi:
see title

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/core/Declarables.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/core/Declarables.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2021 the original author or authors.
+ * Copyright 2018-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -59,7 +59,7 @@ public class Declarables {
 	 * @return the filtered list.
 	 * @since 2.2
 	 */
-	public <T> List<T> getDeclarablesByType(Class<T> type) {
+	public <T extends Declarable> List<T> getDeclarablesByType(Class<? extends T> type) {
 		return this.declarables.stream()
 				.filter(type::isInstance)
 				.map(type::cast)


spring-amqp/src/main/java/org/springframework/amqp/core/Declarables.java
```diff
@@ -59,7 +59,7 @@ public Collection<Declarable> getDeclarables() {
 	 * @return the filtered list.
 	 * @since 2.2
 	 */
-	public <T> List<T> getDeclarablesByType(Class<T> type) {
+	public <T extends Declarable> List<T> getDeclarablesByType(Class<? extends T> type) {
```
artembilan, line 62: I see the point in your change, but I wonder how it is a breaking change to accept it in the current point version?
It is definitely will be safe to postpone this change until the next `3.2` version after May, since I see that code like this `getDeclarablesByType(String.class)` would not be compilable any more.

artembilan (changes requested):
Please, add your name to the `@author` list.
Thanks.
We will come back here when we start version `3.2`.

spring-amqp/src/main/java/org/springframework/amqp/core/Declarables.java
bjmi: Release `3.2` would be perfectly fine.

bjmi:
> Please, add your name to the `@author` list. Thanks. We will come back here when we start version `3.2`.

Do you mean another @author list, because I'm already on the list in `Declarables`.

artembilan (approved):
>  because I'm already on the list in `Declarables`.

Sorry, missed that.
Was note sure that you are contributor already.

So, all good and will be merged when we start `3.2` version.

Thanks

spring-amqp/src/main/java/org/springframework/amqp/core/Declarables.java
artembilan: Why just `Class<T>` is not enough?
Why do you do like that `? extends T`?

artembilan (changes requested):

spring-amqp/src/main/java/org/springframework/amqp/core/Declarables.java
bjmi: This is to increase usage flexibility.
`? extends T` allows
```
@Override
public Collection<Exchange> getExchanges() {
    return declarables.getDeclarablesByType(TopicExchange.class);
}
```
or
```
List<Queue> queues = declarables.getDeclarablesByType(AnonymousQueue.class);
```

PR #2717, opened 2024-05-21
GH-2572 Treat ConfirmType.SIMPLE same as ConfirmType.CORRELATED

bjoernhaeuser:
GH-2572 Do not close channel when `ConfirmType.SIMPLE` is used for publisher confirmations.

See #2571 (discussion) and #2572 (issue) for reference.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
@@ -2344,6 +2344,9 @@ public class RabbitTemplate extends RabbitAccessor // NOSONAR type line count
 				if (channel == null) {
 					throw new IllegalStateException("Connection returned a null channel");
 				}
+				if (!connectionFactory.isPublisherConfirms() && !connectionFactory.isSimplePublisherConfirms()) {
+					RabbitUtils.setPhysicalCloseRequired(channel, true);
+				}
 				this.dedicatedChannels.set(channel);
 			}
 			catch (RuntimeException e) {
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -2344,9 +2344,6 @@ public <T> T invoke(OperationsCallback<T> action, @Nullable com.rabbitmq.client.
 				if (channel == null) {
 					throw new IllegalStateException("Connection returned a null channel");
 				}
-				if (!connectionFactory.isPublisherConfirms()) {
```
artembilan, line 2347: Why remove altogether?
The discussion was about condition improvement:
```
if (!connectionFactory.isPublisherConfirms() && !connectionFactory.isSimplePublisherConfirms()) {
```

artembilan (changes requested):

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
bjoernhaeuser: Thanks for taking the time to review this!

I went through the discussion and thought the code and came to (maybe faulty) conclusion that closing the channel never makes sense for the `invoke` method. Especially for me this totally unexpected and not mentioned in the docs anywhere. 

Please let me know which way you would like to see this proceed!

Thanks.

artembilan: I believe the close is not required when we are in publish-confirm mode.
Therefore that condition and another PR.
Sorry, I don't see anywhere in discussion that we were going to remove this condition altogether.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -2344,7 +2344,7 @@ public <T> T invoke(OperationsCallback<T> action, @Nullable com.rabbitmq.client.
...
+				if (!connectionFactory.isPublisherConfirms() && !connectionFactory.isSimplePublisherConfirms()) {
```
artembilan, line 2347: Isn't the change in #2651 more precise and accurate?

artembilan (changes requested):

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
bjoernhaeuser: Sorry, this was still a bit WIP. Added this change here as well and added also test for this. Hope this covers this completely.

artembilan (changes requested):
Looks like there is failing existing test since we have changed condition, I guess:
```
RabbitTemplateTests > testPublisherConnWithInvokeAndSimpleConfirmations() FAILED
    org.opentest4j.AssertionFailedError: 
    Expecting value to be false but was true
        at app//org.springframework.amqp.rabbit.core.RabbitTemplateTests.testPublisherConnWithInvokeAndSimpleConfirmations(RabbitTemplateTests.java:631)
```

artembilan:
I had to revert the fix to our original discussion: https://github.com/spring-projects/spring-amqp/commit/2bb4f297931ac39a49d320bd6c81ddf4797c5562.
Apparently there are use-cases when listener is added to the `PublisherCallbackChannel` which makes sense only for the `ConfirmType.CORRELATED` mode.

PR #2757, opened 2024-07-16
Escape Asterisk

ngocnhan-tran1996:
I think asterisk character is missing in the black rounded part. That why I create this PR. Please check again and tell me if I miss something.

image

Reference AMQP Docs

https://docs.spring.io/spring-amqp/reference/amqp/abstractions.html#message


artembilan:
@pivotal-cla This is an Obvious Fix

artembilan:
@ngocnhan-tran1996 ,

Thank you for catching that and fixing; looking forward for more!

Issue #2809, opened 2024-09-05
MultiRabbit containerFactory SpEL Resolution Bug


BenEfrati:
**In what version(s) of Spring AMQP are you seeing this issue?**

3.1.6

**Describe the bug**

The `resolveMultiRabbitAdminName` method in `MultiRabbitListenerAnnotationBeanPostProcessor` does not properly resolve SpEL expressions for the containerFactory attribute of `@RabbitListener` annotations. According to the JavaDoc, the containerFactory attribute should support SpEL expressions, but the current implementation does not resolve these expressions, potentially leading to incorrect admin name resolution.

**Additional context**
JavaDoc for containerFactory() method in RabbitListener:
```
 * The bean name of the
 * {@link org.springframework.amqp.rabbit.listener.RabbitListenerContainerFactory} to
 * use to create the message listener container responsible to serve this endpoint.
 * <p>
 * If not specified, the default container factory is used, if any. If a SpEL
 * expression is provided ({@code #{...}}), the expression can either evaluate to a
 * container factory instance or a bean name.
 * @return the
 * {@link org.springframework.amqp.rabbit.listener.RabbitListenerContainerFactory}
 * bean name.
 */
String containerFactory() default "";
```
https://github.com/spring-projects/spring-amqp/blob/d7058bbac660741bbd84cd14904683fc65696015/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListener.java#L124

**To Reproduce**

1. Define a @RabbitListener annotation with a SpEL expression for the containerFactory attribute.
2. Observe that the resolveMultiRabbitAdminName method does not resolve the SpEL expression.

**Expected behavior**

The method should resolve SpEL expressions for both admin and containerFactory attributes. A proposed fix would be:
```
protected String resolveMultiRabbitAdminName(RabbitListener rabbitListener) {
    String admin = super.resolveExpressionAsString(rabbitListener.admin(), "admin");
    if (!StringUtils.hasText(admin) && StringUtils.hasText(rabbitListener.containerFactory())) {
        String containerFactory = super.resolveExpressionAsString(rabbitListener.containerFactory(), "containerFactory");
        admin = containerFactory + RabbitListenerConfigUtils.MULTI_RABBIT_ADMIN_SUFFIX;
    }
    if (!StringUtils.hasText(admin)) {
        admin = RabbitListenerConfigUtils.RABBIT_ADMIN_BEAN_NAME;
    }
    return admin;
}
[...]

PR #2810, opened 2024-09-06
MultiRabbitListenerAnnotationBeanPostProcessor resolveMultiRabbitAdminName

ngocnhan-tran1996:
issue https://github.com/spring-projects/spring-amqp/issues/2809

Condition in `MultiRabbitListenerAnnotationBeanPostProcessor#resolveMultiRabbitAdminName`

1. Get `RabbitListener#admin` if it's not blank else
2. Get `RabbitListener#containerFactory` return `RabbitListener#containerFactory` + `-admin` if it's not blank else
3. Get `RabbitListenerConfigUtils#RABBIT_ADMIN_BEAN_NAME`

in case 2, I think just get bean name from SpEL `#{@...}`  and concate string, no need call `super#resolveExpressionAsString`. Wherther the given name is valid or not, function chain in class `RabbitListenerAnnotationBeanPostProcessor` will attempt to create RabbitAmin with it.

Please give me more advice if I miss something

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitListenerAnnotationBeanPostProcessor.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitListenerAnnotationBeanPostProcessor.java
@@ -23,9 +23,9 @@ import java.lang.reflect.Proxy;
 import java.util.Collection;
 
 import org.springframework.amqp.core.Declarable;
-import org.springframework.amqp.rabbit.config.BaseRabbitListenerContainerFactory;
 import org.springframework.amqp.rabbit.config.RabbitListenerConfigUtils;
 import org.springframework.amqp.rabbit.core.RabbitAdmin;
+import org.springframework.amqp.rabbit.listener.RabbitListenerContainerFactory;
 import org.springframework.util.StringUtils;
 
 /**
@@ -87,7 +87,7 @@ public class MultiRabbitListenerAnnotationBeanPostProcessor extends RabbitListen
 		if (StringUtils.hasText(rabbitListener.containerFactory())) {
 
 			Object resolved = super.resolveExpression(rabbitListener.containerFactory());
-			if (resolved instanceof BaseRabbitListenerContainerFactory<?> rlcf) {
+			if (resolved instanceof RabbitListenerContainerFactory<?> rlcf) {
 
 				return rlcf.getBeanName() + RabbitListenerConfigUtils.MULTI_RABBIT_ADMIN_SUFFIX;
 			}
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitListenerAnnotationBeanPostProcessor.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitListenerAnnotationBeanPostProcessor.java
@@ -74,26 +74,28 @@ public class MultiRabbitListenerAnnotationBeanPostProcessor extends RabbitListen
 	 */
 	protected String resolveMultiRabbitAdminName(RabbitListener rabbitListener) {
 
-		if (StringUtils.hasText(rabbitListener.admin())) {
+		var admin = rabbitListener.admin();
+		if (StringUtils.hasText(admin)) {
 
-			var admin = super.resolveExpression(rabbitListener.admin());
-			if (admin instanceof RabbitAdmin rabbitAdmin) {
+			var resolved = super.resolveExpression(admin);
+			if (resolved instanceof RabbitAdmin rabbitAdmin) {
 
 				return rabbitAdmin.getBeanName();
 			}
 
-			return super.resolveExpressionAsString(rabbitListener.admin(), "admin");
+			return super.resolveExpressionAsString(admin, "admin");
 		}
 
-		if (StringUtils.hasText(rabbitListener.containerFactory())) {
+		var containerFactory = rabbitListener.containerFactory();
+		if (StringUtils.hasText(containerFactory)) {
 
-			Object resolved = super.resolveExpression(rabbitListener.containerFactory());
+			var resolved = super.resolveExpression(containerFactory);
 			if (resolved instanceof RabbitListenerContainerFactory<?> rlcf) {
 
 				return rlcf.getBeanName() + RabbitListenerConfigUtils.MULTI_RABBIT_ADMIN_SUFFIX;
 			}
 
-			return rabbitListener.containerFactory() + RabbitListenerConfigUtils.MULTI_RABBIT_ADMIN_SUFFIX;
+			return containerFactory + RabbitListenerConfigUtils.MULTI_RABBIT_ADMIN_SUFFIX;
 		}
 
 		return RabbitListenerConfigUtils.RABBIT_ADMIN_BEAN_NAME;
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitListenerAnnotationBeanPostProcessor.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitListenerAnnotationBeanPostProcessor.java
@@ -23,9 +23,9 @@ import java.lang.reflect.Proxy;
 import java.util.Collection;
 
 import org.springframework.amqp.core.Declarable;
+import org.springframework.amqp.rabbit.config.BaseRabbitListenerContainerFactory;
 import org.springframework.amqp.rabbit.config.RabbitListenerConfigUtils;
-import org.springframework.amqp.rabbit.listener.RabbitListenerContainerFactory;
-import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
+import org.springframework.amqp.rabbit.core.RabbitAdmin;
 import org.springframework.util.StringUtils;
 
 /**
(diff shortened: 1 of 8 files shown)


artembilan (changes requested):
Thank you for proposed fix!
Unfortunately we cannot go via simple expression parsing where it would simply be just a plain bean name without an expression wrapping. The point of the expression is really let to do more complex logic for that resolution.

See my comment in the respective issue to make it more robust: https://github.com/spring-projects/spring-amqp/issues/2809#issuecomment-2332125923

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -71,13 +73,33 @@ private RabbitListener proxyIfAdminNotPresent(final RabbitListener rabbitListene
...
+			return RabbitListenerConfigUtils.RABBIT_ADMIN_BEAN_NAME;
 		}
-		return admin;
+
+		if (this.beanFactory instanceof ConfigurableListableBeanFactory clbf) {
+
+			Object resolved = super.resolveExpression(rabbitListener.containerFactory());
+			if (resolved instanceof RabbitListenerContainerFactory<?> rlcf) {
+
+				var containerFactoryByName = clbf.getBeansOfType(RabbitListenerContainerFactory.class);
```
artembilan, line 91: I still don't approve this.
See `BeanNameAware` to be implemented on the `RabbitListenerContainerFactory`.
Together with a getter.
The `RabbitAdmin` already comes with the impl.
That `super.resolveExpressionAsString(rabbitListener.admin(), "admin")` would fail if an expression points to the target bean.
So, let's improve that part as well!

artembilan (changes requested):

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitListenerAnnotationBeanPostProcessor.java
ngocnhan-tran1996: @artembilan 

I implement `BeanNameAware`. Please review and tell me if I miss something

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/BaseRabbitListenerContainerFactory.java
```diff
@@ -218,4 +220,13 @@ protected ApplicationContext getApplicationContext() {
 		return this.applicationContext;
 	}
 
+	@Override
+	public void setBeanName(String name) {
+		this.beanName = name;
+	}
+
+	public String getBeanName() {
```
artembilan, line 228: Please, move these two new method into that `RabbitListenerContainerFactory` interface with `default` impl to avoid a breaking change.

The `getBean()` name also would need some simple Javadoc with a `@since 3.2`.

artembilan (changes requested):
And your name to the `@author` section of all the affected classes.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/BaseRabbitListenerContainerFactory.java
ngocnhan-tran1996: I typed a comment in `RabbitListenerContainerFactory`. Please review and tell me if I miss something

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -70,14 +73,30 @@ private RabbitListener proxyIfAdminNotPresent(final RabbitListener rabbitListene
...
+
+		if (StringUtils.hasText(rabbitListener.admin())) {
+
+			var admin = super.resolveExpression(rabbitListener.admin());
+			if (admin instanceof RabbitAdmin rabbitAdmin) {
+
+				return rabbitAdmin.getBeanName();
+			}
+
+			return super.resolveExpressionAsString(rabbitListener.admin(), "admin");
```
artembilan, line 85: I don't think we need to call `(rabbitListener.admin()` so may times.
Why cannot we just extract a local variable for its result?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/MultiRabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -70,14 +73,30 @@ private RabbitListener proxyIfAdminNotPresent(final RabbitListener rabbitListene
...
+
+		if (StringUtils.hasText(rabbitListener.containerFactory())) {
+
+			Object resolved = super.resolveExpression(rabbitListener.containerFactory());
+			if (resolved instanceof RabbitListenerContainerFactory<?> rlcf) {
+
+				return rlcf.getBeanName() + RabbitListenerConfigUtils.MULTI_RABBIT_ADMIN_SUFFIX;
+			}
+
+			return rabbitListener.containerFactory() + RabbitListenerConfigUtils.MULTI_RABBIT_ADMIN_SUFFIX;
```
artembilan, line 96: DITTO

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/RabbitListenerContainerFactory.java
```diff
@@ -48,4 +49,17 @@ default C createListenerContainer() {
 		return createListenerContainer(null);
 	}
 
+	/**
+	 * set a bean name.
+	 * @since 3.2
+	 */
```
artembilan, line 55: We don't need this JavaDoc since the method is inherited from the `BeanNameAware`.
However it still have to come with an `@Override`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/RabbitListenerContainerFactory.java
```diff
@@ -48,4 +49,17 @@ default C createListenerContainer() {
...
+	void setBeanName(String name);
+
+	/**
+	 * Return a bean name of the component or null if not a bean.
+	 * @return the bean name.
+	 * @since 3.2
+	 */
+	String getBeanName();
```
artembilan, line 63: No. They have to be `default` to avoid a breaking change in this minor version.
This way the whole interface would still be a `@FunctionalInterface`.

Plus mark this method as `@Nullable`.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/config/RabbitListenerContainerTestFactory.java
```diff
@@ -63,4 +66,13 @@ public MessageListenerTestContainer createListenerContainer(RabbitListenerEndpoi
 		return container;
 	}
 
+	@Override
+	public void setBeanName(String name) {
+		this.beanName = name;
+	}
+
+	public String getBeanName() {
+		return this.beanName;
```
artembilan, line 75: If those methods are `default` in the interface, we don't need changes to this test entity and, therefore, to the whole class.

artembilan (changes requested):

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/config/RabbitListenerContainerTestFactory.java
ngocnhan-tran1996: I think we should test method `getBeanName` in `AbstractRabbitAnnotationDrivenTests`

artembilan: I saw that there.
Thanks

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/RabbitListenerContainerFactory.java
```diff
@@ -16,18 +16,19 @@
...
 /**
  * Factory of {@link MessageListenerContainer}s.
  * @param <C> the container type.
  * @author Stephane Nicoll
  * @author Gary Russell
+ * @author Ngoc Nhan
  * @since 1.4
  * @see RabbitListenerEndpoint
  */
-@FunctionalInterface
```
artembilan, line 29: You didn't bring this back yet.

artembilan (changes requested):
Oh! I see now what you mean:
```
 EnableRabbitTests > sampleConfiguration() FAILED
    org.opentest4j.AssertionFailedError: 
    expected: "rabbitListenerContainerFactory"
     but was: null
        at app//org.springframework.amqp.rabbit.annotation.AbstractRabbitAnnotationDrivenTests.testSampleConfiguration(AbstractRabbitAnnotationDrivenTests.java:91)
        at app//org.springframework.amqp.rabbit.annotation.EnableRabbitTests.sampleConfiguration(EnableRabbitTests.java:71)
```
Apparently those tests are based on some test entity.
So, yeah: we have to have these new methods implemented in that `RabbitListenerContainerTestFactory` as well.

Why do you push commits for every line of code you change?
Why just don't combine all the requested changes and send a single commit with a proper message?

Thanks

ngocnhan-tran1996:
> And your name to the `@author` section of all the affected classes.



> Oh! I see now what you mean:
> 
> ```
>  EnableRabbitTests > sampleConfiguration() FAILED
>     org.opentest4j.AssertionFailedError: 
>     expected: "rabbitListenerContainerFactory"
>      but was: null
>         at app//org.springframework.amqp.rabbit.annotation.AbstractRabbitAnnotationDrivenTests.testSampleConfiguration(AbstractRabbitAnnotationDrivenTests.java:91)
>         at app//org.springframework.amqp.rabbit.annotation.EnableRabbitTests.sampleConfiguration(EnableRabbitTests.java:71)
> ```
> 
> Apparently those tests are based on some test entity. So, yeah: we have to have these new methods implemented in that `RabbitListenerContainerTestFactory` as well.
> 
> Why do you push commits for every line of code you change? Why just don't combine all the requested changes and send a single commit with a proper message?
> 
> Thanks

Sorry, my network is so lag. After reloading github page, I have seen your newest comment
I bring it back and test is passed

artembilan:
@ngocnhan-tran1996 ,

ngocnhan-tran1996:
@artembilan 

Thanks for your review. I have learn a lot.

Issue #2814, opened 2024-09-11
OpenTelemetry with RabbitMQ add messaging.rabbitmq.destination.routing_key from the semantic convention


vmeunier:
**Expected Behavior**

There are several fields in the [semantic convention for RabbitMQ](https://opentelemetry.io/docs/specs/semconv/messaging/rabbitmq/) and RabbitTemplate is missing the one named `messaging.rabbitmq.destination.routing_key`.

**Current Behavior**

At the moment, in RabbitTemplate, the observationContext concatenates "exchange/routingKey" in the field `messaging.destination.name`

**Context**

If we want to be fully "OTEL compliant" then we should use all the fields we can with the information we have.
IMO it is a mistake to concatenate two different strings in one, it makes the research or filtering in traces/metrics afterwards more complicated.

`message.destination.name` should be the name of the exchange
`messaging.rabbitmq.destination.routing_key` must be the name of the routingKey

```doc
In RabbitMQ, the destination is defined by an exchange and a routing key. messaging.destination.name MUST be set to the name of the exchange. This will be an empty string if the default exchange is used.
```

artembilan:
We do not follow those conventions by default because of `Status: Experimental`.
However additional tags to whatever we have chosen by default would not harm.
In fact we had similar contribution in the past into Spring for Apache Kafka: https://github.com/spring-projects/spring-kafka/issues/2609.

So, feel free to raise respective PR and we will be more than happy to review and merge!

vmeunier:
Ok, i'll try to send you a PR :)

PR #2816, opened 2024-09-13
Add exchange/routingKey to RabbitMessageSenderContext

ngocnhan-tran1996:
I see https://github.com/spring-projects/spring-amqp/issues/2814 and I think `RabbitMessageSenderContext.java` should have field `exchange` and `routingKey`.

But I think it's better if we use `MessageProperties#setReceivedExchange` and `MessageProperties#setReceivedRoutingKey` instead.


artembilan:
The `setReceivedExchange` and `setReceivedRoutingKey` are for consumer side.
Since there is no options in the `org.springframework.amqp.core.Message` fro destination properties, we don't have have unless provide them separately.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitMessageSenderContext.java
```diff
@@ -33,11 +33,17 @@ public class RabbitMessageSenderContext extends SenderContext<Message> {
 
 	private final String destination;
 
-	public RabbitMessageSenderContext(Message message, String beanName, String destination) {
+	private final String exchange;
+
+	private final String routingKey;
+
+	public RabbitMessageSenderContext(Message message, String beanName, String exchange, String routingKey) {
```
artembilan, line 40: This is a breaking change.
Please, consider to deprecate an existing ctor and introduce a new one.
The deprecated one has to call a new one via splitting on that `/`.
Just in case someone uses the ctor we just deprecating.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitMessageSenderContext.java
```diff
@@ -53,4 +59,20 @@ public String getDestination() {
 		return this.destination;
 	}
 
+	/**
+	 * Return the exchange.
+	 * @return the exchange.
```
artembilan, line 64: `@since 3.2`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitMessageSenderContext.java
artembilan, line 40: Some general JavaDoc and `@since 3.2`

artembilan (changes requested):
Also add your name to the `@author` list of all the affected classes.
Don't you think that we need to cover the change somehow in the tests?
Or that one is going to be follow up as a fix for the #2814 ?

Thanks

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitMessageSenderContext.java
ngocnhan-tran1996: I set `null` value if someone use the old ctor

I think we should not split `/` in the old ctor.

For example
`destination = exchange + "/" + routingKey`
exchange = `//`
routingKey = `//`
the destination = `/////`

RabbitMQ does not validate the '/` character when we give a name to the queue or the exchange, and it's so hard to determine our `/`

ngocnhan-tran1996:
This one is going to be follow up as a fix for the https://github.com/spring-projects/spring-amqp/issues/2814

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitMessageSenderContext.java
artembilan: Yes. That makes sense.
Thank you!

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitMessageSenderContext.java
```diff
@@ -37,6 +38,27 @@ public class RabbitMessageSenderContext extends SenderContext<Message> {
...
+		this.exchange = null;
+		this.routingKey = null;
+		this.destination = destination;
+		setRemoteServiceName("RabbitMQ");
+	}
+
+
+	/**
+	 * Create an instance {@code RabbitMessageSenderContext}.
+	 *
```
artembilan, line 55: No blank lines in the method JavaDocs

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/support/micrometer/ObservationTests.java
```diff
@@ -110,7 +111,9 @@ void endToEnd(@Autowired Listener listener, @Autowired RabbitTemplate template,
 
 			@Override
 			public KeyValues getLowCardinalityKeyValues(RabbitMessageSenderContext context) {
-				return super.getLowCardinalityKeyValues(context).and("foo", "bar");
+				return super.getLowCardinalityKeyValues(context).and("foo", "bar")
+					.and("messaging.destination.name", context.getExchange())
+					.and("messaging.rabbitmq.destination.routing_key", context.getRoutingKey());
```
artembilan, line 116: I guess this will go to `TemplateLowCardinalityTags` in the next contribution.

artembilan (changes requested):

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/support/micrometer/ObservationTests.java
ngocnhan-tran1996: I think now it's more easier to determine exchange or routing key for doing traces/metrics without splitting `/` in `RabbitMessageSenderContext`.

ngocnhan-tran1996:
@vmeunier 

I follow your issue and added field exchange/routingkey. I think it will be easier for you to implement metrics/traces, and IMO it's not done yet.

vmeunier:
@ngocnhan-tran1996 ok you're right, I don't know why you needed to add those informations, it confused me since I knew I had that stuff to change :)

artembilan:
I believe the original issue talks about adding  something like:
```
					.and("messaging.destination.name", context.getExchange())
					.and("messaging.rabbitmq.destination.routing_key", context.getRoutingKey());
```
into out-of-the-box convention.
That one is not done yet, so keeping it opened until respective contribution.
Thanks

PR #2839, opened 2024-09-26
Add RabbitTemplate getBeforePublishPostProcessors method

leshalv:


artembilan (changes requested):
I cannot approve this change without any reasoning.
Share with us, please, your experience why would one get access to this property?

leshalv:
The `RabbitTemplate` class currently provides a `setBeforePublishPostProcessors` method to configure `MessagePostProcessor` instances that can modify messages before they are sent. However, there is no corresponding `getBeforePublishPostProcessors` method to retrieve the currently configured processors. This PR introduces the `getBeforePublishPostProcessors` method, enhancing the flexibility and configurability of `RabbitTemplate`.

#### Justification:

1. **Increased Flexibility**:
   - While `setBeforePublishPostProcessors` allows users to configure message processors, there is currently no way to retrieve or modify the existing `MessagePostProcessor` list. Adding `getBeforePublishPostProcessors` gives users the ability to access and modify these processors dynamically, allowing for greater flexibility in scenarios where message processing needs to be altered based on context.

2. **Support for Multiple `RabbitTemplate` Instances**:
   - In more complex applications, it's common to create multiple `RabbitTemplate` instances to handle different business logic. For instance, the `ConfirmCallback` mechanism is globally applied, but by manually creating different `RabbitTemplate` instances, users can configure distinct callbacks for each instance. Introducing the `getBeforePublishPostProcessors` method will allow users to retrieve and reuse message processors between different `RabbitTemplate` instances, enhancing flexibility when handling different message routing and confirmation scenarios.

3. **Improved Consistency**:
   - Spring generally follows a "getter-setter" pattern for its components, and many classes offer both methods for configuring and retrieving values. By providing `getBeforePublishPostProcessors`, the API will become more consistent, following Spring’s design principles and making it easier for developers to interact with the `RabbitTemplate` class.

4. **Better Support for Advanced Use Cases**:
   - In advanced scenarios where users need to dynamically add or remove message processors based on business needs, the ability to retrieve the current processors simplifies the code. Without this getter method, developers need to manually track and manage `MessagePostProcessor` lists, which adds unnecessary complexity and can lead to duplication of logic.

5. **Enhanced Developer Experience**:
   - Having the ability to get the current `MessagePostProcessor` list helps developers debug and analyze the message pipeline more easily. This makes it simpler to understand how messages are being transformed before being sent, improving both the developer experience and troubleshooting process.

---

artembilan (approved):
OK. All those arguments still don't convince me in the importance of this API since all the objects you are going to inject/retrieve from there are already somewhere in your configuration.
Apparently what you are going to be is to get them, modify and set back. So, why just not have a set of those post-processors somewhere outside of `RabbitTemplate` and set them (or override) respectively whenever you need?

Nevertheless since this new API is totally not harmful, I'll accept the change and merge it shortly.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -632,6 +632,17 @@ public void setBeanFactory(BeanFactory beanFactory) throws BeansException {
 		this.evaluationContext.setBeanResolver(new BeanFactoryResolver(beanFactory));
 		this.evaluationContext.addPropertyAccessor(new MapAccessor());
 	}
+			
+	/**
+	 * Return configured before post {@link MessagePostProcessor}s or {@code null}.
+	 * @return configured before post {@link MessagePostProcessor}s or {@code null}.
```
artembilan, line 638: Please, add `@since 3.2` and your name to the `@author` list.

artembilan (changes requested):

artembilan (changes requested):
Also see Checkstyle violations:
```
Error: eckstyle] [ERROR] /home/runner/work/spring-amqp/spring-amqp/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java:635: Line matches the illegal pattern 'Trailing whitespace'. [Regexp]
```
Run `./gradlew clean check` locally.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
artembilan, line 638:
```suggestion
	 * @return configured before post {@link MessagePostProcessor}s or {@code null}.
	 * @since 3.2
```

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -632,6 +632,18 @@ public void setBeanFactory(BeanFactory beanFactory) throws BeansException {
 		this.evaluationContext.setBeanResolver(new BeanFactoryResolver(beanFactory));
 		this.evaluationContext.addPropertyAccessor(new MapAccessor());
 	}
+
```
artembilan, line 635:
```suggestion

```

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitTemplate.java
```diff
@@ -632,6 +632,17 @@ public void setBeanFactory(BeanFactory beanFactory) throws BeansException {
 		this.evaluationContext.setBeanResolver(new BeanFactoryResolver(beanFactory));
 		this.evaluationContext.addPropertyAccessor(new MapAccessor());
 	}
```
artembilan, line 634:
```suggestion
	}

```

PR #2859, opened 2024-10-15
Improve condition for more readability

ngocnhan-tran1996:
I read code in package `org.springframework.amqp.core` and I think some codebase conditions are too complex. I think we can make it simple for readability purpose.

That's why I create draft PR and I wonder whether I can continue to do this, at least in package `org.springframework.amqp.core`. Please let me know your thoughts

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/core/AbstractDeclarable.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/core/AbstractDeclarable.java
@@ -25,9 +25,9 @@ import java.util.Map;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
-
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
+import org.springframework.util.ObjectUtils;
 
 /**
  * Base class for {@link Declarable} classes.
@@ -40,7 +40,7 @@ import org.springframework.util.Assert;
  */
 public abstract class AbstractDeclarable implements Declarable {
 
-	 private final Lock lock = new ReentrantLock();
+	private final Lock lock = new ReentrantLock();
 
 	private boolean shouldDeclare = true;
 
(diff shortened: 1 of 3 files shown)


spring-amqp/src/main/java/org/springframework/amqp/core/AbstractDeclarable.java
```diff
@@ -103,16 +103,18 @@ public void setIgnoreDeclarationExceptions(boolean ignoreDeclarationExceptions)
...
 	public void setAdminsThatShouldDeclare(Object... adminArgs) {
-		Collection<Object> admins = new ArrayList<>();
-		if (adminArgs != null) {
-			if (adminArgs.length > 1) {
-				Assert.noNullElements(adminArgs, "'admins' cannot contain null elements");
-			}
-			if (adminArgs.length > 0 && !(adminArgs.length == 1 && adminArgs[0] == null)) {
-				admins.addAll(Arrays.asList(adminArgs));
-			}
+		this.declaringAdmins = new ArrayList<>();
```
artembilan, line 106: This is not OK. You are breaking an atomicity on the  property and we may and up with a `ConcurrentModificationException` from somewhere.

Another my concern that I don't see a reason in so complex logic here.
The `declaringAdmins` property must be `final` for better performance.

The code of this method should be as simple as:
```
public void setAdminsThatShouldDeclare(Object... adminArgs) {
		Assert.notNull(adminArgs, "'admins' must be provided");
		Assert.noNullElements(adminArgs, "'admins' cannot contain null elements");
		this.declaringAdmins.addAll(Arrays.asList(adminArgs));
}
```

artembilan (changes requested):

spring-amqp/src/main/java/org/springframework/amqp/core/AbstractDeclarable.java
ngocnhan-tran1996: Yes, it's more concise but it will break testcase. Will we have any plan to improve codebase in next version? If not, I am not a big fan of `else/else-if condition` and I will try to contribute that reduce `else/else-if condition` for more readability 

https://github.com/spring-projects/spring-amqp/blob/ef23aa9f7f9485ebc50e45a0eb9ed57083c9ca49/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/core/RabbitAdminDeclarationTests.java#L296-L317

ngocnhan-tran1996: I see this changed file takes more your energy reviewing, so I will revert it and create a new PR if we have a plan for breaking change in next version.

In this PR, I will try to improve `if-else` without breaking change and `equals` methods for performance in module `spring-amqp`

artembilan: OK. Sounds like a plan.

artembilan (approved):
This is OK with me.
Do you plan to add more changes?
Or we are good to turn this into a real PR and merge?
Thanks

ngocnhan-tran1996:
I changed more files and I think it's ready to turn into a real PR.

artembilan:
@ngocnhan-tran1996 ,

Issue #2478, opened 2023-07-05
Using AsyncRabbitTemplate. ConvertSendAndReceive return RabbitConverterFuture display Reply timed out, how to solve?


awen112:
Using AsyncRabbitTemplate. ConvertSendAndReceive return RabbitConverterFuture display Reply timed out, how to solve?

future.whenComplete((result, ex) -> {
            if (ex == null) {
            } else {
            }
        });

awen112:
org.springframework.amqp.core.AmqpReplyTimeoutException: Reply timed out
	at org.springframework.amqp.rabbit.TimeoutTask.run(TimeoutTask.java:56)
	at org.springframework.scheduling.support.DelegatingErrorHandlingRunnable.run(DelegatingErrorHandlingRunnable.java:54)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
	at java.base/java.util.concurrent.FutureTask.run$$$capture(FutureTask.java:264)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:833)

garyrussell:
Yes, it simply means that no reply was received within the timeout.

garyrussell:
Closing due to lack of response.

BenEfrati:
If an exception is thrown by `messageConverter.fromMessage(message) `
https://github.com/spring-projects/spring-amqp/blob/1c49e579064152ad9730d24387336ccb32357f89/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/AsyncRabbitTemplate.java#L611, 
it results in an AmqpReplyTimeoutException.
 This happens because the directReplyToContainer in AsyncRabbitTemplate does not provide an option to set a custom ErrorHandler.

artembilan:
@BenEfrati ,
I believe that is a default behavior.
However you an use different constructor with custom error handler:
```
	/**
	 * Construct an instance using the provided arguments. The first queue the container
[...]

PR #2932, opened 2024-12-18
GH-2478 Handle conversion exception in AsyncRabbitTemplate

BenEfrati:
Previously, conversion errors in AsyncRabbitTemplate lead to AmqpReplyTimeoutException

Fixes https://github.com/spring-projects/spring-amqp/issues/2478

https://github.com/spring-projects/spring-amqp/issues/2478#issuecomment-2543745530

### DIFF
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/AsyncRabbitTemplateTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/AsyncRabbitTemplateTests.java
@@ -18,7 +18,6 @@ package org.springframework.amqp.rabbit;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
-import static org.assertj.core.api.Assertions.fail;
 import static org.awaitility.Awaitility.await;
 import static org.mockito.Mockito.mock;
 
@@ -102,7 +101,7 @@ public class AsyncRabbitTemplateTests {
 	}
 
 	@Test
-	public void testConvert1Arg() throws Exception {
+	public void testConvert1Arg()  {
 		final AtomicBoolean mppCalled = new AtomicBoolean();
 		CompletableFuture<String> future = this.asyncTemplate.convertSendAndReceive("foo", m -> {
 			mppCalled.set(true);
@@ -113,7 +112,7 @@ public class AsyncRabbitTemplateTests {
 	}
 
 	@Test
-	public void testConvert1ArgDirect() throws Exception {
+	public void testConvert1ArgDirect() {
 		this.latch.set(new CountDownLatch(1));
 		CompletableFuture<String> future1 = this.asyncDirectTemplate.convertSendAndReceive("foo");
 		CompletableFuture<String> future2 = this.asyncDirectTemplate.convertSendAndReceive("bar");
@@ -141,19 +140,19 @@ public class AsyncRabbitTemplateTests {
 	}
 
 	@Test
-	public void testConvert2Args() throws Exception {
+	public void testConvert2Args() {
 		CompletableFuture<String> future = this.asyncTemplate.convertSendAndReceive(this.requests.getName(), "foo");
 		checkConverterResult(future, "FOO");
 	}
 
 	@Test
-	public void testConvert3Args() throws Exception {
+	public void testConvert3Args() {
 		CompletableFuture<String> future = this.asyncTemplate.convertSendAndReceive("", this.requests.getName(), "foo");
 		checkConverterResult(future, "FOO");
 	}
 
 	@Test
-	public void testConvert4Args() throws Exception {
+	public void testConvert4Args() {
 		CompletableFuture<String> future = this.asyncTemplate.convertSendAndReceive("", this.requests.getName(), "foo",
 				message -> {
 					String body = new String(message.getBody());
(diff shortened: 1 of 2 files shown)


artembilan (changes requested):
Please, run `./gradlew check` locally before pushing changes to PR.
Right now it even does not compile:
```
 /home/runner/work/spring-amqp/spring-amqp/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/AsyncRabbitTemplate.java:614: error: cannot find symbol
						} catch (MessageConversionException e) {
						         ^
  symbol:   class MessageConversionException
  location: class AsyncRabbitTemplate
```

BenEfrati:
Hi,
Sorry, at first I fork and created branch from old code, missed the import after sync with upstream.
https://github.com/BenEfrati/spring-amqp/commit/50452f89c3c306cc6dcad5b7dee27c036ffa81ac
Pushed the fixed code

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/AsyncRabbitTemplateTests.java
```diff
@@ -394,6 +396,29 @@ public void testStopCancelled() throws Exception {
 		assertThat(callback.result).isNull();
 	}
 
+	@Test
+	@DirtiesContext
```
artembilan, line 400: I wonder if we really need this `@DirtiesContext`.
I see that many other tests in this class have it, but that doesn't mean that it is right.
Would you mind to double check if it works without?
Just possible slight optimization in execution performance!

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/AsyncRabbitTemplateTests.java
```diff
@@ -394,6 +396,29 @@ public void testStopCancelled() throws Exception {
...
+		RabbitConverterFuture<String> replyFuture = this.asyncTemplate.convertSendAndReceive("conversionException");
+
+		final CountDownLatch cdl = new CountDownLatch(1);
+		final AtomicReference<Object> resultRef = new AtomicReference<>();
+        replyFuture.whenComplete((result, ex) -> {
+			resultRef.set(result);
+			cdl.countDown();
+		});
+        assertThat(cdl.await(10, TimeUnit.SECONDS)).isTrue();
+        assertThat(replyFuture).isCompletedExceptionally();
```
artembilan, line 418: May we also verify what exactly exception we got here?

artembilan (changes requested):

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/AsyncRabbitTemplateTests.java
BenEfrati: Every test changes the `AsyncRabbitTemplate` state, so we must reset the context before each test. 
In my test, for example, I'm changing the messageConverter on the `RabbitTemplate`, so if my test will run first, all other tests will fail.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/AsyncRabbitTemplateTests.java
BenEfrati, line 418:
```suggestion
		assertThat(replyFuture).failsWithin(Duration.ofSeconds(10))
				.withThrowableOfType(MessageConversionException.class)
				.withMessageContaining("Failed to convert message");
```
Something like this?

artembilan: Or never mind. I like it.
You have changed the logic, but at the same it is covered with that `failsWithin()` 😄 
Committing and merging.
Thank you!

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/AsyncRabbitTemplateTests.java
```diff
@@ -394,6 +396,29 @@ public void testStopCancelled() throws Exception {
...
+        assertThat(resultRef.get()).isNull();
```
artembilan, line 419: But this line has to be removed then 😄

BenEfrati: sure :) 

Do you want me to align other tests to use this `failsWithin`

```
void testReturn() {
...
		assertThat(future)
				.as("Expected exception")
				.failsWithin(Duration.ofSeconds(10))
				.withThrowableThat()
				.withCauseInstanceOf(AmqpMessageReturnedException.class)
				.extracting("routingKey")
				.isEqualTo(this.requests.getName() + "x");
```

instead 
```
try {
			future.get(10, TimeUnit.SECONDS);
			fail("Expected exception");
		}
		catch (ExecutionException e) {
			assertThat(e.getCause()).isInstanceOf(AmqpMessageReturnedException.class);
			assertThat(((AmqpMessageReturnedException) e.getCause()).getRoutingKey()).isEqualTo(this.requests.getName() + "x");
		}
```

artembilan: I would appreciate.
Thanks.
Those `try..catch` in tests is on my TODO list in favor of `assertThat()` when I touch these classes.
Although still would be with some code.
But now I have learned about `failsWithin()` which looks much cleaner. 😄

artembilan:
@BenEfrati ,

Issue #2949, opened 2025-02-04
Using retryCount results in UnsupportedOperationException


jkaving:
**In what version(s) of Spring AMQP are you seeing this issue?**

From 3.2.0

**Describe the bug**

Following [the retry algorithm example here](https://docs.spring.io/spring-amqp/reference/amqp/resilience-recovering-from-errors-and-broker-failures.html#retry-over-broker) (which uses the new `retryCount` message property) can result in an `UnsupportedOperationException` being thrown from `DefaultMessagePropertiesConverter.fromMessageProperties()`.
This happens if the message doesn't have any other headers.

`DefaultMessagePropertiesConverter.fromMessageProperties()` currently does [this](https://github.com/spring-projects/spring-amqp/blob/2aaa8db7d0b6a16a9c01f3c08e652f6bef2cb6ee/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/DefaultMessagePropertiesConverter.java#L158-L162):
```java
...
Map<String, Object> headers = convertHeadersIfNecessary(source.getHeaders());
long retryCount = source.getRetryCount();
if (retryCount > 0) {
	headers.put(MessageProperties.RETRY_COUNT, retryCount);
}
...
```

The problem is that `convertHeadersIfNecessary()` does [this](https://github.com/spring-projects/spring-amqp/blob/2aaa8db7d0b6a16a9c01f3c08e652f6bef2cb6ee/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/DefaultMessagePropertiesConverter.java#L190-L192) if the source message has no headers:
```java
...
if (CollectionUtils.isEmpty(headers)) {
	return Collections.emptyMap();
}
...
```

`Collections.emptyMap()` returns an unmodifiable `Map` (more specifically an instance of `Collections.EmptyMap`) so `headers.put()` will then result in an `UnsupportedOperationException`.

**To Reproduce**

Create a `MessageProperties` instance, increment the retry count and convert the message using `DefaultMessagePropertiesConverter`:
```java
MessageProperties props = new MessageProperties();
props.incrementRetryCount();
new DefaultMessagePropertiesConverter().fromMessageProperties(props, "UTF8");
```

**Expected behavior**

It should be possible to use `retryCount` even if the message doesn't have other headers.

PR #2950, opened 2025-02-04
GH-2949: Fix retryCount handling

jkaving:
Fixes https://github.com/spring-projects/spring-amqp/issues/2949

The handling of `retryCount` is moved into `convertHeadersIfNecessary()`, so that we can still return a `Collections.emptyMap()` if there are no headers in the source `MessageProperties` and no `retryCount` to add.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/DefaultMessagePropertiesConverter.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/DefaultMessagePropertiesConverter.java
@@ -43,6 +43,7 @@ import org.springframework.util.StringUtils;
  * @author Raylax Grey
  * @author Artem Bilan
  * @author Ngoc Nhan
+ * @author Johan Kaving
  *
  * @since 1.0
  */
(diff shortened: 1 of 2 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/DefaultMessagePropertiesConverter.java
```diff
@@ -155,11 +156,7 @@ else if (MessageProperties.RETRY_COUNT.equals(key)) {
 	@Override
 	public BasicProperties fromMessageProperties(final MessageProperties source, final String charset) {
 		BasicProperties.Builder target = new BasicProperties.Builder();
-		Map<String, Object> headers = convertHeadersIfNecessary(source.getHeaders());
-		long retryCount = source.getRetryCount();
-		if (retryCount > 0) {
-			headers.put(MessageProperties.RETRY_COUNT, retryCount);
-		}
```
jkaving, line 162: An alternative solution could be to keep the `retryCount` code here, but add some kind of check inside the `if (retryCount > 0)` block.

We could e.g., check if `headers` is empty:
```java
...
if (retryCount > 0) {
	if (headers.isEmpty()) {
		headers = new HashMap<>();
	}
	headers.put(MessageProperties.RETRY_COUNT, retryCount);
}
...
```
or catch the exception:
```java
...
if (retryCount > 0) {
	try {
		headers.put(MessageProperties.RETRY_COUNT, retryCount);
	} catch (UnsupportedOperationException e) {
		headers = Map.of(MessageProperties.RETRY_COUNT, retryCount);
	}
}
...
```

artembilan (approved):
I think I'm fine with what you did so far: less objects is better.

Sorry about missing the case when map of headers could be empty.

Will merge when the build is green.

Thank you!

PR #2977, opened 2025-02-16
Improve `build.gradle` for building antora

ngocnhan-tran1996:
When we build document, example with new version (4.0.0-SNAPSHOT) and it will work with Spring Framework 7.x version, but the current document is pointed with Spring `current` (6.2.x) version

>You can read more about that [here](https://docs.spring.io/spring-framework/reference/core/beans/java.html).

Reference
https://docs.spring.io/spring-amqp/reference/4.0-SNAPSHOT/sample-apps.html#hello-world-sync

That's why I think we can use gradle for improving the build document process.

The main points:
- The new Spring AMQP version document will be compatitable with Spring Framework version or Micrometer version document
- Reference javadoc will be easier, use `javadoc:org.springframework.amqp...` instead `{spring-amqp-java-docs}...`


build.gradle
```diff
@@ -1,3 +1,5 @@
+import java.util.regex.Pattern
```
artembilan, line 1: I prefer to not have imports in Gradle configs.
More over this is Groovy, so there is a way to deal with regular expressions without Java API: https://www.baeldung.com/groovy-pattern-matching

build.gradle
```diff
@@ -724,3 +724,32 @@ publishing {
...
 	}
 }
+
+def generateAttributes() {
+	def springDocs = "https://docs.spring.io"
+	def micrometerDocsPrefix = "https://docs.micrometer.io"
+
+	return [
+		'project-version': project.version,
+		'spring-integration-docs': "$springDocs/spring-integration/reference".toString(),
```
artembilan, line 734: Why do we need to call `toString()` on GString, when it is really String in the end after Groovy evaluation.

build.gradle
```diff
@@ -724,3 +724,32 @@ publishing {
...
+		'spring-framework-docs': "$springDocs/spring-framework/reference/%s".formatted(generateVersionWithoutPatch(springVersion)),
```
artembilan, line 735: I believe we don't need call `formatted()` since it is Groovy when its `${}` operator would do that trick for us.

artembilan (changes requested):

build.gradle
ngocnhan-tran1996: Without `toString()`, the antora file will be generated like

```yaml
    spring-integration-docs: !!org.codehaus.groovy.runtime.GStringImpl {
      }
```

This will cause exception when generating document

artembilan:
Thank you for your effort!
Unfortunately, I don't have too much time today to look into this for today's release, so this will be revised for the next milestone.

src/reference/antora/antora-playbook.yml
```diff
@@ -23,6 +23,7 @@ asciidoc:
   extensions:
     - '@asciidoctor/tabs'
     - '@springio/asciidoctor-extensions'
+    - '@springio/asciidoctor-extensions/javadoc-extension'
```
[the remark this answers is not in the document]
ngocnhan-tran1996: This line use for prefix `javadoc:` when generating document

The reference:
https://github.com/spring-io/asciidoctor-extensions?tab=readme-ov-file#javadoc

I think it will be easier when we use javadoc links in document instead prefix `{spring-amqp-java-docs}`

artembilan (approved):
I think this is good so far.
Merging it to give it a spin.

Thank you!

artembilan:
So, that didn't go well: https://docs.spring.io/spring-amqp/reference/4.0-SNAPSHOT/amqp/broker-configuration.html#builder-api

```
See the Javadoc for javadoc:org.springframework.amqp.core.QueueBuilder[org.springframework.amqp.core.QueueBuilder] and javadoc:org.springframework.amqp.core.ExchangeBuilder[org.springframework.amqp.core.ExchangeBuilder] for more information.
```

JavaDocs are not resolved.
Perhaps I was right that `docs-build` branch has to be fixed as well.

ngocnhan-tran1996:
I will checkout from docs-build and create PR for fixing this

PR #2984, opened 2025-02-21
Update antora version

ngocnhan-tran1996:


artembilan:
Now it is better: https://docs.spring.io/spring-amqp/reference/4.0-SNAPSHOT/amqp/broker-configuration.html#builder-api
Thank you!

I wonder if you are OK to do something similar for Spring Kafka project: https://github.com/spring-projects/spring-kafka 😄

PR #3006, opened 2025-03-10
Read x-death count in a safer way

raul-avila-ph:
We had an issue in our system trying to consume a message from RabbitMQ that contained an `x-death` header, with `count` value typed as Integer. This caused a ClassCastException in https://github.com/spring-projects/spring-amqp/blob/355be1ee4c20dd1179c6eed59bb43fac5d18db6b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/DefaultMessagePropertiesConverter.java#L150

The reason the `count` value was an int and not a long is that we were storing headers in an internal database as part of a recovery process, and the typing was slightly changed during serialisation / deserialisation.

We propose this pull request to make the reading process of the `count` field in `x-death` header safer.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/DefaultMessagePropertiesConverter.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/DefaultMessagePropertiesConverter.java
@@ -44,6 +44,7 @@ import org.springframework.util.StringUtils;
  * @author Artem Bilan
  * @author Ngoc Nhan
  * @author Johan Kaving
+ * @author Raul Avila
  *
  * @since 1.0
  */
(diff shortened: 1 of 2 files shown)


artembilan (changes requested):
The build has failed due to Checkstyle violations:
```
 Error: eckstyle] [ERROR] /home/runner/work/spring-amqp/spring-amqp/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/DefaultMessagePropertiesConverter.java:36:1: Wrong order for 'org.springframework.util.NumberUtils' import. [ImportOrder]

> Task :spring-rabbit:checkstyleMain
Error: eckstyle] [ERROR] /home/runner/work/spring-amqp/spring-amqp/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/DefaultMessagePropertiesConverter.java:153:33: 'if' is not followed by whitespace. [WhitespaceAfter]
Error: eckstyle] [ERROR] /home/runner/work/spring-amqp/spring-amqp/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/DefaultMessagePropertiesConverter.java:153:33: 'if' is not followed by whitespace. [WhitespaceAround]
```
Also, please, add your name to the `@author` list of the affected classes.

In other place of the same class I just do:
```
					if (value instanceof Number numberValue) {
						target.setRetryCount(numberValue.longValue());
					}
```
Instead of `NumberUtils`.
I mean you still have problem with Checkstyle, so might be better to not use it at all 😄 

Please, run `gradlew check` command locally before pushing.

raul-avila-ph:
Apologies for the slightly messy commit, I had some restrictions in my company laptop and couldn't do a proper build.

Everything sorted now, any comments let me know.

artembilan:

PR #3015, opened 2025-03-13
Use backtick formatted property

ngocnhan-tran1996:


artembilan:
@ngocnhan-tran1996 ,

Issue #3079, opened 2025-05-29
Further Investigation into Shutdown Anomalies


marcogramy:
Hi,
I am the author of issue #3039.

I am writing again to report that we are still observing anomalies during the shutdown sequence. Specifically, we have noted that the `Recover` command for in-flight messages is being dispatched immediately, without respecting the configured shutdown timeout allocated for the process to finalize its operations.

Upon enabling debug logging, we observe that the "Closing Rabbit Channel" log statement within the `BlockingQueueConsumer.stop()` method occurs directly after the "Waiting for workers to finish" log:
```
2025-05-27T16:23:17.329+02:00  INFO 112517 --- [ionShutdownHook] o.s.a.r.l.SimpleMessageListenerContainer : Waiting for workers to finish.
2025-05-27T16:23:17.330+02:00 DEBUG 112517 --- [ntContainer#0-1] o.s.a.r.listener.BlockingQueueConsumer   : Closing Rabbit Channel: Cached Rabbit Channel: AMQChannel(amqp://javamt@10.122.122.10:5673/,1), conn: Proxy@5e7687f6 Dedicated Rabbit Connection: SimpleConnection@39985b7d [delegate=amqp://javamt@10.122.122.10:5673/, localPort=44504]
2025-05-27T16:23:17.330+02:00 DEBUG 112517 --- [ntContainer#0-2] o.s.a.r.listener.BlockingQueueConsumer   : Closing Rabbit Channel: Cached Rabbit Channel: AMQChannel(amqp://javamt@10.122.122.10:5673/,1), conn: Proxy@4b4be8c7 Dedicated Rabbit Connection: SimpleConnection@3da25e84 [delegate=amqp://javamt@10.122.122.10:5673/, localPort=44494]
```

As a side effect of this behavior, we are seeing messages immediately being returned to the queues following the shutdown event. We have encountered a scenario where a message was returned to the same process that was shutting down.

TCP traffic logs reveal a rapid succession of events (the second column represents the timestamp):
![Image](https://github.com/user-attachments/assets/69627990-6bbb-4e06-ab72-845a9ca3f3a2)

While the root cause of this issue is unclear to us, we have attempted to analyze all points within the libraries where the `stop()` method is invoked. One potential additional point of invocation we identified is the `SimpleMessageListenerContainer.killOrRestart()` function, which generates the "Cancelling Consumer" log entry that we also observe in our logs:
```
[...]

PR #3089, opened 2025-06-12
GH-3079: allow sending ACKs within the ShutdownTimeout period

marcogramy:
Refer to https://github.com/spring-projects/spring-amqp/issues/3079

When the process shutdown, within the `shutdownTimeout` period, we want to allow sending `Basic.Ack` replies to processed in-flight message.


artembilan:
Good., As long as the build is green, I'll merge this.
Thanks

artembilan:
Merge and initiated back-port process.

Thank you for the great contribution!

Issue #3115, opened 2025-07-07
Setter for RabbitStreamTemplateObservationConvention in RabbitStreamTemplate


mangila:
Missing a setter method for RabbitStreamTemplateObservationConvention  in RabbitStreamTemplate

No setter is available

I can't add custom tags for granular observation

Workaround -- create a custom scope around send()

```java
 Span customSpan = tracer.nextSpan().name("custom.name").start();
        try (Tracer.SpanInScope scope = tracer.withSpan(customSpan)) {
            customSpan.tag("custom.tag", "value");
            return rabbitStreamTemplate.send(message);
        } finally {
            customSpan.end();
        }
```

mangila:
Seems all set! Thanks to @ngocnhan-tran1996 
Thanks!

PR #3116, opened 2025-07-07
Add observationConvention Setter for RabbitStreamTemplate

ngocnhan-tran1996:
Closes gh-3115

### DIFF
--- a/spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/producer/RabbitStreamTemplate.java
+++ b/spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/producer/RabbitStreamTemplate.java
@@ -222,9 +222,9 @@ public class RabbitStreamTemplate implements RabbitStreamOperations, Application
 	}
 
 	/**
-	 * Set an observation convention; used to add additional key/values to observations.
+	 * Set a custom {@link RabbitStreamTemplateObservationConvention}.
 	 * @param observationConvention the convention.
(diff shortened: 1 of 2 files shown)


spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/producer/RabbitStreamTemplate.java
```diff
@@ -221,6 +221,14 @@ public void setObservationEnabled(boolean observationEnabled) {
 		this.observationEnabled = observationEnabled;
 	}
 
+	/**
+	 * Set an observation convention; used to add additional key/values to observations.
```
artembilan: I don't think that custom convention is only for key/value.
Let's call it like:
```
* Set a custom {@link RabbitStreamTemplateObservationConvention}.
```
No need to specific and confusion.

spring-rabbit-stream/src/main/java/org/springframework/rabbit/stream/producer/RabbitStreamTemplate.java
```diff
@@ -221,6 +221,14 @@ public void setObservationEnabled(boolean observationEnabled) {
...
+	 * @param observationConvention the convention.
+	 * @since 4.0
```
artembilan: `3.2.6`, please.

We are going to back-port this as a bug.

artembilan (changes requested):

PR #3123, opened 2025-07-16
Use `assertThatException` for exception test

ngocnhan-tran1996:

### DIFF
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java
@@ -143,7 +143,7 @@ import org.springframework.validation.Validator;
 import org.springframework.validation.annotation.Validated;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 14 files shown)


spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/EnableRabbitIntegrationTests.java
```diff
@@ -879,16 +877,15 @@ public void listenerErrorHandler() {
...
-			assertThat(config.errorHandlerChannel).isNotNull();
-		}
+		assertThatException()
+				.isThrownBy(() -> this.rabbitTemplate.convertSendAndReceive("test.pojo.errors2", "foo"))
+				.satisfies(ex -> {
+
+					assertThat(ex.getCause().getMessage()).isEqualTo("from error handler");
+					assertThat(ex.getCause().getCause().getMessage()).isEqualTo("return this");
+					EnableRabbitConfig config = this.context.getBean(EnableRabbitConfig.class);
+					assertThat(config.errorHandlerChannel).isNotNull();
```
artembilan: I think these two last lines could be outside of the exception assertion scope.

PR #3147, opened 2025-08-06
GH-3031: Defer SMLC shutdown for pending replies

### LINKED ISSUE #3031: Request to delay consumer and listener container shutdown to wait for pending reply messages when using Request/Reply Messaging with a fixed name reply queue.
**Expected Behavior**

When using the Request/Reply Messaging feature, a request message is sent, but before the reply message is received, the application starts a graceful shutdown (think of the case when running in a container orchestration platform and it decides that it wants to stop your container).  Would like the consumer and listener container to wait for a reply message -- up to some configurable amount of time (something short, like 30 seconds). We always set an expiration TTL on the request message while calling the template method convertSendAndReceive, so a timeout on that synchronous call is the other possible outcome.

**Current Behavior**

Right now when the graceful shutdown commences, the consumer and listener container shut down quickly, presumably because there are no messages on the reply queue.  The reply message does not yet exist.

**Context**

Our request messages are the result of customer interaction.  They cause processing to happen in a remote location. The reply message is used to indicate successful processing back to a customer.  Our orchestration platform periodically decides to stop a container (or one stops when we deploy a new version of software) and those currently executing instances sometimes have in-flight request messages that are awaiting reply messages.  When the container stops without receiving the reply message, that message is lost, and we have to report a system error to the customer.

Artem noted on stackoverflow.com a possible workaround using a temporary queue name for each request/reply, but I'd really, really like to stick with fixed named queue to solve this issue.

m3k0813:
Closes #3031

### Summary

This pull request introduces a mechanism to defer the shutdown of a `SimpleMessageListenerContainer` to allow in-flight request/reply messages to be received before the consumer is cancelled.

### Problem

Currently, when using request/reply messaging with a fixed reply queue, a graceful shutdown can cause the listener container to stop immediately. If a reply message arrives shortly after, it is lost because the consumer is already gone. This forces the application to treat the operation as a timeout or error, as described in the issue.

### Solution

The solution introduces new configuration on `SimpleMessageListenerContainer` to make it aware of pending replies from a provider, such as a `RabbitTemplate`.

* A new functional interface, `PendingReplyProvider`, is introduced to decouple the container from the template.
* `RabbitTemplate` now exposes a `getPendingReplyCount()` method, which can be used as a `PendingReplyProvider`.
* `SimpleMessageListenerContainer` has two new properties:
    * `setPendingReplyProvider(PendingReplyProvider)`: When set, this enables the shutdown delay logic.
    * `setPendingReplyCheckInterval(long)`: Configures the polling interval while waiting for replies during shutdown.

With this configuration, the container's shutdown logic (`shutdownAndWaitOrCallback`) now checks for pending replies and waits for up to the configured `shutdownTimeout` before proceeding with the normal shutdown process.

### Testing

* A new test case, `testShutdownWithPendingReplies()`, has been added to `SimpleMessageListenerContainerTests` to verify this new behavior.


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -130,6 +131,11 @@ public class SimpleMessageListenerContainer extends AbstractMessageListenerConta
 
 	private long batchReceiveTimeout;
 
+	private long pendingReplyCheckInterval = 200L;
+
+	private @Nullable PendingReplyProvider pendingReplyProvider;
+
+
```
injae-kim: double empty line, remove L138

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/api/PendingReplyProvider.java
```diff
@@ -0,0 +1,20 @@
...
+@FunctionalInterface
+public interface PendingReplyProvider {
+
+	/**
+	 * Return the number of pending replies.
+	 * @return the number of pending replies.
+	 */
+	int getPendingReplyCount();
+
+}
```
injae-kim: image

can you add 1 more empty line at the end of this file?

m3k0813: Thanks for the review, @injae-kim! I'll apply these changes right away.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -357,6 +362,28 @@ public void setBatchReceiveTimeout(long batchReceiveTimeout) {
...
+	/**
+	 * Set a provider for the number of pending replies.
+	 * When set, the container will wait for pending replies during shutdown,
+	 * up to the {@link #setShutdownTimeout(long) shutdownTimeout}.
+	 * @param pendingReplyProvider the pending reply provider.
+	 * @since 4.0
+	 * @see org.springframework.amqp.rabbit.core.RabbitTemplate#getPendingReplyCount()
+	 */
+	public void setPendingReplyProvider(PendingReplyProvider pendingReplyProvider) {
```
artembilan: I don't think we need this extra abstraction.
We already have a `ListenerContainerAware` implemented on the `RabbitTemplate`.
So, that `getPendingReplyCount()` could be moved into that contract.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -652,6 +679,8 @@ private void waitForConsumersToStart(Set<AsyncMessageProcessingConsumer> process
 
 	@Override
 	protected void shutdownAndWaitOrCallback(@Nullable Runnable callback) {
+		waitForPendingReplies();
```
artembilan: I don't think this is a right place for such a logic.
I believe it has to be a part of that `awaitShutdown` in the end of this `shutdownAndWaitOrCallback()` method.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -732,6 +761,39 @@ protected void shutdownAndWaitOrCallback(@Nullable Runnable callback) {
...
+					Thread.sleep(this.pendingReplyCheckInterval);
+				}
+				int remaining = provider.getPendingReplyCount();
+				if (remaining > 0) {
+					logger.warn("Shutdown timeout expired, but " + remaining + " pending replies still remain.");
+				}
+			}
+			catch (InterruptedException e) {
+				Thread.currentThread().interrupt();
+				logger.warn("Interrupted while waiting for pending replies.");
```
artembilan: I don't think we need so much logic here.
Currently, we utilize an `ActiveObjectCounter` for the consumers to be cancelled in this listener container.
I believe we can use it in the `RabbitTemplate` as well for those pending replies to be fulfilled.
Then in that `awaitShutdown`  hook we could take this `ActiveObjectCounter` from the `ListenerContainerAware` and wait on it as we do for consumers.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerTests.java
```diff
@@ -716,6 +718,43 @@ void testWithConsumerStartWhenNotActive() {
...
+		ConnectionFactory connectionFactory = mock(ConnectionFactory.class);
+		Connection connection = mock(Connection.class);
+		Channel channel = mock(Channel.class);
+		given(connectionFactory.createConnection()).willReturn(connection);
+		given(connection.createChannel(false)).willReturn(channel);
+		given(channel.isOpen()).willReturn(true);
+
+		RabbitTemplate template = new RabbitTemplate(connectionFactory);
+		SimpleMessageListenerContainer container = new SimpleMessageListenerContainer(connectionFactory);
+		container.setQueueNames("foo");
```
artembilan: No `foo/bar` language, please, in the project.

artembilan (changes requested):
Let me know if my review makes sense!
Thank you!

m3k0813:
Hi @artembilan,

Thanks for the feedback! I've updated the PR with all the changes you suggested. 
It should be ready for another look now. PTAL.

Thank you!

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/ListenerContainerAware.java
```diff
@@ -39,4 +39,13 @@ public interface ListenerContainerAware {
 	@Nullable
 	Collection<String> expectedQueueNames();
 
+	/**
+	 * Return a counter for pending replies, if any.
+	 * @return the counter, or null.
+	 * @since 4.0
+	 */
+	@Nullable
+	default ActiveObjectCounter<Object> getPendingReplyCounter() {
```
artembilan: This annotation has to be next to the type, not method itself.
Therefore this style is preferred: 
```
default @Nullable ActiveObjectCounter<Object>
```

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -692,6 +692,24 @@ protected void shutdownAndWaitOrCallback(@Nullable Runnable callback) {
 		Runnable awaitShutdown = () -> {
 			logger.info("Waiting for workers to finish.");
 			try {
+				ActiveObjectCounter<Object> replyCounter = null;
+				Object listener = getMessageListener();
+				if (listener instanceof ListenerContainerAware) {
```
artembilan: We can just do this `if (getMessageListener() instanceof ListenerContainerAware listenerContainerAware) {`
Then use that pattern matching variable in the block.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -692,6 +692,24 @@ protected void shutdownAndWaitOrCallback(@Nullable Runnable callback) {
...
+					replyCounter = ((ListenerContainerAware) listener).getPendingReplyCounter();
+				}
+
+				if (replyCounter != null) {
```
artembilan: This should be inside of the previous `if`.
Plus I believe this still has to be conditional on the `replyCounter.getCount() > 0`.
No reason to log anything and wait if there is really nothing to wait for.

And we probably can just extract this into a separate method to this code easier to read.

artembilan (changes requested):
Please, add your name to all the affected classes.
Also, consider to add an entry into the `whats-new.adoc`.

Thanks

m3k0813:
Thanks for the thorough review!
I’ve applied the suggestions and pushed the changes.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerTests.java
```diff
@@ -716,6 +719,45 @@ void testWithConsumerStartWhenNotActive() {
...
+		container.setQueueNames("shutdown.test.queue");
+		container.setMessageListener(template);
+
+		template.setReplyAddress(container.getQueueNames()[0]);
+
+		long shutdownTimeout = 2000L;
+		container.setShutdownTimeout(shutdownTimeout);
+
+		ActiveObjectCounter<Object> replyCounter =
+				(ActiveObjectCounter<Object>) ReflectionTestUtils.getField(template, "pendingRepliesCounter");
```
artembilan: I don't think we need to use a reflection here since `RabbitTemplate` provides for us already `getPendingReplyCounter()`.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerTests.java
```diff
@@ -716,6 +719,45 @@ void testWithConsumerStartWhenNotActive() {
...
+		given(channel.isOpen()).willReturn(true);
+
+		RabbitTemplate template = new RabbitTemplate(connectionFactory);
+		SimpleMessageListenerContainer container = new SimpleMessageListenerContainer(connectionFactory);
+		container.setQueueNames("shutdown.test.queue");
+		container.setMessageListener(template);
+
+		template.setReplyAddress(container.getQueueNames()[0]);
+
+		long shutdownTimeout = 2000L;
```
artembilan: I don't think we need to wait for "never reply" so long.
We deliberately aware that reply never comes back to us in this testing scenario.
So, what is the point to block this test for those 2 seconds?
Now imaging that every single 3000 tests in the project is going to block for similar time.
That would be a disaster for development feedback loop.

I think we should make like `500` millis at most.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerTests.java
```diff
@@ -716,6 +719,45 @@ void testWithConsumerStartWhenNotActive() {
...
+
+		Object pending = new Object();
+		replyCounter.add(pending);
+		assertThat(replyCounter.getCount()).isEqualTo(1);
+
+		container.start();
+
+		long startTime = System.currentTimeMillis();
+		container.stop();
+		long stopDuration = System.currentTimeMillis() - startTime;
```
artembilan: The timing is always problematic for testing.

Let's see if we can `spy` on the `logger` of the `SimpleMessageListenerContainer` and `await().untilAsserted()` for the new `warn` log message you have introduced in this change!

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -36,4 +36,9 @@ There is no need to keep out-dated utilities and recommendation is to migrate to
 The Jackson 2 has been deprecated for removal in whole Spring portfolio.
 Respective new classes have been introduced to support Jackson 3.
 
-See xref:amqp/message-converters.adoc[] for more information.
\ No newline at end of file
+See xref:amqp/message-converters.adoc[] for more information.
+
+[[x40-enhancements]]
+=== Enhancements
```
artembilan: Let's name it as "MessageListenerContainer Changes"!

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -36,4 +36,9 @@ There is no need to keep out-dated utilities and recommendation is to migrate to
...
+
+Defer `SimpleMessageListenerContainer` shutdown for pending `RabbitTemplate` replies.
```
artembilan: This info is not enough.
Let's see if my sentence would be better!

```
The  `SimpleMessageListenerContainer` now awaits at most `shutdownTimeout` for pending replies from the provided `RabbitTemplate` listener on its shutdown.
```

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerTests.java
```diff
@@ -736,26 +736,28 @@ void testShutdownWithPendingReplies() {
...
+		new DirectFieldAccessor(container).setPropertyValue("logger", logger);
+
 		container.start();
 
-		long startTime = System.currentTimeMillis();
 		container.stop();
-		long stopDuration = System.currentTimeMillis() - startTime;
 
-		replyCounter.release(pending);
+		await().atMost(Duration.ofMillis(500)).untilAsserted(() ->
```
artembilan: I think those max 30 seconds by default is fine with us in our tests.
It is really not going to be blocked that long.
Might really maximum those 500 millis.
However, if we have it as exact our given time, there is no guarantee that CPU resources will be available for us in time to satisfy this expectation.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerTests.java
```diff
@@ -736,26 +736,28 @@ void testShutdownWithPendingReplies() {
...
+				verify(logger).warn("Shutdown timeout expired, but 1 pending replies still remain.")
+		);
 
-		assertThat(stopDuration).isGreaterThanOrEqualTo(shutdownTimeout - 500);
+		replyCounter.release(pending);
```
artembilan: I don't think this call makes any difference or sense for the test logic.
Therefore let's remove altogether to avoid extra reading noise!

m3k0813:
Thank you for your detailed and helpful feedback. 
I've pushed the final commit reflecting all of your suggestions, PTAL when you get a chance.

m3k0813:
Thank you for your detailed and helpful feedback!
I've pushed the final commit reflecting all of your suggestions, PTAL when you get a chance.

injae-kim:
Nice work @m3k0813 👍👍👍

artembilan:
Right. So, even that 1 second is not enough.
The test has failed: https://ge.spring.io/s/55ld5jaabojio/tests/task/:spring-rabbit:test/details/org.springframework.amqp.rabbit.listener.SimpleMessageListenerContainerTests/testShutdownWithPendingReplies()?top-execution=1.

I still believe that we have to leave it with default 30 seconds.
Will fix shortly...

m3k0813:
The new test logic makes a lot of sense and looks great.
Thanks for the fix and the explanation!

PR #3167, opened 2025-08-22
Replace Spring Retry usage to core retry

snicoll:
This commit replaces Spring Retry by the core retry support introduced in Spring Framework 7. This is a breaking change mostly in configuration that is detailed below.

Some of the features used from Spring Retry have no equivalent in Spring Framework, and a tailored solution for Spring AMQP has been implemented in this commit. This leads to the change in the following hook points:

* `AbstractAdaptableMessageListener` can be configured by the newly introduced `MessageRecoveryCallback` instead of Spring Retry's `MethodInvocationRecoverer`. The new callback directly provides the Message, replyTo Address, and Exception.
* The `RecoveryCallback` in `RabbitTemplate` is replaced by an interface with the same name.
* `AbstractRetryOperationsInterceptorFactoryBean` should be configured with a `RetryPolicy`, rather than a `RetryTemplate`.

Spring AMQP uses stateless and stateful retry interceptors from Spring Retry. These are replaced by:

* `StatelessRetryOperationsInterceptor` that invokes recovery if provided.
* `StatefulRetryOperationsInterceptor` provides a similar interceptor as the Spring Retry equivalent, except that it is configured with Spring AMQP-specify callbacks.

These interceptors are configured with a `RetryPolicy`, rather than a `RetryTemplate`. The new `RetryTemplate` in Spring Framework is operating on a `RetryPolicy` can can't be further configured. This has an impact on `RetryInterceptorBuilder` where a custom `BackOff` can no longer be specified. Instead, a consumer of the `RetryPolicy.Builder` is provided that offers configuration for it and more.

Using dedicated callbacks also shows inconsistencies with the Null-safety support. The most annoying example is MessageKeyGenerator that should be nullable despite its contract stating that it should not. Spring AMQP has actually a code path to deal with stateful retries when the key is null.

A major difference in default is that Spring Retry has a default policy of 3 attempts: the initial invocation and 2 retries. Spring Framework has a default RetryPolicy of 3 attempts, that does not include the initial invocation. With 4 attempts in total, there is an off-by-one change in `AmqpAppender` and tests to configure a retry policy with 2 attempts.

With Spring Retry being completely removed, this commit also removes the dependency and any further references to it.


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/retry/MessageRecoveryCallback.java
```diff
@@ -0,0 +1,42 @@
...
+
+package org.springframework.amqp.rabbit.retry;
+
+import org.jspecify.annotations.Nullable;
+
+import org.springframework.amqp.core.Address;
+import org.springframework.amqp.core.Message;
+
+/**
+ * Callback for stateful retry after the retry policy has exhausted.
```
artembilan, line 25: Looks like this abstraction is used regardless if retry is stateful or not.
Can we adjust this Javadoc respectively?

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/retry/MissingIdRetryTests.java
```diff
@@ -223,25 +180,19 @@ public void testWithIdAndSuccess() throws Exception {
...
-			ArgumentCaptor getCaptor = ArgumentCaptor.forClass(Object.class);
-			ArgumentCaptor removeCaptor = ArgumentCaptor.forClass(Object.class);
-			verify(cache, times(2)).put(putCaptor.capture(), any(RetryContext.class));
-			verify(cache, times(2)).get(getCaptor.capture());
-			verify(cache, atLeast(2)).remove(removeCaptor.capture());
-			verify(cache, atMost(4)).remove(removeCaptor.capture());
-			logger.debug("puts:" + putCaptor.getAllValues());
-			logger.debug("gets:" + putCaptor.getAllValues());
-			logger.debug("removes:" + removeCaptor.getAllValues());
+			await().atMost(Duration.ofSeconds(3)).until(cache::isEmpty);
```
artembilan, line 183: Might not be enough on CI/CD.
I'm totally OK with those `30 seconds` by default.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/retry/MissingIdRetryTests.java
```diff
@@ -223,25 +180,19 @@ public void testWithIdAndSuccess() throws Exception {
...
 		}
 		finally {
 			container.stop();
 			ctx.close();
 		}
 	}
 
+	@SuppressWarnings("unchecked")
+	private Map<String, Object> getCache(Advice retryInterceptor) {
```
artembilan, line 192: `static`?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/retry/MessageKeyGenerator.java
```diff
@@ -31,10 +32,13 @@ public interface MessageKeyGenerator {
 	 * about assuming uniqueness of any element of the message, especially considering the requirement that it be
 	 * repeatable. A message id is ideal, but may not be present (AMQP does not mandate it), and the message body is a
 	 * byte array whose contents might be repeatable, but its object value is not.
+	 * <p>While returning {@code null} is allowed, this represents a faulty scenario that
+	 * will prevent retry operations to work correctly.
```
artembilan, line 36: In Spring Retry Dave say:
```
		Object generatedKey = defaultKey;
		if (this.keyGenerator != null) {
			generatedKey = this.keyGenerator.getKey(invocation.getArguments());
		}
		if (generatedKey == null) {
			// If there's a generator and he still says the key is null, that means he
			// really doesn't want to retry.
			return null;
		}
```

May it is better to turn it off in favor of non-null contract?
Since we are not relying on Spring Retry any more and I don't see why we would allow nullable path where such a `MessageKeyGenerator` injection would be for nothing in the `StatefulRetryOperationsInterceptor` configuration.

artembilan (changes requested):
This is very smooth.
It also looks like we can remove the `SendRetryContextAccessor` class as well.
I can fix all my concerns on merge.
Will also add your name to all affected classes.
Plus some note into `whats-new.adoc`.
No need to bother you for clean up: just let me WDYT on this my feedback!

Thank you!

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/retry/MissingIdRetryTests.java
snicoll: There is already a wait period before that. The only reason for the 3 seconds is the time by which the processor stops (noop) and the interceptor clears the entry.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/retry/MessageRecoveryCallback.java
snicoll: It is an interceptor for stateful retries. If there is one it is stateful so the javadoc looks good to me.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/retry/MissingIdRetryTests.java
snicoll: Sure...

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/retry/MessageKeyGenerator.java
snicoll: I am not sure what you mean. I want 100% this code to not allow null but Spring AMQP does not allow it. Try to put an assert non null and run the tests and you'll see.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/retry/MessageRecoveryCallback.java
artembilan: But this is a `MessageRecoveryCallback`, not an interceptor.
This can be simply injected alongside with any interceptor into that `BaseRabbitListenerContainerFactory`.
What do I miss?

snicoll: Sorry I was reading this on my phone and assumed the wrong class. Please update the javadoc as you see fit.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/retry/MessageKeyGenerator.java
artembilan: OK. I see now.
We have a fallback implementation in the `StatefulRetryOperationsInterceptorFactoryBean`, based on the `messageId`.
And if that is null we have a logic in the `AbstractMessageListenerContainer` to react respectively:
```
	private void checkStatefulRetry(RuntimeException ex, Message message) {
		if (message.getMessageProperties().isFinalRetryForMessageWithNoId()) {
			if (this.statefulRetryFatalWithNullMessageId) {
				throw new FatalListenerExecutionException(
						"Illegal null id in message. Failed to manage retry for message: " + message, ex);
			}
			else {
				throw new ListenerExecutionFailedException("Cannot retry message more than once without an ID",
						new AmqpRejectAndDontRequeueException("Not retryable; rejecting and not requeuing", ex),
						message);
			}
		}
	}
```

So, sounds like it is OK for some scenarios to return null from this generator.
We probably should repurpose this Javadoc sentence indicating that retry won't happen any more if key is `null`.
Might be the case that message was retried a couple times before, then after some logic it was modified to remote its `messageId` to stop such a retrying on it and just reject.
Make sense?

snicoll: I don't have the history of the projet to know for sure. It'd be nice if that wasn't allowed but I doubt that's going to be easy. 

Feel free to rephrase as you see fit. I tried to document with my limited knowledge of the project.

artembilan:
Merged as is.
Thank you!

PR #3221, opened 2025-11-06
Use Gradle's platform support

wilkinsona:
This PR replaces use of the dependency management plugin with use of Gradle's built-in platform support.


artembilan (approved):
Wow!
If we can do just like that, then it is really not bad to avoid extra plugin and live with some minor Gradle requirements.

Thank you!
Merging...

PR #3233, opened 2025-11-12
Optimize `ArrayList` allocations

artembilan:
Fix `new ArrayList()` calls with explicit size or collection copy whenever it is possible or reasonable


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/MessageListenerAdapter.java
```diff
@@ -205,7 +206,7 @@ private void doSetDelegate(Object delegate) {
 	}
 
 	/**
-	 * @return The target object to delegate message listening to.
+	 * @return The target objects to delegate a message listening to.
```
cppwfs: I think `object` (single) is proper here.

cppwfs (changes requested):
Great optimization.  One nitpick

PR #3281, opened 2026-01-09
Add `AmqpMessageListenerContainer` for AMQP 1.0

artembilan:
* Implement the container a similar way to the `RabbitAmqpListenerContainer`
* Use some reflection for calling internal ProtonJ API
* Therefore, expose `AmqpRuntimeHints` for AOT
* Add useful `setReceivedRoutingKey(message.to())` mapping into the `ProtonUtils.fromProtonMessage()`
* Guard client exceptions parsing against possible nullable `ErrorCondition`
* Document `AmqpMessageListenerContainer`

### DIFF
--- a/spring-amqp-client/src/main/java/org/springframework/amqp/client/aot/package-info.java
+++ b/spring-amqp-client/src/main/java/org/springframework/amqp/client/aot/package-info.java
@@ -1,5 +1,5 @@
 /**
- * Provides classes to support Spring AOT.
+ * Provides classes to support AOT for Spring AMQP 1.0.
  */
 @org.jspecify.annotations.NullMarked
 package org.springframework.amqp.client.aot;
--- a/spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpMessageListenerContainer.java
+++ b/spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpMessageListenerContainer.java
@@ -114,7 +114,7 @@ public class AmqpMessageListenerContainer implements MessageListenerContainer, B
 
 	private boolean taskExecutorSet;
 
-	private boolean autoAccept;
+	private boolean autoAccept = true;
 
 	private int initialCredits = 100;
 
@@ -131,7 +131,7 @@ public class AmqpMessageListenerContainer implements MessageListenerContainer, B
 	public void setupMessageListener(MessageListener messageListener) {
 		this.messageListener = messageListener;
 		this.asyncReplies = messageListener.isAsyncReplies();
-		this.proxy = this.messageListener;
+
 		if (!ObjectUtils.isEmpty(this.adviceChain)) {
 			ProxyFactory factory = new ProxyFactory(messageListener);
 			for (Advice advice : this.adviceChain) {
@@ -139,6 +139,9 @@ public class AmqpMessageListenerContainer implements MessageListenerContainer, B
 			}
 			factory.setInterfaces(messageListener.getClass().getInterfaces());
 			this.proxy = (MessageListener) factory.getProxy(getClass().getClassLoader());
+		}
+		else {
+			this.proxy = messageListener;
 		}
 	}
 
@@ -221,7 +224,6 @@ public class AmqpMessageListenerContainer implements MessageListenerContainer, B
 	 * @param taskExecutor the task executor.
 	 */
 	public void setTaskExecutor(Executor taskExecutor) {
-		Assert.notNull(taskExecutor, "'taskExecutor' cannot be null");
 		this.taskExecutor = taskExecutor;
 		this.taskExecutorSet = true;
 	}
@@ -275,6 +277,7 @@ public class AmqpMessageListenerContainer implements MessageListenerContainer, B
 								// the auto-replenishment for the credit window is disabled.
 								.creditWindow(0)
 								.autoAccept(this.autoAccept);
+
 				for (String queue : this.queues) {
 					for (int i = 0; i < this.consumersPerQueue; i++) {
 						try {
@@ -442,6 +445,7 @@ public class AmqpMessageListenerContainer implements MessageListenerContainer, B
 			ReflectionUtils.makeAccessible(SESSION_WINDOW_FIELD);
 			ReflectionUtils.makeAccessible(WRITE_FLOW_METHOD);
 		}
+
 		private final ClientReceiver receiver;
 
 		private final ProtonReceiver protonReceiver;
--- a/spring-amqp-client/src/test/java/org/springframework/amqp/client/listener/AmqpMessageListenerContainerTests.java
+++ b/spring-amqp-client/src/test/java/org/springframework/amqp/client/listener/AmqpMessageListenerContainerTests.java
@@ -24,6 +24,7 @@ import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.qpid.protonj2.client.Client;
+import org.apache.qpid.protonj2.engine.impl.ProtonReceiver;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 
@@ -32,11 +33,15 @@ import org.springframework.amqp.client.AmqpConnectionFactory;
 import org.springframework.amqp.client.SingleAmqpConnectionFactory;
 import org.springframework.amqp.core.Message;
 import org.springframework.amqp.rabbit.junit.AbstractTestContainerTests;
+import org.springframework.amqp.utils.test.TestUtils;
+import org.springframework.aop.interceptor.DebugInterceptor;
+import org.springframework.aop.support.AopUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.test.annotation.DirtiesContext;
 import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
+import org.springframework.util.MultiValueMap;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
@@ -76,6 +81,8 @@ public class AmqpMessageListenerContainerTests extends AbstractTestContainerTest
 
 	@Test
 	void messagesConsumedFromAllQueues() throws InterruptedException {
+		assertThat(AopUtils.isAopProxy(amqpMessageListenerContainer.getMessageListener())).isTrue();
+
 		Map<String, String> dataToQueue =
 				Map.of(
 						"test_data1", TEST_QUEUE1,
@@ -102,8 +109,19 @@ public class AmqpMessageListenerContainerTests extends AbstractTestContainerTest
 	}
 
 	@Test
+	@SuppressWarnings("unchecked")
 	void pauseAndResumeContainer() throws InterruptedException {
+		MultiValueMap<String, ?> queueToConsumers =
+				TestUtils.getPropertyValue(this.amqpMessageListenerContainer, "queueToConsumers", MultiValueMap.class);
+
+		Object amqpConsumer = queueToConsumers.getFirst(TEST_QUEUE1);
+		ProtonReceiver protonReceiver = TestUtils.getPropertyValue(amqpConsumer, "protonReceiver", ProtonReceiver.class);
+
+		assertThat(protonReceiver.getCredit()).isEqualTo(100);
+
 		this.amqpMessageListenerContainer.pause();
+
+		assertThat(protonReceiver.getCredit()).isEqualTo(0);
 
 		assertThat(this.amqpClient.to(TEST_QUEUE1).body("after resume").send())
 				.succeedsWithin(Duration.ofSeconds(10));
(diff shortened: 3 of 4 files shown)


artembilan:
This is like initial PoC.
There are still a lot of work for "native" listener contract.
Then `@AmqpListener` support.
Separate PRs, though.
Thanks

spring-amqp-client/src/main/java/org/springframework/amqp/client/aot/package-info.java
```diff
@@ -0,0 +1,5 @@
+/**
+ * Provides classes to support Spring AOT.
```
cppwfs: Spring AMQP Client AOT

WDYT?

src/reference/antora/modules/ROOT/pages/amqp10-client.adoc
```diff
@@ -149,8 +149,43 @@ CompletableFuture<TestData> receiveFuture =
 See Javadocs for those methods for more information about the fluent API exposed by the `AmqpClient`.
 
 
-////
 [[amqp10-client-listener]]
 == The AMQP 1.0 Consumer
-////
 
+The event-driven consumer for AMQP 1.o protocol is implemented as an `AmqpMessageListenerContainer` similar to all other xref:amqp/receiving-messages/async-consumer.adoc[listener container implementations] in the framework.
```
cppwfs: 1.0 vs 1.o

src/reference/antora/modules/ROOT/pages/amqp10-client.adoc
```diff
@@ -149,8 +149,43 @@ CompletableFuture<TestData> receiveFuture =
...
+The `AmqpMessageListenerContainer` requires an `AmqpConnectionFactory` and uses `org.apache.qpid.protonj2.client.Receiver` instances internally to consume messages from the provided `queueNames` (essentially, AMQP 1.o addresses).
+The `100` initial credits are used for receiver links by default.
+Every settled delivery then replenishes some number of credits for new upcoming messages.
+The consumed messages are handled by the provided `MessageListener`, and if the `autoAccept` is set to `false`, an `AmqpAcknowledgment` implementation is populated to respective Spring AMQP message property for manual delivery settlement in the target message listener logic.
```
cppwfs, line 159: `autoAccept` is set to false (the default)

spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpMessageListenerContainer.java
```diff
@@ -0,0 +1,556 @@
...
+	@Override
+	public void setBeanName(String name) {
+		this.beanName = name;
+	}
+
+	@Override
+	public void setupMessageListener(MessageListener messageListener) {
+		this.messageListener = messageListener;
+		this.asyncReplies = messageListener.isAsyncReplies();
+		this.proxy = this.messageListener;
```
cppwfs: To be consistent shouldn't this be `messageListener` instead of `this.messageListener`?

spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpMessageListenerContainer.java
```diff
@@ -0,0 +1,556 @@
...
+	public void setErrorHandler(@Nullable ErrorHandler errorHandler) {
+		this.errorHandler = errorHandler;
+	}
+
+	/**
+	 * Set a task executor to run consumers.
+	 * @param taskExecutor the task executor.
+	 */
+	public void setTaskExecutor(Executor taskExecutor) {
+		Assert.notNull(taskExecutor, "'taskExecutor' cannot be null");
```
cppwfs: Do we need this assert for not null ?

spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpMessageListenerContainer.java
```diff
@@ -0,0 +1,556 @@
...
+		try {
+			CompletableFuture<Void>[] completableFutures =
+					this.queueToConsumers.values().stream()
+							.flatMap(List::stream)
+							.map((consumer) ->
+									CompletableFuture.supplyAsync(() -> {
+										consumer.pause();
+										try (consumer) {
+											while (consumer.queuedDeliveries() > 0) {
+												Thread.sleep(100);
```
cppwfs, line 322: Should this be configurable?

spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpMessageListenerContainer.java
```diff
@@ -0,0 +1,556 @@
...
+		private final ProtonSessionIncomingWindow sessionWindow;
+
+		private volatile boolean paused;
+
+		private volatile boolean running = true;
+
+		@SuppressWarnings("NullAway")
+		AmqpConsumer(ClientReceiver receiver) {
+			this.receiver = receiver;
+			this.protonReceiver = (ProtonReceiver) ReflectionUtils.invokeMethod(PROTON_RECEIVER_METHOD, receiver);
```
cppwfs, line 464: I see where we suppress the nulls for this method.  But is it possible that these invokes would return a null?   Just curious.

spring-amqp-client/src/test/java/org/springframework/amqp/client/listener/AmqpMessageListenerContainerTests.java
```diff
@@ -0,0 +1,156 @@
...
+							assertThat(dataToQueue)
+									.containsEntry(new String(m.getBody()), m.getMessageProperties().getReceivedRoutingKey()));
+			message.getMessageProperties().getAmqpAcknowledgment().acknowledge();
+		}
+		Message noMessage = this.testConfig.receivedMessages.poll(1, TimeUnit.SECONDS);
+		assertThat(noMessage).isNull();
+	}
+
+	@Test
+	void pauseAndResumeContainer() throws InterruptedException {
```
cppwfs, line 113: This maybe overkill, but I thought it'd be a good idea to check the credit during the  pause restart process:

```
	@Test
	void pauseAndResumeContainer() throws InterruptedException {
		ProtonReceiver protonReceiver = getFirstProtonReceiver();
		int credit = protonReceiver.getCredit();
		assertThat(credit).isEqualTo(100);
		this.amqpMessageListenerContainer.pause();
		credit = protonReceiver.getCredit();
		assertThat(credit).isEqualTo(0);
		assertThat(this.amqpClient.to(TEST_QUEUE1).body("after resume").send())
				.succeedsWithin(Duration.ofSeconds(10));

		Message message = this.testConfig.receivedMessages.poll(1, TimeUnit.SECONDS);
		assertThat(message).isNull();

		this.amqpMessageListenerContainer.resume();

		message = this.testConfig.receivedMessages.poll(10, TimeUnit.SECONDS);
		assertThat(message)
				.extracting(Message::getBody)
				.isEqualTo("after resume".getBytes());

		credit = protonReceiver.getCredit();
		assertThat(credit).isLessThan(100);
	}

	private ProtonReceiver getFirstProtonReceiver() {
		Field queueToConsumersField = ReflectionUtils.findField(AmqpMessageListenerContainer.class, "queueToConsumers");
		ReflectionUtils.makeAccessible(queueToConsumersField);
		var queueToConsumers = (org.springframework.util.MultiValueMap<String, ?>)
				ReflectionUtils.getField(queueToConsumersField, this.amqpMessageListenerContainer);
		assertThat(queueToConsumers.values()).isNotNull().isNotEmpty();

		// MultiValueMap.values() returns Collection<List<T>>, so get first list then first consumer
		var consumerList = queueToConsumers.values().iterator().next();
		assertThat(consumerList).isNotEmpty();
		Object firstConsumer = consumerList.get(0);

		Field protonReceiverField = ReflectionUtils.findField(firstConsumer.getClass(), "protonReceiver");
		ReflectionUtils.makeAccessible(protonReceiverField);
		return (ProtonReceiver) ReflectionUtils.getField(protonReceiverField, firstConsumer);
	}
```

WDYT?

src/reference/antora/modules/ROOT/pages/amqp10-client.adoc
```diff
@@ -149,8 +149,43 @@ CompletableFuture<TestData> receiveFuture =
...
+
+The `consumersPerQueue` option (`1` by default) implements a concurrency behavior for each provided address to consume.
+
+The `Duration receiveTimeout` option (`1 second` by default) controls the blocking `Receiver.receive()` operation, which is called in a loop until consumer is stopped.
+
+Each consumer (an internal `AmqpMessageListenerContainer.AmqpConsumer` instance) is scheduled for running via `Executor`.
+The default one is `SimpleAsyncTaskExecutor`.
```
cppwfs: `The default one is` should be replaced with `The default is`.

src/reference/antora/modules/ROOT/pages/amqp10-client.adoc
```diff
@@ -149,8 +149,43 @@ CompletableFuture<TestData> receiveFuture =
...
+
+The message processing errors can be handled by the `ErrorHandler` configuration.
+
+The `MessageListener` can be proxied in the `AmqpMessageListenerContainer` if some AOP interceptors are provided, e.g. `TransactionInterceptor`.
+
+The `AmqpMessageListenerContainer` also provides `pause()` and `resume()` API to set AMQP link credits to `0` and replenish, respectively.
+This functionality, essentially, leaves a `Receiver` active, but no delivery is coming from the AMQP peer.
```
cppwfs: possibly remove the word `essentially`.

cppwfs (changes requested):
Great work here.

Helllooooo AmqpMessageListenerContainer!

spring-amqp-client/src/test/java/org/springframework/amqp/client/listener/AmqpMessageListenerContainerTests.java
artembilan: OK. I'll incorporate your suggestion.
Thanks.
However that would be a bit refactored:
1. We don't use `ReflectionUtils` in our test code. Instead we have `TestUtils.getPropertyValue()` with similar functionality.
2. With the `TestUtils` we don't need those `makeAccessible()`
3. The method you suggest is really not a utility method with all those `assertThat()`, which really belongs to the test method itself.
4. The `assertThat(credit).isLessThan(100);` is really could be equal to `0` as it was before `resume()`. So, `isLessThan()` does not reflect reality.

src/reference/antora/modules/ROOT/pages/amqp10-client.adoc
artembilan: This is good catch: it really has to be `true` by default 😄

src/reference/antora/modules/ROOT/pages/amqp10-client.adoc
artembilan: Thanks.
I'll rework the sentence a bit.
But give me, please, English lesson here.
In my language to have a connection to the previous sentence and reflect context of the main subject and don't repeat yourself, we use that `one` word. So, how properly in English to do something similar?
How to state in this sentence that we talk about an `Executor`, without repeating an `Executor` word, which is the last one in the previous sentence?
(You see: I used the `one` again 😄 )

cppwfs: Basically in the original version the word `default` is an adjective to the noun `one` and the verb is `is`. 
In the requested change `default` becomes the noun and the verb is `is`.

Both are correct!    I was going for brevity.   So its a request that can be declined.

spring-amqp-client/src/test/java/org/springframework/amqp/client/listener/AmqpMessageListenerContainerTests.java
cppwfs: Please!   And thank you for the corrections!!!!!  You Rock!

spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpMessageListenerContainer.java
artembilan: Copy/paste artifact from other similar classes 😄 
Thanks

spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpMessageListenerContainer.java
artembilan: I don't think so. This is a busy-wait tick to avoid blocking CPU.
Kinda typical pattern for simple loop-based short condition barriers.

spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpMessageListenerContainer.java
artembilan: See `ClientReceiverLinkType.protonReceiver` property and how it is populated.
Essentially that is done in the `ClientReceiverBuilder.receiver()`.
So, this `protonLink()` never returns null.

PR #3288, opened 2026-01-14
Introduce `@EnableAmqp`

artembilan:
* Import an `AmqpDefaultConfiguration` from the `@EnableAmqp` with convenient infrastructure beans
* Make a default `Client` bean based on the `ClientOptions` built from the `@EnableAmqp` attributes
* Mark `AmqpDefaultConfiguration.protonClient()` bean as `@Conditional` and skip its registration when any `Client` bean is already present in the target project configuration
* Make `Client` injection into the `SingleAmqpConnectionFactory` as optional, and resolve single `Client` from the `BeanFactory` on demand internally
* Change `AmqpClientTests` and `AmqpMessageListenerContainerTests` configurations to rely this or other way on the `@EnableAmqp`
* Document the `@EnableAmqp` feature

### DIFF
--- a/spring-amqp-client/src/main/java/org/springframework/amqp/client/config/AmqpDefaultConfiguration.java
+++ b/spring-amqp-client/src/main/java/org/springframework/amqp/client/config/AmqpDefaultConfiguration.java
@@ -94,9 +94,6 @@ public class AmqpDefaultConfiguration implements ImportAware, EnvironmentAware {
 
 	private static final class OnMissingClientBeanCondition implements Condition {
 
-		OnMissingClientBeanCondition() {
-		}
-
 		@Override
 		@SuppressWarnings("NullAway")
 		public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata) {
--- a/spring-amqp-client/src/main/java/org/springframework/amqp/client/config/EnableAmqp.java
+++ b/spring-amqp-client/src/main/java/org/springframework/amqp/client/config/EnableAmqp.java
@@ -42,7 +42,7 @@ import org.springframework.context.annotation.Import;
 public @interface EnableAmqp {
 
 	/**
-	 * The id for the {@link org.apache.qpid.protonj2.client.ClientOptions#id()}
+	 * The value for the {@link org.apache.qpid.protonj2.client.ClientOptions#id()}
 	 * of the {@link org.apache.qpid.protonj2.client.Client} to be created by the {@link AmqpDefaultConfiguration}.
 	 * Can be specified as a property placeholder, e.g. {@code "${amqp.client-id}"}.
 	 * @return the client id.
@@ -50,10 +50,10 @@ public @interface EnableAmqp {
 	String clientId() default "";
 
 	/**
-	 * The future type for the {@link ClientOptions#futureType()}
+	 * The value for the {@link ClientOptions#futureType()}
 	 * of the {@link org.apache.qpid.protonj2.client.Client} to be created by the {@link AmqpDefaultConfiguration}.
 	 * Can be specified as a property placeholder, e.g. {@code "${amqp.future-type}"}.
-	 * @return the client id.
+	 * @return the future type for the client.
 	 */
 	String clientFutureType() default "";
 
--- a/src/reference/antora/modules/ROOT/pages/amqp10-client.adoc
+++ b/src/reference/antora/modules/ROOT/pages/amqp10-client.adoc
@@ -222,7 +222,7 @@ AmqpMessageListenerContainer protonDeliveryListenerContainer(AmqpConnectionFacto
 == The AMQP 1.0 Annotation Configuration
 
 The `spring-amqp-client` provides a convenient way to configure the AMQP 1.0 infrastructure via annotations.
-One of the `@Configuration` classes in the target project can be marked with an `@EnableAmqp` annotation, which imports an `AmqpDefaultConfiguration`.
-The `AmqpDefaultConfiguration` provides conditional beans (may be configured in the target project manually):
+The `@EnableAmqp` annotation can be set on a `@Configuration` class in the target project to trigger that infrastructure registration.
+The imported by the `@EnableAmqp` annotation `AmqpDefaultConfiguration` provides conditional beans (may be configured in the target project manually):
 
-* The `org.apache.qpid.protonj2.client.Client` with `org.apache.qpid.protonj2.client.ClientOptions` based on the respective attributes of the imported metadata from the `@EnableAmqp`.
+* The `org.apache.qpid.protonj2.client.Client` where its `org.apache.qpid.protonj2.client.ClientOptions` property is based on the respective attributes of the imported metadata from the `@EnableAmqp`.


src/reference/antora/modules/ROOT/pages/amqp10-client.adoc
```diff
@@ -188,4 +191,13 @@ AmqpMessageListenerContainer amqpMessageListenerContainer(AmqpConnectionFactory
...
 }
-----
\ No newline at end of file
+----
+
+[[amqp10-annotation-configuration]]
+== The AMQP 1.0 Annotation Configuration
+
+The `spring-amqp-client` provides a convenient way to configure the AMQP 1.0 infrastructure via annotations.
+One of the `@Configuration` classes in the target project can be marked with an `@EnableAmqp` annotation, which imports an `AmqpDefaultConfiguration`.
```
cppwfs: Replace "One of the `@Configuration` classes" with  "A single `@Configuration` class in the target"

src/reference/antora/modules/ROOT/pages/amqp10-client.adoc
```diff
@@ -214,4 +217,13 @@ AmqpMessageListenerContainer protonDeliveryListenerContainer(AmqpConnectionFacto
...
+The `AmqpDefaultConfiguration` provides conditional beans (may be configured in the target project manually):
+
+* The `org.apache.qpid.protonj2.client.Client` with `org.apache.qpid.protonj2.client.ClientOptions` based on the respective attributes of the imported metadata from the `@EnableAmqp`.
```
cppwfs: The `org.apache.qpid.protonj2.client.Client` with `org.apache.qpid.protonj2.client.ClientOptions` are configured based on the attributes specified in the `@EnableAmqp` annotation.

WDYT?

src/reference/antora/modules/ROOT/pages/amqp10-client.adoc
```diff
@@ -46,6 +46,7 @@ Client protonClient() {
 
 The same `Client` can be used for connecting to different brokers, the connection settings must be provided on a specific connection.
 See below.
+See also
```
artembilan: I see I got the problem.
No merge yet, please.

spring-amqp-client/src/main/java/org/springframework/amqp/client/SingleAmqpConnectionFactory.java
```diff
@@ -105,6 +125,9 @@ public Connection getConnection() {
 			try {
 				connectionToReturn = this.connection;
 				if (connectionToReturn == null) {
+					if (this.protonjClient == null) {
+						this.protonjClient = this.beanFactory.getBean(Client.class);
```
cppwfs, line 129: Should we log debug that we are trying to get the projectjClient as a bean?

spring-amqp-client/src/main/java/org/springframework/amqp/client/config/EnableAmqp.java
```diff
@@ -0,0 +1,60 @@
...
+	 * Can be specified as a property placeholder, e.g. {@code "${amqp.client-id}"}.
+	 * @return the client id.
+	 */
+	String clientId() default "";
+
+	/**
+	 * The future type for the {@link ClientOptions#futureType()}
+	 * of the {@link org.apache.qpid.protonj2.client.Client} to be created by the {@link AmqpDefaultConfiguration}.
+	 * Can be specified as a property placeholder, e.g. {@code "${amqp.future-type}"}.
+	 * @return the client id.
```
cppwfs: String containing the future type.

spring-amqp-client/src/main/java/org/springframework/amqp/client/config/AmqpDefaultConfiguration.java
```diff
@@ -0,0 +1,108 @@
...
+			if (StringUtils.hasText(futureType)) {
+				options.futureType(futureType);
+			}
+		}
+		return Client.create(options);
+	}
+
+	private static final class OnMissingClientBeanCondition implements Condition {
+
+		OnMissingClientBeanCondition() {
```
cppwfs: We probably don't need the default constructor here.

cppwfs (changes requested):
Great addition!

Just my usual annoying comments.

spring-amqp-client/src/main/java/org/springframework/amqp/client/config/AmqpDefaultConfiguration.java
artembilan: You are right.
Java generates a relationship constructor for inner non static classes.
For example, if we remove this ctor and remove `static` from the class definition, then we got this in the bytecode:
```
final class AmqpDefaultConfiguration$OnMissingClientBeanCondition implements Condition {
   // $FF: synthetic field
   final AmqpDefaultConfiguration this$0;

   private AmqpDefaultConfiguration$OnMissingClientBeanCondition(final AmqpDefaultConfiguration this$0) {
      this.this$0 = var1;
   }

   public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata) {
      return context.getBeanFactory().getBeanNamesForType(Client.class).length == 0;
   }
}
```
With just the `static` on class we get:
```
final class AmqpDefaultConfiguration$OnMissingClientBeanCondition implements Condition {
   private AmqpDefaultConfiguration$OnMissingClientBeanCondition() {
   }

   public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata) {
      return context.getBeanFactory().getBeanNamesForType(Client.class).length == 0;
   }
}
```

Which is totally fine for us since `ConditionEvaluator`  does `BeanUtils.instantiateClass(conditionClass)` anyway.
I don't want to make this class `public` since it has a limited usage with its condition just for a `Client.class` .

So, remove this ctor.
Thank you!

spring-amqp-client/src/main/java/org/springframework/amqp/client/SingleAmqpConnectionFactory.java
artembilan: I don't think so.
From this class logic it does not matter how we get a `Client`.
Either way in Spring application, the ctor injection is usually done via bean.
So, giving a debug message about bean would not bring too much value.
I don't see how that could be useful, too.

spring-amqp-client/src/main/java/org/springframework/amqp/client/config/EnableAmqp.java
artembilan: Why do we need to mention the type in the return tag since it is obvious from the return type anyway?

src/reference/antora/modules/ROOT/pages/amqp10-client.adoc
artembilan: "A single" does not compile in my head for my language comprehension. Causes an impression that only one `@Configuration` has to be in my project.
But I'll rephrase the sentence some other way.
Might that would be a compromise for us 😄  
Thanks

src/reference/antora/modules/ROOT/pages/amqp10-client.adoc
artembilan: Well, we really configure a `Client` bean. 
The `ClientOptions`  is something internal to that bean.
But I'll change this sentence a bit.
Thanks

spring-amqp-client/src/main/java/org/springframework/amqp/client/config/EnableAmqp.java
cppwfs: We can remove the return or change it to future type.   But as it is it is incorrect because it does not return a client id.

PR #3289, opened 2026-01-15
Move common APIs from rabbit `module` to `core`

artembilan:
Some exceptions and listener supporting classes
are not only for RabbitMQ integration.
They can be used in any other Spring AMQP implementations.

* Deprecate existing classes in favor of their new counterpart with extensions where possible.
* Still use instances of deprecated classes for proper upper bound of the consumers of the target API, essentially following Postel's Law
* Move `ContainerUtilsTests` to `core` module as we don't test a deprecated class any more
* Add `reactor-core` optional dependency to `core` module to satisfy `ContinuationHandlerMethodArgumentResolver` requirements


spring-amqp/src/main/java/org/springframework/amqp/listener/ContainerUtils.java
```diff
@@ -0,0 +1,117 @@
...
+	 * there is an {@link ImmediateRequeueAmqpException} in the cause chain.
+	 * @param defaultRequeueRejected the default requeue rejected.
+	 * @param throwable the throwable.
+	 * @param logger the logger to use for debug.
+	 * @return true to requeue.
+	 */
+	public static boolean shouldRequeue(boolean defaultRequeueRejected, Throwable throwable, Log logger) {
+		boolean shouldRequeue = defaultRequeueRejected ||
+				throwable instanceof MessageRejectedWhileStoppingException;
+		Throwable t = throwable;
```
cppwfs, line 53: Why not use `throwable` instead of assigning it to `t`?

spring-amqp/src/main/java/org/springframework/amqp/listener/ListenerExecutionFailedException.java
```diff
@@ -0,0 +1,65 @@
...
+	 * @param msg the detail message
+	 * @param cause the exception thrown by the listener method
+	 * @param failedMessage the message(s) that failed
+	 */
+	public ListenerExecutionFailedException(String msg, Throwable cause, Message... failedMessage) {
+		super(msg, cause);
+		this.failedMessages = new ArrayList<>(Arrays.asList(failedMessage));
+	}
+
+	public @Nullable Message getFailedMessage() {
```
cppwfs, line 57: Feels like this should be `getFirstFailedMessage`

spring-amqp/src/main/java/org/springframework/amqp/listener/ContainerUtils.java
artembilan: Because we do `t = cause;` a bit bellow in the logic.
And good practice is not to mutate method argument.
Kinda treat all the method arguments as `final`.

spring-amqp/src/main/java/org/springframework/amqp/listener/ListenerExecutionFailedException.java
artembilan: That is how it was there for for years and I'm fully OK with this contract.
Mostly this exception is created with a single failed message.

PR #3305, opened 2026-01-21
Add `TestUtils.propertyValue()` based on generics

artembilan:
* Deprecate `TestUtils.getPropertyValue()` based on a `Class` param
* Optimize `TestUtils.getPropertyValue()` to not create extra `DirectFieldAccessor` if we are at the end of the nested path
* Refactor project code to use a new `TestUtils.propertyValue()`
* Remove unnecessary now `@SuppressWarnings("unchecked")` from those tests
* Some other code clean up in affected test classes

### DIFF
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/config/AdminParserTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/config/AdminParserTests.java
@@ -76,6 +76,10 @@ public final class AdminParserTests {
 	private void doTest(boolean explicit) {
 		// Create context
 		DefaultListableBeanFactory beanFactory = loadContext();
+		if (beanFactory == null) {
+			// Context was invalid
+			return;
+		}
 
 		// Validate values
 		RabbitAdmin admin = beanFactory.getBean(RabbitAdmin.class);
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactoryTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactoryTests.java
@@ -1901,7 +1901,8 @@ public class CachingConnectionFactoryTests extends AbstractConnectionFactoryTest
 		ccf.setAddressResolver(resolver);
 		Connection con = ccf.createConnection();
 		assertThat(con).isNotNull();
-		assertThat(TestUtils.<Connection>getPropertyValue(con, "target.delegate")).isEqualTo(mockConnection);
+		assertThat(TestUtils.<com.rabbitmq.client.Connection>getPropertyValue(con, "target.delegate"))
+				.isEqualTo(mockConnection);
 		verify(mockConnectionFactory).newConnection(any(ExecutorService.class), eq(resolver), anyString());
 	}
 


artembilan:
A lot of changes in this PR are about migrating from deprecated API.
Thanks

artembilan:
Now, when I look into this from PR perspective, it feels like we don't need that `getPropertyValue()` for `Object` at all.
We need to deprecated it too to avoid a paradox of choice.
It doesn't look like hard to do that, but that would be much more changes.
I can proceed with that after your review, though.

spring-amqp/src/main/java/org/springframework/amqp/utils/test/TestUtils.java
```diff
@@ -62,6 +67,7 @@ private TestUtils() {
 		return value;
 	}
 
+	@Deprecated(since = "4.1", forRemoval = true)
```
cppwfs: Since it is still public static shouldn't it have a javadoc but denoting the `@deprecated` annotation?

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/config/AdminParserTests.java
```diff
@@ -57,28 +58,28 @@ public final class AdminParserTests {
 	private boolean initialisedWithTemplate;
```
cppwfs: Looks like we can remove `initialisedWithTemplate` and `adminBeanName` because they are never set, and thus their `if` blocks can be removed from the code.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/config/AdminParserTests.java
```diff
@@ -101,12 +102,11 @@ private void doTest(boolean explicit) throws Exception {
...
 			assertThat(admin.getRabbitTemplate()).isEqualTo(beanFactory.getBean(RabbitTemplate.class));
 		}
-		assertThat(TestUtils.getPropertyValue(admin, "explicitDeclarationsOnly", Boolean.class)).isEqualTo(explicit);
+		assertThat(TestUtils.<Boolean>propertyValue(admin, "explicitDeclarationsOnly")).isEqualTo(explicit);
 	}
 
 	/**
-	 * Load application context. Fail if tests expects invalid spring-context, but spring-context is valid.
-	 * @return
+	 * Load application context. Fail if tests expect invalid spring-context, but spring-context is valid.
```
cppwfs: Shouldn't we have a `@return` description for the DefaultListableBeanFactory?

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactoryTests.java
```diff
@@ -577,7 +570,7 @@ public void testCheckoutLimitWithRelease() throws Exception {
 
 		final Connection con = ccf.createConnection();
 
-		final AtomicReference<Channel> channelOne = new AtomicReference<Channel>();
+		final AtomicReference<Channel> channelOne = new AtomicReference<>();
```
cppwfs, line 573: Not sure how to report this exactly, but since we are on a cleanup of these classes 😄 ). On lines 587-590 we can collapse the catches.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/SSLConnectionTests.java
```diff
@@ -75,7 +76,8 @@ public void test() throws Exception {
 	public void testAlgNoProps() throws Exception {
 		RabbitConnectionFactoryBean fb = new RabbitConnectionFactoryBean();
 		fb.setMaxInboundMessageBodySize(1000);
-		ConnectionFactory rabbitCf = spy(TestUtils.getPropertyValue(fb, "connectionFactory", ConnectionFactory.class));
+		ConnectionFactory rabbitCf = TestUtils.propertyValue(fb, "connectionFactory");
```
cppwfs: Can these 2 lines be consolidated like below:  i.e. 		

```
ConnectionFactory rabbitCf = spy(TestUtils.<ConnectionFactory>propertyValue(fb, "connectionFactory"));
```

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/config/AdminParserTests.java
artembilan: We do not. This is a `private` method and it is a part of JUnit test.
Removing altogether...

PR #3325, opened 2026-02-06
Add `AmqpMessagingListenerAdapter` for AMQP 1.0 Client

artembilan:
* Add `throws Exception` to the `ProtonDeliveryListener.onDelivery()` contract to avoid any possible misleading transformations in the implementation
* Add `AcknowledgingProtonDeliveryListener` contract as an extension of the `ProtonDeliveryListener`.
Enforce its logic to a new contract with an `AmqpAcknowledgment` argument populated by the `AmqpMessageListenerContainer`.
Such a contract is necessary to properly delegate replenish link credits logic back to the listener container.
This logic is now included in the `AmqpMessageListenerContainer.doInvokeListener()` method. Either way, the target implementation is till free to deal with `Delivery` whatever is possible. However, there is a possibility of inconsistency when the container is paused, but listener calls `addCredit()` on the delivery's receiver
* Add `HandlerAdapter.getMethod()` to avoid extra method parameter propagation to consumers of the `HandlerAdapter`
* Extract `record ReplyExpressionRoot` from the `AbstractAdaptableMessageListener` to make it available for any other not RabbitMQ-based modules
* Implement `AmqpMessagingListenerAdapter` for AMQP 1.0 client. It is similar to the `MessagingMessageListenerAdapter` for RabbitMQ-based modules, and is going to be used for POJO-based event-driven message handling, e.g. future `@AmqpListener`. A lot of `AmqpMessagingListenerAdapter` logic is really a copy/paste from the `MessagingMessageListenerAdapter` and its super class

### DIFF
--- a/spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpListenerErrorHandler.java
+++ b/spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpListenerErrorHandler.java
@@ -23,7 +23,7 @@ import org.springframework.amqp.listener.ListenerExecutionFailedException;
 import org.springframework.messaging.Message;
 
 /**
- * An error handler which is called when a {code @AmqpListener} method
+ * An error handler which is called when a {@code @AmqpListener} method
  * throws an exception. This is invoked higher up the stack than the
  * listener container's error handler.
  *
@@ -37,7 +37,6 @@ public interface AmqpListenerErrorHandler {
 	/**
 	 * Handle the error. If an exception is not thrown, the return value is returned to
 	 * the sender using normal {@code replyTo/@SendTo} semantics.
-	 *
 	 * @param delivery  the ProtonJ delivery that failed.
 	 * @param message   the converted spring-messaging message (if available).
 	 * @param exception the exception the listener threw, wrapped in a {@link ListenerExecutionFailedException}.
--- a/spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpMessagingListenerAdapter.java
+++ b/spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpMessagingListenerAdapter.java
@@ -65,12 +65,13 @@ import org.springframework.messaging.MessagingException;
 /**
  * The {@link ProtonDeliveryListener} implementation for POJO listeners
  * delegating to the provided {@link HandlerAdapter}.
- * <p>Wraps the incoming {@link Delivery} to {@link Message},
+ * <p>
+ * Wraps the incoming {@link Delivery} to {@link Message},
  * and further to the Spring's {@link org.springframework.messaging.Message} abstraction,
  * copying the standard headers using a configurable
  * {@link AmqpHeaderMapper AmqpHeaderMapper}.
- *
- * <p>The original {@link Delivery} and {@link Message Message}
+ * <p>
+ * The original {@link Delivery} and {@link Message Message}
  * are provided as additional arguments so that these can be injected as method arguments if necessary.
  *
  * @author Artem Bilan
@@ -105,11 +106,17 @@ public class AmqpMessagingListenerAdapter implements AcknowledgingProtonDelivery
 
 	private @Nullable Expression responseExpression;
 
+	/**
+	 * Create an instance based on the provided {@link HandlerAdapter}.
+	 * @param handlerAdapter the {@link HandlerAdapter} to invoke as POJO message listener.
+	 */
 	public AmqpMessagingListenerAdapter(HandlerAdapter handlerAdapter) {
 		this.handlerAdapter = handlerAdapter;
 		this.messagingMessageConverter =
 				new MessagingMessageConverterAdapter(handlerAdapter.getBean(), handlerAdapter.getMethod(),
 						false, Delivery.class);
+		this.evalContext.setTypeConverter(new StandardTypeConverter());
+		this.evalContext.addPropertyAccessor(new MapAccessor());
 	}
 
 	@Override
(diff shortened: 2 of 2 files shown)


spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpListenerErrorHandler.java
```diff
@@ -0,0 +1,51 @@
...
+package org.springframework.amqp.client.listener;
+
+import org.apache.qpid.protonj2.client.Delivery;
+import org.jspecify.annotations.Nullable;
+
+import org.springframework.amqp.listener.ListenerExecutionFailedException;
+import org.springframework.messaging.Message;
+
+/**
+ * An error handler which is called when a {code @AmqpListener} method
```
cppwfs: This should be `{@code @AmqpListener}`

spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpMessagingListenerAdapter.java
```diff
@@ -0,0 +1,549 @@
...
+		response.getMessageProperties().setCorrelationId(correlation);
+	}
+
+	/**
+	 * Determine a reply-to address for the given message.
+	 * <p>
+	 * The default implementation first checks the Reply-To address of the supplied request; if that is not
+	 * <code>null</code> it is returned; otherwise, then the configured default response address.
+	 * If result still <code>null</code>, the {@link AmqpException} is thrown.
+	 * @param request the original incoming Rabbit message.
```
cppwfs: Should be `AMQP Message` not `Rabbit Message`.

spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpMessagingListenerAdapter.java
```diff
@@ -0,0 +1,549 @@
...
+	 * Set the {@link AmqpHeaderMapper} implementation to use to map the standard
+	 * AMQP headers. By default, a {@link SimpleAmqpHeaderMapper} is used.
+	 * @param headerMapper the {@link AmqpHeaderMapper} instance.
+	 * @see SimpleAmqpHeaderMapper
+	 */
+	public void setHeaderMapper(AmqpHeaderMapper headerMapper) {
+		this.messagingMessageConverter.setHeaderMapper(headerMapper);
+	}
+
+	public void setMessageConverter(@Nullable MessageConverter messageConverter) {
```
cppwfs: Can we remove the `@Nullable` and the subsequent `		if (messageConverter != null) {`?

spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpMessagingListenerAdapter.java
```diff
@@ -0,0 +1,549 @@
...
+
+	private void handleException(Delivery delivery, Message amqpMessage,
+			org.springframework.messaging.@Nullable Message<?> message,
+			ListenerExecutionFailedException lefe) throws Exception {
+
+		if (this.errorHandler != null) {
+			try {
+				Object errorResult = this.errorHandler.handleError(delivery, message, lefe);
+				if (errorResult != null) {
+					Object payload = message == null ? null : message.getPayload();
```
cppwfs, line 303: nitpick: I usually wrap the conditional in parenthesis.  i.e. `(message == null)` to improve readability.

spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpMessagingListenerAdapter.java
```diff
@@ -0,0 +1,549 @@
...
+					new ListenerExecutionFailedException("Failed to handle message", ex, amqpMessage));
+		}
+	}
+
+	private void invokeHandlerAndProcessResult(Delivery delivery, Message amqpMessage,
+			org.springframework.messaging.Message<?> message) {
+
+		MessageProperties messageProperties = amqpMessage.getMessageProperties();
+		boolean projectionUsed = messageProperties.isProjectionUsed();
+		if (projectionUsed) {
```
cppwfs: why perform the check?  We are resetting the `projectionUsed`  in the `messageProperties` to false regardless.

spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpMessagingListenerAdapter.java
```diff
@@ -0,0 +1,549 @@
...
+
+	public void setReplyPostProcessor(ReplyPostProcessor replyPostProcessor) {
+		this.replyPostProcessor = replyPostProcessor;
+	}
+
+	public void setErrorHandler(@Nullable AmqpListenerErrorHandler errorHandler) {
+		this.errorHandler = errorHandler;
+	}
+
+	public void setReturnExceptions(boolean returnExceptions) {
```
cppwfs, line 168: Can we have javadocs for the public setters and getters?

spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpMessagingListenerAdapter.java
artembilan: Ha! I do exactly opposite for exactly same reason  😄

spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpMessagingListenerAdapter.java
artembilan: This was a copy from existing code in the `MessagingMessageListenerAdapter`.
However digging around I don't see how this property is used at all. 
Removing this logic at all...

spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpMessagingListenerAdapter.java
cppwfs: ok you win 😆

PR #3330, opened 2026-02-09
Introduce `AmqpListenerEndpoint` infrastructure

artembilan:
* Add `AmqpListenerEndpoint` abstraction to connect `MessageListener` with `AmqpMessageListenerContainer` at runtime
* Provide `SimpleAmqpListenerEndpoint` implementation for programmatic registration
* Add `AmqpMessageListenerContainerFactory` for creating `AmqpMessageListenerContainer` at runtime
* The `AmqpListenerEndpointRegistration` configuration container is for convenience when we would like to register several endpoints against same `AmqpMessageListenerContainerFactory`
* Register infrastructure `AmqpListenerEndpointRegistry` bean which processes `AmqpListenerEndpoint`, uses `AmqpMessageListenerContainerFactory`, and registers `AmqpMessageListenerContainer` bean.
The `AmqpListenerEndpointRegistry` bean depends (optionally) on the `AmqpDefaultConfiguration.DEFAULT_AMQP_LISTENER_CONTAINER_FACTORY_BEAN_NAME` bean provided in the target project configuration.
If neither default, nor the specific for the `AmqpListenerEndpointRegistration`, then application context fails
* Some optimization and properties order rearranging in the `AmqpMessageListenerContainer`

### DIFF
--- a/spring-amqp-client/src/main/java/org/springframework/amqp/client/config/AmqpListenerEndpoint.java
+++ b/spring-amqp-client/src/main/java/org/springframework/amqp/client/config/AmqpListenerEndpoint.java
@@ -17,16 +17,16 @@
 package org.springframework.amqp.client.config;
 
 import java.time.Duration;
+import java.util.concurrent.Executor;
 
 import org.aopalliance.aop.Advice;
 import org.jspecify.annotations.Nullable;
 
 import org.springframework.amqp.core.MessageListener;
-import org.springframework.core.task.TaskExecutor;
 
 /**
  * The configuration model to represent a {@link MessageListener}
- * with properties for target @{@link org.springframework.amqp.client.listener.AmqpMessageListenerContainer}
+ * with properties for target {@link org.springframework.amqp.client.listener.AmqpMessageListenerContainer}
  * to be registered as a bean in the application context.
  *
  * @author Artem Bilan
@@ -79,7 +79,7 @@ public interface AmqpListenerEndpoint {
 	 * @return the executor.
 	 */
 	@Nullable
-	TaskExecutor getTaskExecutor();
+	Executor getTaskExecutor();
 
 	/**
 	 * Override the {@code autoAccept} property in the {@link AmqpMessageListenerContainerFactory}.
--- a/spring-amqp-client/src/main/java/org/springframework/amqp/client/config/AmqpListenerEndpointRegistry.java
+++ b/spring-amqp-client/src/main/java/org/springframework/amqp/client/config/AmqpListenerEndpointRegistry.java
@@ -59,6 +59,11 @@ public class AmqpListenerEndpointRegistry implements ApplicationContextAware {
 		this.beanFactory = (DefaultListableBeanFactory) this.applicationContext.getBeanFactory();
 	}
 
+	/**
+	 * Register {@link AmqpMessageListenerContainer} beans based on the provided {@link AmqpListenerEndpointRegistration}.
+	 * @param registration the registration with endpoints to process.
+	 * @return this registry.
+	 */
 	public AmqpListenerEndpointRegistry registerListenerEndpoints(AmqpListenerEndpointRegistration registration) {
 		AmqpMessageListenerContainerFactory containerFactory = registration.getContainerFactory();
 		if (containerFactory == null) {
@@ -76,6 +81,12 @@ public class AmqpListenerEndpointRegistry implements ApplicationContextAware {
 		return this;
 	}
 
+	/**
+	 * Register {@link AmqpMessageListenerContainer} bean based on the provided {@link AmqpListenerEndpoint}.
+	 * The default {@link AmqpMessageListenerContainerFactory} is required in this case.
+	 * @param endpoint the endpoint to process from.
+	 * @return this registry.
+	 */
 	public AmqpListenerEndpointRegistry registerListenerEndpoint(AmqpListenerEndpoint endpoint) {
 		AmqpMessageListenerContainerFactory containerFactory = this.defaultAmqpMessageListenerContainerFactory;
 		Assert.state(containerFactory != null, "No default listener container factory available.");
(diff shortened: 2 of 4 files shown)


spring-amqp-client/src/main/java/org/springframework/amqp/client/config/AmqpMessageListenerContainerFactory.java
```diff
@@ -0,0 +1,200 @@
...
+	 * @see AmqpMessageListenerContainer#setReceiveTimeout(Duration)
+	 */
+	public void setReceiveTimeout(Duration receiveTimeout) {
+		this.receiveTimeout = receiveTimeout;
+	}
+
+	/**
+	 * The default {@code gracefulShutdownPeriod} for container instances created by this factory.
+	 * Can be overridden by {@link AmqpListenerEndpoint#getGracefulShutdownPeriod()}.
+	 * @param gracefulShutdownPeriod the timeout for shoutdown.
```
cppwfs: `shutdown` instead of `shoutdown`

spring-amqp-client/src/main/java/org/springframework/amqp/client/config/AmqpListenerEndpointRegistry.java
```diff
@@ -0,0 +1,115 @@
...
+						"and no default listener container factory available.");
+
+		for (AmqpListenerEndpoint endpoint : registration.getEndpoints()) {
+			registerListenerEndpoint(containerFactory, endpoint);
+		}
+
+		return this;
+	}
+
+	public AmqpListenerEndpointRegistry registerListenerEndpoint(AmqpListenerEndpoint endpoint) {
```
cppwfs, line 90: Probably want javadocs for public methods.

spring-amqp-client/src/main/java/org/springframework/amqp/client/config/AmqpListenerEndpointRegistry.java
```diff
@@ -0,0 +1,115 @@
...
+
+	private void registerListenerContainer(AmqpMessageListenerContainer container, AmqpListenerEndpoint endpoint) {
+		String id = endpoint.getId();
+		if (id == null) {
+			id = endpoint.getMessageListener().getClass().getName() + ".listenerContainer";
+		}
+
+		String generatedBeanName = id;
+		int counter = 0;
+		while (counter == -1 || this.beanFactory.containsBean(id)) {
```
cppwfs: Why do we have a check for `counter == -1`?  `-1` can't be reached.

spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpMessageListenerContainer.java
```diff
@@ -218,7 +217,7 @@ public void setConsumersPerQueue(int consumersPerQueue) {
 		this.consumersPerQueue = consumersPerQueue;
 	}
 
-	public void setErrorHandler(@Nullable ErrorHandler errorHandler) {
+	public void setErrorHandler(ErrorHandler errorHandler) {
```
cppwfs, line 220: Why was the `@Nullable` removed?

spring-amqp-client/src/main/java/org/springframework/amqp/client/config/AmqpListenerEndpoint.java
```diff
@@ -0,0 +1,119 @@
...
+
+import org.aopalliance.aop.Advice;
+import org.jspecify.annotations.Nullable;
+
+import org.springframework.amqp.core.MessageListener;
+import org.springframework.core.task.TaskExecutor;
+
+/**
+ * The configuration model to represent a {@link MessageListener}
+ * with properties for target @{@link org.springframework.amqp.client.listener.AmqpMessageListenerContainer}
```
cppwfs: `@` is outside of `{@link org.springframework.amqp.client.listener.AmqpMessageListenerContainer}`

spring-amqp-client/src/main/java/org/springframework/amqp/client/config/AmqpMessageListenerContainerFactory.java
```diff
@@ -0,0 +1,200 @@
...
+
+import org.springframework.amqp.client.AmqpConnectionFactory;
+import org.springframework.amqp.client.listener.AmqpMessageListenerContainer;
+import org.springframework.amqp.utils.JavaUtils;
+import org.springframework.core.task.TaskExecutor;
+import org.springframework.util.ErrorHandler;
+
+/**
+ * The configuration factory to produce {@link AmqpMessageListenerContainer} instance
+ * based on the provided.
```
cppwfs: Probably should be ended like `based on the provided {@link AmqpConnectionFactory}.`

spring-amqp-client/src/main/java/org/springframework/amqp/client/config/AmqpMessageListenerContainerFactory.java
```diff
@@ -0,0 +1,200 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.amqp.client.config;
+
+import java.time.Duration;
+import java.util.Arrays;
+import java.util.concurrent.Executor;
```
cppwfs, line 21: We can remove this import and use the FQCN for  line 108.

spring-amqp-client/src/test/java/org/springframework/amqp/client/EnableAmqpTests.java
```diff
@@ -0,0 +1,255 @@
...
+		listenerContainer.start();
+
+		this.amqpClient.to(TEST_QUEUE1)
+				.body("test_data")
+				.send();
+
+		Message message = receivedMessages.poll(10, TimeUnit.SECONDS);
+		assertThat(message)
+				.extracting(Message::getBody)
+				.isEqualTo("test_data".getBytes());
```
cppwfs, line 136: I know this is a test but should we use `StandardCharsets.UTF_8` when getting bytes?

cppwfs (changes requested):
This is awesome stuff!
Great work, just a couple of comments.

spring-amqp-client/src/test/java/org/springframework/amqp/client/EnableAmqpTests.java
artembilan: Looks like a default charset for this method falls back exactly to the `sun.nio.cs.UTF_8.INSTANCE`.
So, I don't see a reason specifying it in our tests.

spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpMessageListenerContainer.java
artembilan: That was originally a copy/paste from some other place.
I don't see reason allowing to provide null to our properties.

PR #3356, opened 2026-02-21
feat: add classic queue type option to QueueBuilder

phonty29:
"QueueBuilder" sets the default queue type if no argument is provided. Judging by the fact that "classic()" option doesn't exist, it is believed that the default type is always "classic". But it could be any.

In my project, SRE set the default type to "quorum". And I had to override with "withArguments" method. I think that we also need more convenient "classic" method.

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/core/QueueBuilder.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/core/QueueBuilder.java
@@ -246,7 +246,7 @@ public final class QueueBuilder extends AbstractBuilder {
 	/**
 	 * Set the queue argument to declare a queue of type 'classic' instead of default type.
 	 * @return the builder.
-	 * @since 2.2.2
+	 * @since 3.2.10
 	 */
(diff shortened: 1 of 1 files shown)


spring-amqp/src/main/java/org/springframework/amqp/core/QueueBuilder.java
```diff
@@ -243,6 +243,15 @@ public QueueBuilder singleActiveConsumer() {
 		return withArgument("x-single-active-consumer", true);
 	}
 
+	/**
+	 * Set the queue argument to declare a queue of type 'classic' instead of default type.
+	 * @return the builder.
+	 * @since 2.2.2
```
artembilan: Can we understand the choice of the version here, please?
What we can accept is a back-port down to `3.2.10` according to our support policy: https://spring.io/projects/spring-amqp#support

phonty29: Changed the version

Issue #3366, opened 2026-03-03
DirectMessageListenerContainer discards listener when TimeoutException occurs


AlexSischin:
**In what version(s) of Spring AMQP are you seeing this issue?**

4.1.0-M2, 3.2.9, 3.2.8

**Describe the bug**

`DirectMessageListenerContainer` permanently discards listener when using `CachingConnectionFactory` and a `TimeoutException` occurs in `connect` method:

```
2026-03-04 00:40:16.348 ERROR [org.springframework.amqp.rabbit.RabbitListenerEndpointContainer#0-consumerMonitor-1] o.s.s.s.TaskUtils$LoggingErrorHandler - Unexpected error occurred in scheduled task
org.springframework.amqp.AmqpTimeoutException: java.util.concurrent.TimeoutException
        at org.springframework.amqp.rabbit.support.RabbitExceptionTranslator.convertRabbitAccessException(RabbitExceptionTranslator.java:72)
        at org.springframework.amqp.rabbit.connection.AbstractConnectionFactory.createBareConnection(AbstractConnectionFactory.java:629)
        at org.springframework.amqp.rabbit.connection.CachingConnectionFactory.createConnection(CachingConnectionFactory.java:745)
        at org.springframework.amqp.rabbit.connection.CachingConnectionFactory.createBareChannel(CachingConnectionFactory.java:682)
        at org.springframework.amqp.rabbit.connection.CachingConnectionFactory.getCachedChannelProxy(CachingConnectionFactory.java:655)
        at org.springframework.amqp.rabbit.connection.CachingConnectionFactory.getChannel(CachingConnectionFactory.java:544)
        at org.springframework.amqp.rabbit.connection.CachingConnectionFactory$ChannelCachingConnectionProxy.createChannel(CachingConnectionFactory.java:1509)
        at org.springframework.amqp.rabbit.listener.DirectMessageListenerContainer.consume(DirectMessageListenerContainer.java:813)
        at org.springframework.amqp.rabbit.listener.DirectMessageListenerContainer.doConsumeFromQueue(DirectMessageListenerContainer.java:780)
        at org.springframework.amqp.rabbit.listener.DirectMessageListenerContainer.restartConsumer(DirectMessageListenerContainer.java:610)
        at org.springframework.amqp.rabbit.listener.DirectMessageListenerContainer.lambda$startMonitor$0(DirectMessageListenerContainer.java:533)
[...]

PR #3367, opened 2026-03-07
Fix listener recovery issues (#3366)

AlexSischin:
Fixes https://github.com/spring-projects/spring-amqp/issues/3366

This commit contains several fixes:
1. `CachingConnectionFactory.createConnection` creates new connection if existing one is stale.
2. `DirectMessageListenerContainer` recovers from `AmqpTimeoutException` if it happens during restart.
3. [_unplanned_] Added proper handling of `AmqpTimeoutException` during `DirectMessageListenerContainer` start.
4. [_unplanned_] Fixed consumer duplication when `DirectMessageListenerContainer.startConsumers` fails on channel creation with most exceptions.
5. [_unplanned_] Fixed potential problem when `SimpleConnection.isOpen` does not throw `AutoRecoverConnectionNotCurrentlyOpenException` when expected.
6. [_unplanned_] Fixed `lastRestartAttempt` not being updated during `DirectMessageListenerContainer` initialization attempts.

### DIFF
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainerIntegrationTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainerIntegrationTests.java
@@ -38,6 +38,7 @@ import org.aopalliance.intercept.MethodInterceptor;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Named;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestInfo;
 import org.junit.jupiter.params.ParameterizedTest;
@@ -79,7 +80,6 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 import static org.assertj.core.api.Assertions.fail;
 import static org.awaitility.Awaitility.await;
-import static org.junit.jupiter.api.Named.named;
 import static org.mockito.ArgumentMatchers.anyBoolean;
 import static org.mockito.ArgumentMatchers.anyMap;
 import static org.mockito.ArgumentMatchers.anyString;
@@ -1063,12 +1063,12 @@ public class DirectMessageListenerContainerIntegrationTests {
 
 		return Stream.of(
 				Arguments.of(
-						named("Bidirectional packet loss", suspendMethod),
-						named("Packets reach destination", resumeMethod)
+						Named.named("Bidirectional packet loss", suspendMethod),
+						Named.named("Packets reach destination", resumeMethod)
 				),
 				Arguments.of(
-						named("Connection closed by server", closeMethod),
-						named("Connection is available", startMethod)
+						Named.named("Connection closed by server", closeMethod),
+						Named.named("Connection is available", startMethod)
 				)
 		);
 	}
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainerMockTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainerMockTests.java
@@ -33,6 +33,7 @@ import com.rabbitmq.client.Channel;
 import com.rabbitmq.client.Consumer;
 import com.rabbitmq.client.Envelope;
 import com.rabbitmq.client.impl.recovery.AutorecoveringChannel;
+import org.junit.jupiter.api.Named;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
@@ -54,7 +55,6 @@ import org.springframework.util.backoff.FixedBackOff;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.awaitility.Awaitility.await;
-import static org.junit.jupiter.api.Named.named;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyBoolean;
 import static org.mockito.ArgumentMatchers.anyInt;
@@ -568,13 +568,13 @@ public class DirectMessageListenerContainerMockTests {
 		RuntimeException dummyCause = new RuntimeException("Test cause");
 		String msg = "Test exception";
 		return Stream.of(
-				Arguments.of(named("AmqpTimeoutException",
+				Arguments.of(Named.named("AmqpTimeoutException",
 						(Supplier<? extends Exception>) () -> new AmqpTimeoutException(msg))),
-				Arguments.of(named("AmqpConnectException",
+				Arguments.of(Named.named("AmqpConnectException",
 						(Supplier<? extends Exception>) () -> new AmqpConnectException(msg, dummyCause))),
-				Arguments.of(named("AmqpIOException",
+				Arguments.of(Named.named("AmqpIOException",
 						(Supplier<? extends Exception>) () -> new AmqpIOException(msg, dummyCause))),
-				Arguments.of(named("AutoRecoverConnectionNotCurrentlyOpenException",
+				Arguments.of(Named.named("AutoRecoverConnectionNotCurrentlyOpenException",
 						(Supplier<? extends Exception>) () -> new AutoRecoverConnectionNotCurrentlyOpenException(msg)))
 		);
 	}
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/test/RabbitMQProxy.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/test/RabbitMQProxy.java
@@ -66,7 +66,7 @@ public class RabbitMQProxy implements AutoCloseable, Runnable {
 	private volatile Thread t2cThread;
 
 	public RabbitMQProxy() {
-		this("localhost", 0, "localhost", 5672);
+		this("localhost", 5672);
 	}
 
 	public RabbitMQProxy(String targetHost, int targetPort) {
@@ -190,9 +190,11 @@ public class RabbitMQProxy implements AutoCloseable, Runnable {
 				out.write(buf, 0, n);
 				out.flush();
 			}
-		} catch (IOException ignored) {
+		}
+		catch (IOException ignored) {
 			// Any IO error closes the active connection.
-		} finally {
+		}
+		finally {
 			this.closeActiveConnections();
 		}
 	}
@@ -216,10 +218,13 @@ public class RabbitMQProxy implements AutoCloseable, Runnable {
 	}
 
 	private static void safeClose(Closeable ss) {
-		if (ss == null) return;
+		if (ss == null) {
+			return;
+		}
 		try {
 			ss.close();
-		} catch (IOException ignored) {
+		}
+		catch (IOException ignored) {
 		}
 	}
 


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
```diff
@@ -741,7 +741,7 @@ public final Connection createConnection() throws AmqpException {
 			if (this.cacheMode == CacheMode.CONNECTION) {
 				return connectionFromCache();
 			}
-			if (this.connection.target == null) {
+			if (!this.connection.isOpen()) {
```
g0ddest, line 744: The change from `this.connection.target == null` to `!this.connection.isOpen()` alters behavior for all callers of `createConnection()`.

When the underlying connection is an `AutorecoveringConnection` that is temporarily closed (mid-recovery), the call chain is: `this.connection.isOpen()` → `ChannelCachingConnectionProxy.isOpen()` (line 1593) → `SimpleConnection.isOpen()` (line 134) → throws `AutoRecoverConnectionNotCurrentlyOpenException`.

This exception is not caught in `createConnection()` and will propagate to callers. Previously, callers would silently get the existing (stale) connection back — now they get an unexpected exception.

Inside `DirectMessageListenerContainer.doConsumeFromQueue()` this is safe (all exceptions from `createConnection()` are caught and wrapped as `AmqpConnectException`). But other callers like `RabbitTemplate` may not expect this.

Consider guarding:
`if (this.connection.target == null || !this.connection.isOpen()) {` or catching `AutoRecoverConnectionNotCurrentlyOpenExceptio`n here and returning the existing connection (preserving the auto-recovery contract that `SimpleConnection.isOpen()` Javadoc describes).

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.java
```diff
@@ -869,6 +870,7 @@ protected void shutdownAndWaitOrCallback(@Nullable Runnable callback) {
 		LinkedList<SimpleConsumer> canceledConsumers = null;
 		boolean waitForConsumers = false;
 		this.consumersLock.lock();
+		this.consumersToRestart.clear();
```
g0ddest: `consumersToRestart.clear()` is placed between `lock()` and `try`. If it throws, the lock is never released. In `adjustConsumers()` (line 341) the same call is inside the try block. This should follow the same pattern:
```java
this.consumersLock.lock();
try {
this.consumersToRestart.clear();
if (this.started || this.aborted) {
```

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainerIntegrationTests.java
```diff
@@ -180,6 +190,154 @@ public void testSimple() throws Exception {
...
+		CachingConnectionFactory cf = prerequisites.cf();
+		DirectMessageListenerContainer container = prerequisites.container();
+		AtomicReference<String> expectedMessage = prerequisites.expectedMessage();
+		ThreadPoolTaskExecutor templateExecutor = prerequisites.templateExecutor();
+		RabbitTemplate template = prerequisites.template();
+
+		// Simulate network problems during container start
+		startProblem.accept(rabbitMQProxy);
+		container.start();
+		Thread.sleep(10000);
```
g0ddest: `Thread.sleep(10000)` makes the test slow (2 tests × 2 parameter sets = 40s of sleeping) and fragile on CI. Consider replacing with Awaitility-based polling for the desired state, e.g.:

```java
await().atMost(Duration.ofSeconds(15))
.until(() -> /* consumers detected as failed */);
```

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/test/RabbitMQProxy.java
```diff
@@ -0,0 +1,210 @@
...
+				this.clientSocket = incoming;
+				this.targetSocket = outgoing;
+
+				c2tThread = new Thread(() -> pump(incoming, outgoing), "rabbitmq-socket-proxy-c2t");
+				t2cThread = new Thread(() -> pump(outgoing, incoming), "rabbitmq-socket-proxy-t2c");
+				c2tThread.setDaemon(true);
+				t2cThread.setDaemon(true);
+				c2tThread.start();
+				t2cThread.start();
+			} catch (IOException e) {
```
g0ddest: Minor code style issues — Spring projects use `}\ncatch (` (catch on a new line) rather than `} catch` (. Also, `run()` references `startedLatch` an `serverSocket` without `this.` prefix, inconsistent with the rest of the class.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainerMockTests.java
```diff
@@ -382,6 +400,214 @@ void monitorTaskThreadName() {
...
+		);
+	}
+
+	private static AutoRecoverConnectionNotCurrentlyOpenException buildAutoRecoverConnectionNotCurrentlyOpenException(
+			String message
+	) {
+		try {
+			var constructor = AutoRecoverConnectionNotCurrentlyOpenException.class
+					.getDeclaredConstructor(String.class);
+			constructor.setAccessible(true);
```
g0ddest: `buildAutoRecoverConnectionNotCurrentlyOpenException` uses reflection to access a package-private constructor. This works but is fragile — if the constructor signature changes, it will fail at runtime with an opaque `NoSuchMethodException`. Consider using `Mockito.mock()` or making the constructor accessible from the test package.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/test/RabbitMQProxy.java
AlexSischin: Thanks, fixed both issues)

artembilan: Please, stop saying `thanks` and `fixed` on every single comment.
It is really counterproductive  for both side.
We see an agreement with the next change and don't need extra comment to confirm that.
You would not to spend time commenting on each review.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainerMockTests.java
artembilan: @g0ddest ,

I know that might sound offending, but with all the AI trends these days, it really feels like your review has incorporated some AI agent help.

This is really typical for us to use reflection in our test code against our own production code.
Therefore, what is "runtime" in your meaning in this context.
How is that possible that our own class in our own tests would lead to `NoSuchMethodException`?
That is from where I derivative an AI review impact.
The context of the framework and its tests is lost.

My apologies if that is really your own review.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
artembilan: @g0ddest ,

The analysis and suggestion don't enhance each other.
The  `if (this.connection.target == null || !this.connection.isOpen()) {` is exactly the same as just `if (!this.connection.isOpen()) {`.
See that `ChannelCachingConnectionProxy.isOpen()`  impl:
```
		@Override
		public boolean isOpen() {
			Connection targetToCheck = this.target;
			return targetToCheck != null && targetToCheck.isOpen();
		}
```
So, why do we need extra null check in our `if` when it is done in the method we call?
The suggestion about `AutoRecoverConnectionNotCurrentlyOpenException` is good.
While it is unlikely situation with Spring AMQP, but that would be OK to guard on it anyway.

(This is another argument to my impression that review has done with AI)

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainerIntegrationTests.java
artembilan: I do agree that we don't need to sleep for 10 seconds.
Looks like the next like is a good async barrier to synchronize on.

(Another feeling that review has done with AI: there is a suggestion which is already there in the code. So, the whole context of the code is lost)

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainerIntegrationTests.java
```diff
@@ -180,6 +190,154 @@ public void testSimple() throws Exception {
...
+		// Stop simulating network problems. Container should recover and process messages.
+		resolveProblem.accept(rabbitMQProxy);
+		assertThat(consumersOnQueue(Q1, 1)).isTrue();
+		assertThat(consumersOnQueue(Q2, 1)).isTrue();
+		assertThat(activeConsumerCount(container, 2)).isTrue();
+		expectedMessage.set("foo");
+		template.convertAndSend(Q1, "foo");
+		await().until(() -> Objects.isNull(expectedMessage.get()));
+		expectedMessage.set("bar");
+		template.convertAndSend(Q2, "bar");
```
artembilan: Please, don't use `foo/bar` words in the code.
I know we have a lot of them, but that does not mean we need to introduce more.
We develop here open source, therefore code has to be as professional as possible to purse a goal of great example for community.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainerIntegrationTests.java
```diff
@@ -918,5 +1112,4 @@ public String createConsumerTag(String queue) {
 		}
 
 	}
-
```
artembilan, line 921: Please, revert.
Every member in the class (including last method of the last inner class) has to be surrounded with blank lines.
That way it is much easier to read the code and when we add more methods, it would not cause extra line changes.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainerMockTests.java
```diff
@@ -382,6 +400,214 @@ void monitorTaskThreadName() {
 				.startsWith("aBean-consumerMonitor");
 	}
 
+	@MethodSource("transientExceptions")
+	@ParameterizedTest
+	public void keepsConnectionAttemptsRegardlessExceptionAtChannelCreationAtStart(
+			Supplier<? extends Exception> exceptionSupplier
+	) throws Exception {
+		ConnectionFactory connectionFactory = mock(ConnectionFactory.class);
```
artembilan: When code block head is multi-line (method declaration, `if`, loops), then we prefer to put a blank line before that code block body.
So, in this case I want you to press Enter key just after that opening brace.
In addition: consider to use simplified Mockito api. 
Just `ConnectionFactory connectionFactory = mock();` is enough.

All of that would make the code more readable.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainerMockTests.java
artembilan: I think `mock()` is sufficient in this context. Not saying that it is better than direct reflection (how exactly Mockito creates those proxies if not via reflection?), but that definitely would make our code shorter.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainerMockTests.java
```diff
@@ -382,6 +400,214 @@ void monitorTaskThreadName() {
...
+			return constructor.newInstance(message);
+		} catch (Exception e) {
+			throw new RuntimeException(e);
+		}
+	}
+
+	private void assertActiveConsumerCount(AbstractMessageListenerContainer container, int expected) {
+		String propertyPath = "consumers";
+		Collection<?> consumers = TestUtils.getPropertyValue(container, propertyPath);
+		if (consumers == null) {
```
artembilan: I think these null checks are redundant.
I also think that these support methods have to be `static`.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/test/RabbitMQProxy.java
```diff
@@ -0,0 +1,215 @@
...
+/**
+ * RabbitMQ proxy for simulating network issues.
+ * <p>
+ * Accepts only a single connection. Can be started and closed multiple times.
+ * </p>
+ *
+ * @author Alexei Sischin
+ * @since 4.1.0
+ */
+public class RabbitMQProxy implements AutoCloseable, Runnable {
```
artembilan, line 40: I wonder if we really need so complex solution with network interaction.
Cannot we just simulate failures via mocks on the connection objects?

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/test/MockingThrowableUtils.java
```diff
@@ -0,0 +1,50 @@
...
+ * <p>
+ * It is useful when one needs to create an instance of {@link Throwable} with no public constructors or factories.
+ * In other cases it is recommended to create a real object because {@link Throwable}'s may be processed
+ * by logger in unexpected ways.
+ * </p>
+ *
+ * @author Alexei Sischin
+ * @since 4.1.0
+ */
+public final class MockingThrowableUtils {
```
artembilan: I'm not sure that we need this utility at all.
It really can be achieve with plain `Mockito`, and it really not always a case when we need all those properties on the target exception.
Can you revise this, please, in favor of local mock API in that test?
Plus, also revise if we definitely need to stub all those properties .

Why just `mock(AutoRecoverConnectionNotCurrentlyOpenException.class)` is not enough for that supplier in the test configuration?

AlexSischin: Just mocking is not enough because logger needs those properties. Well, except `.getMessage()`: I added it because it is commonly used in general. And I moved this code to a separate class because if logger implementation changes, this way it will be easier to update all mocks at once.

artembilan: OK. We have a `TestUtils` for that matter in the shared module as a place where this could live.
But again: doesn't looks like we need it at all.
Let's just make `AutoRecoverConnectionNotCurrentlyOpenException` ctor as `public`!

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/test/RabbitMQProxy.java
AlexSischin: I genuinely find this solution the most simple. Integration tests use a fully fledged RabbitMQ server and a real client. Adding a proxy between them in order to simulate network issues looks like the most straightforward approach. And it is less prone to mock misconfiguration.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainerIntegrationTests.java
AlexSischin: These tests verify that container doesn't abandon consumers on the second, third or fourth connection attempt. And this takes time, especially against a real rabbit with network involved. I cannot set very strict timeouts because test may randomly fail due to network or CPU lag. So, the test needs to wait for a few seconds anyway.

`await().until(container::isActive);` - this barrier is just a container state sanity check. `isActive` returns `true` even if consumer initialization has failed (the test proves it). I don't know if it is intended or not.

It would be better, however, to count connection attempts directly, and not just wait for a magic time. But it is tricky because the only state that could be used for this purpose is `lastRestartAttempt`, which is not updated if restart attempts happen during initialization.

For this purpose, I am also going to fix the issue with `lastRestartAttempt`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
g0ddest: I agree, the null guard looks like redundant in the case.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainerMockTests.java
g0ddest: It was just a style proposal, I really try to evade direct reflection in my code. Using reflection, as previously said is fragile, don't give you an instant feedback on class change on compile stage. Also it is breaking the test principle when when behavior is testing, not the internals.  But as I previously said it is just a test robustness note.

I understand your impression of ai based review, I really used it  to have my comments to be better formulated. I wish I hadn't done this, judging by your reaction, although I see that you more or less agree with the comments.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
AlexSischin: I did not guard `AutoRecoverConnectionNotCurrentlyOpenException` on purpose. I checked all the usages of `CachingConnectionFactory.createConnection()` and `ConnectionFactory#createConnection` and most of them currently would fail on the same exception, for instance, when calling `Connection.createChannel` right after creating connection. Or their behavior would not change at all. Moreover, other container implementations would fail on the same exception as well, so it doesn't break the base interface contract.

What could potentially be affected when using `CachingConnectionFactory`:
* `RabbitUtils.getMaxFrame` - this method will now return -1, meaning that `frame_max` cannot be determined, which is valid because the connection is not open.
* `RabbitTemplate.invoke` - it simply won't execute `RabbitUtils.closeConnection(connection)` any more, which is an improvement since the connection is auto recoverable.

Both examples may have other containers injected, which would throw the error anyway.

Am I missing something?

g0ddest: I understand that this is a bit outside the scope of the issue and the probability of the situation is too small, so I suggest you ignore this comment.

Issue #3384, opened 2026-03-18
EnableAmqp is not semantically equivalent to EnableRabbit


snicoll:
I am working on integrating AMQP 1.0 Generic in Spring Boot. In doing so, I am trying to align with [`RabbitAnnotationDrivenConfiguration`](https://github.com/spring-projects/spring-boot/blob/568b2fdb4bd2ae7068cb416b0ab0a04f72ea3338/module/spring-boot-amqp/src/main/java/org/springframework/boot/amqp/autoconfigure/RabbitAnnotationDrivenConfiguration.java#L49).

I know that the Javadoc of `@EnableAmqp` has a different scope than `@EnableRabbit` but it is problematic for the following reasons:

* I am trying to find something that'd match "Enable Rabbit listener annotated endpoints that are created under the cover by a RabbitListenerContainerFactory" for AMQP 1.0.  This boils down to a bean post processor and a registry. I'd need something that does that onlty.
* `RabbitListenerConfigUtils.RABBIT_LISTENER_ANNOTATION_PROCESSOR_BEAN_NAME` is a constant that we can use to backoff if the BPP is already created. There's no such equivalent for AMQP.

Perhaps the biggest problem of all is the creation of the proton client with what seems to attempt to implement `@ConditionalOnMissingBean`. Foundational projects should not attempt to do that as they don't have the necessary lifecycle to do it reliably. 

Can `@EnableAmqp` be reworked to be equivalent to `@EnableRabbit`?

snicoll:
Not the auto-configuration, the creation of the infrastructure. See [the RabbitMQ version](https://github.com/spring-projects/spring-boot/blob/16639dfb947c456c29054971f8a5f205135a7484/module/spring-boot-amqp/src/main/java/org/springframework/boot/amqp/autoconfigure/RabbitAnnotationDrivenConfiguration.java#L148-L153) for another example.

PR #3385, opened 2026-03-19
GH-3384: Rework `@EnableAmqp` just for `@AmqpListener`

artembilan:
Fixes: https://github.com/spring-projects/spring-amqp/issues/3384

* Remove ProtonJ `Client` as required bean from the `SingleAmqpConnectionFactory`: if not provided, create one internally.
Therefore, remove bean definition for the `Client` from the `AmqpDefaultConfiguration`
* Remove `@EnableAmqp` attributes since they are out of use from now on
* Expose `AmqpDefaultConfiguration.AMQP_LISTENER_ANNOTATION_PROCESSOR_BEAN_NAME` for an `AmqpListenerAnnotationBeanPostProcessor`
* Adjust tests to not check for the ProtonJ `Client` bean
* Fix `amqp10-client.adoc` to not talk about removed already ProtonJ `Client` bean, and explain how it is not mandatory for the `SingleAmqpConnectionFactory`

### DIFF
--- a/spring-amqp-client/src/main/java/org/springframework/amqp/client/SingleAmqpConnectionFactory.java
+++ b/spring-amqp-client/src/main/java/org/springframework/amqp/client/SingleAmqpConnectionFactory.java
@@ -43,6 +43,8 @@ public class SingleAmqpConnectionFactory implements AmqpConnectionFactory, Dispo
 
 	private final Client protonClient;
 
+	private final boolean internalClient;
+
 	private String host = "localhost";
 
 	private int port = -1;
@@ -54,6 +56,7 @@ public class SingleAmqpConnectionFactory implements AmqpConnectionFactory, Dispo
 	 */
 	public SingleAmqpConnectionFactory() {
 		this.protonClient = Client.create();
+		this.internalClient = true;
 	}
 
 	/**
@@ -62,6 +65,7 @@ public class SingleAmqpConnectionFactory implements AmqpConnectionFactory, Dispo
 	 */
 	public SingleAmqpConnectionFactory(Client protonClient) {
 		this.protonClient = protonClient;
+		this.internalClient = false;
 	}
 
 	public SingleAmqpConnectionFactory setHost(String host) {
@@ -137,6 +141,10 @@ public class SingleAmqpConnectionFactory implements AmqpConnectionFactory, Dispo
 		if (connectionToClose != null) {
 			connectionToClose.close();
 			this.connection = null;
+		}
+
+		if (this.internalClient) {
+			this.protonClient.close();
 		}
 	}
 


spring-amqp-client/src/main/java/org/springframework/amqp/client/SingleAmqpConnectionFactory.java
```diff
@@ -25,31 +25,23 @@
...
-import org.springframework.beans.BeansException;
-import org.springframework.beans.factory.BeanFactory;
-import org.springframework.beans.factory.BeanFactoryAware;
 import org.springframework.beans.factory.DisposableBean;
 
 /**
  * The {@link AmqpConnectionFactory} implementation to hold a single, shared {@link Connection} instance.
- * If instance is created without a {@link Client}, it will be resolved from the {@link BeanFactory} on demand
- * internally from the {@link #getConnection()} call.
+ * If instance is created without a {@link Client}, a default {@link Client#create()} is used internally.
```
snicoll: The second link could be "a {@link Client#create() default client} is used internally".

spring-amqp-client/src/main/java/org/springframework/amqp/client/SingleAmqpConnectionFactory.java
```diff
@@ -58,20 +50,20 @@ public class SingleAmqpConnectionFactory implements AmqpConnectionFactory, BeanF
 	private volatile @Nullable Connection connection;
 
 	/**
-	 * Create an instance based on a {@link Client} bean resolved from the {@link BeanFactory}.
+	 * Create an instance based on a default {@link Client#create()}.
```
snicoll: Ditto.

spring-amqp-client/src/main/java/org/springframework/amqp/client/SingleAmqpConnectionFactory.java
```diff
@@ -25,31 +25,23 @@
...
 	private final Lock instanceLock = new ReentrantLock();
 
 	private ConnectionOptions connectionOptions = new ConnectionOptions();
 
-	@SuppressWarnings("NullAway.Init")
-	private Client protonjClient;
-
-	@SuppressWarnings("NullAway.Init")
-	private BeanFactory beanFactory;
+	private final Client protonjClient;
```
snicoll: I've seen the use of `protonClient` and `protonjClient` but I wonder if the J is very useful in this context. Perhaps name that `protonClient`everywhere?

spring-amqp-client/src/main/java/org/springframework/amqp/client/SingleAmqpConnectionFactory.java
```diff
@@ -25,31 +25,23 @@
...
- * If instance is created without a {@link Client}, it will be resolved from the {@link BeanFactory} on demand
- * internally from the {@link #getConnection()} call.
+ * If instance is created without a {@link Client}, a default {@link Client#create()} is used internally.
  *
  * @author Artem Bilan
  *
  * @since 4.1
  */
-public class SingleAmqpConnectionFactory implements AmqpConnectionFactory, BeanFactoryAware, DisposableBean {
+public class SingleAmqpConnectionFactory implements AmqpConnectionFactory, DisposableBean {
```
snicoll, line 38: If you create the client internally, I think you should close it in the `DisposableBean` implementation. If it's provided to you, I suppose the caller is responsible of closing it. 

Perhaps you could change the constructor so that you take a `ClientOptions`. This way, it's obvious you create the client, and you are responsible of closing it. Someone who insists to provide a client should then be responsible of closing it.

spring-amqp-client/src/test/java/org/springframework/amqp/client/listener/AmqpListenerAnnotationTests.java
```diff
@@ -221,19 +220,19 @@ MessageConverter jsonMessageConverter() {
 			return new JacksonJsonMessageConverter();
 		}
 
+		@Bean
+		AmqpConnectionFactory amqpConnectionFactory() {
+			return new SingleAmqpConnectionFactory()
```
snicoll, line 225: `new SingleAmqpConnectionFactory(new ClientOptions())` is a bit gross so I suppose keeping the no-arg constructor that does that is better (cfr. my comment above).

Another constructor for the ClientOptions allow to keep the convenience of closing the client while providing the necessary configuration. I am not 100% sure it's worth it though as Spring Boot would expose the client as a bean.

artembilan: Yes, it is not.
We close our connection in this `SingleAmqpConnectionFactory.destroy()`.
And maximum what default client does is it iterates over connections created with it and close it.
Therefore I see no reason in taking care about internal client instance clean up since one-to-one connection here is going to be closed anyway.
Thus, no need in expose ctor for `ClientOptions`.
I thought about that, but it is so limited that if we go with custom options, we would simply go with custom `Client`.
It is more obvious for users to have `Client` configuration separated from connection factory and have it just injected.
The Inversion of Control in its best 🤷

snicoll: I am trying to understand what you're saying. Do you mean that because you close the connection you don't need to close the client? Isn't that an internal implementation detail of the client?

If the client doesn't need to be closed, I don't understand the benefit of exposing it as a bean before...

artembilan: Yes. It is an internal detail of the client.
And I see that default one (we use internally) just closes a connection we are managing here in this factory.
Here is that client code:
```
    public synchronized Future<Client> closeAsync() {
        if (CLOSED_UPDATER.compareAndSet(this, 0, 1)) {
            if (connections.isEmpty()) {
                closedFuture.complete(this);
            } else {
                // Make a copy as the connection close will modify the connections
                // Map as each connection closes and removes itself from the container.
                List<Connection> connectionsView = new ArrayList<>(connections.values());
                connectionsView.forEach((connection) -> {
                    try {
                        connection.closeAsync();
                    } catch (Throwable ignored) {
                        LOG.trace("Error while closing connection, ignoring", ignored);
                    }
                });
            }
        }

        return closedFuture;
    }
```
So, nothing more than closing managed connections.
And since we close that one ourselves in our factory on destroy, calling that client `close()` is redundant.

Or do you mean that independently of those client internal details we shouldn't care about them and still be as robust as possible in our code to guard it from possible additions into that `Client.close()` impl even if right now it is idempotent?

src/reference/antora/modules/ROOT/pages/amqp10-client.adoc
```diff
@@ -45,7 +45,8 @@ Client protonClient() {
 ----
 
 The same `Client` can be used for connecting to different brokers, the connection settings must be provided on a specific connection.
-See also <<amqp10-client-connection-factory>> and <<amqp10-annotation-configuration>> below.
+However, this bean is not mandatory, `SingleAmqpConnectionFactory` creates instance internally.
```
cppwfs, line 48: Should we say `client` instead of `bean`?

artembilan: I made this sentence in the context of the `Client` bean declared before this paragraph.
So, I feel like saying "bean" makes sense here.

spring-amqp-client/src/test/java/org/springframework/amqp/client/listener/AmqpListenerAnnotationTests.java
snicoll: The latter. Let's assume the qpid client evolves in such a way that something else needs to be closed. 

If it does, you may have to call close so we can keep things as they are or adapt if necessary in the future. However the client implements Closable so it kinds of inferring in its api it should be closed. 

In practice, I don't know if that is that important.

cppwfs (approved):
LGTM

If @snicoll agrees I'll merge.

PR #3407, opened 2026-04-16
GH-2305: Add immediate scale-down to SimpleMessageListenerContainer

### LINKED ISSUE #2305: SimpleMessageListenerContainer should be offer more aggressive scale down options [AMQP-745]
**[Yannick](https://jira.spring.io/secure/ViewProfile.jspa?name=yenki)** opened **[AMQP-745](https://jira.spring.io/browse/AMQP-745?redirect=false)** and commented

This is somehow related to #2259

Unless we set 'consecutiveIdleTrigger' to 1 and an ridiculously low receiveTimeout, it's very hard to have consumers removed from a queue having reasonable amount of activity given the current SimpleMessageListenerContainer implementation.

For instance I have a queue with 100 msg/s and a consumer handling 1 msg each millisecond. This means that one consumer should easily be able to sustain the load all by itself.
Now even if I set 'consecutiveIdleTrigger' to 1, I could theoretically have 100 consumers and none of them would be removed. That make it 1000 If I were to use the default 'consecutiveIdleTrigger' value of 10.

#2259 sounds like a very good way to solve the problem but more short term I think having a base class with template and/or overridable methods would do the trick for us to hook and inject a more aggressive scale down strategy. I can contribute something I just don't know exactly which solution is more likely to be merged upstream. Please advise.

---

**Affects:** 1.6.8

**Reference URL:** #2259

joyfulviper:
Fixes https://github.com/spring-projects/spring-amqp/issues/2305

When `setConcurrentConsumers()` is invoked dynamically (for example, from within a listener) to scale down, the current implementation arbitrarily cancels a consumer via `adjustConsumers()`, which may kill a consumer that is still actively processing a message while another consumer that just finished keeps running.

This change introduces an `immediateScaleDown` flag. When enabled:

* Scale-down requests are deferred to `checkAdjust()` so that a consumer
  stops itself only after it has finished processing the current message.
* `consecutiveIdleTrigger` and `stopConsumerMinInterval` are bypassed for
  scale-down decisions — the intent is immediate reaction.
* A `markedForStop` flag lets the consumer exit its main loop cleanly,
  preventing it from processing an additional prefetched message between
  the ack and the basicCancel.

Scale-up behavior is unchanged.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
@@ -89,6 +89,7 @@ import org.springframework.util.backoff.BackOffExecution;
  * @author Java4ye
  * @author Thomas Badie
  * @author Jeongjun Min
+ * @author DoYeon Kim
  *
  * @since 1.0
  */
@@ -146,6 +147,8 @@ public class SimpleMessageListenerContainer extends AbstractMessageListenerConta
 
 	private boolean enforceImmediateAckForManual;
 
+	private boolean immediateScaleDown;
+
 	private volatile int concurrentConsumers = 1;
 
 	private volatile @Nullable Integer maxConcurrentConsumers;
@@ -196,7 +199,16 @@ public class SimpleMessageListenerContainer extends AbstractMessageListenerConta
 			int delta = this.concurrentConsumers - concurrentConsumers;
 			this.concurrentConsumers = concurrentConsumers;
 			if (isActive()) {
-				adjustConsumers(delta);
+				if (this.immediateScaleDown && delta > 0) {
+					// Defer scale-down to checkAdjust() so that a consumer stops itself
+					// after it finishes processing the current message, instead of
+					// adjustConsumers() arbitrarily cancelling a consumer that may still
+					// be processing.
+					logger.debug("Immediate scale-down: deferring consumer removal to checkAdjust");
+				}
+				else {
+					adjustConsumers(delta);
+				}
 			}
 		}
 		finally {
@@ -529,6 +541,36 @@ public class SimpleMessageListenerContainer extends AbstractMessageListenerConta
 	 */
 	public void setEnforceImmediateAckForManual(boolean enforceImmediateAckForManual) {
 		this.enforceImmediateAckForManual = enforceImmediateAckForManual;
+	}
+
+	/**
+	 * Set to {@code true} to enable immediate scale-down behavior when the number of
+	 * concurrent consumers is reduced dynamically (for example, by invoking
+	 * {@link #setConcurrentConsumers(int)} or {@link #setConcurrency(String)}
+	 * from within a listener).
+	 * <p>
+	 * When enabled, a scale-down request does not arbitrarily cancel one of the existing
+	 * consumers. Instead, each consumer checks after every message whether it is in
+	 * excess and, if so, stops itself once the current message has been processed. This
+	 * avoids cancelling a consumer that is still actively processing a message while a
+	 * consumer that just finished keeps running.
+	 * <p>
+	 * When enabled, the following properties are bypassed for scale-down decisions
+	 * because the intent is immediate reaction:
+	 * {@link #setConsecutiveIdleTrigger(int) consecutiveIdleTrigger} and
+	 * {@link #setStopConsumerMinInterval(long) stopConsumerMinInterval}.
+	 * Scale-up related properties are not affected.
+	 * <p>
+	 * Has no effect unless {@link #setMaxConcurrentConsumers(int) maxConcurrentConsumers}
+	 * is also configured, because {@code checkAdjust()} is only invoked in that case.
+	 * Default is {@code false}.
+	 * @param immediateScaleDown true to enable immediate scale-down.
+	 * @since 4.1
+	 * @see #setConcurrentConsumers(int)
+	 * @see #setMaxConcurrentConsumers(int)
+	 */
+	public void setImmediateScaleDown(boolean immediateScaleDown) {
+		this.immediateScaleDown = immediateScaleDown;
 	}
 
 	/**
@@ -1297,6 +1339,8 @@ public class SimpleMessageListenerContainer extends AbstractMessageListenerConta
 
 		private boolean failedExclusive;
 
+		private volatile boolean markedForStop;
+
 		AsyncMessageProcessingConsumer(BlockingQueueConsumer consumer) {
 			this.consumer = consumer;
 			this.start = new CountDownLatch(1);
@@ -1355,7 +1399,9 @@ public class SimpleMessageListenerContainer extends AbstractMessageListenerConta
 			try {
 				initialize();
 				SimpleMessageListenerContainer.this.processorThreadsToInterrupt.add(Thread.currentThread());
-				while (isActive(this.consumer) || this.consumer.hasDelivery() || !this.consumer.cancelled()) {
+				while (!this.markedForStop
+						&& (isActive(this.consumer) || this.consumer.hasDelivery() || !this.consumer.cancelled())) {
+
 					mainLoop();
 				}
 			}
@@ -1507,14 +1553,47 @@ public class SimpleMessageListenerContainer extends AbstractMessageListenerConta
 						considerAddingAConsumer();
 						this.consecutiveMessages = 0;
 					}
+					if (SimpleMessageListenerContainer.this.immediateScaleDown) {
+						markForStopIfExcess();
+					}
 				}
 			}
 			else {
 				this.consecutiveMessages = 0;
-				if (this.consecutiveIdles++ > SimpleMessageListenerContainer.this.consecutiveIdleTrigger) {
+				if (SimpleMessageListenerContainer.this.immediateScaleDown) {
+					markForStopIfExcess();
+					this.consecutiveIdles = 0;
+				}
+				else if (this.consecutiveIdles++ > SimpleMessageListenerContainer.this.consecutiveIdleTrigger) {
 					considerStoppingAConsumer(this.consumer);
 					this.consecutiveIdles = 0;
 				}
+			}
+		}
+
+		/**
+		 * When immediate scale-down is enabled, stop this consumer itself if the
+		 * number of active consumers exceeds the configured concurrency. This runs
+		 * after a message has been processed (acked), so the consumer terminates
+		 * cleanly rather than being cancelled mid-processing by another thread.
+		 */
+		private void markForStopIfExcess() {
+			SimpleMessageListenerContainer.this.consumersLock.lock();
+			try {
+				SimpleMessageListenerContainer container = SimpleMessageListenerContainer.this;
+				if (container.consumers != null
+						&& container.consumers.size() > container.concurrentConsumers) {
+
+					this.consumer.basicCancel(true);
+					container.consumers.remove(this.consumer);
+					this.markedForStop = true;
+					if (logger.isDebugEnabled()) {
+						logger.debug("Immediate scale-down: consumer marked for stop: " + this.consumer);
+					}
+				}
+			}
+			finally {
+				SimpleMessageListenerContainer.this.consumersLock.unlock();
 			}
 		}
 
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerTests.java
@@ -194,9 +194,9 @@ public class SimpleMessageListenerContainerTests {
 	 */
 	@Test
 	public void testTxSizeAcks() throws Exception {
-		ConnectionFactory connectionFactory = mock(ConnectionFactory.class);
-		Connection connection = mock(Connection.class);
-		Channel channel = mock(Channel.class);
+		ConnectionFactory connectionFactory = mock();
+		Connection connection = mock();
+		Channel channel = mock();
 		given(connectionFactory.createConnection()).willReturn(connection);
 		given(connection.createChannel(false)).willReturn(channel);
 		final AtomicReference<Consumer> consumer = new AtomicReference<>();
@@ -246,9 +246,9 @@ public class SimpleMessageListenerContainerTests {
 	 */
 	@Test
 	public void testTxSizeAcksWIthShortSet() throws Exception {
-		ConnectionFactory connectionFactory = mock(ConnectionFactory.class);
-		Connection connection = mock(Connection.class);
-		Channel channel = mock(Channel.class);
+		ConnectionFactory connectionFactory = mock();
+		Connection connection = mock();
+		Channel channel = mock();
 		given(connectionFactory.createConnection()).willReturn(connection);
 		given(connection.createChannel(false)).willReturn(channel);
 		final AtomicReference<Consumer> consumer = new AtomicReference<>();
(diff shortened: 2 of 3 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -196,7 +199,18 @@ public void setConcurrentConsumers(final int concurrentConsumers) {
 			int delta = this.concurrentConsumers - concurrentConsumers;
 			this.concurrentConsumers = concurrentConsumers;
 			if (isActive()) {
-				adjustConsumers(delta);
+				if (this.immediateScaleDown && delta > 0) {
+					// Defer scale-down to checkAdjust() so that a consumer stops itself
+					// after it finishes processing the current message, instead of
+					// adjustConsumers() arbitrarily cancelling a consumer that may still
+					// be processing.
+					if (logger.isDebugEnabled()) {
```
artembilan: We don't need `isDebugEnabled()` if message is a string constant.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -531,6 +545,45 @@ public void setEnforceImmediateAckForManual(boolean enforceImmediateAckForManual
...
+		this.immediateScaleDown = immediateScaleDown;
+	}
+
+	/**
+	 * Return whether immediate scale-down is enabled.
+	 * @return true if immediate scale-down is enabled.
+	 * @since 4.1
+	 * @see #setImmediateScaleDown(boolean)
+	 */
+	public boolean isImmediateScaleDown() {
```
artembilan: We don't need getter.
Even if you see for other props, that doesn't mean there is any legit justification to get access to to such an internal property.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerTests.java
```diff
@@ -799,6 +800,74 @@ void listenerIsInterruptedOnUnsuccessfulShutdown() throws Exception {
...
 	}
 
+	@Test
+	public void testImmediateScaleDownGetterSetter() {
+		SimpleMessageListenerContainer container = new SimpleMessageListenerContainer();
+		assertThat(container.isImmediateScaleDown()).isFalse();
+		container.setImmediateScaleDown(true);
+		assertThat(container.isImmediateScaleDown()).isTrue();
+		container.setImmediateScaleDown(false);
+		assertThat(container.isImmediateScaleDown()).isFalse();
```
artembilan: I don't see a reason in this test since its functionality is covered by other tests.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerTests.java
```diff
@@ -799,6 +800,74 @@ void listenerIsInterruptedOnUnsuccessfulShutdown() throws Exception {
...
+	}
+
+	@Test
+	public void testImmediateScaleDownDefersConsumerRemoval() throws Exception {
+		ConnectionFactory connectionFactory = mock(ConnectionFactory.class);
+		Connection connection = mock(Connection.class);
+		Channel channel = mock(Channel.class);
```
artembilan: Consider to use `mock()` with Class arg.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerTests.java
```diff
@@ -799,6 +800,74 @@ void listenerIsInterruptedOnUnsuccessfulShutdown() throws Exception {
...
+		given(connectionFactory.createConnection()).willReturn(connection);
+		given(connection.createChannel(false)).willReturn(channel);
+		given(channel.isOpen()).willReturn(true);
+		given(connection.isOpen()).willReturn(true);
+		willAnswer(invocation -> "consumerTag")
+				.given(channel).basicConsume(anyString(), anyBoolean(), anyString(), anyBoolean(), anyBoolean(),
+						anyMap(), any(Consumer.class));
+
+		SimpleMessageListenerContainer container = new SimpleMessageListenerContainer(connectionFactory);
+		container.setQueueNames("foo");
```
artembilan: We prefer do not use `foo/bar` language in our code.
This goes to the same category as we discussed `aggressive` word with you.
The language in our code should be as official and professional as possible.
In the end this is an Open Source and Community is looking at us.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -146,6 +147,8 @@ public class SimpleMessageListenerContainer extends AbstractMessageListenerConta
 
 	private boolean enforceImmediateAckForManual;
 
+	private volatile boolean immediateScaleDown;
```
artembilan: I don't think this property is intended to be changed at runtime.
Let's remove that `volatile` which is not for properties to be set only once during Spring application context initialization!

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -531,6 +545,45 @@ public void setEnforceImmediateAckForManual(boolean enforceImmediateAckForManual
...
+	 * avoids cancelling a consumer that is still actively processing a message while a
+	 * consumer that just finished keeps running.
+	 * <p>
+	 * When enabled, the following properties are bypassed for scale-down decisions
+	 * because the intent is immediate reaction:
+	 * {@link #setConsecutiveIdleTrigger(int) consecutiveIdleTrigger} and
+	 * {@link #setStopConsumerMinInterval(long) stopConsumerMinInterval}.
+	 * Scale-up related properties are not affected.
+	 * <p>
+	 * Has no effect unless {@link #setMaxConcurrentConsumers(int) maxConcurrentConsumers}
```
artembilan, line 564: ... which is also called from the `setConcurrency(String)`.
I mean we need to adjust this Javadoc somehow to mention that as well.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -1355,7 +1412,8 @@ public void run() { // NOSONAR - line count
 			try {
 				initialize();
 				SimpleMessageListenerContainer.this.processorThreadsToInterrupt.add(Thread.currentThread());
-				while (isActive(this.consumer) || this.consumer.hasDelivery() || !this.consumer.cancelled()) {
+				while (!this.markedForStop
+						&& (isActive(this.consumer) || this.consumer.hasDelivery() || !this.consumer.cancelled())) {
```
artembilan, line 1403: Now this is going to be nit-pick just for the code style to make it more readable and easier to digest.

If block header (method declaration, loops, ifs, tries etc.) is multi-line, then body of the block should be separate with blank line.

So, in this case you need to press Enter key after this `while()` definition.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -1507,17 +1565,49 @@ private void checkAdjust(boolean receivedOk) {
...
+		 * after a message has been processed (acked), so the consumer terminates
+		 * cleanly rather than being cancelled mid-processing by another thread.
+		 */
+		private void markForStopIfExcess() {
+			SimpleMessageListenerContainer.this.consumersLock.lock();
+			try {
+				SimpleMessageListenerContainer container = SimpleMessageListenerContainer.this;
+				if (container.consumers != null
+						&& container.consumers.size() > container.concurrentConsumers) {
+					this.consumer.basicCancel(true);
```
artembilan, line 1587: Same here: the `if` is multi-line - please blank line before this line of code.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerTests.java
```diff
@@ -799,6 +800,74 @@ void listenerIsInterruptedOnUnsuccessfulShutdown() throws Exception {
...
+	public void testImmediateScaleDownDefersConsumerRemoval() throws Exception {
+		ConnectionFactory connectionFactory = mock(ConnectionFactory.class);
+		Connection connection = mock(Connection.class);
+		Channel channel = mock(Channel.class);
+		given(connectionFactory.createConnection()).willReturn(connection);
+		given(connection.createChannel(false)).willReturn(channel);
+		given(channel.isOpen()).willReturn(true);
+		given(connection.isOpen()).willReturn(true);
+		willAnswer(invocation -> "consumerTag")
+				.given(channel).basicConsume(anyString(), anyBoolean(), anyString(), anyBoolean(), anyBoolean(),
```
artembilan: I think `basicConsume` call should go to a new line to preserve much readable method chain style.
Probably then all args of `basicConsume` would go to the same line.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerTests.java
```diff
@@ -799,6 +800,74 @@ void listenerIsInterruptedOnUnsuccessfulShutdown() throws Exception {
...
+		container.afterPropertiesSet();
+		container.start();
+
+		Set<?> consumers = TestUtils.getPropertyValue(container, "consumers");
+		await().until(() -> consumers.size() == 2);
+
+		// With immediate scale-down enabled, reducing concurrentConsumers must NOT
+		// immediately cancel a consumer (that is deferred to checkAdjust()).
+		container.setConcurrentConsumers(1);
+		assertThat(consumers).hasSize(2);
```
artembilan: I don't think this is robust.
Since `checkAdjust()` is called from the other thread there is no guarantee that CPU would go immediately to the next execution in this test method just after `container.setConcurrentConsumers(1);`.
The CPU could switch to do something else and time might be spend there giving a race that `checkAdjust()` could be called on other thread before we come back here for this `assertThat()`.

We could mitigate the race condition via some `CountDownLatch` injected into this or that stub of some mock.
And ensure in the end that some method was never invoked.
I  know, so much reflection has to be involved.
I don't see any other way...

Looks like `considerStoppingAConsumer()` is never called in case of `immediateScaleDown`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java
```diff
@@ -907,11 +960,13 @@ private void considerStoppingAConsumer(BlockingQueueConsumer consumer) {
 		try {
 			if (this.consumers != null && this.consumers.size() > this.concurrentConsumers) {
 				long now = System.currentTimeMillis();
-				if (this.lastConsumerStopped + this.stopConsumerMinInterval < now) {
+				if (this.immediateScaleDown || this.lastConsumerStopped + this.stopConsumerMinInterval < now) {
```
artembilan: The `considerStoppingAConsumer()` is called only once, in that `checkAdjust()`.
And looks like it is called only in case of `immediateScaleDown`.
Therefore, I think changes in this method are redundant.

artembilan (changes requested):
This looks promising.
Not saying that I have too many concerns, but still there is some review which might be helpful.

Thanks

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerTests.java
joyfulviper: Could you clarify what you mean? The current code already uses mock(ConnectionFactory.class) with the Class argument. Should I change it to a different style?

artembilan: The existing code was written years ago.
And of course we don't just go to refine it because we should.
But that doesn't mean that new code should follow old pattern when more modern API could be beneficial.
If you have cycles, feel free to fix other usages of Class-based `mock()` method in this class.
But no obligation and no pushing.
Thanks

joyfulviper: Thanks for the suggestion! I've updated the new test to use `mock()` (type-inference style).

Regarding the other `mock(Class)` usages in this class — I'm happy to clean those up as well. I also noticed there are about 30+ usages of `foo`/`bar` as queue names throughout this test class, which seems to fall under the same category as the language concern you mentioned earlier. Would it be appropriate to clean those up too — in this PR as a separate commit, or as a follow-up PR?

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerTests.java
joyfulviper: I've changed the approach to use `CountDownLatch` to block consumer threads inside the `basicConsume` mock answer, and also verify the self-removal after releasing them.

```java
CountDownLatch consumersReadyLatch = new CountDownLatch(2);
CountDownLatch proceedLatch = new CountDownLatch(1);
willAnswer(invocation -> {
    Consumer callback = invocation.getArgument(6);
    callback.handleConsumeOk("consumerTag");
    consumersReadyLatch.countDown();
    proceedLatch.await(10, TimeUnit.SECONDS);
    return "consumerTag";
}).given(channel)
        .basicConsume(anyString(), anyBoolean(), anyString(), anyBoolean(), anyBoolean(),
                anyMap(), any(Consumer.class));
```

Both consumer threads block on `proceedLatch` inside `basicConsume()`, so they never enter `mainLoop()`. After `consumersReadyLatch` confirms both are ready, `setConcurrentConsumers(1)` and the assertion run with no race.

Then after releasing consumers, I verify the self-removal:

```java
// 1. Verify deferral — no immediate removal
container.setConcurrentConsumers(1);
assertThat(consumers).hasSize(2);

// 2. Release consumers — they enter mainLoop, receiveTimeout expires,
//    checkAdjust() calls markForStopIfExcess(), and the excess consumer
//    removes itself.
proceedLatch.countDown();
await().until(() -> consumers.size() == 1);

container.stop();
```

This way, one test verifies both: the deferral (no immediate cancel) and the eventual self-removal via `markForStopIfExcess()`.

Does this approach look reasonable?

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerTests.java
artembilan: > I've updated the new test to use 

I think you have not pushed the change yet.

If that `foo/bar` and `mock(Class)` is too much for you, please, don't spend time on them.
That is really minor and not warrant contribution requirements.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainerTests.java
artembilan: Thank you for looking into this!
Unfortunately, I still see a flaw.
Yes, with that `proceedLatch` we ensure that our test is passing, but condition was made up.
We never reach `checkAdjust()` after calling `setConcurrentConsumers()`  due to that latch in `basicConsume`.
Therefore it is obvious that both consumers are not cancelled because they just don't move on in their `mainLoop`.
This condition would be `true` even without a new option you are introducing in this contribution.


Let's consider to inject a `spy` for `Log` into the container!
And in the end after `stop()` verify that `Idle consumer terminating` message was never logged but `Immediate scale-down: consumer marked for stop` was instead.

Not sure yet how to do that, but latch in the `basicConsume` doesn't help to see the whole picture.

artembilan (changes requested):

joyfulviper:
Thank you for the thorough review! I've addressed all the feedback and pushed the changes. Please take another look when you have a chance.

artembilan (approved):
Love it 👍 !

Will merge when PR build is green.

artembilan:
@joyfulviper ,

thank you very much for contribution; looking forward for more!

joyfulviper:
Thanks for the thorough review, @artembilan 🙌 Will contribute again when the opportunity comes up 🙂

Issue #3418, opened 2026-04-20
Double-ack regression when using MANUAL ack mode with ErrorHandler


aramperes:
**In what version(s) of Spring AMQP are you seeing this issue?**

4.0.1 and above

Since [this commit](https://github.com/spring-projects/spring-amqp/commit/c6519b274fece59579176ef54f5ca9804236e418)

**Describe the bug**

When an Error Handler acknowledges or rejects a message and returns a `null` response, the framework is automatically acknowledging the message again, which causes client & broker errors with unknown delivery tags.

**To Reproduce**

Implement a `RabbitListenerErrorHandler#handleError(Message, Channel, Message, ListenerExecutionFailedException)` that calls:

```
// ... Logic here ...
channel.basicReject(message.getMessageProperties().getDeliveryTag(), true);
return null;
``` 

This causes an error when the message is later acknowledged automatically by the framework, even though the ack mode is manual:

https://github.com/spring-projects/spring-amqp/blob/310c536e4dec5cd050746196290ef8516cdb4af2/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapter.java#L217-L219

**Expected behavior**

It should be possible to disable this auto-acknowledgment behaviour for error handlers especially when the is already set to `MANUAL`.

For our use-case, we handle errors by logging them, and deciding whether they should be requeued or not (based on external logic).

Perhaps the framework can check if the message was already acknowledged in the error handler?

**Sample**

N/A

artembilan:
There is no way to check if the `deliveryTag` has been ack'ed this or other way before we call the next settlement for it.
The logic there is like this:
```
	protected void basicAck(Message request, @Nullable Channel channel) {
		if (channel != null) {
			try {
				channel.basicAck(request.getMessageProperties().getDeliveryTag(), false);
			}
			catch (IOException e) {
				this.logger.error("Failed to ack message", e);
			}
		}
	}
```
That `this.logger.error()` is harmful and just goes to the logs without any other system effect.
I think we will change this to warning since it is really not an error from the application perspective, and something may happen with that  `deliveryTag` out of our application control.

aramperes:
@artembilan Thank you for taking a look at this issue.

The double-acknowledgement causes unwanted side effects that I think the library user should be able to prevent:

- The error log on the client side (now warning) is misleading and clearly not intended to happen in MANUAL mode.
[...]

Issue #3247, opened 2025-11-26
Error handler for async RabbitListeners (Mono, Future, suspend) not acking message when ErrorHandler returns `null`


jensbaitingerbosch:
**In what version(s) of Spring AMQP are you seeing this issue?**

3.2.8

**Describe the bug**

When having an async Rabbit listener which throws an exception and the error handler returns null (assuming that the exception had been somehow mitigated (e.g. triggering a retry), the message is not acked but 'hangs' unacked for all eternity (or until the rabbit server times out). 

In contrast using the same logic with a sync listener, the message is acked. 

**To Reproduce**

Send a message to an the queue `theQueue` and the message stays anacked. Send a message to `theOtherQueue` and the message is consumed. 

```
@Component
class Consumer() {

    @RabbitListener(queues = ["theQueue"], errorHandler = "errorHandler" )
    suspend fun consumeRabbitMessage(message: Message): Unit? {
        throw RuntimeException("something went wrong while consuming the message")
    }

       @RabbitListener(queues = ["theOtherQueue"], errorHandler = "errorHandler" )
   fun consumeRabbitMessageSync(message: Message): Unit? {
        throw RuntimeException("something went wrong while consuming the message")
    }
}

@Component
class ErrorHandler : RabbitListenerErrorHandler {
    override fun handleError(
        msg: Message?,
        ch: Channel?,
        msg2: org.springframework.messaging.Message<*>?,
        ex: ListenerExecutionFailedException?
    ): Any? {

        //asume the error was handled internally
        return null
    }
}
```

**Expected behavior**

The message should be acked (analog behaviour to a sync listener)

PR #3441, opened 2026-05-20
GH-3418: Option to disable acknowledgements on error

aramperes:
This is another approach to fixing GH-3418.

In my opinion the `MANUAL` acknowledge mode should not be _automatically_ acknowledging messages. It should be up to the developer to acknowledge messages whether they are successfully handled or not.

Evidently since this change was intentional for GH-3247 I propose adding an option on the Container Factory to disable the new behavior, returning control to the library user to acknowledge (or reject) messages in error handlers, without causing unintended effects.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/BaseRabbitListenerContainerFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/BaseRabbitListenerContainerFactory.java
@@ -47,6 +47,7 @@ import org.springframework.util.Assert;
  * @author Ngoc Nhan
  * @author Artem Bilan
  * @author Stephane Nicoll
+ * @author Aram Peres
  *
  * @since 2.4
  *
@@ -232,6 +233,7 @@ public abstract class BaseRabbitListenerContainerFactory<C extends MessageListen
 	 * acknowledgment itself (e.g. via {@code channel.basicReject()}) without causing
 	 * a duplicate-ack error. Default is {@code true}.
 	 * @param acknowledgeOnError false to skip auto-ack on null return from error handler.
+	 * @since 3.2.11
 	 * @see MessagingMessageListenerAdapter#setAcknowledgeOnError(boolean)
 	 */
 	public void setAcknowledgeOnError(boolean acknowledgeOnError) {
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/BaseRabbitListenerContainerFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/BaseRabbitListenerContainerFactory.java
@@ -29,6 +29,7 @@ import org.springframework.amqp.rabbit.listener.AbstractMessageListenerContainer
 import org.springframework.amqp.rabbit.listener.RabbitListenerContainerFactory;
 import org.springframework.amqp.rabbit.listener.RabbitListenerEndpoint;
 import org.springframework.amqp.rabbit.listener.adapter.AbstractAdaptableMessageListener;
+import org.springframework.amqp.rabbit.listener.adapter.MessagingMessageListenerAdapter;
 import org.springframework.amqp.rabbit.retry.MessageRecoverer;
 import org.springframework.amqp.utils.JavaUtils;
 import org.springframework.beans.BeansException;
@@ -68,6 +69,8 @@ public abstract class BaseRabbitListenerContainerFactory<C extends MessageListen
 	private @Nullable Boolean micrometerEnabled;
 
 	private @Nullable Boolean observationEnabled;
+
+	private @Nullable Boolean acknowledgeOnError;
 
 	@SuppressWarnings("NullAway.Init")
 	private ApplicationContext applicationContext;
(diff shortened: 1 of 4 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/BaseRabbitListenerContainerFactory.java
```diff
@@ -218,6 +225,19 @@ public void setObservationEnabled(boolean observationEnabled) {
 		return this.observationEnabled;
 	}
 
+	/**
+	 * Set to {@code false} to prevent auto-acknowledgment of messages when a
+	 * {@link org.springframework.amqp.rabbit.listener.api.RabbitListenerErrorHandler}
+	 * returns {@code null} in MANUAL ack mode. This lets the error handler manage
+	 * acknowledgment itself (e.g. via {@code channel.basicReject()}) without causing
+	 * a duplicate-ack error. Default is {@code true}.
+	 * @param acknowledgeOnError false to skip auto-ack on null return from error handler.
```
artembilan, line 235: `@since 3.2.11`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -108,6 +110,19 @@ protected MessagingMessageConverterAdapter newMessagingMessageConverterAdapter(
...
 		return new MessagingMessageConverterAdapter(bean, method, batch, Channel.class);
 	}
+
+	/**
+	 * Set to {@code false} to prevent the framework from automatically acknowledging
+	 * messages when the error handler returns {@code null} in MANUAL ack mode. This
+	 * allows the error handler to manage acknowledgment itself (e.g. by calling
+	 * {@code channel.basicReject()}) without causing a duplicate ack error.
+	 * Default is {@code true} (auto-acknowledge on null return from error handler).
+	 * @param acknowledgeOnError false to skip the auto-ack on null return from error handler.
```
artembilan, line 121: `@since 3.2.11`

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapterTests.java
```diff
@@ -336,6 +338,29 @@ public Object fromMessage(org.springframework.amqp.core.Message message) throws
 		assertThat(ehCalled.get()).isTrue();
 	}
 
+	@Test
+	void acknowledgeOnErrorDefaultAutoAcksInManualMode() throws Exception {
+		org.springframework.amqp.core.Message message = MessageTestUtils.createTextMessage("foo");
```
artembilan: Let's consider not use unprofessional `foo/bar` language!
I know we have many of those already, but rules have been changed since then and it is better to not introduce more of those

artembilan (changes requested):
This is great!

Please, add your name to the `@author` list of the effected classes.

Thank you!

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapterTests.java
aramperes: Done in [`be0cc48` (this PR)](https://github.com/spring-projects/spring-amqp/pull/3441/changes/be0cc48f24f3edafb44cc2137b16f7a3683feb35), let me know if you have a preference.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -108,6 +111,20 @@ protected MessagingMessageConverterAdapter newMessagingMessageConverterAdapter(
...
 		return new MessagingMessageConverterAdapter(bean, method, batch, Channel.class);
 	}
+
+	/**
+	 * Set to {@code false} to prevent the framework from automatically acknowledging
+	 * messages when the error handler returns {@code null} in MANUAL ack mode. This
+	 * allows the error handler to manage acknowledgment itself (e.g. by calling
+	 * {@code channel.basicReject()}) without causing a duplicate ack error.
+	 * Default is {@code true} (auto-acknowledge on null return from error handler).
+	 * @since 3.2.11
```
artembilan: This is wrong place for this tag.
Must be after `@param` and `@return`.
Essentially this is our order of Javadoc tags:
```
		<module name="AtclauseOrder">
			<property name="target" value="CLASS_DEF, INTERFACE_DEF, ENUM_DEF"/>
			<property name="tagOrder" value="@param, @author, @since, @see, @version, @serial, @deprecated"/>
		</module>
		<module name="AtclauseOrder">
			<property name="target" value="METHOD_DEF, CTOR_DEF, VARIABLE_DEF"/>
			<property name="tagOrder" value="@param, @return, @throws, @since, @deprecated, @see"/>
		</module>
```
Will try to fix it on merge.
Thanks

artembilan (changes requested):

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapter.java
artembilan:
```suggestion
```

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -108,6 +111,20 @@ protected MessagingMessageConverterAdapter newMessagingMessageConverterAdapter(
...
+	 * @param acknowledgeOnError false to skip the auto-ack on null return from error handler.
```
artembilan, line 121:
```suggestion
	 * @param acknowledgeOnError false to skip the auto-ack on null return from error handler.
	 * @since 3.2.11
```

artembilan (approved):
Merging after applying suggestions re. `@since` position.

Thanks

PR #3483, opened 2026-06-18
GH-3438: Add routing `AmqpConnectionFactory` support

### LINKED ISSUE #3438: AbstractRoutingConnectionFactory equivalent for AMQP 1.0/AmqpConnectionFactory
### Discussed in https://github.com/spring-projects/spring-amqp/discussions/3437

<div type='discussions-op-text'>

<sup>Originally posted by **vdotjansen** May 12, 2026</sup>
In AMQP 0-9-1 we use the AbstractRoutingConnectionFactory to identify which ConnectionFactory to use at Runtime, are there plans to create a equivalent for the AMQP 1.0/AmqpConnectionFactory variant?</div>

mcmxcivr:
Introduce a `RoutingAmqpConnectionFactory` strategy interface with an `AbstractRoutingAmqpConnectionFactory` base class that routes `getConnection()` calls to one of several target `AmqpConnectionFactory` instances based on a runtime lookup key, with optional lenient fallback to a default factory.

* Add `SimpleRoutingAmqpConnectionFactory` resolving the lookup key from a thread-bound `SimpleResourceHolder` resource
* Make `RabbitAmqpTemplate.getPublisher()` aware of a `RoutingAmqpConnectionFactory` so the target connection is resolved per call

Closes gh-3438

### DIFF
--- a/spring-rabbitmq-client/src/main/java/org/springframework/amqp/rabbitmq/client/AbstractRoutingAmqpConnectionFactory.java
+++ b/spring-rabbitmq-client/src/main/java/org/springframework/amqp/rabbitmq/client/AbstractRoutingAmqpConnectionFactory.java
@@ -1,0 +1,159 @@
+/*
+ * Copyright 2026-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.amqp.rabbitmq.client;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+import com.rabbitmq.client.amqp.Connection;
+import org.jspecify.annotations.Nullable;
+
+import org.springframework.beans.factory.InitializingBean;
+import org.springframework.util.Assert;
+
+/**
+ * Abstract {@link AmqpConnectionFactory} implementation that routes {@link #getConnection()}
+ * calls to one of various target {@link AmqpConnectionFactory AmqpConnectionFactories} based
+ * on a lookup key. The latter is usually (but not necessarily) determined through some
+ * thread-bound context.
+ *
+ * @author Robin Collard
+ *
+ * @since 4.2
+ */
+public abstract class AbstractRoutingAmqpConnectionFactory
+		implements AmqpConnectionFactory, RoutingAmqpConnectionFactory, InitializingBean {
+
+	private final Map<Object, AmqpConnectionFactory> targetConnectionFactories = new ConcurrentHashMap<>();
+
+	private @Nullable AmqpConnectionFactory defaultTargetConnectionFactory;
+
+	private boolean lenientFallback = true;
+
+	/**
+	 * Specify the map of target {@link AmqpConnectionFactory AmqpConnectionFactories},
+	 * with the lookup key as key.
+	 * <p>The key can be of arbitrary type; this class implements the generic lookup process
+	 * only. The concrete key representation will be handled by {@link #determineCurrentLookupKey()}.
+	 * @param targetConnectionFactories the target connection factories and lookup keys.
+	 */
+	public void setTargetConnectionFactories(Map<Object, AmqpConnectionFactory> targetConnectionFactories) {
+		Assert.notNull(targetConnectionFactories, "'targetConnectionFactories' must not be null.");
+		Assert.noNullElements(targetConnectionFactories.values().toArray(),
+				"'targetConnectionFactories' cannot have null values.");
+		this.targetConnectionFactories.putAll(targetConnectionFactories);
+	}
+
+	/**
+	 * Specify the default target {@link AmqpConnectionFactory}, if any.
+	 * <p>This {@link AmqpConnectionFactory} will be used as target if none of the keyed
+	 * {@link #targetConnectionFactories} match the {@link #determineCurrentLookupKey()}
+	 * current lookup key.
+	 * @param defaultTargetConnectionFactory the default target connection factory.
+	 */
+	public void setDefaultTargetConnectionFactory(AmqpConnectionFactory defaultTargetConnectionFactory) {
+		this.defaultTargetConnectionFactory = defaultTargetConnectionFactory;
+	}
+
+	/**
+	 * Specify whether to apply a lenient fallback to the default {@link AmqpConnectionFactory}
+	 * if no specific {@link AmqpConnectionFactory} could be found for the current lookup key.
+	 * <p>Default is "true", accepting lookup keys without a corresponding entry in the
+	 * {@link #targetConnectionFactories} - simply falling back to the default
+	 * {@link AmqpConnectionFactory} in that case.
+	 * <p>Switch this flag to "false" if you would prefer the fallback to only apply if the
+	 * lookup key was {@code null}. Lookup keys without an {@link AmqpConnectionFactory} entry
+	 * will then lead to an {@link IllegalStateException}.
+	 * @param lenientFallback true to fall back to the default, if specified.
+	 * @see #setTargetConnectionFactories
+	 * @see #setDefaultTargetConnectionFactory
+	 * @see #determineCurrentLookupKey()
+	 */
+	public void setLenientFallback(boolean lenientFallback) {
+		this.lenientFallback = lenientFallback;
+	}
+
+	protected boolean isLenientFallback() {
+		return this.lenientFallback;
+	}
+
+	@Override
+	public void afterPropertiesSet() {
+		Assert.isTrue(this.defaultTargetConnectionFactory != null || !this.targetConnectionFactories.isEmpty(),
+				"At least one target factory (or default) is required");
+	}
+
+	@Override
+	public Connection getConnection() {
+		return determineTargetConnectionFactory().getConnection();
+	}
+
+	/**
+	 * Retrieve the current target {@link AmqpConnectionFactory}. Determine the
+	 * {@link #determineCurrentLookupKey() current lookup key}, perform a lookup in the
+	 * {@link #targetConnectionFactories} map, fall back to the specified
+	 * {@link #defaultTargetConnectionFactory} if necessary.
+	 * @return the connection factory.
+	 * @see #determineCurrentLookupKey()
+	 */
+	protected AmqpConnectionFactory determineTargetConnectionFactory() {
+		Object lookupKey = determineCurrentLookupKey();
+		AmqpConnectionFactory connectionFactory = null;
+		if (lookupKey != null) {
+			connectionFactory = this.targetConnectionFactories.get(lookupKey);
+		}
+		if (connectionFactory == null && (this.lenientFallback || lookupKey == null)) {
+			connectionFactory = this.defaultTargetConnectionFactory;
+		}
+		if (connectionFactory == null) {
+			throw new IllegalStateException("Cannot determine target AmqpConnectionFactory for lookup key ["
+					+ lookupKey + "]");
+		}
+		return connectionFactory;
+	}
+
+	@Override
+	public @Nullable AmqpConnectionFactory getTargetConnectionFactory(Object key) {
+		return this.targetConnectionFactories.get(key);
+	}
+
+	/**
+	 * Add the given {@link AmqpConnectionFactory} and associate it with the given lookup key.
+	 * @param key the lookup key.
+	 * @param connectionFactory the {@link AmqpConnectionFactory}.
+	 */
+	public void addTargetConnectionFactory(Object key, AmqpConnectionFactory connectionFactory) {
+		this.targetConnectionFactories.put(key, connectionFactory);
+	}
+
+	/**
+	 * Remove the {@link AmqpConnectionFactory} associated with the given lookup key and return it.
+	 * @param key the lookup key.
+	 * @return the {@link AmqpConnectionFactory} that was removed.
+	 */
+	public @Nullable AmqpConnectionFactory removeTargetConnectionFactory(Object key) {
+		return this.targetConnectionFactories.remove(key);
+	}
+
+	/**
+	 * Determine the current lookup key. This will typically be implemented to check a
+	 * thread-bound context.
+	 * @return the lookup key.
+	 */
+	protected abstract @Nullable Object determineCurrentLookupKey();
+
+}
--- a/spring-rabbitmq-client/src/main/java/org/springframework/amqp/rabbitmq/client/RoutingAmqpConnectionFactory.java
+++ b/spring-rabbitmq-client/src/main/java/org/springframework/amqp/rabbitmq/client/RoutingAmqpConnectionFactory.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2026-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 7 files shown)


artembilan:
This is a new feature which we will consider when we start new minor, somewhere in July.

Meanwhile, please, ensure that project build is clean for your changes locally: `./gradlew clean check`

Thank you!

spring-rabbitmq-client/src/main/java/org/springframework/amqp/rabbitmq/client/AbstractRoutingAmqpConnectionFactory.java
```diff
@@ -0,0 +1,161 @@
...
+
+/**
+ * Abstract {@link AmqpConnectionFactory} implementation that routes {@link #getConnection()}
+ * calls to one of various target {@link AmqpConnectionFactory AmqpConnectionFactories} based
+ * on a lookup key. The latter is usually (but not necessarily) determined through some
+ * thread-bound context.
+ *
+ * @author Robin Collard
+ *
+ * @since 4.1.1
```
artembilan: `4.2`

spring-rabbitmq-client/src/main/java/org/springframework/amqp/rabbitmq/client/AbstractRoutingAmqpConnectionFactory.java
```diff
@@ -0,0 +1,161 @@
...
+	 * @param lenientFallback true to fall back to the default, if specified.
+	 * @see #setTargetConnectionFactories
+	 * @see #setDefaultTargetConnectionFactory
+	 * @see #determineCurrentLookupKey()
+	 */
+	public void setLenientFallback(boolean lenientFallback) {
+		this.lenientFallback = lenientFallback;
+	}
+
+	public boolean isLenientFallback() {
```
artembilan: I don't think we need to have it as `public`.
I may understand why `protected` though...

spring-rabbitmq-client/src/main/java/org/springframework/amqp/rabbitmq/client/AbstractRoutingAmqpConnectionFactory.java
```diff
@@ -0,0 +1,161 @@
...
+	protected void addTargetConnectionFactory(Object key, AmqpConnectionFactory connectionFactory) {
+		this.targetConnectionFactories.put(key, connectionFactory);
+	}
+
+	/**
+	 * Removes the {@link AmqpConnectionFactory} associated with the given lookup key and returns it.
+	 * @param key the lookup key.
+	 * @return the {@link AmqpConnectionFactory} that was removed.
+	 */
+	protected @Nullable AmqpConnectionFactory removeTargetConnectionFactory(Object key) {
```
artembilan: And in my opinion these two should be `public`.
Feels like target application API not something to hind into inheritors only.

spring-rabbitmq-client/src/main/java/org/springframework/amqp/rabbitmq/client/RoutingAmqpConnectionFactory.java
```diff
@@ -0,0 +1,41 @@
...
+package org.springframework.amqp.rabbitmq.client;
+
+import org.jspecify.annotations.Nullable;
+
+/**
+ * Implementations select an {@link AmqpConnectionFactory} based on a supplied key.
+ *
+ * @author Robin Collard
+ *
+ * @since 4.1.1
```
artembilan: `4.2`

spring-rabbitmq-client/src/main/java/org/springframework/amqp/rabbitmq/client/RoutingAmqpConnectionFactory.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+@FunctionalInterface
+public interface RoutingAmqpConnectionFactory {
+
+	/**
+	 * Returns the {@link AmqpConnectionFactory} bound to given lookup key, or {@code null}
```
artembilan: The method Javadocs must be imperative.
This is a method and we give it an instruction what to do.
So, `Return the {@link AmqpConnectionFactory} `

spring-rabbitmq-client/src/main/java/org/springframework/amqp/rabbitmq/client/AbstractRoutingAmqpConnectionFactory.java
```diff
@@ -0,0 +1,161 @@
...
+		return connectionFactory;
+	}
+
+	@Override
+	public @Nullable AmqpConnectionFactory getTargetConnectionFactory(Object key) {
+		return this.targetConnectionFactories.get(key);
+	}
+
+	/**
+	 * Adds the given {@link AmqpConnectionFactory} and associates it with the given lookup key.
```
artembilan: Imperative, not infinitive.

spring-rabbitmq-client/src/main/java/org/springframework/amqp/rabbitmq/client/AbstractRoutingAmqpConnectionFactory.java
```diff
@@ -0,0 +1,161 @@
...
+	 * @param key the lookup key.
+	 * @param connectionFactory the {@link AmqpConnectionFactory}.
+	 */
+	protected void addTargetConnectionFactory(Object key, AmqpConnectionFactory connectionFactory) {
+		this.targetConnectionFactories.put(key, connectionFactory);
+	}
+
+	/**
+	 * Removes the {@link AmqpConnectionFactory} associated with the given lookup key and returns it.
```
artembilan: DITTO

spring-rabbitmq-client/src/main/java/org/springframework/amqp/rabbitmq/client/SimpleRoutingAmqpConnectionFactory.java
```diff
@@ -0,0 +1,41 @@
...
+import org.springframework.amqp.rabbit.connection.SimpleResourceHolder;
+
+/**
+ * An {@link AbstractRoutingAmqpConnectionFactory} implementation which gets a {@code lookupKey}
+ * for the current {@link AmqpConnectionFactory} from a thread-bound resource by the key of the
+ * instance of this {@link AmqpConnectionFactory}.
+ *
+ * @author Robin Collard
+ *
+ * @since 4.1.1
```
artembilan: DITTO

spring-rabbitmq-client/src/test/java/org/springframework/amqp/rabbitmq/client/RoutingAmqpConnectionFactoryTests.java
```diff
@@ -0,0 +1,163 @@
...
+import static org.assertj.core.api.Assertions.assertThatIllegalStateException;
+import static org.assertj.core.api.Assertions.assertThatNoException;
+import static org.mockito.BDDMockito.given;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
+/**
+ * @author Robin Collard
+ *
+ * @since 4.1.1
```
artembilan: DITTO

spring-rabbitmq-client/src/test/java/org/springframework/amqp/rabbitmq/client/RoutingAmqpConnectionFactoryTests.java
```diff
@@ -0,0 +1,163 @@
...
+ */
+public class RoutingAmqpConnectionFactoryTests {
+
+	@Test
+	void routesToTargetByLookupKey() {
+		AmqpConnectionFactory cf1 = mock(AmqpConnectionFactory.class);
+		AmqpConnectionFactory cf2 = mock(AmqpConnectionFactory.class);
+		Connection connection1 = mock(Connection.class);
+		Connection connection2 = mock(Connection.class);
```
artembilan: Please, use a `mock()` API without arguments.

src/reference/antora/modules/ROOT/pages/rabbitmq-amqp-client.adoc
```diff
@@ -80,6 +80,61 @@ AmqpConnectionFactory connectionFactory(Environment environment) {
 
 See `SingleAmqpConnectionFactory` setters for all connection-specific setting.
 
+[[amqp-client-routing-connection-factory]]
+== Routing Connection Factory
+
+Similar to the `AbstractRoutingConnectionFactory` for the AMQP 0.9.1 protocol (see xref:amqp/connections.adoc#routing-connection-factory[Routing Connection Factory]), the `spring-rabbitmq-client` module provides an `AbstractRoutingAmqpConnectionFactory` for the AMQP 1.0 protocol.
```
artembilan: Say `(since version 4.2)` after `AbstractRoutingAmqpConnectionFactory`

src/reference/antora/modules/ROOT/pages/rabbitmq-amqp-client.adoc
```diff
@@ -80,6 +80,61 @@ AmqpConnectionFactory connectionFactory(Environment environment) {
...
+It is important to unbind the resource after use.
+
+The routing algorithm is as follows: the `determineCurrentLookupKey()` is consulted for a key; if it resolves to a registered target `AmqpConnectionFactory`, that one is used.
+Otherwise, if there is no target for the `lookupKey` (or the key is `null`) and the factory is configured with `lenientFallback = true` (the default), the `defaultTargetConnectionFactory` is used.
+If `lenientFallback = false` and no target matches a non-`null` key, an `IllegalStateException` is thrown.
+
+To implement a custom routing strategy, extend `AbstractRoutingAmqpConnectionFactory` and provide a `determineCurrentLookupKey()` implementation.
+For more information, see the javadoc:org.springframework.amqp.rabbitmq.client.AbstractRoutingAmqpConnectionFactory[JavaDoc] for `AbstractRoutingAmqpConnectionFactory`.
+
+Similar to the AMQP 0.9.1 listener containers, the `RabbitAmqpListenerContainer` is aware of a `RoutingAmqpConnectionFactory`.
```
artembilan: Looks like everything said in this paragraph is missed being implemented.
Are you going to address that as a part of this PR?

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -9,4 +9,8 @@
 === AMQP 1.0 Client
 
 A new `spring-amqp-client` module has been introduced to provide support for interacting with generic AMQP 1.0 protocol.
-See xref:amqp10-client.adoc[] for more information.
\ No newline at end of file
+See xref:amqp10-client.adoc[] for more information.
+
+The `spring-rabbitmq-client` module now provides an `AbstractRoutingAmqpConnectionFactory` (and a `SimpleRoutingAmqpConnectionFactory` implementation) to route to one of several `AmqpConnectionFactory` instances based on a runtime `lookupKey`.
+The `RabbitAmqpListenerContainer` is aware of a `RoutingAmqpConnectionFactory` and resolves the target factory from its queue names (optionally prefixed via `setLookupKeyQualifier`).
```
artembilan: This is bogus in this change set.
More over it is redundant.
Just saying in the whats-new that we have a new `RoutingAmqpConnectionFactory` contract is enough, and give a link to that target chapter.

artembilan (changes requested):
Make sure your name is present in all the affected classes.
Please, make sure all the changes are in.
I'm a bit frustrated to see docs changes but not code to enjoy review.
Please, also rebase your branch to the latest upstream main which is `4.2` already.
There are going to be some conflicts, especially in the whats-new.

Thanks

spring-rabbitmq-client/src/main/java/org/springframework/amqp/rabbitmq/client/AbstractRoutingAmqpConnectionFactory.java
artembilan: Was not addressed.
Or do you mean that my review mitigation is in progress?

spring-rabbitmq-client/src/main/java/org/springframework/amqp/rabbitmq/client/AbstractRoutingAmqpConnectionFactory.java
```diff
@@ -0,0 +1,161 @@
...
+ * Abstract {@link AmqpConnectionFactory} implementation that routes {@link #getConnection()}
+ * calls to one of various target {@link AmqpConnectionFactory AmqpConnectionFactories} based
+ * on a lookup key. The latter is usually (but not necessarily) determined through some
+ * thread-bound context.
+ *
+ * @author Robin Collard
+ *
+ * @since 4.1.1
+ *
+ * @see RoutingAmqpConnectionFactory
```
artembilan: This `@see` is redundant because we are implementing this contract.
Maybe you mean to mention `org.springframework.amqp.rabbit.connection.RoutingConnectionFactory`?

spring-rabbitmq-client/src/main/java/org/springframework/amqp/rabbitmq/client/RoutingAmqpConnectionFactory.java
artembilan: DITTO

spring-rabbitmq-client/src/main/java/org/springframework/amqp/rabbitmq/client/RoutingAmqpConnectionFactory.java
artembilan: Missed to fix method Javadoc to imperative.

spring-rabbitmq-client/src/main/java/org/springframework/amqp/rabbitmq/client/SimpleRoutingAmqpConnectionFactory.java
artembilan: DITTO

spring-rabbitmq-client/src/test/java/org/springframework/amqp/rabbitmq/client/RoutingAmqpConnectionFactoryTests.java
artembilan: DITTO

spring-rabbitmq-client/src/test/java/org/springframework/amqp/rabbitmq/client/RoutingAmqpConnectionFactoryTests.java
artembilan: I asked to use `mock()` API without arguments.

src/reference/antora/modules/ROOT/pages/rabbitmq-amqp-client.adoc
artembilan: This is misleading part you have not addressed yet.

Sorry, but this our collaboration is a bit counter productive: I leave review, you push something but without addressing review.
And you even don't comment why you have not addressed review this or other way 😢

artembilan (changes requested):

mcmxcivr:
First of all, thank you for reviewing this PR.

Looking at this again and based on the comments, I don't think it should go in. The problem is that only one of the two places that need routing has it, and that one place leaks.

The routing path in getPublisher() leaks two publishers on every send.
RabbitAmqpTemplate.doSend calls getPublisher() twice, The new routing code makes a new Publisher each call and never keeps it. So every send makes two publishers, closes neither. The destroy() only closes this.publisher, which the routing path never sets.

Consumers can't route at all. RabbitAmqpListenerContainer just calls this.connectionFactory.getConnection() once, in start(), on the thread that starts it. No lookup key is set on that thread. So a routing factory finds null and hands back the defaultTargetConnectionFactory, or fails at startup with IllegalStateException. The 0.9.1 containers can do this so people will expect the same here and won't get it.

Last but not least, replies go to the wrong broker, quietly.

The documentation mentioned a part of code which I removed because it registered a connection factory for each tenant which I think should not be part of the spring-amqp code base.

I'm not sure if I'm going to be able to create a solution for those problems any time soon.

src/reference/antora/modules/ROOT/pages/rabbitmq-amqp-client.adoc
mcmxcivr: Sorry, this is my first PR. The last commit was to update the branch only.

artembilan:
> RabbitAmqpTemplate.doSend calls getPublisher() twice,

We can fix it with local variable.
Then we can close it in the `finally`  if it is not the same as `this.publisher`.

> Consumers can't route at all. 

So, let's don't mention that then at all!

mcmxcivr:
I addressed the comments in the review. Also added tests for the two methods that are now public and updated the documentation. I also cleaned up history which, in hindsight, was not a smart thing to do for the review.

Is the `RabbitAmqpTemplate` change as expected?

PR #3490, opened 2026-06-24
GH-3487: Apply Nullability to MessageProperties

### LINKED ISSUE #3487: Kotlin: Inconstistent nullability in all properties of `MessageProperties`

For example:

4.1.0 

**Describe the bug**

The getter and setters of some properties in `org.springframework.amqp.core.MessageProperties`

Example:

https://github.com/spring-projects/spring-amqp/blob/44aec8b2051ff87483d3c97a62535a9754e9af81/spring-amqp/src/main/java/org/springframework/amqp/core/MessageProperties.java#L228-L235

The getter returns a nullable String, the setter is not marked as Nullable. For Kotlin nullable Strings and non-nullable Strings are different types, therefore it does not recognize the setter `setAppId()` as the setter for the nullable property `appId`. As the property does not have a setter, for kotlin the property is a `val` (not changeable)

The workaround calling the setter directly does not work in our setup as ~~ktlint~~ IntelliJ recognizes it as setter and changes the format automatically to `appId = <value>`

Affected are (list might not be complete):
- `type`
- `timestamp`
- `messageId`
- `userId`
- `clusterId`
- `expiration`

Non affected (counterexamples)
- `correlationId` (getter and setter nullable)
- `replyTo` (getter and setter nullable)
- `contentEncoding` (getter and setter nullable)
- `priority` (both getter and setter non-nullable)

**To Reproduce**

```kotlin
val mp = MessageProperties()
mp.appId = "myApp"
```
produces the compiler error `val cannot be reasigned`

**Expected behavior**

the property can be (re)assigned.

ngocnhan-tran1996:
For Kotlin, nullable and non-nullable types are different types. Since the setter method in the Java class is not annotated with `@Nullable`, Kotlin treats the property as a val (read-only). This commit fixes that issue.

Closes gh-3487

### DIFF
--- a/spring-amqp/src/test/kotlin/org/springframework/amqp/core/KotlinMessagePropertiesTests.kt
+++ b/spring-amqp/src/test/kotlin/org/springframework/amqp/core/KotlinMessagePropertiesTests.kt
@@ -21,25 +21,30 @@ import assertk.assertions.isEqualTo
 import assertk.assertions.isNotNull
 import assertk.assertions.isTrue
 import org.junit.jupiter.api.Test
+import org.mockito.Mockito.mock
 import java.lang.reflect.Type
 import java.util.Date
 
 /**
  * @author Ngoc Nhan
+ *
+ * @since 4.0.5
  */
 class KotlinMessagePropertiesTests {
 
 	@Test
 	fun testSetterForNullabilityFields () {
 		val currentDate = Date()
+		val amqpAcknowledgment = mock(AmqpAcknowledgment::class.java)
+		val inferredArgumentType = mock(Type::class.java)
 		val messageProperties = MessageProperties()
-			messageProperties.amqpAcknowledgment = AmqpAcknowledgmentImpl()
+			messageProperties.amqpAcknowledgment = amqpAcknowledgment
 			messageProperties.appId = "appId"
 			messageProperties.clusterId = "clusterId"
 			messageProperties.consumerQueue = "consumerQueue"
 			messageProperties.consumerTag = "consumerTag"
 			messageProperties.expiration = "expiration"
-			messageProperties.inferredArgumentType = TypeImpl()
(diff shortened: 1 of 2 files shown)


spring-amqp/src/test/kotlin/org/springframework/amqp/core/KotlinMessagePropertiesTests.kt
```diff
@@ -0,0 +1,101 @@
...
+		override fun equals(other: Any?): Boolean {
+			if (this === other) return true
+			if (other !is AmqpAcknowledgmentImpl) return false
+
+			return true
+		}
+
+	}
+
+	class TypeImpl: Type {
```
artembilan: Cannot we go with `mock()` for these object properties instead of some specific type impls?

spring-amqp/src/test/kotlin/org/springframework/amqp/core/KotlinMessagePropertiesTests.kt
```diff
@@ -0,0 +1,101 @@
...
+import assertk.assertThat
+import assertk.assertions.isEqualTo
+import assertk.assertions.isNotNull
+import assertk.assertions.isTrue
+import org.junit.jupiter.api.Test
+import java.lang.reflect.Type
+import java.util.Date
+
+/**
+ * @author Ngoc Nhan
```
artembilan, line 29: `@since 4.0.5`

artembilan (changes requested):
Looks good!

Please, add your name to the `@author` list of that changed by you `MessageProperties` class.

Thank you!

ngocnhan-tran1996:
I had already added my name to the `@author` list before.

https://github.com/spring-projects/spring-amqp/blob/7fbdcc213fab90085522664d5bc4076fa5fe723d/spring-amqp/src/main/java/org/springframework/amqp/core/MessageProperties.java#L43-L45

PR #3508, opened 2026-07-10
Update project to use TOML for version management

cppwfs:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -31,10 +31,6 @@ ext {
 	linkScmConnection = 'git://github.com/spring-projects/spring-amqp.git'
 	linkScmDevConnection = 'git@github.com:spring-projects/spring-amqp.git'
 
-	micrometerVersion = libs.versions.micrometerVersion.get()
-	micrometerTracingVersion = libs.versions.micrometerTracingVersion.get()
-	springVersion = libs.versions.springVersion.get()
-
 	javaProjects = subprojects - project(':spring-amqp-bom')
 }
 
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -31,32 +31,9 @@ ext {
...
-	rabbitmqAmqpClientVersion = '0.8.0'
-	rabbitmqStreamVersion = '1.3.0'
-	rabbitmqVersion = '5.27.1'
-	reactorVersion = '2025.0.6'
-	springDataVersion = '2025.1.6'
-	springVersion = '7.0.8'
-	testcontainersVersion = '2.0.5'
+	micrometerVersion = libs.versions.micrometerVersion.get()
+	micrometerTracingVersion = libs.versions.micrometerTracingVersion.get()
+	springVersion = libs.versions.springVersion.get()
```
artembilan: Please, revise the reasoning behind these variables.
Totally unclear why would we use them instead of lib entries.
And on the other hand they are out of use at all.

artembilan (changes requested):
Cool beans!

Just one concern about unused variables.

Thanks

build.gradle
cppwfs: These are used by the `generateAttributes` function for `spring-framework-javadoc`, `micrometer-docs`, and  `micrometer-tracing-docs`.   
Line 700-708

artembilan: Got it!
Thanks.
Cannot we scope then them inside of that `generateAttributes()` function?
Looks like we do something similar already on `main`.
Not saying that it is ideal over there, since local vars might be better, than evaluation is already so complex links.

PR #3519, opened 2026-07-15
GH-3518: RepublishMessageRecoverer: no stack trace by default

### LINKED ISSUE #3518: `RepublishMessageRecoverer`: don't publish error stack trace by default
Current logic is:
```
		if (this.maxStackTraceLength < 0) {
			int maxStackTraceLen = RabbitUtils
					.getMaxFrame(((RabbitTemplate) this.errorTemplate).getConnectionFactory());
			if (maxStackTraceLen > 0) {
				maxStackTraceLen -= this.frameMaxHeadroom;
				this.maxStackTraceLength = maxStackTraceLen;
			}
		}
```
Expose a flag `includeStackTrace` (`false` by default) to opt-in for this information.

artembilan:
Fixes: https://github.com/spring-projects/spring-amqp/issues/3518

The standard use-cases is to not deal with a whole stack trace.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/retry/RepublishMessageRecoverer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/retry/RepublishMessageRecoverer.java
@@ -38,8 +38,9 @@ import org.springframework.util.Assert;
 
 /**
  * {@link MessageRecoverer} implementation that republishes recovered messages
- * to a specified exchange with the exception stack trace stored in the
- * message header x-exception.
+ * to a specified exchange with the exception message and stack trace
+ * (if opted-in via {@link #includeStackTrace(boolean)}) stored in the
+ * message headers: {@value X_EXCEPTION_MESSAGE} and {@code X_EXCEPTION_STACKTRACE}.
  * <p>
  * If no routing key is provided, the original routing key for the message,
  * prefixed with {@link #setErrorRoutingKeyPrefix(String)} (default "error.")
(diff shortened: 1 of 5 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/retry/RepublishMessageRecoverer.java
```diff
@@ -160,6 +157,18 @@ public RepublishMessageRecoverer frameMaxHeadroom(int headroom) {
 		return this;
 	}
 
+	/**
+	 * Set {@code true} to add header with error stack trace.
+	 * @param includeStackTrace {@code true} to add header with error stack trace.
+	 * @return this.
+	 * @since 4.0.5
+	 */
+	public RepublishMessageRecoverer includeStackTrace(boolean includeStackTrace) {
```
cppwfs, line 167: Line 41 states that the stack trace is included by default.   This is no longer true.

cppwfs:
Line 160 in the  resilience-recovering-from-errors-and-broker-failures.adoc indicates that stack-trace is included by default.   We may want to say its opt-in.

cppwfs:
Do we want to change the behavior in 4.0.x, this seems to be a breaking change and that is ok for the 4.1.x but for 4.0.x should it be left alone?

cppwfs (commented):
Looks great!   Nice addition to allow users to opt-in.    

Since this breaking change do we want to add a what's new entry?

artembilan:
> Do we want to change the behavior in 4.0.x, this seems to be a breaking change and that is ok for the 4.1.x but for 4.0.x should it be left alone?

See https://github.com/spring-projects/spring-amqp/issues/3518#issuecomment-4993007320
and respective label on the issue.

The fix is, essentially, a security hardening to not expose application internals via publishing error details.

cppwfs (approved):
LGTM!

Thanks for working on this!

PR #3524, opened 2026-07-18
GH-3434: `RetryListener` on stateless interceptor

### LINKED ISSUE #3434: Enable registering custom RetryListeners
**Expected Behavior**

The RetryTemplate's RetryListeners, which is instantiated in StatelessRetryOperationsInterceptor should be customizable

**Current Behavior**

Only its RetryPolicy is customizable, not the listeners.

**Context**

I am in a shared-library implementing a RabbitListenerErrorHandler.
In the handleError() method I cannot find a way to differentiate between
1. spring.rabbitmq.listener.simple.retry.enabled=false or (true AND all retries are exhausted)
  -> like in there is no hope
2. spring.rabbitmq.listener.simple.retry.enabled=true, but not all retries are exhausted.
  -> like there is still hope

The MessageRecoverer is also no help, since i am unable to detect within handleError(), if there even will be a recoverer later.
Actually i dont event want to rely on "spring.rabbitmq.listener.simple.retry.enabled", since one could have overwritten the RetryTemplate on the container.
but in the end, maybe even a RetryListener wont help me :/

xxxxxxjun:
Fixes #3434

The `RetryTemplate` created inside `StatelessRetryOperationsInterceptor` could only be
configured with a `RetryPolicy`, so there was no way to react to retry events, e.g. to detect
that the retry policy has been exhausted.

Expose a `setRetryListener(RetryListener)` option on `StatelessRetryOperationsInterceptor` and
`StatelessRetryOperationsInterceptorFactoryBean`, plus a `retryListener()` option on
`RetryInterceptorBuilder.stateless()`. `StreamRetryOperationsInterceptorFactoryBean` inherits
the new option.

The API takes a single listener, mirroring `RetryTemplate#setRetryListener()`; multiple
listeners can be combined with a `CompositeRetryListener`. The stateful interceptor is left
out since it manages retry state itself and doesn't use a `RetryTemplate`.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/RetryInterceptorBuilder.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/RetryInterceptorBuilder.java
@@ -25,6 +25,7 @@ import org.jspecify.annotations.Nullable;
 import org.springframework.amqp.rabbit.retry.MessageKeyGenerator;
 import org.springframework.amqp.rabbit.retry.MessageRecoverer;
 import org.springframework.amqp.rabbit.retry.NewMessageIdentifier;
+import org.springframework.core.retry.RetryListener;
 import org.springframework.core.retry.RetryPolicy;
 import org.springframework.util.Assert;
 
@@ -63,6 +64,7 @@ import org.springframework.util.Assert;
  * @author Gary Russell
  * @author Artem Bilan
  * @author Stephane Nicoll
+ * @author Jun Cho
  *
  * @since 1.3
  *
@@ -230,12 +232,30 @@ public abstract class RetryInterceptorBuilder<B extends RetryInterceptorBuilder<
 		private final StatelessRetryOperationsInterceptorFactoryBean factoryBean =
 				new StatelessRetryOperationsInterceptorFactoryBean();
 
+		private @Nullable RetryListener retryListener;
+
 		StatelessRetryInterceptorBuilder() {
+		}
+
+		/**
+		 * Apply a {@link RetryListener} to the stateless interceptor. If multiple
+		 * listeners are needed, use a
+		 * {@link org.springframework.core.retry.support.CompositeRetryListener}.
+		 * @param retryListener the retry listener to use.
+		 * @return this.
+		 * @since 4.1.1
+		 */
+		public StatelessRetryInterceptorBuilder retryListener(RetryListener retryListener) {
+			this.retryListener = retryListener;
+			return this;
 		}
 
 		@Override
 		public StatelessRetryOperationsInterceptor build() {
 			this.applyCommonSettings(this.factoryBean);
+			if (this.retryListener != null) {
+				this.factoryBean.setRetryListener(this.retryListener);
+			}
 			return this.factoryBean.getObject();
 		}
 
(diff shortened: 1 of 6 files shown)


artembilan:
We will review this next month, just before `4.2.0-M1`!

Thanks

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/RetryInterceptorBuilder.java
```diff
@@ -230,12 +232,30 @@ public static final class StatelessRetryInterceptorBuilder
...
 		StatelessRetryInterceptorBuilder() {
 		}
 
+		/**
+		 * Apply a {@link RetryListener} to the stateless interceptor. If multiple
+		 * listeners are needed, use a
+		 * {@link org.springframework.core.retry.support.CompositeRetryListener}.
+		 * @param retryListener the retry listener to use.
+		 * @return this.
+		 * @since 4.1.1
```
artembilan, line 246: Since `4.2`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/StatelessRetryOperationsInterceptor.java
```diff
@@ -35,26 +36,38 @@
...
 				new RetryTemplate(retryPolicy != null ? retryPolicy : RetryPolicy.builder().delay(Duration.ZERO).build());
 		this.recoverer = recoverer;
 	}
 
+	/**
+	 * Set a {@link RetryListener} to be registered on the internal {@link RetryTemplate}.
+	 * If multiple listeners are needed, use a
+	 * {@link org.springframework.core.retry.support.CompositeRetryListener}.
+	 * @param retryListener the retry listener to use
+	 * @since 4.1.1
```
artembilan, line 61: DITTO

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/StatelessRetryOperationsInterceptorFactoryBean.java
```diff
@@ -42,16 +43,35 @@
...
 	protected final Log logger = LogFactory.getLog(getClass()); // NOSONAR
 
+	private @Nullable RetryListener retryListener;
+
+	/**
+	 * Set a {@link RetryListener} for the target stateless interceptor. If multiple
+	 * listeners are needed, use a
+	 * {@link org.springframework.core.retry.support.CompositeRetryListener}.
+	 * @param retryListener the retry listener to use
+	 * @since 4.1.1
```
artembilan, line 61: DITTO

src/reference/antora/modules/ROOT/pages/amqp/resilience-recovering-from-errors-and-broker-failures.adoc
```diff
@@ -92,6 +92,9 @@ Only a subset of retry capabilities can be configured this way.
 More advanced features would need the configuration of a `RetryPolicy`.
 See the {spring-framework-javadoc}/org/springframework/core/retry/RetryPolicy.html[`RetryPolicy` Javadoc] for more information about available policies and their configuration.
 
+Starting with version 4.1.1, a `RetryListener` can be registered with the stateless interceptor (via the `RetryInterceptorBuilder.stateless().retryListener()` option or the respective `StatelessRetryOperationsInterceptorFactoryBean` property) to react to retry events such as retry policy exhaustion.
```
artembilan, line 95: DITTO

artembilan (changes requested):
Mention this new addition in the `whats-new.adoc`, please.

Thank you!

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/3e2764a54bde613a33ca4f8766cb669611c4f2c6 after minor clean up as we discussed in the review.

PR #3530, opened 2026-07-21
GH-3529: Fix concurrency in RabbitAmqpListenerContainer.ConsumerBatch

### LINKED ISSUE #3529: Concurrency race condition around `RabbitAmqpListenerContainer.ConsumerMessageHandler.ConsumerBatch`
Two independent bugs, both confirmed in the code:

  1. The `releaseInternal()` does a non-atomic check-then-act on the volatile field `batchReleaseFuture.` The `release()` (called from the consumer-dispatch thread) calls `cancel(true)` then `releaseInternal()` directly; the `TaskScheduler` thread can already be inside `releaseInternal()` past the null-check when this happens (cancel returns `false` if already running — it doesn't prevent the race). Both threads can invoke `invokeBatchListener` on the same batch → duplicate delivery.
  2. `ConsumerMessageHandler.handle()` checks `currentBatch.batchReleaseFuture == null` and, if the scheduler thread hasn't nulled it yet, proceeds to mutate the plain `ArrayList.add()` while the scheduler thread may be handing
  that same list to `invokeBatchListener` — an unsynchronized shared mutable collection, so `ConcurrentModificationException` or a lost message are both plausible.

artembilan:
Fixes: https://github.com/spring-projects/spring-amqp/issues/3529

The `ConsumerBatch.releaseInternal()` could be called concurrently from a scheduled task and regular batch fulfillment.

* Add `ConsumerMessageHandler.lock` to guard in concurrent operations
* Add `ConsumerBatch.released` to avoid second undesired release

### DIFF
--- a/spring-rabbitmq-client/src/main/java/org/springframework/amqp/rabbitmq/client/listener/RabbitAmqpListenerContainer.java
+++ b/spring-rabbitmq-client/src/main/java/org/springframework/amqp/rabbitmq/client/listener/RabbitAmqpListenerContainer.java
@@ -586,7 +586,9 @@ public class RabbitAmqpListenerContainer
 
 	private class ConsumerMessageHandler implements Consumer.MessageHandler {
 
-		private volatile @Nullable ConsumerBatch consumerBatch;
+		private final Lock lock = new ReentrantLock();
+
+		private @Nullable ConsumerBatch consumerBatch;
 
 		ConsumerMessageHandler() {
 		}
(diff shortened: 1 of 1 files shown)


cppwfs (changes requested):
Looks good.    

Do we need a test for validation?

artembilan:
> Do we need a test for validation?

Concurrency is hard to test.
Nothing fails.
Claude double checked everything.
Good to go 😄 
Thanks

PR #3533, opened 2026-07-24
GH-3532: Add connection name support

### LINKED ISSUE #3532: Feature Request: Add support for setting a connection name in SingleAmqpConnectionFactory
**Expected Behavior**

I would like to propose exposing a method in `SingleAmqpConnectionFactory` to define the connection name, delegating it to the underlying `ConnectionBuilder`. 

Something similar to this:

```java
public SingleAmqpConnectionFactory setConnectionName(String connectionName) {
    this.connectionBuilder.name(connectionName);
    return this;
}
```

**Current Behavior**

Currently, SingleAmqpConnectionFactory only accepts com.rabbitmq.client.amqp.Environment , which the factory uses connectionBuilder, that returns a new fresh instance of ConnectionBuilder. Because of the behavior, it is impossible to use set a connection name beforehand, and SingleAmqpConnectionFactory does not expose any method to do so.

**Context**

It will be nice to not need to go into reflection way or create a new class just to set the connection name :)

MovinduJay:
Adds support for configuring a connection name on
SingleAmqpConnectionFactory.

The new fluent setConnectionName(String) method delegates to the
underlying RabbitMQ AMQP ConnectionBuilder.

Also adds focused unit coverage verifying that:

- the connection name is delegated to ConnectionBuilder.name(...)
- the setter returns the same SingleAmqpConnectionFactory instance

Fixes #3532

### DIFF
--- a/spring-rabbitmq-client/src/main/java/org/springframework/amqp/rabbitmq/client/SingleAmqpConnectionFactory.java
+++ b/spring-rabbitmq-client/src/main/java/org/springframework/amqp/rabbitmq/client/SingleAmqpConnectionFactory.java
@@ -60,6 +60,17 @@ public class SingleAmqpConnectionFactory implements AmqpConnectionFactory, Dispo
 		return this;
 	}
 
+	/**
+	 * Set the name used to identify the connection.
+	 * @param connectionName the connection name
+	 * @return this connection factory
+	 * @since 4.0.1
+	 */
+	public SingleAmqpConnectionFactory setConnectionName(String connectionName) {
+		this.connectionBuilder.name(connectionName);
+		return this;
+	}
+
 	public SingleAmqpConnectionFactory setPort(int port) {
 		this.connectionBuilder.port(port);
 		return this;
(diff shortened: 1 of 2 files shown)


artembilan:
Thank you!
We will look into this in a couple weeks.

spring-rabbitmq-client/src/main/java/org/springframework/amqp/rabbitmq/client/SingleAmqpConnectionFactory.java
```diff
@@ -60,6 +60,17 @@ public SingleAmqpConnectionFactory setHost(String host) {
 		return this;
 	}
 
+	/**
+	 * Set the name used to identify the connection.
+	 * @param connectionName the connection name
+	 * @return this connection factory
+	 * @since 4.0.1
```
artembilan: Please, fix `@since` for `4.0.5`.
Please, add your name to the `@author` list of this class.

spring-rabbitmq-client/src/test/java/org/springframework/amqp/rabbitmq/client/SingleAmqpConnectionFactoryTests.java
```diff
@@ -0,0 +1,54 @@
...
+ * @author Movindu Jayathilake
+ *
+ * @since 4.0.1
+ */
+class SingleAmqpConnectionFactoryTests {
+
+	@Test
+	void setConnectionNameDelegatesToConnectionBuilder() {
+		Environment environment = mock(Environment.class);
+		ConnectionBuilder connectionBuilder = mock(ConnectionBuilder.class);
```
artembilan: Please, use empty `mock()` API instead.

artembilan (changes requested):
Just couple nit-picks.
Thanks

PR #3536, opened 2026-07-27
Deprecate RabbitConnectionFactoryBean#enableHostnameVerification for amqp-client 5.33

### LINKED ISSUE #3525: RabbitConnectionFactoryBean#enableHostnameVerification is no longer relevant: Upgade to RabbitMQ Client 5.34.x

4.1.0

**Describe the bug**

It's not really a bug, but a change in the RabbitMQ Client. In `RabbitConnectionFactoryBean#enableHostnameVerification` there is a piece of code that setup the `useSslProtocol` and  conditionally enables hostname verification: https://github.com/spring-projects/spring-amqp/blob/945c148e7e2f96b55c86913171a398e92b901eef/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java#L905-L908

However, `useSslProtocol(sslContext)`  calls `enableHostnameVerification` so it'll always enabled, even if the property is `false`.

arimu1:
## Summary

Since `amqp-client` 5.33, `ConnectionFactory#useSslProtocol()` enables hostname verification automatically. The existing `RabbitConnectionFactoryBean` logic that conditionally called `enableHostnameVerification()` after `useSslProtocol()` was therefore redundant when `true`, and misleading when `false` (verification was still enabled by the client).

This change:

- Deprecates `RabbitConnectionFactoryBean#setEnableHostnameVerification(boolean)` since **4.1** (`forRemoval = true`); setter is now a no-op
- Removes the obsolete explicit `enableHostnameVerification()` calls from SSL setup
- Routes `skipServerCertificateValidation` through `ConnectionFactory#useTlsWithNoVerification()` (when skip is true, `sslAlgorithm` is ignored)
- Bumps `amqp-client` to 5.33.0 and adjusts `setUri` exception handling for the updated client API
- Updates reference documentation and what's new

Fixes #3525

## Test plan

- [x] `./gradlew :spring-rabbit:test --tests "org.springframework.amqp.rabbit.connection.SSLConnectionTests"`
- [x] `./gradlew :spring-rabbit:test --tests "org.springframework.amqp.rabbit.logback.AmqpAppenderTests"`
- [x] Assert factory bean no longer calls `enableHostnameVerification()` explicitly
- [x] Updated skip-server-certificate tests to expect `useTlsWithNoVerification()`


arimu1:
Addressed review feedback:

1. `@Deprecated(since = "4.1")` (was incorrectly `4.2`) + javadoc; setter is now a no-op and the unused field was removed.
2. Docs note that when `skipServerCertificateValidation` is `true`, `sslAlgorithm` is ignored (`useTlsWithNoVerification()` uses the client default protocol).
3. Removed “Made with Cursor” from the PR body; what’s-new mentions the `amqp-client` 5.33.0 dependency.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/RabbitConnectionFactoryBean.java
```diff
@@ -678,10 +676,14 @@ public void setChannelRpcTimeout(int channelRpcTimeout) {
 	 * <code>useSslProtocol</code> methods. Requires amqp-client 5.4.0 or later.
 	 * @param enable false to disable.
 	 * @since 2.0.6
+	 * @deprecated since 4.1 in favor of relying on {@code amqp-client} secure-by-default
```
artembilan, line 679: If you don’t do any code review and engineering here, consider to teach your Cursor what is breaking change and when that could be done.
We definitely cannot upgrade to a new Client in a patch version, and therefore, cannot even deprecate the property.
I’m not even talking about unrelated changes for those catch blocks .
We will really  revise this feature when start version 4.2

artembilan (changes requested):

arimu1:
Thanks @artembilan — understood. Upgrading `amqp-client` / deprecating `enableHostnameVerification` in a patch isn't appropriate; that belongs in 4.2.

Closing this PR. Happy to revisit when 4.2 work starts.

PR #3537, opened 2026-07-30
Align with commercial build support

artembilan:
Align spring-amqp's Gradle and CI setup with spring-amqp-commercial so
the same workflow files and build scripts can run, largely unmodified,
in both repositories.

- Relocate gradle/publish-maven.gradle to the repo root and wrap its
  content in `allprojects`, collapsing the three separate `apply from`
  call sites in build.gradle into one, matching the commercial repo's
  layout.
- Add commercial-settings.gradle (copied from spring-amqp-commercial)
  with its `apply from` left commented out in build.gradle; it must
  only be uncommented in the commercial repository.
- Recognize `*.x-internal` maintenance branches alongside `main` and
  `*.x` in the CI-triggering workflows.
- Introduce a `vars.COMMERCIAL` repository variable as the single
  source of truth for repo-specific behavior (runner labels,
  Artifactory URLs/repositories/credentials, OS test matrix) in
  pr-build.yml, ci.yml and ci-snapshot.yml, replacing ad-hoc
  `endsWith(github.repository, '-commercial')` checks.
- Add release-train automation workflows (cut-release-branch,
  update-release-train-versions, release-train-build/join/leave/ready/
  retry/test, post-release), gating the commercial-only ones on
  `vars.COMMERCIAL`.
- Drop the unused SPRING_RELEASE_CHAT_WEBHOOK_URL secret from release.yml.


artembilan:
Now, when we switch to the private commercial branch, we would need to change only one line - uncomment that `apply from` 😄

.github/workflows/release-train-build.yml
```diff
@@ -0,0 +1,92 @@
+# This file was auto-generated by github-actions-workflow-generator 0.0.6. Do not edit.
+# To update it, modify .github/workflow-generator.yml as needed and re-run the generator.
+
+name: "Release Train – Build"
```
cppwfs, line 4: Do we want to add generated code to the repo?

publish-maven.gradle
```diff
@@ -0,0 +1,83 @@
...
+							name = 'Mark Fisher'
+							email = 'mark.ryan.fisher@gmail.com'
+							roles = ['project founder']
+						}
+						developer {
+							id = 'markpollack'
+							name = 'Mark Pollack'
+							email = 'mark.pollack@broadcom.com'
+							roles = ['project founder']
+						}
```
cppwfs, line 71: What about Glenn Renfro 😢 ?

cppwfs (changes requested):
Looks Great!   One question and a playful request.  
Weird that we have to blend commercial with OSS .

publish-maven.gradle
artembilan: Here you are, Glenn: https://github.com/spring-projects/spring-amqp/commit/b938847bebf834f9dbdc80ccd86cda6f2502551e !
While making this PR I noticed that we are missing you.
Wanted to do that in a separate commit to emphasize.
Sorry about that.
🤗

.github/workflows/release-train-build.yml
artembilan: Yes, we want right now.
For a clean commit history and smooth rebase in the future.
Look, when release is done, the tag is going to have these workflows anyway.
So, even if we don't merge it back to base branch from tag, the commit with these workflows is going to be leaked anyway.
Therefore, I don't see any harm to add them right now and carry over for release branch and tag merge.

artembilan:
Changed my mind and made:
```
if (System.getenv('ARTIFACTORY_USERNAME')) {
	apply from: "${rootProject.projectDir}/commercial-settings.gradle"
}
```
Now we don't need to do any changes when we cut private branches.

artembilan:
> Weird that we have to blend commercial with OSS .

Looks like that is there in Spring Boot for a while already: https://github.com/spring-projects/spring-boot/blob/main/.github/workflows/build-and-deploy-snapshot.yml#L23.
No harm and questions from there.

I do all of this to simplify our DevOps around release train.

Thank you for review!

.github/workflows/ci.yml
```diff
@@ -20,14 +21,20 @@ jobs:
...
       matrix:
         testJavaVersion: [ 17, 21, 25, 26 ]
+        # Uncomment when we switch to hosted runners
+        # os: ${{ fromJSON(vars.COMMERCIAL && '["ubuntu22-2-8", "win22-8-32"]' || '["ubuntu-latest", "macos-latest", "windows-latest"]') }}
         os: [ ubuntu-latest, macos-latest, windows-latest ]
         exclude:
-        - os: ubuntu-latest
-          testJavaVersion: 25
+          # Uncomment when we switch to hosted runners
+          # - os: ${{ vars.COMMERCIAL && 'ubuntu22-2-8' || 'ubuntu-latest' }}
```
cppwfs, line 29: Why were the `vars.COMMERCIAL` checks commented out?  If we are proceeding with commented strategy  why even have the check just use the hosted runners ?

cppwfs (changes requested):
Just one more comment.

.github/workflows/ci.yml
artembilan: I might not be clear.

The `ubuntu22-2-8` and `ubuntu-latest` both hosted by GH.
However, only `ubuntu22-2-8` can get access to private repo.
We don't need private repo in this WF, and `ubuntu-latest` has much more instances and somehow a bit faster.
We want to switch to self-hosted runner in commercial repos when we have them.
Therefore, this commented out code is for the future consideration, not a strategy for the current environment.

With the solution we have in this PR I anticipate not changes in the private branches.
Probably just private version change of the project, nothing more.
But we'll see.
That's why I'm asking for review and merge, so we can proceed with the private part.

Thanks

cppwfs (approved):
Based on our discussion below it LGTM!

PR #3544, opened 2026-08-02
GH-3540: Deprecate `RestTemplateNodeLocator`

### LINKED ISSUE #3540: Deprecate `RestTemplateNodeLocator`
Spring Framework `7.1` has the `RestTemplate` as deprecated already: https://github.com/spring-projects/spring-framework/issues/36574

renechoi:
Fixes #3540

`RestTemplate` is deprecated for removal in Spring Framework `7.1` (spring-projects/spring-framework#36574) and `RestClient` is the designated replacement.

`RestTemplateNodeLocator` is not only one of the choices a user can make: it is what `ConnectionFactoryUtils.nodeLocator()` returns for the `LocalizedQueueConnectionFactory` when `spring-webflux` is not on the class path. Deprecating it on its own would leave non-reactive users without a supported `NodeLocator`, so this adds the `RestClient` variant first and makes it the new non-reactive default.

* `RestClientNodeLocator` (new) mirrors `RestTemplateNodeLocator`: basic authentication via `BasicAuthenticationInterceptor` and the same `/api/queues/{vhost}/{queue}` call. `RestClient` ships in `spring-web` just like `RestTemplate`, so the dependency profile does not change. One incidental improvement: `RestTemplateNodeLocator` wires `HttpComponentsClientHttpRequestFactory` explicitly and therefore requires the optional `httpclient5` dependency, while `RestClient` uses Apache HttpComponents when it is present and otherwise falls back to whichever client is available, down to the JDK `HttpClient`.
* `ConnectionFactoryUtils.nodeLocator()` returns `RestClientNodeLocator` when `WebClient` is absent.
* `RestTemplateNodeLocator` is annotated `@Deprecated(since = "4.2", forRemoval = true)`. A class level `@SuppressWarnings("removal")` keeps the build clean of the warnings its own use of `RestTemplate` now emits.
* Reference manual and What's New updated.

Verified locally with the JDK 25 toolchain, against RabbitMQ 4.3.4 with the management plugin (local broker plus the Testcontainers one the test class starts):

* new `RestClientNodeLocatorTests` covers the URI construction, including the `%2F` encoding of the default vhost, and the JSON deserialization, through `MockRestServiceServer`
* `LocalizedQueueConnectionFactoryIntegrationTests.findLocal()` now exercises `RestClientNodeLocator` against a live broker and keeps a leg on `RestTemplateNodeLocator` while that one is still present
* `./gradlew :spring-rabbit:test --tests "org.springframework.amqp.rabbit.connection.*"`: 135 tests, 0 failures
* `./gradlew :spring-rabbit:checkstyleMain :spring-rabbit:checkstyleTest :spring-rabbit:javadoc`: clean

### DIFF
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/LocalizedQueueConnectionFactoryIntegrationTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/LocalizedQueueConnectionFactoryIntegrationTests.java
@@ -102,6 +102,7 @@ public class LocalizedQueueConnectionFactoryIntegrationTests extends AbstractTes
 	}
 
 	@Test
+	@SuppressWarnings("removal")
 	void findLocal() {
 		ConnectionFactory defaultCf = mock(ConnectionFactory.class);
 		BrokerRunningSupport brokerRunning = RabbitAvailableCondition.getBrokerRunning();
@@ -112,9 +113,12 @@ public class LocalizedQueueConnectionFactoryIntegrationTests extends AbstractTes
 		ConnectionFactory cf = lqcf.getTargetConnectionFactory("[local]");
 		RabbitAdmin admin = new RabbitAdmin(cf);
 		assertThat(admin.getQueueProperties("local")).isNotNull();
-		lqcf.setNodeLocator(new RestTemplateNodeLocator());
+		lqcf.setNodeLocator(new RestClientNodeLocator());
 		ConnectionFactory cf2 = lqcf.getTargetConnectionFactory("[local]");
 		assertThat(cf2).isSameAs(cf);
+		lqcf.setNodeLocator(new RestTemplateNodeLocator());
+		ConnectionFactory cf3 = lqcf.getTargetConnectionFactory("[local]");
+		assertThat(cf3).isSameAs(cf);
 		lqcf.destroy();
 	}
 
--- a/src/reference/antora/modules/ROOT/pages/amqp/connections.adoc
+++ b/src/reference/antora/modules/ROOT/pages/amqp/connections.adoc
@@ -624,7 +624,8 @@ Notice that the first three parameters are arrays of `addresses`, `adminUris`, a
 These are positional in that, when a container attempts to connect to a queue, it uses the admin API to determine which node is the lead for the queue and connects to the address in the same array position as that node.
 
 IMPORTANT: Starting with version 3.0, the RabbitMQ `http-client` is no longer used to access the Rest API.
-Instead, by default, the `WebClient` from Spring Webflux is used if `spring-webflux` is on the class path; otherwise a `RestTemplate` is used.
+Instead, by default, the `WebClient` from Spring Webflux is used if `spring-webflux` is on the class path; otherwise a `RestClient` is used.
+Before version 4.2, the non-reactive default was a `RestTemplate`, which is deprecated in Spring Framework `7.1`.
 
 To add `WebFlux` to the class path:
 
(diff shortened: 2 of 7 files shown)


spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/LocalizedQueueConnectionFactoryIntegrationTests.java
```diff
@@ -112,9 +113,12 @@ void findLocal() {
 		ConnectionFactory cf = lqcf.getTargetConnectionFactory("[local]");
 		RabbitAdmin admin = new RabbitAdmin(cf);
 		assertThat(admin.getQueueProperties("local")).isNotNull();
-		lqcf.setNodeLocator(new RestTemplateNodeLocator());
+		lqcf.setNodeLocator(new RestClientNodeLocator());
 		ConnectionFactory cf2 = lqcf.getTargetConnectionFactory("[local]");
 		assertThat(cf2).isSameAs(cf);
+		lqcf.setNodeLocator(new RestTemplateNodeLocator());
```
artembilan, line 119: I prefer to not test already deprecated API.
Please, consider to remove this logic.

src/reference/antora/modules/ROOT/pages/amqp/connections.adoc
```diff
@@ -624,7 +624,8 @@ Notice that the first three parameters are arrays of `addresses`, `adminUris`, a
 These are positional in that, when a container attempts to connect to a queue, it uses the admin API to determine which node is the lead for the queue and connects to the address in the same array position as that node.
 
 IMPORTANT: Starting with version 3.0, the RabbitMQ `http-client` is no longer used to access the Rest API.
-Instead, by default, the `WebClient` from Spring Webflux is used if `spring-webflux` is on the class path; otherwise a `RestTemplate` is used.
+Instead, by default, the `WebClient` from Spring Webflux is used if `spring-webflux` is on the class path; otherwise a `RestClient` is used.
+Before version 4.2, the non-reactive default was a `RestTemplate`, which is deprecated in Spring Framework `7.1`.
```
artembilan, line 628: I don't think we need to say this here.
This doc is specifically for 4.2 version, so we are free to talk about RestClient` and fully forget `RestTemplate`.

src/reference/antora/modules/ROOT/pages/amqp/connections.adoc
```diff
@@ -661,7 +662,8 @@ lqcf.setNodeLocator(new NodeLocator<MyClient>() {
 });
 ----
 
-The framework provides the `WebFluxNodeLocator` and `RestTemplateNodeLocator`, with the default as discussed above.
+The framework provides the `WebFluxNodeLocator` and `RestClientNodeLocator`, with the default as discussed above.
```
artembilan, line 665: Instead of talking about `RestTemplateNodeLocator` as deprecated in this doc, just say:

> `RestClientNodeLocator` (since version 4.2)

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -4,3 +4,10 @@
 
 [[changes-in-4-2-since-4-1]]
 == Changes in 4.2 Since 4.1
+
+[[x42-rest-client-node-locator]]
+=== `RestClientNodeLocator`
+
+A new `RestClientNodeLocator` is now the default non-reactive `NodeLocator` implementation for the `LocalizedQueueConnectionFactory`.
+The `RestTemplateNodeLocator` is deprecated since the `RestTemplate` is deprecated in Spring Framework `7.1`.
```
artembilan, line 12: This part is good since we are in the place where we would like to say what was changed and why.

artembilan (changes requested):
Please, double check that your name is in all the affected classes.

Thank you!

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/995781af1b4852dbd72e40005a22e73544850a2e after some clean up.

@renechoi ,

thank you for the contribution; looking forward for more!

PR #3549, opened 2026-08-04
GH-1545: Add batchSize and batchReceiveTimeout to @RabbitListener

### LINKED ISSUE #1545: RabbitListener Add batch(), batchSize(), and receiveTimeout() Overrides
Similar to https://github.com/spring-projects/spring-kafka/issues/1872

Related to https://github.com/spring-projects/spring-amqp/issues/1544

MartinFerret:
Fixes #1545

Adds `batchSize()` and `batchReceiveTimeout()` overrides to `@RabbitListener`, resolved onto the endpoint and applied in `SimpleRabbitListenerContainerFactory` and `RabbitAmqpListenerContainerFactory`, with the endpoint value taking precedence over the factory. Both are `String` attributes so SpEL and property placeholders are supported, consistent with `priority()` and `concurrency()`.
`batch()` already existed since 3.0, so only these two were needed.

`AbstractRabbitListenerContainerFactory` is untouched, so `DirectRabbitListenerContainerFactory` silently ignores both attributes as discussed.

These attributes override values only; they do not enable `consumerBatchEnabled`, which stays a container factory property. That is called out explicitly in the reference documentation.

The two getters are `default` methods on `RabbitListenerEndpoint`, since the endpoint is the only channel between the BPP and the factory and `batchListener` already lives there. 

Happy to move them to `AbstractRabbitListenerEndpoint` with an `instanceof` check in the factories instead, if you'd rather leave the shared interface untouched.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListener.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListener.java
@@ -95,6 +95,7 @@ import org.springframework.messaging.handler.annotation.MessageMapping;
  *
  * @author Stephane Nicoll
  * @author Gary Russell
+ * @author Martin Ferret
  * @since 1.4
  * @see EnableRabbit
  * @see RabbitListenerAnnotationBeanPostProcessor
@@ -345,5 +346,32 @@ public @interface RabbitListener {
 	 * @see Boolean#parseBoolean(String)
 	 */
 	String batch() default "";
+
+	/**
+	 * Override the container factory's {@code batchSize} property. Ignored by container
+	 * factories that have no such property, such as the
+	 * {@code DirectRabbitListenerContainerFactory}. This overrides the value only; it
+	 * does not enable {@code consumerBatchEnabled}, which stays a container factory
+	 * property - refer to the reference documentation.
+	 * @return the batch size for this listener. If not set, the container factory setting
+	 * is used. SpEL {@code #{...}} and property place holders {@code ${...}} are
+	 * supported.
+	 * @since 4.2
+	 * @see #batchReceiveTimeout()
+	 */
+	String batchSize() default "";
+
+	/**
+	 * Override the container factory's {@code batchReceiveTimeout} property; the number
+	 * of milliseconds a batch gathering process should go on before a partial batch is
+	 * released. Ignored by container factories that have no such property, such as the
+	 * {@code DirectRabbitListenerContainerFactory}.
+	 * @return the batch receive timeout for this listener. If not set, the container
+	 * factory setting is used. SpEL {@code #{...}} and property place holders
+	 * {@code ${...}} are supported.
+	 * @since 4.2
+	 * @see #batchSize()
+	 */
+	String batchReceiveTimeout() default "";
 
 }
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/SimpleRabbitListenerContainerFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/SimpleRabbitListenerContainerFactory.java
@@ -193,10 +193,14 @@ public class SimpleRabbitListenerContainerFactory
 
 		super.initializeContainer(instance, endpoint);
 
-		// endpoint settings override the default factory settings
-		Integer endpointBatchSize = endpoint != null ? endpoint.getBatchSize() : null;
-		Long endpointBatchReceiveTimeout = endpoint != null ? endpoint.getBatchReceiveTimeout() : null;
-		String concurrency = endpoint != null ? endpoint.getConcurrency() : null;
+		Integer endpointBatchSize = null;
+		Long endpointBatchReceiveTimeout = null;
+		String concurrency = null;
+		if (endpoint != null) { // endpoint settings overriding default factory settings
+			endpointBatchSize = endpoint.getBatchSize();
+			endpointBatchReceiveTimeout = endpoint.getBatchReceiveTimeout();
+			concurrency = endpoint.getConcurrency();
+		}
 		JavaUtils javaUtils = JavaUtils.INSTANCE
 				.acceptOrElseIfNotNull(endpointBatchSize, this.batchSize, instance::setBatchSize)
 				.acceptIfNotNull(concurrency, instance::setConcurrency);
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/SimpleRabbitListenerContainerFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/SimpleRabbitListenerContainerFactory.java
@@ -35,6 +35,7 @@ import org.springframework.amqp.utils.JavaUtils;
  * @author Artem Bilan
  * @author Dustin Schultz
  * @author Jeonggi Kim
+ * @author Martin Ferret
  *
  * @since 1.4
  */
@@ -192,13 +193,13 @@ public class SimpleRabbitListenerContainerFactory
 
 		super.initializeContainer(instance, endpoint);
 
+		// endpoint settings override the default factory settings
+		Integer endpointBatchSize = endpoint != null ? endpoint.getBatchSize() : null;
+		Long endpointBatchReceiveTimeout = endpoint != null ? endpoint.getBatchReceiveTimeout() : null;
+		String concurrency = endpoint != null ? endpoint.getConcurrency() : null;
 		JavaUtils javaUtils = JavaUtils.INSTANCE
-				.acceptIfNotNull(this.batchSize, instance::setBatchSize);
-		String concurrency = null;
-		if (endpoint != null) {
-			concurrency = endpoint.getConcurrency();
-			javaUtils.acceptIfNotNull(concurrency, instance::setConcurrency);
-		}
+				.acceptOrElseIfNotNull(endpointBatchSize, this.batchSize, instance::setBatchSize)
+				.acceptIfNotNull(concurrency, instance::setConcurrency);
 		javaUtils
 				.acceptIfCondition(concurrency == null && this.concurrentConsumers != null, this.concurrentConsumers,
 						instance::setConcurrentConsumers)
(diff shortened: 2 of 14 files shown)


spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/SimpleRabbitListenerContainerFactory.java
```diff
@@ -192,13 +193,23 @@ protected void initializeContainer(SimpleMessageListenerContainer instance,
...
+			}
+			Long endpointBatchReceiveTimeout = endpoint.getBatchReceiveTimeout();
+			if (endpointBatchReceiveTimeout != null) {
+				batchReceiveTimeoutToUse = endpointBatchReceiveTimeout;
+			}
 			concurrency = endpoint.getConcurrency();
-			javaUtils.acceptIfNotNull(concurrency, instance::setConcurrency);
 		}
+		JavaUtils javaUtils = JavaUtils.INSTANCE
+				.acceptIfNotNull(batchSizeToUse, instance::setBatchSize)
```
artembilan: How about to try `JavaUtils.acceptOrElseIfNotNull()` API instead of manual `if..else`?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractRabbitListenerEndpoint.java
```diff
@@ -444,6 +480,8 @@ protected StringBuilder getEndpointDescription() {
 				append("' | queueNames='").append(this.queueNames).
 				append("' | exclusive='").append(this.exclusive).
 				append("' | priority='").append(this.priority).
+				append("' | batchSize='").append(this.batchSize).
+				append("' | batchReceiveTimeout='").append(this.batchReceiveTimeout).
```
artembilan, line 485: Good. 
May we see why `batchListener` is not included here?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListener.java
```diff
@@ -346,4 +347,31 @@
 	 */
 	String batch() default "";
 
+	/**
+	 * Override the container factory's {@code batchSize} property. Only applied by
+	 * container factories that support a batch size; it is ignored by others, such as
+	 * the {@code DirectRabbitListenerContainerFactory}.
+	 * This attribute only overrides the value; it does not enable consumer side batching
+	 * on its own - refer to the reference documentation.
```
artembilan: Isn't that just enough to say that `batch` attribute should be `true` to have an effect of this option, too?

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessorTests.java
```diff
@@ -334,6 +335,45 @@ public void concurrency() throws InterruptedException, ExecutionException {
...
+	public void batchOverridesNotSetByDefault() {
+		ConfigurableApplicationContext context = new AnnotationConfigApplicationContext(
+				getConfigClass(), SimpleMessageListenerTestBean.class);
+
+		RabbitListenerContainerTestFactory factory = context.getBean(RabbitListenerContainerTestFactory.class);
+		RabbitListenerEndpoint endpoint = factory.getListenerContainers().get(0).getEndpoint();
+		assertThat(endpoint.getBatchSize()).isNull();
+		assertThat(endpoint.getBatchReceiveTimeout()).isNull();
+
+		context.close();
```
artembilan: Let's see `try-with-resource` instead!

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/config/RabbitListenerContainerFactoryTests.java
```diff
@@ -84,6 +85,57 @@ public void createSimpleContainer() {
...
+
+		SimpleRabbitListenerEndpoint endpoint = new SimpleRabbitListenerEndpoint();
+		endpoint.setMessageListener(this.messageListener);
+		endpoint.setQueueNames("myQueue");
+		endpoint.setBatchSize(5);
+		endpoint.setBatchReceiveTimeout(2500L);
+
+		SimpleMessageListenerContainer container = this.factory.createListenerContainer(endpoint);
+
+		DirectFieldAccessor fieldAccessor = new DirectFieldAccessor(container);
```
artembilan: See `TestUtils.getPropertyValue()` instead.
Right, that is also based on the `DirectFieldAccessor`, and feels like not effective.
However we would have a consistency.
And we can revise `TestUtils` to cache objects for their `DirectFieldAccessor` to avoid some small overhead.
But different story indeed.

spring-rabbitmq-client/src/test/java/org/springframework/amqp/rabbitmq/client/config/RabbitAmqpListenerContainerFactoryTests.java
```diff
@@ -0,0 +1,104 @@
...
+import org.springframework.amqp.utils.test.TestUtils;
+import org.springframework.beans.factory.support.StaticListableBeanFactory;
+import org.springframework.messaging.handler.annotation.support.DefaultMessageHandlerMethodFactory;
+import org.springframework.util.ReflectionUtils;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+
+/**
+ * @author Martin Ferret
```
artembilan, line 36: `@since 4.2`
Over blank line after `@author`

spring-rabbitmq-client/src/test/java/org/springframework/amqp/rabbitmq/client/config/RabbitAmqpListenerContainerFactoryTests.java
```diff
@@ -0,0 +1,104 @@
...
+		MethodRabbitListenerEndpoint endpoint = createEndpoint();
+		endpoint.setBatchSize(5);
+
+		RabbitAmqpListenerContainer container = this.factory.createListenerContainer(endpoint);
+
+		assertThat(endpoint.getBatchListener()).isTrue();
+		assertThat(TestUtils.<Integer>getPropertyValue(container, "batchSize")).isEqualTo(5);
+	}
+
+	private MethodRabbitListenerEndpoint createEndpoint() {
```
artembilan: `static` ?

src/reference/antora/modules/ROOT/pages/amqp/receiving-messages/batch.adoc
```diff
@@ -88,5 +88,28 @@ When `consumerBatchEnabled` is `true`, the listener **must** be a batch listener
...
+@RabbitListener(queues = "batch.1", containerFactory = "consumerBatchContainerFactory",
+        batchSize = "10", batchReceiveTimeout = "${my.batch.timeout}")
+public void consumerBatch(List<Invoice> invoices) {
+    ...
+}
+----
+
+IMPORTANT: These attributes only *override values*; they do not enable consumer side batching on their own.
+`consumerBatchEnabled` remains a container factory property, and it is what turns on consumer side batching in the `SimpleMessageListenerContainer`; `batchReceiveTimeout` only has an effect in that mode.
+Setting `batchSize()` on a listener whose container factory does not have `consumerBatchEnabled` set to `true` therefore only changes the number of messages processed per acknowledgment or transaction, not the size of the list passed to the listener.
```
artembilan: Right. Same question here.
Isn't `batch = true` what makes it `consumerBatchEnabled` ?

artembilan (changes requested):
This is great!
I don't have to many concerns.

Thank you!

src/reference/antora/modules/ROOT/pages/amqp/receiving-messages/batch.adoc
MartinFerret: `batch = "true"` never sets `consumerBatchEnabled`. The implication runs the other way: `SimpleRabbitListenerContainerFactory.setConsumerBatchEnabled(true)` forces `setBatchListener(true)`, never the reverse.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListener.java
MartinFerret: Tightened it, thanks!

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/SimpleRabbitListenerContainerFactory.java
```diff
@@ -192,13 +193,13 @@ protected void initializeContainer(SimpleMessageListenerContainer instance,
...
 		super.initializeContainer(instance, endpoint);
 
+		// endpoint settings override the default factory settings
+		Integer endpointBatchSize = endpoint != null ? endpoint.getBatchSize() : null;
+		Long endpointBatchReceiveTimeout = endpoint != null ? endpoint.getBatchReceiveTimeout() : null;
+		String concurrency = endpoint != null ? endpoint.getConcurrency() : null;
 		JavaUtils javaUtils = JavaUtils.INSTANCE
-				.acceptIfNotNull(this.batchSize, instance::setBatchSize);
-		String concurrency = null;
-		if (endpoint != null) {
```
artembilan, line 198: I'm not sure in such a logic change.
So, we have here `if (endpoint != null) {` which covers all the property we would call from that endpoint.
Therefore, I would expect something like this inside of this `if` block:
```
.acceptOrElseIfNotNull(endpoint.getBatchSize(), this.batchSize, instance::setBatchSize)
```
All those ternary operators feel like a step back in the code flow.

artembilan (changes requested):
Looks promising!

Please, don't squash commits.
That will be easier for us to review just new change set.
We will squash them on merge.

Thank you!

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/SimpleRabbitListenerContainerFactory.java
MartinFerret: Good point, the ternaries were definitely a step back! I've put the `if (endpoint != null)` block back, and it now just reads the endpoint values so  the `acceptOrElseIfNotNull` calls right after it stay nice and flat.                                                                                                                                                                                                                                                                          
                                                                                                                                                                       
The one thing I didn't move inside the block is the application itself, and I wanted to flag why. 

`createListenerContainer()` delegates to `createListenerContainer(null)`, so `initializeContainer` genuinely can be handed a null endpoint, and if the `acceptOrElseIfNotNull` call lives inside the block the factory's own `batchSize` quietly stops reaching the container in that case.                                                                                                                                                                                                  
 
Happy to reshape it further if you had something different in mind!

artembilan: I see.
Can we put then that logic in the `else` block then?
One logic when `endpoint != null` and another otherwise.

artembilan: OK. Never mind.
I believe we can live with the code you have contributed.
Merging.
Thank you!

MartinFerret: Thank you Artem, what an amazing work you do here!

PR #3550, opened 2026-08-05
Fix error-spin in AMQP 1.0 listener container

artembilan:
`AmqpMessageListenerContainer.AmqpConsumer.run()` caught every exception from `Receiver.receive()`, handed it to the container `ErrorHandler` and retried immediately, with no backoff and no attempt to re-open the receiver link.
ProtonJ's `reconnectEnabled(true)` restores the *connection* but not the user-facing `Receiver`, so a broken link never healed.

Measured against a real broker:

| trigger | measured behaviour before this change |
|---|---|
| queue deleted (link detached) | ~2.2M loop iterations/sec, forever, `isRunning() == true` |
| `rabbitmqctl close_all_connections` | ~1.8M iterations/sec, forever; messages published afterwards are never delivered, while `isRunning()` still returns `true` |

So a single broker blip cost one burned CPU core plus a silently dead consumer that reported itself healthy.

## Container

* A `receive()` failure is now distinguished from a listener failure - only the former means the receiver link is broken, so a poison message no longer triggers link recovery and legit throughput is unaffected.
* The receiver link is re-opened according to a new `recoveryBackOff` option, a `FixedBackOff` with a `5 seconds` interval and unlimited attempts by default. `setRecoveryInterval(Duration)` is exposed for the interval alone, consistent with the other time-based options of this container.
* When the `BackOff` returns `BackOffExecution.STOP`, the consumer stops and removes itself from the container, so `isRunning()` reports `false` once no consumers are left instead of lying.
* Recovery sleeps in 100 ms slices, so a container `stop()` stays responsive rather than blocking for a whole recovery interval.
* `queuedDeliveries()` and `close()` tolerate an already dead link, so a container stop is not derailed by it.
* A recovery attempt which creates the link locally but then fails to open it closes that link before backing off. Otherwise, with the default unlimited attempts against a queue that stays absent, every attempt would leave another half-open link behind in the ProtonJ session.
* `recoveryBackOff` / `recoveryInterval` are also exposed on `AmqpMessageListenerContainerFactory`.

Two behaviour changes worth calling out:

1. A failing `receive()` with **no** `ErrorHandler` configured is now logged at `ERROR` and recovered, instead of throwing out of `run()` and killing the consumer thread.
2. Listener exceptions keep their dispatch semantics - `ErrorHandler` if present, rethrow otherwise - but the rethrow now unwinds into the new consumer cleanup in the `finally` block, so the consumer deregisters itself from the container. Previously such a listener exception left a dead consumer thread behind a container whose `isRunning()` still returned `true`; now `isRunning()` reports `false` once no consumers are left. The dispatch itself is unchanged, but this consequence is new.

## Tests

The same spin was turning `:spring-amqp-client:test` into an intermittent `OutOfMemoryError` (~50% of commercial CI runs) rather than a test failure. `EnableAmqpTests` wired a plain Mockito mock as the container `ErrorHandler`; a non-`stubOnly` mock retains every invocation together with its `AmqpException` argument and stack trace (~1KB each), so with the spin feeding it the heap grew ~70MB/s and the 2GB test worker was gone in under 30 seconds:

```
t= 5s errorHandler invocations=299352  usedHeapMB=417
t=10s errorHandler invocations=648927  usedHeapMB=739
t=15s errorHandler invocations=994561  usedHeapMB=1189
t=20s errorHandler invocations=1337893 usedHeapMB=1507
t=25s errorHandler invocations=1669201 usedHeapMB=1792
java.lang.OutOfMemoryError: Java heap space
```

A control run with a plain lambda `ErrorHandler` kept the heap flat at 36-186MB at the same spin rate, which is what identified the mock as the amplifier.

* The container `ErrorHandler` mock is created with `withSettings().stubOnly()` - such mocks do not record invocations. The existing `mockingDetails(errorHandler).isMock()` assertions still hold.
* Dropped the `rabbitmqctl set_log_level debug` calls, which reconfigured the *shared static* broker for the whole test JVM.
* Dropped the failure-path `LOG.error("LOGS FROM RABBITMQ CONTAINER: " + RABBITMQ.getLogs())` dumps, which materialised the entire container log into a `String` exactly when the heap was already exhausted - and one of them sat inside `@RetryingTest(10)`.
* Lowered the `org.apache.qpid.protonj2.client` logger from `trace` to `warn` and removed `ConnectionOptions.traceFrames(true)`.

New `AmqpMessageListenerContainerRecoveryTests` covers both paths against a real broker: a consumer that recovers its receiver after the queue is deleted and re-declared, and a consumer that gives up under a bounded `FixedBackOff` so that `isRunning()` flips to `false`. Both assert the error handler was invoked fewer than 100 times, which is the regression guard against the spin.

## Verification

`./gradlew :spring-amqp-client:check` is green locally, including `checkstyle`, `javadoc` and the Testcontainers-based tests.

One caveat for the reviewer: during these runs `AmqpListenerAnnotationTests.requestReplyAndContainerFactoryOverrides` failed once, on all 10 retries, with `ClientSessionRemotelyClosedException: function_clause` - a broker-side Erlang crash. I could not reproduce it in three subsequent full runs, and the pre-existing tests pass with this change applied while the new test class is excluded, so it looks like unrelated broker flakiness rather than a regression. Flagging it in case it is already known.

**Auto-cherry-pick to `4.1.x`**

### DIFF
--- a/spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpMessageListenerContainer.java
+++ b/spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpMessageListenerContainer.java
@@ -60,6 +60,9 @@ import org.springframework.util.MultiValueMap;
 import org.springframework.util.ObjectUtils;
 import org.springframework.util.ReflectionUtils;
 import org.springframework.util.StringUtils;
+import org.springframework.util.backoff.BackOff;
+import org.springframework.util.backoff.BackOffExecution;
+import org.springframework.util.backoff.FixedBackOff;
 
 /**
  * The {@link MessageListenerContainer} implementation for AMQP 1.0 protocol.
@@ -79,6 +82,10 @@ import org.springframework.util.StringUtils;
 public class AmqpMessageListenerContainer implements MessageListenerContainer, BeanNameAware {
 
 	private static final LogAccessor LOG = new LogAccessor(AmqpMessageListenerContainer.class);
+
+	private static final Duration DEFAULT_RECOVERY_INTERVAL = Duration.ofSeconds(5);
+
+	private static final long RECOVERY_SLEEP_SLICE = 100;
 
 	private final Lock lock = new ReentrantLock();
 
@@ -107,6 +114,9 @@ public class AmqpMessageListenerContainer implements MessageListenerContainer, B
 	private Duration receiveTimeout = Duration.ofSeconds(1);
 
 	private Duration gracefulShutdownPeriod = Duration.ofSeconds(30);
+
+	private BackOff recoveryBackOff =
+			new FixedBackOff(DEFAULT_RECOVERY_INTERVAL.toMillis(), FixedBackOff.UNLIMITED_ATTEMPTS);
 
 	private Executor taskExecutor = new SimpleAsyncTaskExecutor();
 
@@ -239,6 +249,32 @@ public class AmqpMessageListenerContainer implements MessageListenerContainer, B
 		this.receiveTimeout = receiveTimeout;
 	}
 
+	/**
+	 * Set an interval between recovery attempts of a failed consumer.
+	 * Default 5 seconds.
+	 * @param recoveryInterval the interval between recovery attempts.
+	 * @since 4.1.1
+	 * @see #setRecoveryBackOff(BackOff)
+	 */
+	public void setRecoveryInterval(Duration recoveryInterval) {
+		setRecoveryBackOff(new FixedBackOff(recoveryInterval.toMillis(), FixedBackOff.UNLIMITED_ATTEMPTS));
+	}
+
+	/**
+	 * Specify the {@link BackOff} for a failed consumer recovery.
+	 * The receiver link is not restored by the ProtonJ client itself, even when the connection
+	 * is re-established, so the consumer re-opens its receiver according to this {@link BackOff}.
+	 * The consumer is stopped when the {@link BackOff} returns {@link BackOffExecution#STOP},
+	 * and then the {@link #isRunning()} returns {@code false}
+	 * as soon as no consumers are left in this container.
+	 * The default is {@link FixedBackOff} with a 5-second interval and unlimited attempts.
+	 * @param recoveryBackOff the {@link BackOff} to use for the consumer recovery.
+	 * @since 4.1.1
+	 */
+	public void setRecoveryBackOff(BackOff recoveryBackOff) {
+		this.recoveryBackOff = recoveryBackOff;
+	}
+
 	@Override
 	public void afterPropertiesSet() {
 		Assert.state(this.queues != null, "At least one queue has to be provided for consuming.");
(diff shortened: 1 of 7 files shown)


spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpMessageListenerContainer.java
```diff
@@ -239,6 +248,33 @@ public void setReceiveTimeout(Duration receiveTimeout) {
...
 	}
 
+	/**
+	 * Specify the interval between recovery attempts of a failed consumer, in milliseconds.
+	 * The default is 5000 ms, that is, 5 seconds.
+	 * @param recoveryInterval the interval between recovery attempts.
+	 * @since 4.1.1
+	 * @see #setRecoveryBackOff(BackOff)
+	 */
+	public void setRecoveryInterval(long recoveryInterval) {
```
cppwfs: Should use seconds to be consistent with the class.

spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpMessageListenerContainer.java
```diff
@@ -239,6 +248,33 @@ public void setReceiveTimeout(Duration receiveTimeout) {
 		this.receiveTimeout = receiveTimeout;
 	}
 
+	/**
+	 * Specify the interval between recovery attempts of a failed consumer, in milliseconds.
+	 * The default is 5000 ms, that is, 5 seconds.
```
cppwfs: If we decide to leave this at `ms` then we don't need to put the conversion to seconds.

spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpMessageListenerContainer.java
```diff
@@ -239,6 +248,33 @@ public void setReceiveTimeout(Duration receiveTimeout) {
...
+	 * is re-established, so the consumer re-opens its receiver according to this {@link BackOff}.
+	 * The consumer is stopped when the {@link BackOff} returns {@link BackOffExecution#STOP},
+	 * and then the {@link #isRunning()} returns {@code false}
+	 * as soon as no consumers are left in this container.
+	 * The default is {@link FixedBackOff} with a 5 seconds interval and unlimited attempts.
+	 * @param recoveryBackOff the {@link BackOff} to use for the consumer recovery.
+	 * @since 4.1.1
+	 */
+	public void setRecoveryBackOff(BackOff recoveryBackOff) {
+		Assert.notNull(recoveryBackOff, "'recoveryBackOff' must not be null.");
```
cppwfs: Don't need to do the `assertNotNull`

spring-amqp-client/src/main/java/org/springframework/amqp/client/listener/AmqpMessageListenerContainer.java
```diff
@@ -306,6 +322,56 @@ public void start() {
...
+	}
+
+	/**
+	 * Remove a consumer which has stopped without a chance to recover,
+	 * so the {@link #isRunning()} does not report this container as running
+	 * when all its consumers are gone.
+	 * @param queue the queue the consumer was consuming from.
+	 * @param consumer the consumer to remove.
+	 */
+	private void removeConsumer(String queue, AmqpConsumer consumer) {
```
cppwfs: Shouldn't this removeConsumer go below the `run` method?

cppwfs (changes requested):

PR #3554, opened 2026-08-09
Upgrade Log4j to 2.26.1 and replace deprecated APIs

### LINKED ISSUE #3539: Upgrade to Log4J `2.26.x`
It reports something like:
```
Note: ~\spring-amqp\spring-rabbit\src\main\java\org\springframework\amqp\rabbit\log4j2\AmqpAppender.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
```

Amlan2000:
Fixes #3539

Upgrade Log4j from 2.25.5 to 2.26.1.

Replace the deprecated Log4j PatternLayout builder methods in `AmqpAppender` with their non-deprecated `set...` equivalents.

The following were updated:

- `withPattern` → `setPattern`
- `withCharset` → `setCharset`
- `withAlwaysWriteExceptions` → `setAlwaysWriteExceptions`
- `withNoConsoleNoAnsi` → `setNoConsoleNoAnsi`

Validation:
- `./gradlew :spring-rabbit:test`
- `./gradlew build`


artembilan (changes requested):
Hey, @Amlan2000 !

looking good!
Please, add your name to the `@author` list of the affected class.
I don’t see a technical or logical reason to rename from `with` to `set`: I would prefer without any suffix since this is a builder pattern , but that is not our project to decide. Can you find the discussion about changing to `set`?
And answering to your question: yes, we want the latest, but stable. If you would leave it as `2.26.0’, the next weekend Dependabot would pick it up and bump 😅.
Thank you for your contribution !
I’ll merge tomorrow when you address my minor feedback

Amlan2000:
Thanks @artembilan! I have added myself to the @author list.

Also, regarding the 'set' method changes, I agree with you that 'set' does not really make much sense here either. Just the method name with the builder pattern would probably have been cleaner.

However, I did some digging into the Log4j repository, and found this [discussion](https://github.com/apache/logging-log4j2/issues/3750). It looks like Log4j made this change for consistency and clarity, and deprecated the with methods as part of the transition toward Log4j 3.x. So I believe this is the reason for the change here.

Issue #3435, opened 2026-05-08
RabbitMQ 4.3 upgrade breaks retry/dead-letter expectations due to delivery-count semantics change


honnel:
**In what version(s) of Spring AMQP are you seeing this issue?**

Observed with Spring AMQP 4.0.3 (via Spring Boot 4.0.6).

**Describe the bug**

After upgrading RabbitMQ from 4.2.x to 4.3.x, retry/dead-letter behavior for quorum queues changes in a way that breaks existing assumptions in our Spring AMQP integration tests.

RabbitMQ 4.3 introduced separate counters:
- `x-acquired-count`: increments on every requeue/return
- `x-delivery-count`: increments only when a delivery attempt is considered failed

In our scenario, poison handling is effectively tied to `delivery-limit` and therefore to `x-delivery-count`. On RabbitMQ 4.3, returns that are not considered failed delivery attempts no longer advance `x-delivery-count`, which can prevent dead-lettering at the expected attempt threshold.

Reference:
- https://www.rabbitmq.com/blog/2026/04/23/rabbitmq-4.3-release#unlimited-returns

**To Reproduce**

1. Configure a quorum queue with `delivery-limit=4`.
2. Use a listener that always fails message processing.
3. Run the same retry/dead-letter integration test against:
   - RabbitMQ 4.2.5
   - RabbitMQ 4.3.0
4. Compare headers and dead-letter outcome (`x-delivery-count`, `x-acquired-count`).

**Expected behavior**

Retry/dead-letter behavior should remain consistent across RabbitMQ 4.2 and 4.3, or Spring AMQP should account for RabbitMQ 4.3 semantics so that effective requeue attempts are handled correctly.

In practice, Spring AMQP likely needs to consider `x-acquired-count` in addition to `x-delivery-count` for quorum queues on RabbitMQ 4.3+.

**Sample**

Minimal reproducible sample :
- https://github.com/honnel/spring-amqp-delivery-count-example

The sample contains integration tests that compare RabbitMQ 4.2.5 and 4.3.0 behavior for retry/dead-letter handling with quorum queues.

artembilan:
I'm not sure what is the question for Spring AMQP.
There is no where the logic checking that `x-delivery-count` either way.
According to your test application that is something what really should be done in that target application.
Both those flags are changed on the broker, not in the framework.

Please, elaborate what exactly you see should be fixed from Spring AMQP perspective.

Thanks

davidcoppens1:
[...]

PR #3559, opened 2026-08-14
GH-3507: Reject deliveries individually

### LINKED ISSUE #3507: RabbitMQ 4.3 upgrade still breaks retry/dead-letter expectations for synchronous listeners
4.1.0

**Describe the bug**

The bug is the same as described in https://github.com/spring-projects/spring-amqp/issues/3435

After upgrading RabbitMQ from 4.2.x to 4.3.x, the `x-delivery-count` is increased for asynchronous listeners (as expected) but it is still not increased for synchronous listeners. This means that the delivery-limit is never reached, resulting in endless retries.

**To Reproduce**

1. Configure a quorum queue with delivery-limit=4.
2. Use RabbitMQ 4.3.0 and spring-amqp 4.1.0
4. Send a message to:
  * a synchronous RabbitListener that always throws an exception
  * an asychronous RabbitListener that always returns a failed future

Compare headers and dead-letter outcome (x-delivery-count, x-acquired-count).

**Expected behavior**

Both messages should only be re-delivered four times and then sent to the dead-letter queue.

**Actual behavior**

The message to the synchronous RabbitListener is re-delivered infinitely and is never sent to the dead-letter queue. The message to the asynchronous RabbitListener is re-delivered four times and then sent to the dead-letter queue.

**Sample**

https://github.com/rstief-i/spring-amqp-delivery-count-example# based on the example from @honnel

artembilan:
Fixes #3507

## The problem

RabbitMQ bumps `x-delivery-count` only for deliveries rejected **individually**; it does not for a `basic.nack` with `multiple=true`.
Both listener containers issued exactly such a bulk nack for a listener that is not returning an async reply:

* `BlockingQueueConsumer.rollbackOnExceptionIfNecessary()` nacked the highest of the outstanding tags with `multiple=true` — the `tag < 0` branch, which `SimpleMessageListenerContainer` always takes unless `isAsyncReplies()`.
* `DirectMessageListenerContainer.SimpleConsumer.rollback()` nacked with `multiple = !isAsyncReplies()`.

So a quorum queue `x-delivery-limit` was never reached and a failing message was redelivered forever instead of being dead-lettered.
This is the report in #3507, and the same symptom as #3435 for the synchronous case.

## The change

* `DirectMessageListenerContainer.SimpleConsumer.rollback()` now rejects the failed delivery individually.
  This consumer handles a single delivery at a time — a producer-created batch is one delivery whose de-batched messages all carry its tag — hence there is exactly one delivery to reject.
* `BlockingQueueConsumer.rollbackOnExceptionIfNecessary()` rejects the outstanding delivery individually when there is exactly one of them: the default configuration, which includes a de-batched producer batch, again a single delivery tag for many messages.
  Consumer batching, or a `batchSize` greater than one, accumulate several outstanding deliveries instead and keep the cumulative nack.

Accumulated deliveries keep the cumulative nack because listener code may have settled some of them itself — `ConsumerBatchingTests` acks or rejects part of a batch and then throws — and re-settling an already settled delivery individually is a protocol violation rather than a per-call error, so the broker answers it with `PRECONDITION_FAILED - unknown delivery tag` and closes the whole channel, failing everything else in progress on it.
A cumulative nack tolerates such a delivery instead — the broker settles whatever is still outstanding up to the given tag — hence it is retained for that case.
A batch listener which needs each of its deliveries counted can reject them individually itself, which is already the way to handle the batch messages one by one.

With `AcknowledgeMode.NONE` there is nothing to settle at all — the broker settles a delivery as it sends it — so `SimpleConsumer` now skips the rejection in that mode.
It used to nack unconditionally whenever `ContainerUtils.isRejectManual()`, which the cumulative nack turned into a silent no-op, while an individual reject would close the channel.
That matters in particular for `DirectReplyToMessageListenerContainer`: a `NONE` container which tracks its in-use consumers by channel.

The success path bulk `basicAck(max, multiple=true)` is left alone — `x-delivery-count` only concerns rejects.

## Verification

Against a real RabbitMQ **4.3.0** — new `DeliveryLimitRedeliveryTests`, a quorum queue with `x-delivery-limit=3` and a listener that always throws:

* `simpleContainerRejectionIsCountedTowardsDeliveryLimit()` and `directContainerRejectionIsCountedTowardsDeliveryLimit()` — the message is dead-lettered after `limit + 1` deliveries. Both fail against the previous bulk nack with *"the message was redelivered indefinitely instead of being dead-lettered"*.

`o.s.a.rabbit.listener`, `annotation`, `config` and `core` are green against a live broker, and `./gradlew build -x test` passes across modules.

This also answers the "confirm with the RabbitMQ team" question from the issue thread empirically: on 4.3.0 individual rejects do bump the count, the bulk nack does not.

## Known follow-up, not in this PR

Accumulated deliveries (consumer batching, `batchSize > 1`) still use the cumulative nack, so `x-delivery-count` is not bumped for those individually — only the default single-delivery case is fixed here. Extending this to per-message counting in batch mode would need a way to know which deliveries a listener already settled itself without re-wrapping the channel, which is deliberately left out of this PR to keep it small and backportable.

## Backport

No `whats-new.adoc` entry (this is an internal fix, not a new capability), and confirmed to cherry-pick onto `4.1.x` and `4.0.x` without conflicts, so this is fit to backport down to `4.0.x`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
@@ -66,6 +66,7 @@ import org.springframework.amqp.rabbit.connection.ConnectionFactoryUtils;
 import org.springframework.amqp.rabbit.connection.RabbitResourceHolder;
 import org.springframework.amqp.rabbit.connection.RabbitUtils;
 import org.springframework.amqp.rabbit.connection.RoutingConnectionFactory;
+import org.springframework.amqp.rabbit.listener.adapter.AbstractAdaptableMessageListener;
 import org.springframework.amqp.rabbit.listener.api.ChannelAwareBatchMessageListener;
 import org.springframework.amqp.rabbit.listener.api.ChannelAwareMessageListener;
 import org.springframework.amqp.rabbit.support.DefaultMessagePropertiesConverter;
@@ -463,6 +464,28 @@ public abstract class AbstractMessageListenerContainer extends ObservableListene
 	@Override
 	public @Nullable MessageListener getMessageListener() {
 		return this.messageListener;
+	}
+
+	/**
+	 * Return true when the listener may settle a delivery itself, which happens only if it
+	 * is given the consumer {@link Channel}. The container cannot then reject a failed
+	 * delivery individually: re-settling an already settled delivery is a protocol
+	 * violation which closes the whole channel.
+	 * @return true if the listener may settle a delivery itself.
+	 * @since 4.2
+	 * @see AbstractAdaptableMessageListener#maySettleDelivery()
+	 */
+	boolean listenerMaySettleDelivery() {
+		MessageListener listener = getMessageListener();
+		if (!(listener instanceof ChannelAwareMessageListener) || !isExposeListenerChannel()) {
+			// A plain MessageListener is not given any channel; a not exposed one is not the consumer channel.
+			return false;
+		}
+		if (listener instanceof AbstractAdaptableMessageListener adaptableListener) {
+			return adaptableListener.maySettleDelivery();
+		}
+		// An arbitrary ChannelAwareMessageListener does whatever it wants with the channel.
+		return true;
 	}
 
 	/**
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
@@ -472,7 +472,7 @@ public abstract class AbstractMessageListenerContainer extends ObservableListene
 	 * delivery individually: re-settling an already settled delivery is a protocol
 	 * violation which closes the whole channel.
 	 * @return true if the listener may settle a delivery itself.
-	 * @since 4.2
+	 * @since 4.0.6
 	 * @see AbstractAdaptableMessageListener#maySettleDelivery()
 	 */
 	boolean listenerMaySettleDelivery() {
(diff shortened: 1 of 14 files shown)


artembilan:
Pushed a rework of this fix (force-push, so the diff against the previous revision isn't meaningful — see the new commits directly):

* The individual-reject fix is now conditioned on whether the listener could have settled the delivery itself. Re-settling an already-settled delivery individually is a protocol violation (`PRECONDITION_FAILED - unknown delivery tag`), which closes the whole channel — so a delivery is only rejected individually when the container is the sole party settling it.
  - `AbstractAdaptableMessageListener.maySettleDelivery()` (new) is the listener-side answer: `true` by default, narrowed in `MessagingMessageListenerAdapter` for methods that receive a `Channel`, async replies, `RabbitListenerErrorHandler`, and manual-ack error handling; `BatchMessagingMessageListenerAdapter` and unresolved `@RabbitHandler` methods are assumed to settle.
  - `AbstractMessageListenerContainer.listenerMaySettleDelivery()` combines that with the container's own configuration (`exposeListenerChannel`, listener type).
  - `BlockingQueueConsumer.rollbackOnExceptionIfNecessary()` and `DirectMessageListenerContainer.SimpleConsumer.rollback()` reject individually only when the listener cannot have settled the delivery; otherwise the cumulative nack (which tolerates an already-settled delivery) is kept, same as before this PR.
* Added `DeliveryLimitRedeliveryTests` against a live broker (RabbitMQ 4.3.0) confirming the dead-lettering behavior for both containers, plus mock-based coverage in `BlockingQueueConsumerTests`, `DirectMessageListenerContainerMockTests`, `ExternalTxManagerTests`, and `MessagingMessageListenerAdapterTests`.
* Added a further test in `LocallyTransactedTests` covering a `ChannelAwareMessageListener` (which is handed the channel and may settle its own delivery): confirms the cumulative nack is kept for it, not the individual reject.

Compiles clean and the full affected test surface passes, including repeated runs of the new/changed tests to rule out flakiness.

artembilan:
Merged as: https://github.com/spring-projects/spring-amqp/commit/110bbbb8d35c4a0a35a5467c709906ad91f8f466

PR #3560, opened 2026-08-16
GH-3448: Close channels if confirm wait fails

### LINKED ISSUE #3448: CachingConnectionFactory channels leak in publisher confirms and checkoutTimeout equal to 0

3.2.9
Spring boot: 3.5.13

**Describe the bug**

A channels leak has been detected when the App lost the connection with RabbitMQ cluster after a crash. When the RabbitMQ is restored, the App reconnects and the **message publishing is totally stuck**.

In my research I saw that 2 years ago a [fix](https://github.com/spring-projects/spring-amqp/discussions/2637#discussioncomment-9497292) was applied for a similar scenario but it seems that it has not been completely solved.

**To Reproduce**

I have created a public repository [[link](https://github.com/miguele86/rabbit-channels-leak-demo)] with all the infrastructure and tools to reproduce and monitor the issue. All the steps are described in the [README](https://github.com/miguele86/rabbit-channels-leak-demo#rabbitmq-channel-leak-demo) file.

I only was able to reproduce this issue in this scenario with high concurrency during the RabbitMQ disaster recovery and network degradation through toxiproxy to force messages to be sent to the DLQ.

App configuration key points:
- publisher-confirm-type: correlated
- high prefetch value of 200
- low cache size of 25 (default)
- checkout-timeout of 0 (default). when this value **is higher than zero** there are **not channels leak**

**Observations**

I can see in the memory dump when the RabbitMQ cluster goes down and the app lost the connection, there are a lot of **ChannelN** instances when they should be cleaned.

Image

When RabbitMQ cluster is restored and the app reconnects, 200 messages are consumed (by prefetch value) and when the **streambridge** client tries to get a channel is **totally blocked** and **the ACK is never sent** to the broker to process the consumed messages. This picture shows the point where is blocked:

Image

When the **channel-rpc-timeout** is reached (10 minutes by default), a **ChannelContinuationTimeoutException** is thrown and the app tries to process the 200 prefetched messages again. This causes channels to accumulate until the 
  ? maximum allowed per connection (2047) is reached:

Image

And this exception is thrown:
 org.springframework.amqp.**AmqpResourceNotAvailableException**: The channelMax limit is reached. Try later.

This issue **blocks totally the app** and must be restarted to work properly again.

In grafana can be observed how the app tries to create new channels every 10 minutes:

Image

How the messages from the consumer queue are not being processed:

Image

Image

In the sample repository I have attached the **log and memory dumps (hprof) files** generated when this issue occurs in **tar.gz files.**

**Expected behavior**

The **channels created on demand** when the channel size is exceeded **should be cleaned and closed** correctly like when the **checkout-timeout is higher than zero** and the channels are being reused **to avoid blocking message publishing**

Thanks in advance.
Regards.

arimu1:
Fixes #3448

When publisher confirms are correlated and `checkoutTimeout` is 0 (the default), `returnToCache()` parks overflow channels until `waitForConfirms()` completes. After an abrupt broker crash that wait can fail with exceptions other than `ShutdownSignalException` or `TimeoutException`. Those channels were left in `channelsAwaitingAcks` and never physically closed, so `ChannelN` instances accumulated and publishing stayed blocked after reconnect.

This closes the channel for any failed confirm wait (including interrupt) and drops it from `channelsAwaitingAcks`, matching the existing timeout/shutdown path.

## Test plan
- [x] `./gradlew :spring-rabbit:test --tests org.springframework.amqp.rabbit.connection.CachingConnectionFactoryTests --tests org.springframework.amqp.rabbit.core.RabbitTemplatePublisherCallbacksIntegration1Tests.testPublisherConfirmNotReceived`
- [x] `:spring-rabbit:checkstyleMain` and `:spring-rabbit:checkstyleTest`
- [x] New unit test: unexpected `waitForConfirms()` failure with `checkoutTimeout=0` physically closes the channel and a later publish proceeds

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
@@ -1328,21 +1328,30 @@ public class CachingConnectionFactory extends AbstractConnectionFactory
 								}
 								catch (InterruptedException ex) {
 									Thread.currentThread().interrupt();
+									closeAfterFailedConfirmWait();
 								}
-								catch (ShutdownSignalException | TimeoutException ex) {
+								catch (Exception ex) {
 									// The channel didn't handle confirms, so close it altogether to avoid
 									// memory leaks for pending confirms
-									try {
-										physicalClose();
-									}
-									catch (@SuppressWarnings(UNUSED) Exception e) {
-									}
+									closeAfterFailedConfirmWait();
 								}
 							});
 				}
 			}
 			else {
 				doReturnToCache(proxy);
+			}
+		}
+
+		private void closeAfterFailedConfirmWait() {
+			Channel targetChannel = this.target;
+			if (targetChannel != null) {
+				this.theConnection.channelsAwaitingAcks.remove(targetChannel);
+			}
+			try {
+				physicalClose();
+			}
+			catch (@SuppressWarnings(UNUSED) Exception e) {
 			}
 		}
 
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
@@ -1312,7 +1312,8 @@ public class CachingConnectionFactory extends AbstractConnectionFactory
 					&& this.publisherConfirms
 					&& proxy instanceof PublisherCallbackChannel publisherCallbackChannel) {
 
-				this.theConnection.channelsAwaitingAcks.put(this.target, proxy);
+				Channel channelAwaitingAcks = this.target;
+				this.theConnection.channelsAwaitingAcks.put(channelAwaitingAcks, proxy);
 				AtomicBoolean ackCallbackCalledImmediately = new AtomicBoolean();
 				publisherCallbackChannel
 						.setAfterAckCallback(c -> {
@@ -1328,12 +1329,12 @@ public class CachingConnectionFactory extends AbstractConnectionFactory
 								}
 								catch (InterruptedException ex) {
 									Thread.currentThread().interrupt();
-									closeAfterFailedConfirmWait();
+									closeAfterFailedConfirmWait(channelAwaitingAcks);
 								}
 								catch (Exception ex) {
 									// The channel didn't handle confirms, so close it altogether to avoid
 									// memory leaks for pending confirms
-									closeAfterFailedConfirmWait();
+									closeAfterFailedConfirmWait(channelAwaitingAcks);
 								}
 							});
 				}
@@ -1343,13 +1344,12 @@ public class CachingConnectionFactory extends AbstractConnectionFactory
 			}
 		}
 
-		private void closeAfterFailedConfirmWait() {
-			Channel targetChannel = this.target;
-			if (targetChannel != null) {
-				this.theConnection.channelsAwaitingAcks.remove(targetChannel);
+		private void closeAfterFailedConfirmWait(@Nullable Channel channelAwaitingAcks) {
+			if (channelAwaitingAcks != null) {
+				this.theConnection.channelsAwaitingAcks.remove(channelAwaitingAcks);
 			}
 			try {
-				physicalClose();
+				physicalClose(channelAwaitingAcks);
 			}
 			catch (@SuppressWarnings(UNUSED) Exception e) {
 			}
@@ -1413,13 +1413,17 @@ public class CachingConnectionFactory extends AbstractConnectionFactory
 			}
 		}
 
-		@SuppressWarnings("NullAway") // Dataflow analysis limitation
 		private void physicalClose() throws IOException, TimeoutException {
+			physicalClose(this.target);
+		}
+
+		@SuppressWarnings("NullAway") // Dataflow analysis limitation
+		private void physicalClose(@Nullable Channel channelToClose) throws IOException, TimeoutException {
 			if (logger.isDebugEnabled()) {
-				logger.debug("Closing cached Channel: " + this.target);
+				logger.debug("Closing cached Channel: " + channelToClose);
 			}
 			RabbitUtils.clearPhysicalCloseRequired();
-			if (this.target == null) {
+			if (channelToClose == null) {
 				return;
 			}
 			boolean async = false;
@@ -1428,22 +1432,24 @@ public class CachingConnectionFactory extends AbstractConnectionFactory
 						(ConfirmType.CORRELATED.equals(CachingConnectionFactory.this.confirmType) ||
 								CachingConnectionFactory.this.publisherReturns)) {
 					async = true;
-					asyncClose();
+					asyncClose(channelToClose);
 				}
 				else {
-					this.target.close();
-					if (this.target instanceof AutorecoveringChannel auto) {
+					channelToClose.close();
+					if (channelToClose instanceof AutorecoveringChannel auto) {
 						ClosingRecoveryListener.removeChannel(auto);
 					}
 				}
 			}
 			catch (AlreadyClosedException e) {
 				if (logger.isTraceEnabled()) {
-					logger.trace(this.target + " is already closed", e);
+					logger.trace(channelToClose + " is already closed", e);
 				}
 			}
 			finally {
-				this.target = null;
+				if (this.target == channelToClose) {
+					this.target = null;
+				}
 				if (!async) {
 					releasePermitIfNecessary();
 				}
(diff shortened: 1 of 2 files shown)


artembilan (commented):
Thanks for tracking this down — the leak/blocked-publishing symptom is real and the general direction (closing the channel and dropping it from `channelsAwaitingAcks` on any failed confirm wait) is right. One correctness concern before merge, though.

`closeAfterFailedConfirmWait()` re-reads `this.target` at execution time instead of using the channel reference that was actually inserted into `channelsAwaitingAcks`:

```java
private void closeAfterFailedConfirmWait() {
    Channel targetChannel = this.target;   // re-read, not necessarily the failed channel
    if (targetChannel != null) {
        this.theConnection.channelsAwaitingAcks.remove(targetChannel);
    }
    physicalClose();                        // closes this.target, not necessarily the failed channel
}
```

`publisherCallbackChannel.waitForConfirms(...)` is invoked through the dynamic proxy (`CachedChannelInvocationHandler.invoke`). When that call fails **and** `this.target.isOpen()` is already `false` — which is exactly the realistic broker-crash case this PR targets — `invoke()`'s own `InvocationTargetException` handler reassigns `this.target` to a freshly created bare channel *before* rethrowing, i.e. before this `catch (Exception ex)` block in `returnToCache` ever runs.

So in that scenario:
- `channelsAwaitingAcks.remove(targetChannel)` removes with the wrong key — the entry that's actually keyed by the old, failed channel is never removed, so the leak this PR is meant to fix can persist.
- `physicalClose()` tears down `this.target`, which by now is the brand-new replacement channel `invoke()` just created for future use — potentially reproducing the "publishing stays blocked after reconnect" symptom.

The new test doesn't exercise this path because `mockChannel.isOpen()` is stubbed to unconditionally return `true`, so `invoke()`'s reconnection branch never triggers — not representative of a real broker crash, where `isOpen()` would flip to `false`.

Suggested fix: capture the channel reference in a local variable at the point it's `put` into `channelsAwaitingAcks` (before scheduling the async task), and thread that captured reference through to the failure handler for both the map removal and the close, instead of re-reading `this.target`.

artembilan (changes requested):
Please, follow up with my review.

Thanks

arimu1:
Thanks @artembilan — that reconnect race is real.

`waitForConfirms()` goes through the proxy. When it fails with `isOpen() == false`, `invoke()` replaces `this.target` *before* the `returnToCache` catch runs, so re-reading `this.target` removed/closed the replacement and left the failed channel in `channelsAwaitingAcks`.

Follow-up `cf48d5e65`:
- capture the `Channel` at `channelsAwaitingAcks.put` and use that same ref for `remove` + `physicalClose`
- `physicalClose` only nulls `this.target` if it is still that captured channel
- test now flips `mockChannel.isOpen()` to `false` inside `waitForConfirms()`, and asserts the replacement channel is not closed

`:spring-rabbit:test` CachingConnectionFactoryTests 48/0 fail/1 skip (JDK 25); checkstyle main+test clean.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
```diff
@@ -1346,6 +1344,17 @@ private void returnToCache(ChannelProxy proxy) {
 			}
 		}
 
+		private void closeAfterFailedConfirmWait(@Nullable Channel channelAwaitingAcks) {
+			if (channelAwaitingAcks != null) {
+				this.theConnection.channelsAwaitingAcks.remove(channelAwaitingAcks);
```
artembilan: **Possible race: the ack path and the failed-wait path can both claim the same channel.**

`closeAfterFailedConfirmWait` discards the return value of `channelsAwaitingAcks.remove(channelAwaitingAcks)`. Compare with the success path at line 1321, which uses `doReturnToCache(channelsAwaitingAcks.remove(c))` — the `remove()` result is the ownership token that decides whether the channel goes back to the cache.

Scenario: the confirm ack fires and `doReturnToCache` puts the channel back in `channelList` (or it gets checked out again) at roughly the same time `waitForConfirms` throws for an unrelated reason (e.g. the connection drops right after the ack was already processed). With the `catch (Exception ex)` now widened from the previous `ShutdownSignalException | TimeoutException`, that path fires more often. Because the `remove()` result here is ignored, `physicalClose(channelAwaitingAcks)` still runs and closes/async-closes the raw channel even though it may already be back in the cache or in active use by another publisher on a different thread — surfacing as a spurious `AlreadyClosedException`/IOException to an unrelated caller.

Suggest gating on the remove result, mirroring the ack path:
```java
if (channelAwaitingAcks == null || this.theConnection.channelsAwaitingAcks.remove(channelAwaitingAcks) == null) {
    return;
}
```

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
```diff
@@ -1312,7 +1312,8 @@ private void returnToCache(ChannelProxy proxy) {
 					&& this.publisherConfirms
 					&& proxy instanceof PublisherCallbackChannel publisherCallbackChannel) {
 
-				this.theConnection.channelsAwaitingAcks.put(this.target, proxy);
+				Channel channelAwaitingAcks = this.target;
+				this.theConnection.channelsAwaitingAcks.put(channelAwaitingAcks, proxy);
```
artembilan, line 1320: `channelAwaitingAcks` is read from `this.target` without any lock and passed straight into `ConcurrentHashMap.put(channelAwaitingAcks, proxy)`. If `this.target` happens to be `null` here, this throws `NullPointerException` immediately (`ConcurrentHashMap` rejects null keys), which would propagate out of `returnToCache`/`logicalClose` from a channel-close call. This is pre-existing (the line only wraps the same `this.target` read into a local var), but it's now inconsistent with the rest of the diff: `closeAfterFailedConfirmWait` and `physicalClose` were both changed to accept `@Nullable Channel` specifically to guard against a null channel — while the one place that actually stores the channel as a map key has no such guard.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
```diff
@@ -1328,15 +1329,12 @@ private void returnToCache(ChannelProxy proxy) {
 								}
 								catch (InterruptedException ex) {
 									Thread.currentThread().interrupt();
+									closeAfterFailedConfirmWait(channelAwaitingAcks);
```
artembilan, line 1336: On `InterruptedException` this now also calls `closeAfterFailedConfirmWait`, which is good when the channel truly needs cleanup — but if the interrupt is caused by the channels-executor shutting down (`shutdownNow()`), `physicalClose` → `asyncClose` will try to `executorService.execute(...)` on that same (now-shutting-down) executor and get a `RejectedExecutionException`. `asyncClose`'s `catch (RuntimeException e)` only does `inFlightAsyncCloses.release(channel)`; it never calls `channel.close()` and never calls `releasePermitIfNecessary()` (which `physicalClose` skipped because `async` was set to `true` before the rejection). So in the exact scenario most likely to interrupt this wait (executor shutdown), the newly-added cleanup silently does nothing — reintroducing the channel/permit leak this PR is meant to fix.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java
```diff
@@ -1346,6 +1344,17 @@ private void returnToCache(ChannelProxy proxy) {
 			}
 		}
 
+		private void closeAfterFailedConfirmWait(@Nullable Channel channelAwaitingAcks) {
+			if (channelAwaitingAcks != null) {
+				this.theConnection.channelsAwaitingAcks.remove(channelAwaitingAcks);
+			}
+			try {
+				physicalClose(channelAwaitingAcks);
```
artembilan: Minor efficiency note: when we get here, `waitForConfirms` has already thrown (channel didn't handle confirms). If the confirm type is `CORRELATED`, `physicalClose` routes to `asyncClose`, which calls `channel.waitForConfirmsOrDie(getCloseTimeout())` again before closing. That's a second full `closeTimeout` wait on an already-broken channel, tying up a channels-executor thread for the duration, before the channel is actually closed.

artembilan (changes requested):
LGTM.
Will be merged when PR build is green.

Thank you!

artembilan (approved):
Sorry.
Meant to approve.

artembilan (changes requested):
We cannot pass the build for 30 mins.
Re-runs don't help.
Feels like the change you suggest is braking something.
Please, run the whole project locally `gradlew clean check` to see if something comes up.

Thanks

artembilan:
Merged as: https://github.com/spring-projects/spring-amqp/commit/f4ae4dbe07153777d0c9b05281dbf7cfaaf2b4e8.

PR #3610, opened 2026-08-29
GH-3604: Apply default container factory name before configurers

### LINKED ISSUE #3604: RabbitListenerEndpointRegistrar containerFactoryBeanName can not be set to null in Kotlin
### Discussed in https://github.com/spring-projects/spring-amqp/discussions/3600

<div type='discussions-op-text'>

<sup>Originally posted by **JMsand** August 28, 2026</sup>
Hey folks, 

While migrating some RabbitMQ code from 3  to 4.1 i had problems with code relying on setting the defaultContainerFactoryBeanName in RabbitListenerAnnotationBeanPostProcessor.

The code tries to prevent  that in the RabbitListenerAnnotationBeanPostProcessor afterSingletonsInstantiated() method  setContainerFactoryBeanName(String)  is called with a default value which overrides a possible previously set factory bean name.

The change seems to be in this commit :
https://github.com/spring-projects/spring-amqp/commit/0173186

This is the code  in spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
Screenshot 2026-08-28 205603

And with the JSpecify migration it made that field cannot be set to null in Kotlin. which makes afterSingletonsInstantiated overwrite the bean name.

So my question is was this intentional? and if not, is it possible that this can be reversed? I think i canmake an PR for this but let me know what is preferred.

Regards and thanks

JM</div>

kalayciburak:
Fixes #3604

`afterSingletonsInstantiated` copied the default container factory bean name onto the registrar after `RabbitListenerConfigurer` ran, so a name set from a configurer was overwritten.

Apply the default name first. Configurers can still override it. The bean name stays required; null is not supported.

## Test plan

- `./gradlew :spring-rabbit:test --tests "org.springframework.amqp.rabbit.annotation.RabbitListenerAnnotationBeanPostProcessorTests"`
- `./gradlew :spring-rabbit:test --tests "org.springframework.amqp.rabbit.listener.RabbitListenerEndpointRegistrarTests"`

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
@@ -98,7 +98,7 @@ import org.springframework.validation.Validator;
  * @author Artem Bilan
  * @author Ngoc Nhan
  * @author Martin Ferret
- * @author Burak KALAYCI
+ * @author Burak Kalayci
  *
  * @since 1.4
  *
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
@@ -98,6 +98,7 @@ import org.springframework.validation.Validator;
  * @author Artem Bilan
  * @author Ngoc Nhan
  * @author Martin Ferret
+ * @author Burak KALAYCI
  *
  * @since 1.4
  *
@@ -198,6 +199,7 @@ public class RabbitListenerAnnotationBeanPostProcessor
 	public void afterSingletonsInstantiated() {
 		super.afterSingletonsInstantiated();
 		this.registrar.setBeanFactory(getBeanFactory());
+		this.registrar.setContainerFactoryBeanName(this.defaultContainerFactoryBeanName);
 
 		if (getBeanFactory() instanceof ListableBeanFactory lbf) {
 			Map<String, RabbitListenerConfigurer> instances =
@@ -216,8 +218,6 @@ public class RabbitListenerAnnotationBeanPostProcessor
 			}
 			this.registrar.setEndpointRegistry(this.endpointRegistry);
 		}
-
-		this.registrar.setContainerFactoryBeanName(this.defaultContainerFactoryBeanName);
 
 		// Set the custom handler method factory once resolved by the configurer
 		MessageHandlerMethodFactory handlerMethodFactory = this.registrar.getMessageHandlerMethodFactory();
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessorTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessorTests.java
@@ -45,9 +45,12 @@ import org.springframework.amqp.rabbit.config.RabbitListenerContainerTestFactory
 import org.springframework.amqp.rabbit.listener.AbstractRabbitListenerEndpoint;
 import org.springframework.amqp.rabbit.listener.MethodRabbitListenerEndpoint;
 import org.springframework.amqp.rabbit.listener.RabbitListenerEndpoint;
+import org.springframework.amqp.rabbit.listener.RabbitListenerEndpointRegistrar;
 import org.springframework.amqp.rabbit.listener.RabbitListenerEndpointRegistry;
 import org.springframework.amqp.rabbit.listener.SimpleMessageListenerContainer;
+import org.springframework.amqp.utils.test.TestUtils;
 import org.springframework.beans.factory.BeanCreationException;
+import org.springframework.beans.factory.support.StaticListableBeanFactory;
 import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.context.annotation.AnnotationConfigApplicationContext;
 import org.springframework.context.annotation.Bean;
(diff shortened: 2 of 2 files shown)


artembilan (changes requested):
>  so setting the default container
factory bean name to null does not overwrite a value already
set on RabbitListenerEndpointRegistrar

I think we need to step back and revise the target user con figuration instead.
The `RabbitListenerAnnotationBeanPostProcessor` states it clear:
```
	/**
	 * Set the name of the {@link RabbitListenerContainerFactory} to use by default.
	 * <p>If none is specified, "rabbitListenerContainerFactory" is assumed to be defined.
	 * @param containerFactoryBeanName the {@link RabbitListenerContainerFactory} bean name.
	 */
	public void setContainerFactoryBeanName(String containerFactoryBeanName) {
```
I don't see how `null` can slip here, even if Kotlin.
So, if you migrating code provides null, it has to stop doing that since we don't support `null` for such a bean name.

Now another part of the claim.
You see the `RabbitListenerEndpointRegistrar` is internal to the `RabbitListenerAnnotationBeanPostProcessor`:
```
private final RabbitListenerEndpointRegistrar registrar = new RabbitListenerEndpointRegistrar();
```

Does not have access and cannot be modified.
Therefore, it is absolutely not clear how something different could be set into the registrar without post-processor.

Please, elaborate, but I feel like this fix and ticket should be rejected.

Thank you for understanding!

kalayciburak:
RabbitListenerConfigurer.configureRabbitListeners gets that same registrar instance (EnableRabbit even documents it). afterSingletonsInstantiated runs those first, then copies defaultContainerFactoryBeanName over whatever they set.

3.x only did that copy when the BPP field was non-null, so some kotlin code used setContainerFactoryBeanName(null) as a skip. JSpecify made the setter non-null and dropped the guard.

null as a bean name is a weird API though, agreed. the normal paths still work: set the name on the BPP, or registrar.setContainerFactory(instance) (instance is resolved first, so the later name copy doesnt clobber it).

i can close this if you want to reject the ticket. or drop the nullable setter and skip the copy only when the registrar already has a name.

artembilan (changes requested):
> `RabbitListenerConfigurer.configureRabbitListeners` gets that same registrar instance (`EnableRabbit` even documents it). `afterSingletonsInstantiated` runs those first

I see. Thank you for the pointer!
So, the logic now is like this:
```
for (RabbitListenerConfigurer configurer : instances.values()) {
				configurer.configureRabbitListeners(this.registrar);
			}
		}
...
		this.registrar.setContainerFactoryBeanName(this.defaultContainerFactoryBeanName);
```
Where `this.registrar.setContainerFactoryBeanName()` could be called in one of those `RabbitListenerConfigurer` instances.
So, I think we should just flip the order, and have `this.registrar.setContainerFactoryBeanName` before calling configurers.
No any other annotations must be there. 
The bean name must not be null. 
That's it.
If Kotlin app does that, then it is wrong and should revise its logic.

kalayciburak:
flipped the order, default name goes on the registrar before configurers. no nullable, bean name stays required.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
```diff
@@ -98,6 +98,7 @@
  * @author Artem Bilan
  * @author Ngoc Nhan
  * @author Martin Ferret
+ * @author Burak KALAYCI
```
artembilan: I think I see you are using your last name as normal just capitalized name in other your contributions, not fully in upper case.
Why really fully uppercase and why not consistent?
For example: https://github.com/spring-projects/spring-integration/pull/11270/files#diff-3c11e3526541242e4325fb9e67efc1910d75a1b7fa15e40e78508dbb5dc3c499R41

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessorTests.java
```diff
@@ -371,6 +375,24 @@ public void batchOverridesResolvedFromExpressionAndPlaceholder() {
...
+		postProcessor.setBeanFactory(beanFactory);
+		postProcessor.setEndpointRegistry(new RabbitListenerEndpointRegistry());
+
+		postProcessor.afterSingletonsInstantiated();
+
+		RabbitListenerEndpointRegistrar registrar =
+				TestUtils.getPropertyValue(postProcessor, "registrar");
+		assertThat(registrar).isNotNull();
+		String containerFactoryBeanName = TestUtils.getPropertyValue(registrar, "containerFactoryBeanName");
+		assertThat(containerFactoryBeanName).isEqualTo("customFactory");
```
artembilan: We don't need so much code.
I think just this would be enough:
```
assertThat((String) TestUtils.getPropertyValue(postProcessor, "registrar.containerFactoryBeanName")).isEqualTo("customFactory");
```

Plus I see you have ignored my comment about explicit cast since we want to back-port the fix down to `7.0.x` where nor generic argument on the `TestUtils.getPropertyValue()` yet.

artembilan (changes requested):

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessor.java
kalayciburak: yep leftover habit, switched to Kalayci

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/annotation/RabbitListenerAnnotationBeanPostProcessorTests.java
kalayciburak: simplified, plus the cast for 7.0.x

PR #3616, opened 2026-09-03
Fix `includeStackTrace` when frame_max is unknown

renechoi:
`includeStackTrace(true)` stores `-1` in `maxStackTraceLength` ("resolve the limit from frame_max") when the error template is a `RabbitTemplate`. `RabbitUtils.getMaxFrame()` returns `-1` when it cannot determine the negotiated frame_max, and `processStackTrace()` only overwrites the field when the resolved value is positive. The field therefore stays `-1`, and `recover()` gates the header on `maxStackTraceLength > 0`, so the `x-exception-stacktrace` header that was opted in for is dropped silently.

One field cannot be both the on/off switch and the truncation limit. This gates the header on a dedicated flag; the resolved-limit path is unchanged.

New test `shouldIncludeTheStacktraceWhenTheFrameMaxCannotBeDetermined` fails on `main`, passes here. `:spring-rabbit:test`: 1215 tests, 0 failures.

### DIFF
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/retry/RepublishMessageRecoverer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/retry/RepublishMessageRecoverer.java
@@ -84,7 +84,7 @@ public class RepublishMessageRecoverer implements MessageRecoverer {
 
 	private int frameMaxHeadroom = DEFAULT_FRAME_MAX_HEADROOM;
 
-	private volatile boolean includeStackTrace;
+	private boolean includeStackTrace;
 
 	private volatile Integer maxStackTraceLength = 0;
 
@@ -163,6 +163,8 @@ public class RepublishMessageRecoverer implements MessageRecoverer {
 
 	/**
 	 * Set {@code true} to add header with error stack trace.
+	 * When the truncation limit cannot be resolved from the negotiated {@code frame_max},
+	 * the stack trace is published untruncated.
 	 * @param includeStackTrace {@code true} to add header with error stack trace.
 	 * @return this.
 	 * @since 4.0.5
--- a/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/retry/RepublishMessageRecoverer.java
+++ b/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/retry/RepublishMessageRecoverer.java
@@ -50,6 +50,7 @@ import org.springframework.util.Assert;
  * @author James Carr
  * @author Gary Russell
  * @author Artem Bilan
+ * @author Rene Choi
  *
  * @since 1.3
  */
@@ -82,6 +83,8 @@ public class RepublishMessageRecoverer implements MessageRecoverer {
 	private String errorRoutingKeyPrefix = "error.";
 
 	private int frameMaxHeadroom = DEFAULT_FRAME_MAX_HEADROOM;
+
+	private volatile boolean includeStackTrace;
 
 	private volatile Integer maxStackTraceLength = 0;
 
@@ -165,6 +168,7 @@ public class RepublishMessageRecoverer implements MessageRecoverer {
 	 * @since 4.0.5
 	 */
 	public RepublishMessageRecoverer includeStackTrace(boolean includeStackTrace) {
+		this.includeStackTrace = includeStackTrace;
 		this.maxStackTraceLength =
 				includeStackTrace ? (!(this.errorTemplate instanceof RabbitTemplate) ? Integer.MAX_VALUE : -1) : 0;
 		return this;
@@ -210,7 +214,7 @@ public class RepublishMessageRecoverer implements MessageRecoverer {
 			if (truncatedExceptionMessage != null) {
 				exceptionMessage = truncatedExceptionMessage;
 			}
-			if (this.maxStackTraceLength > 0) {
+			if (this.includeStackTrace) {
 				headers.put(X_EXCEPTION_STACKTRACE, stackTraceAsString);
 			}
 			headers.put(X_EXCEPTION_MESSAGE, exceptionMessage);
--- a/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/retry/RepublishMessageRecovererTests.java
+++ b/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/retry/RepublishMessageRecovererTests.java
@@ -30,15 +30,21 @@ import org.springframework.amqp.core.AmqpTemplate;
 import org.springframework.amqp.core.Message;
 import org.springframework.amqp.core.MessageDeliveryMode;
 import org.springframework.amqp.core.MessageProperties;
+import org.springframework.amqp.rabbit.connection.Connection;
+import org.springframework.amqp.rabbit.connection.ConnectionFactory;
+import org.springframework.amqp.rabbit.core.RabbitTemplate;
 import org.springframework.expression.spel.standard.SpelExpressionParser;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.BDDMockito.given;
+import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.verify;
 
 /**
  * @author James Carr
  * @author Gary Russell
  * @author Artem Bilan
+ * @author Rene Choi
  *
  * @since 1.3
  */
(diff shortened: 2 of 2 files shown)


artembilan (commented):
Thanks, Rene — the diagnosis is correct and the fix is the right shape. One field cannot be both the opt-in switch and the truncation limit, and splitting them is exactly right.

I verified both claims in the description rather than taking them on trust:

* `:spring-rabbit:test --tests RepublishMessageRecovererTests` on this branch — 12 tests, 0 failures.
* Reverting only `RepublishMessageRecoverer.java` to `main` and re-running — `shouldIncludeTheStacktraceWhenTheFrameMaxCannotBeDetermined()` fails at line 132, the other 11 pass. So the new test genuinely pins the regression.
* `RabbitUtils.getMaxFrame()` does return `-1` when `connection.getDelegate()` is `null`, so the test's premise (and its comment) is accurate.

For the record, the regression came in with 39a69a0be (#3519, 4.0.5). That commit changed the default from `-1` to `0` and added the `maxStackTraceLength > 0` gate around a `headers.put` that had previously been unconditional. Your statement that "when the limit cannot be resolved the whole stack trace is published, which is what happened before the header became opt-in" is precisely right — that is the pre-4.0.5 behaviour, so this restores it rather than inventing a new one. My commit, my bug.

A few things I checked for breakage, all clean:

* `maxStackTraceLength` has no public setter — `includeStackTrace(boolean)` is its only external mutator — so gating on the new flag cannot change behaviour for anyone who was setting the length directly.
* There is exactly one `headers.put(X_EXCEPTION_STACKTRACE, ...)` site in the main sources, and `RepublishMessageRecovererWithConfirms` overrides only `doSend`, not `recover()`. So the single-site change is complete.
* Defaults are unchanged: `includeStackTrace=false` / `maxStackTraceLength=0` still means no header.

Worth noting a case the PR fixes that isn't called out: a broker with `frame_max=0` (AMQP "no limit") also makes `getMaxFrame()` return a non-positive value, so the header was being dropped there too. That is now correct.

---

Two follow-ups. **Neither blocks this PR** — both are pre-existing on `main` and fixing them here would be scope creep.

**1. `frame_max` at or below `frameMaxHeadroom` yields a negative limit.**

```java
if (maxStackTraceLen > 0) {
    maxStackTraceLen -= this.frameMaxHeadroom;   // 4096 - 20_000 = -15_904
    this.maxStackTraceLength = maxStackTraceLen;
}
```

The `> 0` guard is applied *before* the subtraction, so a broker with `frame_max` below the 20 000 default headroom leaves the field negative. I confirmed the consequence with a throwaway test on this branch: `frame_max=4096` publishes a **10 108-character** stack-trace header, untruncated, against a 4 096-byte frame limit. On `main` that header was simply dropped.

So this PR does shift that case from "silently dropped" to "published oversized". I don't think that argues against merging — it is exactly the pre-4.0.5 behaviour, and `frame_max <= 20000` is well off the beaten path — but it is the residual `-1` sentinel still living in `maxStackTraceLength` that produces it. Your two-field split is what makes a proper fix possible later; I'd rather track it separately than widen this PR.

**2. The limit is re-resolved on every `recover()`.**

Because the field stays negative in both the unresolvable and the below-headroom cases, `processStackTrace()` calls `RabbitUtils.getMaxFrame()` on every single recovered message instead of once. With a `CachingConnectionFactory` the connection itself is cached, so this is wasted work rather than connection churn — but it should still resolve once and stick. Same root cause as (1).

I'll open an issue for both.

---

Before I merge, three small things:

1. **Please add a backport marker.** This is a regression in 4.0.5, so it needs to reach both maintenance lines. Add to the commit message:

   ```
   **Auto-cherry-pick to `4.1.x` & `4.0.x`**
   ```

2. **Please open an issue and reference it.** Per CONTRIBUTING the commit subject should be `GH-nnnn: Fix ...` with a `Fixes https://github.com/spring-projects/spring-amqp/issues/nnnn` line in the body. For a regression that gets backported this isn't bureaucracy — it's what drives the changelog entry on all three branches.

3. **Minor:** consider a sentence on `includeStackTrace(boolean)` documenting that when the limit cannot be resolved from `frame_max` the full stack trace is published untruncated. It is the non-obvious half of the contract, and it is what someone hitting (1) will go looking for.

Style, tabs and the `@author` additions all look right; no comments there.

One caveat on my verification: I ran the unit tests only. `RepublishMessageRecovererIntegrationTests` is broker-gated and I did not run it. I think the risk is negligible — both gates agree at the defaults — but I'd rather say so than imply coverage I don't have.

artembilan (changes requested):
Please, disregards couple my AI agent suggestions in the review:

>  **Auto-cherry-pick to `4.1.x` & `4.0.x`**

I'll do that on merge. No need for you as an external contributor to worry about that.
You cannot definitely know what versions the fix would be aimed for.

>  Please open an issue and reference it.

No need to do this since we simply can mark this PR for specific milestone.
And go from there.

My AI agent just followed strictly with the Contributing Guidelines.
I'll teach not being so picky next time 😄 

Thank you and let me to review your fix myself!

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/retry/RepublishMessageRecoverer.java
```diff
@@ -83,6 +84,8 @@ public class RepublishMessageRecoverer implements MessageRecoverer {
 
 	private int frameMaxHeadroom = DEFAULT_FRAME_MAX_HEADROOM;
 
+	private volatile boolean includeStackTrace;
```
artembilan: We don't need `volatile` since this is one-time configuration setting.
Unlike `maxStackTraceLength` which is re-calculated at runtime and might be reached even concurrently.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/retry/RepublishMessageRecovererTests.java
```diff
@@ -106,6 +112,26 @@ void shouldIncludeTheStacktraceInTheHeaderOfThePublishedMessage() {
 		assertThat(message.getMessageProperties().getHeaders().get("x-exception-stacktrace")).isEqualTo(expectedHeaderValue);
 	}
 
+	@Test
+	void shouldIncludeTheStacktraceWhenTheFrameMaxCannotBeDetermined() {
+		RabbitTemplate rabbitTemplate = mock(RabbitTemplate.class);
+		ConnectionFactory connectionFactory = mock(ConnectionFactory.class);
+		Connection connection = mock(Connection.class);
```
artembilan: Please, consider to use `mock()` API without arguments instead.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/retry/RepublishMessageRecovererTests.java
```diff
@@ -106,6 +112,26 @@ void shouldIncludeTheStacktraceInTheHeaderOfThePublishedMessage() {
...
+		recoverer =
+				new RepublishMessageRecoverer(rabbitTemplate)
+						.includeStackTrace(true);
+		ByteArrayOutputStream baos = new ByteArrayOutputStream();
+		cause.printStackTrace(new PrintStream(baos));
+		final String expectedHeaderValue = baos.toString();
+
+		recoverer.recover(message, cause);
+
+		assertThat(message.getMessageProperties().getHeaders().get("x-exception-stacktrace")).isEqualTo(expectedHeaderValue);
```
artembilan: Consider to use AssertJ `MapAssert` API for convenience:
```
assertThat(message.getMessageProperties().getHeaders()).containsEntry(RepublishMessageRecoverer.X_EXCEPTION_STACKTRACE, expectedHeaderValue);
```

artembilan (changes requested):
Very cool!

Just couple nit-picks.

Thank you again!

renechoi:
All three applied in a61f65b: `volatile` dropped (configuration-time only), no-arg `mock()`, `containsEntry` with the constant.

I also took the Javadoc sentence from the earlier review, since only the other two were withdrawn. Say the word and I'll drop it.

After the assertion change: `RepublishMessageRecovererTests` 12/12, reverting only `RepublishMessageRecoverer.java` to `main` still fails that one test and nothing else. Full `:spring-rabbit:test` 1215 tests, 0 failures; Checkstyle and Javadoc clean.

Agreed the residual `-1` sentinel is what drives both follow-ups.

PR #3624, opened 2026-09-06
 Improve null validation for AbstractRouting(Amqp)ConnectionFactory

ngocnhan-tran1996:
Use the Collection returned by `Map#values()` directly when validating `targetConnectionFactories`, avoiding an unnecessary array allocation.


renechoi:
`AbstractRoutingAmqpConnectionFactory` in `spring-rabbitmq-client` carries the byte-identical `Assert.noNullElements(...values().toArray(), ...)`; those are the only 2 of the repo's 20 `noNullElements` call sites that wrap a map view. Your call whether it belongs here or in a follow-up.

The twin's `RoutingAmqpConnectionFactoryTests#nullTargetValuesAreRejected` also asserts `.withMessageContaining("cannot have null values")`. Without a message assertion the new test can't tell the null-map `Assert` from the null-value one, since both throw `IllegalArgumentException`.

Issue #3520, opened 2026-07-15
The `AbstractJacksonMessageConverter` may got confused with wrong `content-type` from the message


artembilan:
The current logic is:
```
String contentType = properties.getContentType();
		if (this.assumeSupportedContentType && contentType.equals(MessageProperties.DEFAULT_CONTENT_TYPE)
				|| contentType.contains(this.supportedContentType.getSubtype())) {
```
The `contains()` is too wide and may end-up accepting something like `multipart/form-data; boundary=json`.

Consider to fix using `MimeType.isCompatibleWith()`, e.g. set an `application/*+json` from the `JacksonJsonMessageConverter`. That will make it compatible with any JSON-based content-type.

PR #3637, opened 2026-09-10
Apply `assumeSupportedContentType` only to the default C/T

kdelay:
`AbstractJacksonMessageConverter.fromMessage()` wraps its content-type check in `assumeSupportedContentType`, so `setAssumeSupportedContentType(false)` turns the converter off for every content type, including an explicit `application/json`.

The setter Javadoc and `message-converters.adoc` scope it to an absent or default `contentType`, and `AbstractJackson2MessageConverter` still reads `(assumeSupportedContentType && isDefault) || matches`. The parentheses came in with 9e06e026 (GH-3520), whose stated goal was only `contains()` -> `MimeType.isCompatibleWith()`.

A test is added to each converter test class; with the production line reverted both fail with `but was instance of: byte[]` and the other 42 pass. `:spring-amqp:test` (176), `:spring-rabbit:test` (1229, 663 skipped), checkstyle and javadoc pass on JDK 25.

### DIFF
--- a/spring-amqp/src/main/java/org/springframework/amqp/support/converter/AbstractJacksonMessageConverter.java
+++ b/spring-amqp/src/main/java/org/springframework/amqp/support/converter/AbstractJacksonMessageConverter.java
@@ -275,9 +275,8 @@ public abstract class AbstractJacksonMessageConverter extends AbstractMessageCon
 		Object content = null;
 		MessageProperties properties = message.getMessageProperties();
 		String contentType = properties.getContentType();
-		if (this.assumeSupportedContentType &&
-				(contentType.equals(MessageProperties.DEFAULT_CONTENT_TYPE)
-						|| this.supportedContentType.isCompatibleWith(MimeType.valueOf(contentType)))) {
+		if ((this.assumeSupportedContentType && contentType.equals(MessageProperties.DEFAULT_CONTENT_TYPE))
+				|| this.supportedContentType.isCompatibleWith(MimeType.valueOf(contentType))) {
 
 			String encoding = determineEncoding(properties, contentType);
 			content = doFromMessage(message, conversionHint, properties, encoding);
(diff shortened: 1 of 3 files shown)


spring-amqp/src/test/java/org/springframework/amqp/support/converter/JacksonJsonMessageConverterTests.java
```diff
@@ -313,6 +313,21 @@ public void testMissingContentType() {
 		assertThat(foo).isSameAs(bytes);
 	}
 
+	@Test
+	public void testExplicitContentTypeWhenNotAssumingSupportedContentType() {
+		byte[] bytes = "{\"name\" : \"foo\" }".getBytes();
```
artembilan: We prefer to not use `foo/bar` language anymore. Revise your new tests, please.

artembilan (changes requested):
Add you name to the `@author` list of all the affected classes.

kdelay:
Done in cb9a6c5.

- Added the `@author` tag to `AbstractJacksonMessageConverter` and to both test classes.
- The two new tests no longer use the `Foo` fixture. They now deserialize into a `TestData` record and assert the value, so the `foo`/`bar` naming is gone from what I added. The pre-existing `Foo`/`Bar` cases in those files are untouched.

`:spring-amqp:test` is 176 tests, 0 failures; `checkstyleMain`, `checkstyleTest` and `javadoc` pass.

spring-amqp/src/main/java/org/springframework/amqp/support/converter/AbstractJacksonMessageConverter.java
```diff
@@ -43,6 +43,7 @@
  * Abstract Jackson 3 message converter.
  *
  * @author Artem Bilan
+ * @author kdelay
```
artembilan, line 46: I think this is not your official name .
We cannot accept nick-names.
Not sure if this was reviewed but I see that comments are in AI style.

artembilan (changes requested):

PR #3640, opened 2026-09-11
Use supported content type when assumeSupportedContentType is disabled

ngocnhan-tran1996:
- Clarify Jackson 3 JSON-compatible content type requirement in message-converters.adoc
- Add Jackson2JsonMessageConverterTests to preserve the behavior between Jackson 2 and Jackson 3

I also opened a discussion at https://github.com/spring-projects/spring-amqp/discussions/3639 in case my understanding is incorrect.


artembilan (changes requested):
Please, remove `Jackson2JsonMessageConverterTests`.
I really don't care about deprecated API.

Let's see if my answer in your discussion does not cover what you have tried here to do!

artembilan (changes requested):
Can you explain, please, what is going on here?
I don't see a reasonable sentence and arguments why should I accept this change.
That hard-coded `text/x-json` looks suspicious.

Thank you!

ngocnhan-tran1996:
My bad, I forgot to explain the reasoning in the commit body.

From https://github.com/spring-projects/spring-amqp/issues/3520:

> Consider fixing this by using `MimeType.isCompatibleWith()`, e.g. setting an `application/*+json` content type from the `JacksonJsonMessageConverter`. That will make it compatible with any JSON-based content type.

This suggests that `text/x-json` can also be considered a JSON-based content type and should therefore be converted accordingly.

The [documentation ](https://docs.spring.io/spring-amqp/reference/amqp/message-converters.html#JacksonJsonMessageConverter-from-message) also mentions that, to meet the `JacksonJsonMessageConverter` requirements on the consumer side, the producer must set the `contentType` message property, for example to `application/json` or `text/x-json`.

However, there is no wildcard that can cover both `application/json` and `text/x-json`. Therefore, `text/x-json` needs to be explicitly included as a compatible content type rather than being hard-coded as a special case.

artembilan:
>  Therefore, text/x-json needs to be explicitly included as a compatible content type rather than being hard-coded as a special case.

If you need `text/x-json`, then you need configure the converter respectively, e.g. `text/*-json`.
We never claimed supporting both `application/*` and `text/*`.

I think I need to take my comment back on the discussion and state that we support only if `type` is the same and sub-type starts with wildcard. See that `MimeType.isCompatibleWith()`.

The biggest point is to not allow to convert something what is not expected. Kinda security concern.

Let's revise your PR as harmful and close as invalid!

ngocnhan-tran1996:
Thanks for the clarification. I’ll close this PR request.

PR #3641, opened 2026-09-11
Use `supportedContentType` when `assumeSupportedContentType` is disabled

ngocnhan-tran1996:
When `assumeSupportedContentType` is disabled, `contentType` should still be checked against `supportedContentType`. This commit fixes the check to use `supportedContentType` instead of skipping the check entirely.

Reference: https://docs.spring.io/spring-amqp/reference/amqp/message-converters.html#JacksonJsonMessageConverter-from-message


artembilan (approved):
LGTM.

Will merge when PR build is green.

Thank you!

Issue #2729, opened 2024-06-10
ShutdownSignalException when sending message to RabbitMQ after app restart/launch for the first time


purbarunc:
I am getting the `ShutdownSignalException ` exception when attempting to send message to the RabbitMQ. The weird thing is this happens the very first time when the spring boot app is launched or restarted. This automatically gets fixed in the second and subsequent attempts and hence I don't believe there shouldn't be any configuration issue. Hence, I am not able to debug the issue

> Below given is the error log.

com.rabbitmq.client.ShutdownSignalException: channel error; protocol method: #method<channel.close>(reply-code=406, reply-text=PRECONDITION_FAILED - inequivalent arg 'type' for exchange 'MyTopicExchange' in vhost '/': received 'topic' but current is 'direct', class-id=40, method-id=10)
	at com.rabbitmq.client.impl.ChannelN.asyncShutdown(ChannelN.java:528) ~[amqp-client-5.21.0.jar:5.21.0]
	at com.rabbitmq.client.impl.ChannelN.processAsync(ChannelN.java:349) ~[amqp-client-5.21.0.jar:5.21.0]
	at com.rabbitmq.client.impl.AMQChannel.handleCompleteInboundCommand(AMQChannel.java:193) ~[amqp-client-5.21.0.jar:5.21.0]
	at com.rabbitmq.client.impl.AMQChannel.handleFrame(AMQChannel.java:125) ~[amqp-client-5.21.0.jar:5.21.0]
	at com.rabbitmq.client.impl.AMQConnection.readFrame(AMQConnection.java:761) ~[amqp-client-5.21.0.jar:5.21.0]
	at com.rabbitmq.client.impl.AMQConnection.access$400(AMQConnection.java:48) ~[amqp-client-5.21.0.jar:5.21.0]
	at com.rabbitmq.client.impl.AMQConnection$MainLoop.run(AMQConnection.java:688) ~[amqp-client-5.21.0.jar:5.21.0]
	at java.base/java.lang.Thread.run(Thread.java:1583) ~[na:na]

![image](https://github.com/spring-projects/spring-amqp/assets/28572738/6cf72812-05fe-4f98-9520-4da2799f836f)

> My RabbitMQ Exchange and Queue configuration in the Management console is as below:

![image](https://github.com/spring-projects/spring-amqp/assets/28572738/9d14dee0-842e-4aca-a8a5-5af8c300816f)
[...]

PR #3645, opened 2026-09-12
GH-2480: Avoid ThreadLocal

### LINKED ISSUE #2480: Examine ThreadLocal Usage
Consider replacing with some other mechanism if possible.

arnabnandy7:
### Description

This change examines and refactors the usage of `ThreadLocal` across the codebase.

In modern execution environments (such as thread pools and virtual threads), retaining state via `ThreadLocal` can lead to unintended resource retention or lifecycle complexities. While JDK Scoped Values represent the long-term solution for thread-scoped state, they require preview features beyond the current Java 17 baseline.

Following the precedent set in Spring for Apache Kafka (GH-2729), internal `ThreadLocal` variables have been transitioned to `ConcurrentHashMap` structures keyed by `Thread`. This strategy provides explicit, deterministic thread-bound state management and facilitates clean resource reclamation while maintaining existing public API contracts and runtime behavior.

### Key Areas of Change

- **Channel Caching & Resource Management**: Migrated thread-bound channels in connection factories and templates, as well as transaction and resource synchronization holders, to concurrent thread-keyed maps with lifecycle cleanup upon unbinding.
- **Parser & Testing Contexts**: Transitioned thread-scoped state in XML bean definition parsers and JUnit extension conditions to concurrent maps.
- **Documentation & Tests**: Aligned reference documentation terminology with the thread-bound cache model and adjusted internal test assertions accordingly.

Resolves #2480.

### DIFF
--- a/spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/RabbitAvailableCondition.java
+++ b/spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/RabbitAvailableCondition.java
@@ -17,7 +17,9 @@
 package org.springframework.amqp.rabbit.junit;
 
 import java.lang.reflect.AnnotatedElement;
+import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
 
 import com.rabbitmq.client.ConnectionFactory;
 import org.junit.jupiter.api.extension.AfterAllCallback;
@@ -50,7 +52,7 @@ public class RabbitAvailableCondition
 	private static final ConditionEvaluationResult ENABLED = ConditionEvaluationResult.enabled(
 			"@RabbitAvailable is not present");
 
-	private static final ThreadLocal<BrokerRunningSupport> BROKER_RUNNING_HOLDER = new ThreadLocal<>();
+	private static final Map<Thread, BrokerRunningSupport> BROKER_RUNNING_HOLDER = new ConcurrentHashMap<>();
 
 	@Override
 	public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext context) {
@@ -73,7 +75,7 @@ public class RabbitAvailableCondition
 				}
 				brokerRunning.setPurgeAfterEach(rabbit.purgeAfterEach());
 				brokerRunning.test();
-				BROKER_RUNNING_HOLDER.set(brokerRunning);
+				BROKER_RUNNING_HOLDER.put(Thread.currentThread(), brokerRunning);
 				Store store = getStore(context);
 				store.put(BROKER_RUNNING_BEAN, brokerRunning);
 				store.put("queuesToDelete", queues);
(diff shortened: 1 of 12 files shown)


spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/RabbitAvailableCondition.java
```diff
@@ -50,7 +52,7 @@ public class RabbitAvailableCondition
 	private static final ConditionEvaluationResult ENABLED = ConditionEvaluationResult.enabled(
 			"@RabbitAvailable is not present");
 
-	private static final ThreadLocal<BrokerRunningSupport> BROKER_RUNNING_HOLDER = new ThreadLocal<>();
+	private static final Map<Thread, BrokerRunningSupport> BROKER_RUNNING_HOLDER = new ConcurrentHashMap<>();
```
artembilan, line 55: I don’t think this makes sense.
See how `ThreadLocal` is implemented: https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/ThreadLocal.java. It is already a map based on a current thread. So, you are not fixing here anything , just replacing API with home-made code copy.

many confirmation that this was not done by AI, and was reviewed by you including running tests on the project ?

thanks

artembilan (changes requested):

spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/RabbitAvailableCondition.java
arnabnandy7: Hi @artembilan,

Thank you for the review and feedback.

To clarify the background: this PR was opened to explore #2480 following the pattern previously used by Gary Russell in Spring for Apache Kafka (GH-2729 / PR #2791), where `ThreadLocal` instances were replaced with thread-keyed maps in early virtual thread discussions. 

All unit tests and Checkstyle were executed locally and passed before opening the PR. However, your point is completely valid: `ConcurrentHashMap<Thread, ...>` holds strong references to `Thread` objects (which can risk retaining thread instances if unbinding is missed), whereas `ThreadLocal` is already the JDK's built-in thread-scoped table and is cleaned up with thread lifecycle.

Without Java 21+ Scoped Values (`java.lang.ScopedValue`) available on the Java 17 baseline, replacing `ThreadLocal` with an external map does not provide an improvement over standard `ThreadLocal`. 

Given this, I am happy to close this PR and defer the issue until Scoped Values can be targeted once the baseline allows it. Let me know if you would prefer closing the PR. Thanks!

arnabnandy7: Hi @artembilan,

You make a fair point. `ThreadLocal` is already backed by a thread-bound map internally, and `ConcurrentHashMap<Thread, ...>` introduces strong reference retention without real benefits over standard `ThreadLocal` on Java 17. 

But agree that replacing `ThreadLocal` with an external map doesn't make architectural sense without Scoped Values. 

I'll go ahead and close this PR. Thanks for the feedback!

PR #3646, opened 2026-09-12
GH-2608: Fix traceId propagation on async replies

### LINKED ISSUE #2608: TraceId propagation to the new thread local
The scenario and details are located here: https://github.com/micrometer-metrics/tracing/issues/510#issue-2066891159

Version: 3.1.1

arnabnandy7:
### Description

When `@RabbitListener` methods return asynchronous types (`CompletableFuture<?>` or `Mono<?>`), distributed tracing context (`traceId`, `spanId`) was not propagated to reply messages.

This was caused by two issues:
1. **Premature Receiver Observation Completion**: The listener container stopped the receiver observation immediately after synchronous invocation returned, before the asynchronous task finished on another thread.
2. **Missing Reply Sender Observation**: Reply messages were published directly over the channel without a sender observation, preventing Micrometer tracing handlers from injecting propagation headers (such as `traceparent`) into the outbound reply message.

### Proposed Changes

- Keep the receiver observation active for asynchronous listener results until the `CompletableFuture` or `Mono` completes, restoring the observation scope on the completing thread.
- Observe reply message publishing using `RabbitTemplateObservation.TEMPLATE_OBSERVATION`, ensuring tracing headers are injected and linked to the active receiver observation.
- Add `replyObservationConvention` configuration to listener containers and container factories to allow customization of reply observations.
- Document reply observation behavior and add unit tests verifying distributed trace propagation across async executions.

Fixes #2608


renechoi (commented):
The test does not exercise an incoming trace ID: `extract()` ignores the carrier and the request has no B3 headers. Adding `assertThat(sentTraceId).isNotBlank()` made the test fail at line 491; the injected `X-B3-TraceId` was empty, so the all-match assertion compares empty IDs. Please seed/extract a known inbound ID and assert it on the reply.

`setObservationEnabled(false)` is also ineffective because `sendResponse()` checks only the registry. Setting false in the sync test and expecting no B3 header failed at line 556. Please honor the flag or remove the API.

arnabnandy7:
> The test does not exercise an incoming trace ID: `extract()` ignores the carrier and the request has no B3 headers. Adding `assertThat(sentTraceId).isNotBlank()` made the test fail at line 491; the injected `X-B3-TraceId` was empty, so the all-match assertion compares empty IDs. Please seed/extract a known inbound ID and assert it on the reply.
> 
> `setObservationEnabled(false)` is also ineffective because `sendResponse()` checks only the registry. Setting false in the sync test and expecting no B3 header failed at line 556. Please honor the flag or remove the API.

Addressed all concerns, please re-review the changes. Thanks for your effort. @renechoi

renechoi (approved):
Both requested fixes are covered by the updated tests: the propagator now extracts the known inbound B3 IDs and the reply asserts the same trace ID; disabling observation now bypasses reply header injection.

I also ran `:spring-rabbit:test` (1,232 tests, 666 skipped), `:spring-rabbit:checkstyleMain`, `:spring-rabbit:checkstyleTest`, and `:spring-rabbit:javadoc` on `8b82e68a`; all passed.

artembilan (changes requested):
This looks like discussion between two AI agents. I don’t think that I need review from external AI in the public PRs. Can that be done in isolation before it made it into the PR for my review? That review and approval don’t guarantee that I’ll accept a contribution. So, might be even better to not run your review agent since this project needs approvals from the committee anyway .

now to the concept. I don’t think that we need a dedicated option for the reply. It is really a part of the request. So, indeed if we want observation for the request, we definitelly assume reply to be included as well. Request-n-reply is a unit of work.

thank you for understanding!

arnabnandy7:
> now to the concept. I don’t think that we need a dedicated option for the reply. It is really a part of the request. So, indeed if we want observation for the request, we definitelly assume reply to be included as well. Request-n-reply is a unit of work.

Thanks for the feedback, @artembilan! Request and reply are indeed a single unit of work, so if observation is enabled for the container, the reply should naturally be observed without needing dedicated options.

I've removed the separate `replyObservationConvention` options from the container and factory, so replies now automatically follow the container's observation configuration using the default convention.

Also, apologies for the noise on the PR from that external review thread earlier! Please take another look when you get a chance.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
```diff
@@ -668,35 +798,59 @@ protected void sendResponse(@Nullable Channel channel, Address replyTo, Message
...
+			}
+		}
+		else {
+			try {
+				this.retryTemplate.<@Nullable Object>execute(() -> {
+					doPublish(channel, replyTo, message);
+					return null;
+				});
+			}
+			catch (RetryException ex) {
```
artembilan, line 749: `recoveryCallback.recover(...)` swallows the `RetryException` and returns normally, so `publish()` completes without throwing. But `publish()` is invoked from `sendResponse()` inside `RabbitTemplateObservation.TEMPLATE_OBSERVATION...observe(() -> publish(...))` (line ~809). Since `observe()` never sees an exception, the reply-send observation is reported as a **successful** span/metric even though the reply was actually dropped and handed to the recovery callback. Consider calling `observation`-aware error reporting here, or restructuring so a recovered failure still marks the observation as an error.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -1542,10 +1563,27 @@ protected void executeListener(Channel channel, Object data) {
...
 					DefaultRabbitListenerObservationConvention.INSTANCE,
-					() -> new RabbitMessageReceiverContext(message, getListenerId()), registry);
-			observation.observe(() -> executeListenerAndHandleException(channel, data));
+					() -> context, registry);
+			observation.start();
+			Observation.Scope scope = observation.openScope();
+			boolean async = false;
+			try {
+				executeListenerAndHandleException(channel, data);
+				async = context.isAsync();
```
artembilan: `async` is only set from `context.isAsync()` after `executeListenerAndHandleException` returns *normally*. But `context.isAsync()` can already be `true` before that call returns (`AbstractAdaptableMessageListener.handleResult` sets it as soon as it sees a `CompletableFuture`/`Mono`, before the async task completes). If anything throws after that point — e.g. a second de-batched fragment failing, or post-invoke bookkeeping/commit failing in `cleanUpAfterInvoke` — the `catch` block is taken with `async` still `false`, so `finally` stops the observation immediately. Later, when the future/Mono actually completes, `AbstractAdaptableMessageListener`'s callback calls `observation.stop()` again on the same `Observation` — a double-stop. Reading `context.isAsync()` inside the `finally` block (instead of relying on the post-call assignment) would close this window.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
```diff
@@ -425,10 +507,58 @@ else if (JavaUtils.MONO_PRESENT && MonoHandler.isMono(resultArg.getReturnValue()
 					this.logger.warn("Container AcknowledgeMode must be MANUAL for a Mono<?> return type" +
 							"(or Kotlin suspend function); otherwise the container will ack the message immediately");
 				}
+				Observation observation = (isObservationEnabled() && this.observationRegistry != null)
+						? this.observationRegistry.getCurrentObservation() : null;
+				if (observation != null
+						&& observation.getContext() instanceof RabbitMessageReceiverContext receiverContext) {
+					receiverContext.setAsync(true);
+				}
 				MonoHandler.subscribe(resultArg.getReturnValue(),
```
artembilan: `MonoHandler.subscribe` delegates to `Mono#subscribe(Consumer, Consumer<Throwable>, Runnable)`, which has no cancellation callback. If the returned `Mono` is cancelled without a terminal `onNext`/`onError`/`onComplete` signal (e.g. an upstream `timeout()`/`take()`, or the subscription being disposed on container shutdown), none of the three lambdas below run, and `observation.stop()` is never called — since the container already skipped stopping it (via `context.isAsync()`), this observation is now leaked for the lifetime of the process.

artembilan (commented):
Solid fix for the core problem (async replies losing trace context). A few correctness gaps and some duplicated boilerplate worth addressing before merge; nothing here blocks the overall approach.

One more non-inline note: `AbstractMessageListenerContainer.setObservationEnabled(true)` on an already-*running* container pushes `getObservationRegistry()` into the adapter, but that registry is only resolved from the `ApplicationContext` inside `checkObservation()`, which only runs from `start()`. So flipping the flag at runtime on a live container is a silent no-op for reply tracing until the container is stopped/started again. Worth a javadoc caveat at minimum.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
```diff
@@ -410,13 +467,38 @@ protected void handleResult(@Nullable InvocationResult resultArg, Message reques
...
+				}
 				completable.whenComplete((r, t) -> {
-					if (t == null) {
-						asyncSuccess(resultArg, request, channel, source, r);
-						basicAck(request, channel);
+					if (observation != null) {
+						Observation.Scope asyncScope = observation.openScope();
+						try {
+							if (t == null) {
+								asyncSuccess(resultArg, request, channel, source, r);
```
artembilan, line 458: The success branches here (and the Mono `onNext`/`onComplete` equivalents below) call `asyncSuccess(...)`/`basicAck(...)` inside only a `try/finally`, with no `catch` to report to `observation`. Every *other* path this PR adds (the `t != null` branch here, and the Mono `onError` branch) calls `observation.error(...)` before stopping — but if `asyncSuccess` → `sendResponse` → `publish` throws on the success path, that exception is silently lost (nothing observes the `CompletableFuture` returned by `whenComplete`), and the observation is stopped as if the reply succeeded. Worth wrapping these calls the same way the error branches do.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
```diff
@@ -410,13 +467,38 @@ protected void handleResult(@Nullable InvocationResult resultArg, Message reques
 					this.logger.warn("Container AcknowledgeMode must be MANUAL for a Future<?> return type; "
 							+ "otherwise the container will ack the message immediately");
 				}
+				Observation observation = (isObservationEnabled() && this.observationRegistry != null)
```
artembilan: The `if (observation != null) { scope = openScope(); try { ... } finally { scope.close(); } }` shape is hand-rolled four times across this method (here, and the three Mono callbacks below) instead of using `Observation.tryScoped(observation, runnable)` / `observation.scoped(...)`, which already exist in Micrometer for exactly this null-safe-scope pattern. This codebase already prefers the delegating form elsewhere (e.g. `RabbitTemplate.java` around line 2467: `observation.observe(() -> sendToRabbit(...))`). Collapsing these four copies into one helper would remove the duplication and the risk of the four copies drifting apart on a future fix.

arnabnandy7:
> Solid fix for the core problem (async replies losing trace context). A few correctness gaps and some duplicated boilerplate worth addressing before merge; nothing here blocks the overall approach.
> 
> One more non-inline note: `AbstractMessageListenerContainer.setObservationEnabled(true)` on an already-_running_ container pushes `getObservationRegistry()` into the adapter, but that registry is only resolved from the `ApplicationContext` inside `checkObservation()`, which only runs from `start()`. So flipping the flag at runtime on a live container is a silent no-op for reply tracing until the container is stopped/started again. Worth a javadoc caveat at minimum.

Thanks for the detailed review and pointers, @artembilan!

I've addressed all the feedback in the latest push:
 - **Runtime observation enablement**: `setObservationEnabled(true)` now invokes `checkObservation()` if the container is already running to resolve the `ObservationRegistry` from the `ApplicationContext` if not yet initialized, and added a Javadoc caveat explaining this.
 - **Double-stop prevention**: In `executeListener()`, `!context.isAsync()` is now checked directly in the `finally` block to close the window where an exception after `setAsync(true)` caused premature observation stopping.
 - **Mono cancellation leak**: Added an overloaded `MonoHandler.subscribe` accepting a cancellation callback (`doOnCancel`), and tied observation completion to an `AtomicBoolean` guard so cancellation cleanly terminates the observation without leaking.
 - **Async reply exception handling**: Enclosed success callback paths in `try/catch` to record `observation.error(ex)` and delegate to `asyncFailure` if reply dispatching or acking fails.
 - **Scope boilerplate cleanup**: Consolidated the repeated try-finally scope handling into a `scoped(@Nullable Observation, Runnable)` helper.
 - **Reply recovery error reporting**: Passed the reply `Observation` to `publish()` and explicitly record `observation.error(replyFailureException)` when a retry failure is routed to `recoveryCallback`.

artembilan (commented):
The review was from my own AI agent.
I’m not very sure though that everything is so solid as it claims.
 I’ll review it myself on Monday.

thanks

artembilan:
Follow-up on finding #4 from the earlier review (async success path never calling `observation.error()`): that asymmetry isn't specific to this PR — it's the same pattern in Spring Kafka's `MessagingMessageListenerAdapter.handleResult`, so I filed it separately there rather than blocking this PR on it: spring-projects/spring-kafka#4677.

That said, Spring Kafka's `handleResult` is a good reference for a cleaner shape than what this PR currently has here. Instead of branching into separate `CompletableFuture`/`Mono` handling with four near-duplicate `openScope()/try/finally` blocks, it normalizes both cases into a single `CompletableFuture` up front (`Mono#toFuture()` when the result is a `Mono`) and then has exactly **one** `whenComplete` callback with one scope/finally block for the whole async completion:

https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java#L565-L603

@arnabnandy7 worth restructuring `AbstractAdaptableMessageListener.handleResult` to follow that same normalize-then-single-callback shape — it would resolve the "4x duplicated scope handling" comment above in one pass instead of extracting a separate helper for it.

---
🤖 I am @artembilan's AI agent, posting on his behalf.

arnabnandy7:
> I’ll review it myself on Monday.

I'll wait for your review.

artembilan (changes requested):
Any updates according similarities with Spring Kafka? So, I’ll have less and more concise code to review tomorrow.
Also , consider to remove those new getters.
I also, don’t think that runtime reaction to the enable observation flag change is unnecessary.

arnabnandy7:
> Any updates according similarities with Spring Kafka? So, I’ll have less and more concise code to review tomorrow. Also , consider to remove those new getters. I also, don’t think that runtime reaction to the enable observation flag change is unnecessary.

I was waiting for this only, will amend now. Needed a confirmation whether to follow these suggestions and make changes.

src/reference/antora/modules/ROOT/pages/amqp/receiving-messages/micrometer-observation.adoc
```diff
@@ -15,6 +15,9 @@ The default implementations add the `name` tag for template observations and `li
 
 You can either subclass `DefaultRabbitTemplateObservationConvention` or `DefaultRabbitListenerObservationConvention` or provide completely new implementations.
 
+Starting with version 4.2, listener reply messages are observed using `RabbitTemplateObservation.TEMPLATE_OBSERVATION` when observation is enabled, propagating distributed tracing context (such as `traceparent`) to reply messages.
+For asynchronous listener return types (`CompletableFuture` or `Mono`), the listener receiver observation remains open until the asynchronous computation completes, ensuring trace context is propagated across threads to the reply message.
```
artembilan: A short sentence should be added into `whats-new.adoc` with a link to this chapter.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -440,6 +440,33 @@ public void setMessageListener(MessageListener messageListener) {
...
+
+	/**
+	 * {@inheritDoc}
+	 * <p>
+	 * If enabled on an already running container, an attempt will be made to resolve
+	 * the {@link ObservationRegistry} from the {@link ApplicationContext} if not already resolved,
+	 * and configure the message listener accordingly.
+	 */
+	@Override
+	public void setObservationEnabled(boolean observationEnabled) {
```
artembilan: As I said before: I don't believe into mutation at runtime.
If that happens, then restart.
Therefore, this should be as simple setter: the scope of its goal must not be extended.
The listener adapter should be configured in the `start()` logic.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -440,6 +440,33 @@ public void setMessageListener(MessageListener messageListener) {
...
+		}
+		else {
+			configureListenerObservation(getMessageListener());
+		}
+	}
+
+	private void configureListenerObservation(@Nullable MessageListener messageListener) {
+		if (messageListener instanceof AbstractAdaptableMessageListener adaptable) {
+			adaptable.setObservationRegistry(getObservationRegistry());
+			adaptable.setObservationEnabled(isObservationEnabled());
```
artembilan: I think just one `adaptable.setObservationRegistry` is enough.
And that should be conditional on `isObservationEnabled()`.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -440,6 +440,33 @@ public void setMessageListener(MessageListener messageListener) {
...
+			adaptable.setBeanName(getListenerId());
```
artembilan: This call does not reflect reality. You really don't provide a bean name for the `AbstractAdaptableMessageListener`.
Consider to use `setListenerId()` instead.
Or even just `listenerId()` to not confuse with public setters.
Or... Even better: `populateObservationRegistry(getObservationRegistry(), getListenerId())`.
I wish this to not be a `public` API, but we'll see when we reach that `AbstractAdaptableMessageListener` for review...

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -1542,10 +1575,25 @@ protected void executeListener(Channel channel, Object data) {
...
+			try {
+				executeListenerAndHandleException(channel, data);
+			}
+			catch (Throwable ex) {
+				observation.error(ex);
+				throw ex;
+			}
+			finally {
+				scope.close();
+				if (!context.isAsync()) {
```
artembilan: Why do we extend `RabbitMessageReceiverContext` API and responsibility?
I don't think it is really its scope to be aware of async listener.
Let's also see if we really need to defer this `consumer` observation stop until after async reply is sent!
What is wrong with closing consumer observation, and then attach a producer one in when its thread is doing reply?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/ObservableListenerContainer.java
```diff
@@ -103,6 +103,15 @@ public void setObservationEnabled(boolean observationEnabled) {
 		this.observationEnabled = observationEnabled;
 	}
 
+	/**
+	 * Return whether observation is enabled.
+	 * @return true if observation is enabled.
+	 * @since 4.2
+	 */
+	public boolean isObservationEnabled() {
```
artembilan: I think we already have `getObservationRegistry()` API.
Plus I don't see a reason to have it as `public` anyway.
Please, consider to remove and rework. 
We do have over there in the inheritors a logic like `!registry.isNoop()`.
Should be enough so far.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
```diff
@@ -334,6 +347,51 @@ protected boolean isManualAck() {
...
+	public void setBeanName(@Nullable String beanName) {
+		this.beanName = beanName;
+	}
+
+	/**
+	 * Return the bean name or listener id to associate with reply observations.
+	 * @return the bean name.
+	 * @since 4.2
+	 */
+	public @Nullable String getBeanName() {
```
artembilan: I asked to remove all the getter.
Plus, we have just discussed in the container that new API might not be a setter at all.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
artembilan: Looks like a lot of our over weekend exchanges were not addressed.
See Spring Kafka for concise solution: https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java#L567

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/micrometer/RabbitMessageReceiverContext.java
```diff
@@ -62,4 +62,24 @@ public String getListenerId() {
...
+	public void setAsync(boolean async) {
+		this.async = async;
+	}
+
+	/**
+	 * Return whether this context is being processed asynchronously.
+	 * @return true if asynchronous
+	 * @since 4.2
+	 */
+	public boolean isAsync() {
```
artembilan: I'm against this new API addition to the context.
It is really not a receiver observation responsibility to be aware of an async listener.

artembilan (changes requested):
I'm stopping here.
Let's see if we can come to the same page before we make a final decision about this effort!

Thanks

arnabnandy7:
Apologies for the frustration, @artembilan. I completely understand and appreciate you taking the time to steer this in the right direction.

I agree with your points:
1. Normalize `Mono` to `CompletableFuture` up front via `toFuture()` and handle everything with a single `whenComplete` block, eliminating all duplicated callback handling.
2. Removal of 
  a. `setAsync` / `isAsync` completely so the context remains focused purely on receiver metadata.
  b. `getBeanName()` and avoid introducing any unnecessary public APIs.
  c. the dynamic runtime check in `setObservationEnabled()`.

 If this matches what you have in mind, I will make these changes and push a clean update. Thanks again for your patience!

artembilan:
Yes, @arnabnandy7 .
Sounds like a plan!
I also what you to double check if sync reply is also covered.
I mean it was there indeed, since consumer observation was blocked until we send reply.
However, since we introduce a producer observation I think it would be natural to fold an async reply handling into the same producer model from now on.

Thanks again!

arnabnandy7:
Thanks, @artembilan!

I have pushed the updated commit with all these changes in place.

To confirm regarding sync replies: **both sync and async replies now share the exact same producer observation model**. Both flow through `sendResponse()`, which creates and starts the `RabbitTemplateObservation.TEMPLATE_OBSERVATION` to observe the reply publish and inject the tracing headers (as a child span of the consumer observation).

In addition to `asyncObservationPropagation`, we have `syncObservationPropagation` in `MessagingMessageListenerAdapterTests` verifying that synchronous replies produce the reply observation span and propagate the trace ID as expected.

Please take another look when you get a chance!

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -440,6 +440,17 @@ public void setMessageListener(MessageListener messageListener) {
 		this.isBatchListener = messageListener instanceof BatchMessageListener
 				|| messageListener instanceof ChannelAwareBatchMessageListener;
 		this.asyncReplies = messageListener.isAsyncReplies();
+		configureListenerObservation(messageListener);
```
artembilan: This setter also happens only once during configuration phase.
We do have `configureListenerObservation()` in the `checkObservation()` which is called from the `start()`.
So, I don't see a reason in this runtime adjustment.
I also don't see now a reason in the `configureListenerObservation()` at all.
Why that `checkObservation()` is not enough for us?
Why do we need to propagate `MessageListener` argument at all if we can just use `this.messageListener` over there in the `checkObservation()`?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
```diff
@@ -334,6 +344,24 @@ protected boolean isManualAck() {
 		return this.isManualAck;
 	}
 
+	/**
+	 * Set the {@link ObservationRegistry} to use for observing reply messages.
+	 * @param observationRegistry the observation registry.
+	 * @since 4.2
+	 */
+	public void setObservationRegistry(@Nullable ObservationRegistry observationRegistry) {
```
artembilan: Why do we accept `@Nullable` here?
Doesn't look like we would allow null to be passed here at all.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
```diff
@@ -334,6 +344,24 @@ protected boolean isManualAck() {
...
+		this.observationRegistry = observationRegistry;
+	}
+
+	/**
+	 * Set the listener id to associate with reply observations.
+	 * @param listenerId the listener id.
+	 * @since 4.2
+	 */
+	public void setListenerId(@Nullable String listenerId) {
```
artembilan: DITTO.

Can you explain me, please, why don't you go my other suggested API instead of setters?
Like `setupObservationRegistry(ObservationRegistry observationRegistry, String listenerId)` 🤷

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
```diff
@@ -405,37 +433,54 @@ protected void handleResult(@Nullable InvocationResult resultArg, Message reques
...
 				}
-				MonoHandler.subscribe(resultArg.getReturnValue(),
-						r -> asyncSuccess(resultArg, request, channel, source, r),
-						t -> asyncFailure(request, channel, t, source),
-						() -> basicAck(request, channel));
+				completableFutureResult = MonoHandler.toCompletableFuture(resultArg.getReturnValue());
+			}
+
+			if (completableFutureResult != null) {
+				Observation observation = (this.observationRegistry != null && !this.observationRegistry.isNoop())
```
artembilan: See your logic in the listener container: you don't propagate an `ObservationRegistry` if it is `NOOP`.
So, condition is redundant.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
```diff
@@ -405,37 +433,54 @@ protected void handleResult(@Nullable InvocationResult resultArg, Message reques
...
+						? this.observationRegistry.getCurrentObservation() : null;
+				completableFutureResult.whenComplete((r, t) -> {
+					scoped(observation, () -> {
```
artembilan: I'm against this `scoped()` approach.
See https://github.com/spring-projects/spring-kafka/pull/4678 for a proper solution.
(Pay attention: that one is done with Claude Code, too).

First of all it creates a new object internally, in addition to the `Runnable` lambda.
Secondly you have a bug in the `cause` block missing to report error back into the observation.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
```diff
@@ -668,35 +713,62 @@ protected void sendResponse(@Nullable Channel channel, Address replyTo, Message
...
 					+ replyTo.getRoutingKey() + "]");
-			if (this.retryTemplate == null) {
-				doPublish(channel, replyTo, message);
+			if (this.observationRegistry != null && !this.observationRegistry.isNoop()) {
+				Message messageToSend = message;
+				Observation replyObservation = RabbitTemplateObservation.TEMPLATE_OBSERVATION.observation(null,
+								DefaultRabbitTemplateObservationConvention.INSTANCE,
+								() -> new RabbitMessageSenderContext(messageToSend,
+										this.listenerId != null ? this.listenerId : "rabbitListenerReply",
+										replyTo.getExchangeName(), replyTo.getRoutingKey()),
```
artembilan: Consider to extract these repetitive getter calls into local variables.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
```diff
@@ -668,35 +713,62 @@ protected void sendResponse(@Nullable Channel channel, Address replyTo, Message
...
 			this.logger.debug("Publishing response to exchange = [" + replyTo.getExchangeName() + "], routingKey = ["
 					+ replyTo.getRoutingKey() + "]");
-			if (this.retryTemplate == null) {
-				doPublish(channel, replyTo, message);
+			if (this.observationRegistry != null && !this.observationRegistry.isNoop()) {
+				Message messageToSend = message;
+				Observation replyObservation = RabbitTemplateObservation.TEMPLATE_OBSERVATION.observation(null,
+								DefaultRabbitTemplateObservationConvention.INSTANCE,
+								() -> new RabbitMessageSenderContext(messageToSend,
+										this.listenerId != null ? this.listenerId : "rabbitListenerReply",
```
artembilan: I don't think this "default value" logic should be here.
Consider to move it into the property definition.
That would be good addition to our agreement that no `@Nullable` in the setters.
(Or even better if you move to my proposal)

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapterTests.java
```diff
@@ -409,6 +426,174 @@ void maySettleDeliveryOnlyWhenTheListenerHasTheChannel() {
 		assertThat(getBatchInstance("withFooBatch").maySettleDelivery()).isTrue();
 	}
 
+	@Test
+	public void asyncObservationPropagation() throws Exception {
+		SimpleTracer tracer = new SimpleTracer();
+		Propagator propagator = createPropagator(tracer);
+		TestObservationRegistry registry = createObservationRegistry(tracer, propagator);
+
+		Channel channel = mock(Channel.class);
```
artembilan: Use `mock()` API without arguments.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapterTests.java
```diff
@@ -409,6 +426,174 @@ void maySettleDeliveryOnlyWhenTheListenerHasTheChannel() {
...
+		MessagingMessageListenerAdapter listener = getSimpleInstance("asyncEcho", String.class);
+		listener.setObservationRegistry(registry);
+		listener.setListenerId("testListener");
+
+		org.springframework.amqp.core.Message message = MessageTestUtils.createTextMessage("foo");
```
artembilan: No `foo/bar` language in our code, please.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapterTests.java
```diff
@@ -409,6 +426,174 @@ void maySettleDeliveryOnlyWhenTheListenerHasTheChannel() {
...
+		Observation.Scope scope = receiverObservation.openScope();
+		try {
+			listener.onMessage(message, channel);
+		}
+		finally {
+			scope.close();
+			receiverObservation.stop();
+		}
+
+		ArgumentCaptor<BasicProperties> propsCaptor = ArgumentCaptor.forClass(BasicProperties.class);
```
artembilan: Use `ArgumentCaptor.captor()` API instead.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapterTests.java
```diff
@@ -409,6 +426,174 @@ void maySettleDeliveryOnlyWhenTheListenerHasTheChannel() {
...
+							.build();
+					return tracer.spanBuilder().setParent(parent);
+				}
+				return tracer.spanBuilder();
+			}
+
+		};
+	}
+
+	private TestObservationRegistry createObservationRegistry(SimpleTracer tracer, Propagator propagator) {
```
artembilan: `static`?

artembilan (changes requested):

arnabnandy7:
Thanks for the detailed review, @artembilan

I've pushed an update addressing all your feedback:

- Replaced the separate setters with `setupObservationRegistry(ObservationRegistry, String)` with non-null checks, and defaulted `listenerId` directly on the field definition.
- Removed `configureListenerObservation()` and its call in `setMessageListener()`, inlining the adapter setup directly into `checkObservation()`.
- Dropped `scoped()` in favor of the Spring Kafka #4678 pattern using `observation.openScope()` with try-with-resources. Now reporting errors via `observation.error(cause)` on the failure path, and wrapping the success path in `try/catch` so reply publish / ack failures are also captured by the observation.
- Extracted local variables for `exchange` and `routingKey` and simplified the `isNoop()` check.
- Switched to argument-less `mock()`, `ArgumentCaptor.captor()`, made helper methods `static`, replaced all `foo/bar` test names with descriptive ones, and added a test verifying async error recording on the observation context.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
```diff
@@ -1459,6 +1459,16 @@ public void start() {
 		}
 	}
 
+	@Override
+	protected void checkObservation() {
+		super.checkObservation();
+		ObservationRegistry registry = getObservationRegistry();
+		if (!registry.isNoop() && this.messageListener instanceof AbstractAdaptableMessageListener adaptable) {
+			String listenerId = getListenerId();
+			adaptable.setupObservationRegistry(registry, listenerId != null ? listenerId : "rabbitListenerReply");
```
artembilan: I don't think this is correct.
The `getListenerId()` is never `null`.
I mean you to have `rabbitListenerReply` as a default value in the `AbstractAdaptableMessageListener.listenerId`  property.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
```diff
@@ -334,6 +344,24 @@ protected boolean isManualAck() {
...
 
+	/**
+	 * Set up the {@link ObservationRegistry} and listener id to use for observing reply messages.
+	 * @param observationRegistry the observation registry.
+	 * @param listenerId the listener id.
+	 * @since 4.2
+	 */
+	public void setupObservationRegistry(ObservationRegistry observationRegistry, String listenerId) {
+		Assert.notNull(observationRegistry, "'observationRegistry' must not be null");
+		Assert.notNull(listenerId, "'listenerId' must not be null");
```
artembilan: With Nullability on board we don't need these checks.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapterTests.java
```diff
@@ -409,6 +426,200 @@ void maySettleDeliveryOnlyWhenTheListenerHasTheChannel() {
...
+	public void asyncObservationPropagation() throws Exception {
+		SimpleTracer tracer = new SimpleTracer();
+		Propagator propagator = createPropagator(tracer);
+		TestObservationRegistry registry = createObservationRegistry(tracer, propagator);
+
+		Channel channel = mock();
+		MessagingMessageListenerAdapter listener = getSimpleInstance("asyncEcho", String.class);
+		listener.setupObservationRegistry(registry, "testListener");
+
+		org.springframework.amqp.core.Message message = MessageTestUtils.createTextMessage("testPayload");
```
artembilan: See `MessageBuilder` instead.
That fluent API covers properties and headers, too.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapterTests.java
```diff
@@ -409,6 +426,200 @@ void maySettleDeliveryOnlyWhenTheListenerHasTheChannel() {
...
+		message.getMessageProperties().setHeader("X-B3-TraceId", "4bf92f3577b34da6a3ce929d0e0e4736");
+		message.getMessageProperties().setHeader("X-B3-SpanId", "00f067aa0ba902b7");
+
+		RabbitMessageReceiverContext receiverContext = new RabbitMessageReceiverContext(message, "testListener");
+		Observation receiverObservation = RabbitListenerObservation.LISTENER_OBSERVATION.observation(null,
+				RabbitListenerObservation.DefaultRabbitListenerObservationConvention.INSTANCE,
+				() -> receiverContext, registry);
+		receiverObservation.start();
+		Observation.Scope scope = receiverObservation.openScope();
+		try {
```
artembilan: I don't think we need low-level Observation API.
I believe just `observe(() -> )` is enough.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapterTests.java
```diff
@@ -409,6 +426,200 @@ void maySettleDeliveryOnlyWhenTheListenerHasTheChannel() {
...
+		});
+
+		BasicProperties props = propsCaptor.getValue();
+		assertThat(props.getHeaders()).containsKey("X-B3-TraceId");
+		String sentTraceId = props.getHeaders().get("X-B3-TraceId").toString();
+		assertThat(sentTraceId).isNotBlank();
+		assertThat(sentTraceId).isEqualTo("4bf92f3577b34da6a3ce929d0e0e4736");
+
+		Deque<SimpleSpan> spans = tracer.getSpans();
+		assertThat(spans).hasSizeGreaterThanOrEqualTo(2);
```
artembilan: Well, with that `verify(channel).basicPublish()` assertion there is a risk of race condition that a `replyObservation` is not closed yet and not reported back to the tracer.
I mean look into the original code:
```
replyObservation.observe(() -> publish(channel, replyTo, messageToSend, replyObservation));
```
You catch here only `basicPublish()` interaction, but an `observe()` is not done yet.

I'm also not sure that `hasSizeGreaterThan` is OK here.
Why would be accept more than `2` as a success in this test?

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapterTests.java
```diff
@@ -409,6 +426,200 @@ void maySettleDeliveryOnlyWhenTheListenerHasTheChannel() {
...
+		assertThat(spans.stream().allMatch(s -> s.context().traceId().equals("4bf92f3577b34da6a3ce929d0e0e4736"))).isTrue();
```
artembilan: I think this `traceId` literal deserves to be extracted into local variable. It is used in the code too much.

src/reference/antora/modules/ROOT/pages/amqp/receiving-messages/micrometer-observation.adoc
```diff
@@ -15,6 +15,8 @@ The default implementations add the `name` tag for template observations and `li
 
 You can either subclass `DefaultRabbitTemplateObservationConvention` or `DefaultRabbitListenerObservationConvention` or provide completely new implementations.
 
+Starting with version 4.2, listener reply messages are observed using `RabbitTemplateObservation.TEMPLATE_OBSERVATION` when observation is enabled, propagating distributed tracing context (such as `traceparent`) to reply messages.
```
artembilan, line 18: Can you also mention here that reply (`PRODUCER`) observation is a child one of the `CONSUMER` one from the listener itself?

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
```diff
@@ -405,30 +433,44 @@ protected void handleResult(@Nullable InvocationResult resultArg, Message reques
...
-						r -> asyncSuccess(resultArg, request, channel, source, r),
-						t -> asyncFailure(request, channel, t, source),
-						() -> basicAck(request, channel));
+				completableFutureResult = MonoHandler.toCompletableFuture(resultArg.getReturnValue());
+			}
+
+			if (completableFutureResult != null) {
+				Observation observation = getCurrentObservation();
+				completableFutureResult.whenComplete((r, t) -> {
+					try (Observation.Scope ignored = observation.openScope()) {
```
artembilan: Oh! I see now.
Apparently `observation.scoped()` is right API to use here.
I thought we don't need to open scope at all.
But that is exactly what make an inheritance working.
Why didn't you object when I asked opposite?
Please, bring back `scoped()`, but still keep that `observation.error()` in the `else` block.

artembilan (changes requested):
I think this is

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
arnabnandy7: I was replying in accumulation, to avoid cluttering in inline threads, that's why I didn't reply in any of inline by far. @artembilan, I'll amend these as per your suggestion.

artembilan: Right. I saw that:

> Dropped scoped() in favor of the Spring Kafka #4678 pattern using observation.openScope() with try-with-resources.

My point is that original idea for `scoped()` was good. You shouldn't insisted to keep it instead.
That was my bad reviewing Spring Kafka PR just for diff. I also should've look deeper into the scope of the `whenComplete()` in the `MessagingMessageListenerAdapter` of Spring Kafka 😄

arnabnandy7: restored `scoped()` and kept the explicit error reporting in place

src/reference/antora/modules/ROOT/pages/amqp/receiving-messages/micrometer-observation.adoc
arnabnandy7: yes sure, added

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapterTests.java
arnabnandy7: I’ve updated the test to wait until exactly two spans have been reported before inspecting them, so it no longer assumes that `basicPublish()` means the reply observation has already completed. 

I also replaced `hasSizeGreaterThanOrEqualTo(2)` with an exact size assertion.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapterTests.java
arnabnandy7: extracted the repeated trace ID into a constant. I did the same for the span ID since it is also reused across these tests.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapterTests.java
arnabnandy7: replaced the manual `start()`, `openScope()`, and `stop()` sequence with `observeChecked(...)`. I used the checked variant because `listener.onMessage(...)` declares a checked exception.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapterTests.java
arnabnandy7: Updated, the new tests now use the AMQP MessageBuilder fluent API to construct the message body, reply-to property, and tracing headers.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
arnabnandy7: removed both explicit `Assert.notNull(...)` checks

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer.java
arnabnandy7: removed the redundant null fallback and now pass `getListenerId()` directly. The listenerId property in `AbstractAdaptableMessageListener` retains rabbitListenerReply as its default value.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
```diff
@@ -405,30 +431,43 @@ protected void handleResult(@Nullable InvocationResult resultArg, Message reques
...
+			if (completableFutureResult != null) {
+				Observation observation = getCurrentObservation();
+				completableFutureResult.whenComplete((r, t) ->
+					observation.scoped(() -> {
+						if (t == null) {
+							try {
+								asyncSuccess(resultArg, request, channel, source, r);
+								basicAck(request, channel);
+							}
+							catch (RuntimeException ex) {
```
artembilan: Yeah... With the `scoped()` we don't need this `catch` here any more.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapterTests.java
```diff
@@ -409,6 +430,176 @@ void maySettleDeliveryOnlyWhenTheListenerHasTheChannel() {
...
+		ArgumentCaptor<BasicProperties> propsCaptor = ArgumentCaptor.captor();
+		await().untilAsserted(() -> {
+			verify(channel).basicPublish(eq("testExchange"), eq("testRoutingKey"), eq(false), propsCaptor.capture(),
+					any(byte[].class));
+		});
+
+		BasicProperties props = propsCaptor.getValue();
+		assertThat(props.getHeaders()).containsKey("X-B3-TraceId");
+		String sentTraceId = props.getHeaders().get("X-B3-TraceId").toString();
+		assertThat(sentTraceId).isNotBlank();
```
artembilan: If we do `isEqualTo()`, then `isNotBlank()` is redundant.

artembilan (changes requested):

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
arnabnandy7: removed the unnecessary try/catch

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/adapter/MessagingMessageListenerAdapterTests.java
arnabnandy7: removed the redundant `isNotBlank()`

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
```diff
@@ -666,37 +699,64 @@ protected void sendResponse(@Nullable Channel channel, Address replyTo, Message
 		postProcessChannel(channel, message);
 
 		try {
-			this.logger.debug("Publishing response to exchange = [" + replyTo.getExchangeName() + "], routingKey = ["
-					+ replyTo.getRoutingKey() + "]");
-			if (this.retryTemplate == null) {
-				doPublish(channel, replyTo, message);
+			String exchange = replyTo.getExchangeName();
+			String routingKey = replyTo.getRoutingKey();
+			this.logger.debug("Publishing response to exchange = [" + exchange + "], routingKey = ["
```
artembilan: Wrap this into `if (this.logger.isDebugEnabled())` for optimization.
I know simple string concatenation is compiled into a `StringBuilder`, but still sub-optimal if we are not going to log this message in 99% of prods.

artembilan (approved):
LGTM.
Just one last nit-pick I see from here.
I know, existing code, but still would be better to fix since we are here already.

I'm pulling this locally for closer look, that simple fix and possible merge.

Nothing to do from your side any more.

Thank you very much!

It is impressive if you have used some AI agent to help you with the fix 😄 
What was that if not a secret (or case at all)?

arnabnandy7:
> It is impressive if you have used some AI agent to help you with the fix 😄 What was that if not a secret (or case at all)?

Not entirely but since I try to contribute to new libraries wherever I find anything that interests me, I use Antigravity to understand repo structure, code quality, good practices that being followed etc. Also, conceptually how to get started as a first timer. 

Once I know these I can get upto speed. Relying on full generation won't be good for my understanding of problem I'm solving either. So, it's a bit pointless to do so. But things like documentation etc generation I ask it to verify and suggest whether it follows standards.

spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/adapter/AbstractAdaptableMessageListener.java
arnabnandy7: @artembilan wrapped, done and dusted :) thanks for all the effort

artembilan:
Merged as https://github.com/spring-projects/spring-amqp/commit/fc404f095cbd4f894f638b16c68e5834770eb2da.

@arnabnandy7 ,

thank you again for a great contribution; looking forward for more!

arnabnandy7:
> Merged as https://github.com/spring-projects/spring-amqp/commit/fc404f095cbd4f894f638b16c68e5834770eb2da.
> 
> @arnabnandy7 ,
> 
> thank you again for a great contribution; looking forward for more!

Sure @artembilan I'll wait for upcoming collaboration as well.

PR #3648, opened 2026-09-13
Update docs

ngocnhan-tran1996:
* Use `JacksonJsonMessageConverter` class instead of `Jackson2JsonMessageConverter`.
* Update `supportedContentType` and its related methods to replace `supportedMediaType`.
* Document the JSON-compatible content type requirement for Jackson 3.x.


src/reference/antora/modules/ROOT/pages/amqp/message-converters.adoc
```diff
@@ -218,6 +218,10 @@ Starting with version 2.2, `application/json` is assumed if there is no `content
 To revert to the previous behavior (return an unconverted `byte[]`), set the converter's `assumeSupportedContentType` property to `false`.
 If the content type is not supported, a `WARN` log message `Could not convert incoming message with content-type [...]`, is emitted and `message.getBody()` is returned as is -- as a `byte[]`.
 So, to meet the `JacksonJsonMessageConverter` requirements on the consumer side, the producer must add the `contentType` message property -- for example, as `application/json` or `text/x-json` or by using the `JacksonJsonMessageConverter`, which sets the header automatically.
+
+NOTE: With the Jackson 3.x library, when converting from the `Message`, an incoming `MessageProperties.getContentType()` must be JSON-compliant, `"application/{asterisk}+json".isCompatibleWith(contentType)` is used by `MimeType` to perform this check.
+See https://github.com/spring-projects/spring-amqp/issues/3520[gh-3520]
```
artembilan: We don't use links to GH issues from our docs.
That's really not how technical documentation works.
If you'd like to add more info, just express it here directly.

artembilan (changes requested):
Please, revise my minor concern.

Thank you!

renechoi (commented):
The paragraph you are adding to already opens with a NOTE saying `contentType.contains("json")` is used. That is `AbstractJackson2MessageConverter`; `AbstractJacksonMessageConverter` gates on `supportedContentType.isCompatibleWith(...)` with `application/*+json`. Two rules six lines apart, so rewriting the first NOTE may beat appending a second.

That paragraph also still offers `text/x-json` as an example contentType. I ran it: `JacksonJsonMessageConverter` hands the body back as raw `byte[]`, while `Jackson2JsonMessageConverter` converts it.

src/reference/antora/modules/ROOT/pages/amqp/message-converters.adoc
```diff
@@ -218,6 +218,9 @@ Starting with version 2.2, `application/json` is assumed if there is no `content
 To revert to the previous behavior (return an unconverted `byte[]`), set the converter's `assumeSupportedContentType` property to `false`.
 If the content type is not supported, a `WARN` log message `Could not convert incoming message with content-type [...]`, is emitted and `message.getBody()` is returned as is -- as a `byte[]`.
 So, to meet the `JacksonJsonMessageConverter` requirements on the consumer side, the producer must add the `contentType` message property -- for example, as `application/json` or `text/x-json` or by using the `JacksonJsonMessageConverter`, which sets the header automatically.
+
+NOTE: With the Jackson 3.x library, when converting from the `Message`, an incoming `MessageProperties.getContentType()` must be JSON-compliant, `"application/{asterisk}+json".isCompatibleWith(contentType)` is used by `MimeType` to perform this check.
```
artembilan: See top-level comment: https://github.com/spring-projects/spring-amqp/pull/3648#pullrequestreview-5211044079

I think he is right: it is better to revise already existing NOTE instead of adding one more contradicting the previous one.
I would even say that the fist NOTE is a bug according to the current behavior.
Still fine to fix it here.

Thanks

artembilan (changes requested):

src/reference/antora/modules/ROOT/pages/amqp/message-converters.adoc
```diff
@@ -213,11 +213,12 @@ To handle this situation, set the `TypePrecedence` property on the `JacksonJsonM
...
 
-NOTE: When converting from the `Message`, an incoming `MessageProperties.getContentType()` must be JSON-compliant (`contentType.contains("json")` is used to check).
+NOTE: When converting from the `Message`, an incoming `MessageProperties.getContentType()` must be JSON-compliant, `"application/{asterisk}+json".isCompatibleWith(contentType)` is used by `MimeType` to perform this check.
 Starting with version 2.2, `application/json` is assumed if there is no `contentType` property, or it has the default value `application/octet-stream`.
 To revert to the previous behavior (return an unconverted `byte[]`), set the converter's `assumeSupportedContentType` property to `false`.
 If the content type is not supported, a `WARN` log message `Could not convert incoming message with content-type [...]`, is emitted and `message.getBody()` is returned as is -- as a `byte[]`.
 So, to meet the `JacksonJsonMessageConverter` requirements on the consumer side, the producer must add the `contentType` message property -- for example, as `application/json` or `text/x-json` or by using the `JacksonJsonMessageConverter`, which sets the header automatically.
```
artembilan: Thanks for the update!
I think this last sentence in the NOTE deserves more thinking.
As we know with from our previous collaboration, the mentioned `text/x-json` is not going to work since it is "not compatible with" `"application/*+json`.
And that is correct and expected by security hardening approach.

artembilan (changes requested):

PR #3654, opened 2026-09-16
GH-3653: Propagate addressShuffleMode to publisher factory

### LINKED ISSUE #3653: Propagate addressShuffleMode to the default publisher connection factory
## Problem
When `CachingConnectionFactory` uses its default publisher connection factory, configuring `addressShuffleMode` on the main factory does not configure the publisher factory.
Observed with Spring AMQP 4.1.0 and still present in 4.1.1.
## Reproduction
CachingConnectionFactory factory = new CachingConnectionFactory();
factory.setAddresses("rabbit-1:5672,rabbit-2:5672");
factory.setAddressShuffleMode(AddressShuffleMode.NONE);
The main connection uses NONE, but the publisher connection retains its default RANDOM mode and may connect to a different RabbitMQ node.

## Expected behavior
setAddressShuffleMode(...) should propagate the value to the default publisher connection factory, consistent with the documentation stating that properties configured on the main factory are propagated to the publisher factory.

Alternatively, expose a public getter or customization hook for the configured shuffle mode.

Current workaround requires configuring the publisher factory separately or accessing the main factory’s private field.

Gaurav1112:
Fixes #3653

`setAddressShuffleMode` on the main factory never reached the default publisher connection factory, so with `setAddresses("rabbit-1:5672,rabbit-2:5672")` and `AddressShuffleMode.NONE` the publisher sub-factory silently kept `RANDOM` and could connect to a different node. Every sibling setter that carries Spring-level state (`setAddresses`, `setConnectionNameStrategy`, `setCloseTimeout`, ...) already does the guarded `publisherConnectionFactory` delegation; this adds the same three lines to `setAddressShuffleMode`.

The new test mirrors the adjacent `channelListenersPropagateToDefaultPublisherFactory` and fails on `main` with `expected: NONE but was: RANDOM`. `:spring-rabbit:test --tests "*ConnectionFactory*"` and checkstyle are green.

While verifying the sibling idiom I noticed `setAddressResolver`, `setConnectionCreatingBackOff`, and `CachingConnectionFactory.setPublisherChannelFactory` also don't propagate — left untouched here; happy to file a follow-up if you consider any of them bugs rather than intended.

Developed with AI-agent assistance and human review.


spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactoryTests.java
```diff
@@ -2149,6 +2149,21 @@ void channelListenersPropagateToDefaultPublisherFactory() {
 				.isEqualTo(1);
 	}
 
+	// GH-3653
```
artembilan: We don't use this approach any more.
It is very easy to lose context of the original issue with time when we do some other (not related) changes into the code.

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactoryTests.java
```diff
@@ -2149,6 +2149,21 @@ void channelListenersPropagateToDefaultPublisherFactory() {
...
+	@Test
+	void addressShuffleModePropagatesToDefaultPublisherFactory() {
+		CachingConnectionFactory ccf = new CachingConnectionFactory("someHost", 1234);
+		ccf.setAddresses("host1:5672,host2:5672");
+		ccf.setAddressShuffleMode(AddressShuffleMode.NONE);
+
+		AbstractConnectionFactory publisher = (AbstractConnectionFactory) ccf.getPublisherConnectionFactory();
```
artembilan: Why do we need an explicit cast if we don't use an `AbstractConnectionFactory` API downstream?

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactoryTests.java
```diff
@@ -2149,6 +2149,21 @@ void channelListenersPropagateToDefaultPublisherFactory() {
...
+		assertThat(publisher).isNotNull();
+
+		assertThat(TestUtils.getPropertyValue(publisher, "addressShuffleMode", AddressShuffleMode.class))
```
artembilan: This is already a deprecated API:
```
	@Deprecated(since = "4.1", forRemoval = true)
	public static <T> @Nullable T getPropertyValue(Object root, String propertyPath, Class<T> type) {
```

Consider to use `assertThatObject()` instead - and your change will be compatible down to `4.0.x`!

artembilan (changes requested):
And to confirm my review, here is an excerpt from the logs:
```
/home/runner/work/spring-amqp/spring-amqp/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactoryTests.java:2162: warning: [removal] <T>getPropertyValue(Object,String,Class<T>) in TestUtils has been deprecated and marked for removal

		assertThat(TestUtils.getPropertyValue(publisher, "addressShuffleMode", AddressShuffleMode.class))
> Task :spring-rabbit:compileTestJava
		                    ^
  where T is a type-variable:
    T extends Object declared in method <T>getPropertyValue(Object,String,Class<T>)
```

So, consider to teach your AI to not rely on the deprecated API.
Or really try to be involved more in review of the code yourself.

Thanks

spring-rabbit/src/test/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactoryTests.java
```diff
@@ -2149,6 +2149,20 @@ void channelListenersPropagateToDefaultPublisherFactory() {
 				.isEqualTo(1);
 	}
 
+	@Test
+	void addressShuffleModePropagatesToDefaultPublisherFactory() {
```
artembilan: Sorry, I missed this one with the previous review.
The "addressShuffleMode propagates to"  what?
I mean this feels from sentence from the English perspective.

Wouldn't it be better sound like: `addressShuffleModeIsPropagatedToDefaultPublisherFactory` ?

Thanks

artembilan (changes requested):

